PR #309, opened 2014-11-07
Add URLs without slash at start of string

tommyp:
Due to an error in Specialist Publisher, we were sending URLs without a slash at the start as the ID to Rummager. The sitemap uses these IDs to build all the URLs and if they start with a slash assumes they are local and prepends https://www.gov.uk to them.

This commit changes the check to do this unless the URL starts with http, in which case it is an external URL. Previously this assumed that if it started with a slash, it was an internal URL and appended the slash.

[Ticket](https://govuk.zendesk.com/agent/tickets/856504).


edds:
I have taken the liberty of adding a simple unit test for this method. This should be good to merge now right?

PR #329, opened 2015-01-14
Allow facet example scope of "query"

rboulton:
Previously, when requesting examples for facet values, the only allowed
scope has been "global" - ie, return examples for the facet field
regardless of whether they match the query or not.

This adds the option of "query" scope - ie, return only examples which
match the query and filters.

The motivation of doing this now is to allow us to make services and
information pages only list examples which are produced by the
organisation the services and information page is for.

For example, on the HMRC services and information page, with global
scope we have a "Maritime Safety" section for which none of the 4
examples come from HMRC.  If the "query" scope is requested, it would contain only
examples which are tagged with HMRC.

(Note - this won't change anything on the site immediately: we'd also need to modify whitehall to request "query" scope to turn this option on on the services and information pages.)

### DIFF
--- a/lib/facet_example_fetcher.rb
+++ b/lib/facet_example_fetcher.rb
@@ -1,9 +1,10 @@
 # Fetch example values for facets
 class FacetExampleFetcher
-  def initialize(index, es_response, params)
+  def initialize(index, es_response, params, search_builder)
     @index = index
     @response_facets = es_response["facets"]
     @params = params
+    @search_builder = search_builder
   end
 
   # Fetch all requested example facet values
@@ -26,10 +27,18 @@ class FacetExampleFetcher
   end
 
 private
-  # Fetch facet examples for a given field
   def fetch_for_field(field_name, facet_params)
     example_count = facet_params[:examples]
     example_fields = facet_params[:example_fields]
+    scope = facet_params[:example_scope]
+
+    if scope == :query
+      query = @search_builder.query
+      filter = @search_builder.filter
+    else
+      query = nil
+      filter = nil
+    end
 
     facet_options = @response_facets.fetch(field_name, {}).fetch("terms", [])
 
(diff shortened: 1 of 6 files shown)


lib/facet_example_fetcher.rb
```diff
@@ -26,10 +27,18 @@ def fetch
...
 private
-  # Fetch facet examples for a given field
   def fetch_for_field(field_name, facet_params)
     example_count = facet_params[:examples]
     example_fields = facet_params[:example_fields]
+    scope = facet_params[:example_scope]
+
+    if scope == :global
+      query = nil
+      filter = nil
```
elliotcm: This branch isn't necessary.  `example_scope` isn't a required parameter to this object, so it's possible that neither branch of this conditional will execute.  In that case `fetch_by_slug` will be handed `nil` and `nil` anyway, same as if it was `:global`.  May as well not bother explicitly setting to `nil`, or if you're keen on being explicit, setting them to `nil` whenever the scope isn't `:query`.

lib/unified_search_builder.rb
```diff
@@ -25,8 +25,8 @@ def payload
     Hash[{
       from: @params[:start],
       size: @params[:count],
-      query: query_hash_with_best_bets,
-      filter: filters_hash,
+      query: query,
+      filter: filter,
```
elliotcm, line 29: What's the benefit of this change?  Memoization?

test/unit/search_parameter_parser_test.rb
```diff
@@ -528,17 +528,35 @@ def date_filter(field_name, values)
...
+    p = SearchParameterParser.new("facet_organisations" => ["10,examples:5,example_fields:slug:title,example_scope:query"])
+    assert p.valid?
+    assert_equal(expected_params({
+      facets: {
+        "organisations" => expected_facet_params({
+          requested: 10,
+          examples: 5,
+          example_fields: ["slug", "title"],
+          example_scope: :query,
+      })}}), p.parsed_params)
```
elliotcm: `})}})`.  Oof.

elliotcm:
I'd love some more whitespace in the tests to avoid Big Block of Text syndrome but that's subjective.

lib/unified_search_builder.rb
rboulton, line 29: Yes.

PR #330, opened 2015-01-15
Add mainstream_browse_pages field

rboulton:
Mainstream browse pages were modelled as three separate fields: section,
subsection and subsubsection.  This is unhelpful in many ways (can't
search directly for all content in a subsection, can't facet usefully,
can't represent tagging to multiple browse pages), so model them instead
as a single field containing the full path.

If the mainstream_browse_pages field isn't populated in supplied data,
build it from the "section", "subsection" and "subsubsection" fields.

In future, we'll get them in this form directly, at which point we'll
also be able to there may be multiple browse pages tagged to a piece of
content.

### DIFF
--- a/lib/elasticsearch/index.rb
+++ b/lib/elasticsearch/index.rb
@@ -476,6 +476,7 @@ module Elasticsearch
     def index_doc(doc_hash, popularities)
       if @is_content_index
         doc_hash = prepare_popularity_field(doc_hash, popularities)
+        doc_hash = prepare_mainstream_browse_page_field(doc_hash)
         doc_hash = prepare_tag_field(doc_hash)
       end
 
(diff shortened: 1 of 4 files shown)


lib/elasticsearch/index.rb
```diff
@@ -492,6 +493,32 @@ def prepare_popularity_field(doc_hash, popularities)
...
+      # Mainstream browse pages were modelled as three separate fields:
+      # section, subsection and subsubsection.  This is unhelpful in many ways,
+      # so model them instead as a single field containing the full path.
+      #
+      # In future, we'll get them in this form directly, at which point we'll
+      # also be able to there may be multiple browse pages tagged to a piece of
+      # content.
+      if doc_hash["mainstream_browse_pages"]
+        doc_hash
+      else
```
elliotcm: The double-nested `if` gets a little harder to follow by the end.  This branch could be replaced by a guard condition:
`return doc_hash if doc_hash["mainstream_browse_pages"]`.

test/unit/elasticsearch_index_test.rb
```diff
@@ -282,6 +282,35 @@ def test_should_populate_tags_field
...
+{"took":5,"items":[
+  { "index": { "_index":"test-index", "_type":"edition", "_id":"/foo/bar", "ok":true } }
+]}
+    eos
+    stub_request(:post, "http://example.com:9200/test-index/_bulk").with(
+        body: payload,
+        headers: {"Content-Type" => "application/json"}
+    ).to_return(body: response)
+    @wrapper.add [document]
+    assert_requested(:post, "http://example.com:9200/test-index/_bulk")
```
elliotcm: I believe `stub_request` returns an object you can use for assertions.  I think this would work:

```
bulk_request = stub_request(:post, "http://example.com:9200/test-index/_bulk").with(
    body: payload,
    headers: {"Content-Type" => "application/json"}
).to_return(body: response)

@wrapper.add([document])

assert_requested(bulk_request)
```

PR #332, opened 2015-01-16
Make RUMMAGER_INDEX=all refer to all indexes

rboulton:
Currently, it only refers to all indexes holding content.  This is
frequently confusing people.  Instead, make RUMMAGER_INDEX=all refer to
all indexes used by rummager.

This means that auxiliary indexes are also migrated when a
deploy:migrations is done - currently, they're not being.  This means
that if we changed the schemas of the auxiliary indexes, they wouldn't
be updated automatically, which also seems undesirable.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -58,14 +58,10 @@ end
 def index_names
   case ENV["RUMMAGER_INDEX"]
   when "all"
-    content_index_names
+    search_config.index_names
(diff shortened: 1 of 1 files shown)


mattbostock:
LGTM, though I'll leave it to someone more familiar with Rummager to merge.

PR #335, opened 2015-01-20
Identify entities during indexing

heathd:
This is part of the [firebreak work to use named entities to improve search results](https://gov-uk.atlassian.net/wiki/pages/viewpage.action?pageId=20185307). 

This PR changes rummager to use the new [entity-extractor](https://github.com/alphagov/entity-extractor) service when indexing documents. Any entities found in the document are stored in the `entities` field of the document in the elasticsearch index.

The `rummager:migrate_index` task has been modified so that it disables entity extraction by default. This is because entitiy extraction could slow down the migration by up to 20 minutes. There is now an option to that rake task to enable entity extraction should it be required, e.g. when first populating the field.

The entity extraction feature is controlled by a config setting `enable_entity_extraction`. This is disabled by default (except if `RACK_ENV=='development'`). It can be overridden by adding an initializer in the deploy scripts.

For the time being we want this disabled in staging/production and only enabled in preview.

### DIFF
--- a/lib/entity_extractor_client.rb
+++ b/lib/entity_extractor_client.rb
@@ -11,18 +11,37 @@ class EntityExtractorClient
     @service_base_url = URI.parse(service_base_url)
     @options = default_options.merge(options)
     @logger = options[:logger] || Logging.logger[self]
+    @swallow_connection_errors = options[:swallow_connection_errors] || false
+    @had_connection_error = false
   end
 
   def call(document)
-    response = RestClient.post(extract_url, document, @options)
-    JSON.parse(response)
-  rescue RestClient::Exception => e
-    logger.error(e)
-    nil
+    if swallow_connection_errors? && had_connection_error?
+      nil
+    else
+      response = RestClient.post(extract_url, document, @options)
+      JSON.parse(response)
+    end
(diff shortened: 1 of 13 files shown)


lib/entity_extractor_client.rb
```diff
@@ -0,0 +1,39 @@
...
+  def initialize(service_base_url, options = {})
+    @service_base_url = URI.parse(service_base_url)
+    @options = default_options.merge(options)
+    @logger = options[:logger] || Logging.logger[self]
+  end
+
+  def call(document)
+    response = RestClient.post(extract_url, document, @options)
+    JSON.parse(response)
+  rescue RestClient::Exception => e
```
rboulton: I don't think we should be swallowing this exception.  We discussed maybe swallowing connection error in development environments (to allow developers to run migrations if the entity extractor isn't running).

test/unit/elasticsearch_index_test.rb
```diff
@@ -3,6 +3,7 @@
 require "search_config"
 require "webmock"
 
+
```
rboulton: This extra whitespace smells a bit pythony to me.

rboulton:
Happy to merge this other than the exception swallowing bit.

PR #340, opened 2015-01-23
Use entities in searches

rboulton:
When a search is performed, make a call to the entity extractor service to see if any known named entities are found in the query.  If they are, boost results containing those entities.  The boost is probably much too high at present, but this is deliberate so that we can clearly see its effect.

The entity extractor is only currently enabled in preview. It is currently disabled in production and staging, so this will have no effect there.

In development, if the service isn't running this will report an error to the console, but proceed as if no entities were found, so this shouldn't make it harder to develop other applications which depend on calls to rummager.

@heathd and @rboulton paired on this.

### DIFF
--- a/lib/unified_search_builder.rb
+++ b/lib/unified_search_builder.rb
@@ -11,7 +11,7 @@ class UnifiedSearchBuilder
   GOVERNMENT_BOOST_FACTOR = 0.4
   POPULARITY_OFFSET = 0.001
 
-  def initialize(params, metaindex)
+  def initialize(params, metaindex, entity_extractor)
     @params = params
     @query = params[:query]
     if @params[:debug][:disable_best_bets]
@@ -19,6 +19,7 @@ class UnifiedSearchBuilder
     else
       @best_bets_checker = BestBetsChecker.new(metaindex, @query)
     end
+    @entity_extractor = entity_extractor
   end
 
   def payload
@@ -76,7 +77,7 @@ class UnifiedSearchBuilder
   end
 
   def boost_filters
-    format_boosts + [time_boost] + [closed_org_boost] + [devolved_org_boost]
+    (format_boosts + [time_boost, closed_org_boost, devolved_org_boost, entities_boost]).compact
   end
 
   def best_bets
@@ -397,4 +398,14 @@ class UnifiedSearchBuilder
     }
   end
 
+  def entities_boost
+    {
+      filter: { terms: { entities: find_entities} },
+      boost: 20
+    } unless find_entities.empty?
+  end
+
+  def find_entities
+    @entities ||= @entity_extractor.call(@query) || []
+  end
 end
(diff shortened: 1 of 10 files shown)


lib/unified_search_builder.rb
```diff
@@ -397,4 +398,14 @@ def devolved_org_boost
...
   end
 
+  def entities_boost
+    {
+      filter: { terms: { entities: find_entities} },
+      boost: 20
+    } unless find_entities.empty?
+  end
+
+  def find_entities
```
elliotcm: This could just as easily be `entities` instead of `find_entities`, especially since it's memoized.

lib/unified_search_builder.rb
```diff
@@ -397,4 +398,14 @@ def devolved_org_boost
...
+    @entities ||= @entity_extractor.call(@query) || []
```
elliotcm: Is there any situation in which one of the extractors would return `nil`?

test/unit/unified_search_builder_test.rb
```diff
@@ -96,7 +101,8 @@ def with_base_filters(filter)
       stub_zero_best_bets
       @builder = UnifiedSearchBuilder.new(
         query_options,
-        @metasearch_index
+        @metasearch_index,
+        stub_entity_extractor,
```
elliotcm: This could do with some DRYing up.

test/unit/unified_search_builder_test.rb
```diff
@@ -502,12 +517,16 @@ def with_base_filters(filter)
     end
   end
 
+  context "" do
+  end
```
elliotcm: Empty context needs deleting.

test/unit/unified_searcher_test.rb
```diff
@@ -163,7 +168,7 @@ def with_base_filters(filter)
     setup do
       @combined_index = stub("unified index")
       mock_best_bets("cheese")
-      @searcher = UnifiedSearcher.new(@combined_index, @metasearch_index, {}, {}, stub_suggester)
+      @searcher = UnifiedSearcher.new(@combined_index, @metasearch_index, {}, {}, stub_suggester, stub_entity_extractor)
```
elliotcm: Also worth DRYing up.

elliotcm:
Some minor twiddlings desired, otherwise top quality beans :+1:

lib/unified_search_builder.rb
heathd: The 'error swallowing' one used in development will return nil if it can't connect to the entity extractor on its second (or subsequent) attempts.

elliotcm: OK

lib/unified_search_builder.rb
rboulton: Done

test/unit/unified_search_builder_test.rb
rboulton: done

PR #345, opened 2015-01-30
Remove all entity extraction related code

rboulton:
This removes the code which calls the entity-extractor service, the
schema entries for the entities fields, the code that indexes the
entities, and the code that uses entities at search time.

This code was experimental, and we're not planning on taking the
experiment further at present.

Cleaning this code up in a single commit rather than reverting the
commits where it was added should make it easier to see the full extent
of the changes we made to support entity extraction (for future
reference if we want to try this again), and probably makes for a
cleaner history too.  Additionally, there was some minor refactoring
performed as part of the entity extraction work which we want to retain,
so we don't want to just revert all the commits.


heathd:
all looked good, just noticed that the `EntityExtractorClient` implementation was still there, so have removed, maybe should squash that commit down

elliotcm:
The `entity_extraction.rb` initializer is also still there, though empty.

elliotcm:
Correction, it isn't, Atom's index was out of date.

heathd:
@elliotcm thanks for checking all the same! grateful for another pair of eyes on this

heathd:
I've squashed the commit

elliotcm:
Happy to merge once green.

rboulton:
Thanks @heathd - missed that one.

PR #347, opened 2015-02-03
Group facet example searches

rboulton:
When calculating examples for facet searches (eg, with queries such as
facet_organisations=100,example_scope:query,examples:10), we use
elasticsearch's `msearch` API to send a query for each facet value to
gather the examples. This is fine when there's a reasonably small number
of facet values to fetch examples for, but can cause problems if too
many facet values are found.

There are two problems which may occur:
- Elasticsearch has an internal queue limit on the number of searches
  to be performed: this defaults to 1000.  If we go close to this
  limit, we risk getting error responses saying that the queue is full.
  I believe the 1000 limit is shared across all nodes, so each query
  submitted by a user counts several times against the 1000 limit. I
  heave experimentally verified that it is possible to hit the limit
  without actually sending as many as 1000 queries simultaneously.
- Elasticsearch may take too long to perform the individual queries
  involved in the `msearch` call and time out.

For our current database, the most pathological field is the
organisations field, which can trigger nearly 900 searches for examples
in the worst case.  This takes 9 seconds to run on my dev VM, and times
out on the live site (when called by accessing the publicly exposed
search API directly).  There are no "user-facing" pages on the live site
which trigger such pathological cases, though, because facet examples
are only requested for highly filtered searches.

To avoid these problems, instead of sending all the searches at once, we
send them in batches of 50.  This means that we wait for 50 queries to
have been fully completed before sending the next 50, which avoids both
the queue size problem and the timeout problem.

In future, this workaround will be insufficient to return results fast
enough.  An alternative approach would be to use a custom elasticsearch
"aggregation" to gather information on the facet examples in a single
pass through the index.  This ought to be many times faster, but is
dependent on us upgrading the version of elasticsearch in use in
production (we're currently on a version which doesn't support
aggregations).

### DIFF
--- a/lib/facet_example_fetcher.rb
+++ b/lib/facet_example_fetcher.rb
@@ -48,7 +48,7 @@ private
     if slugs.empty?
       {}
     else
-      fetch_by_slug(field_name, slugs, example_count, example_fields, query, filter)
+      batched_fetch_by_slug(field_name, slugs, example_count, example_fields, query, filter)
     end
   end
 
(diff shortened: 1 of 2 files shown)


dhwthompson:
I’m concerned that this change didn’t require any changes in the unit tests. Even if existing behaviour hasn’t changed, presumably there’s some case not covered by the tests where this new code would behave differently?

dhwthompson:
Ah, wait, I think I understand this more clearly: this is purely about the _timing_ of requests, isn’t it? I can see how that’d be a pain to test.

dhwthompson:
Oh no, wait…

My brain hurts

evilstreak:
I think this probably _could_ have a test, since the effect of the change is to have rummager make multiple `msearch` requests to elasticsearch where it previously would have had a single request.

Given that the original method already did all of the result fetching and formatting and just needs the batching added, can we do something like this?

``` ruby
def batched_fetch_by_slug(field_name, slugs, example_count, example_fields, query, filter)
  some_results = slugs.each_slice(50).map { |fewer_slugs|
    fetch_by_slug(field_name, fewer_slugs, example_count, example_fields, query, filter)
  }
  some_results.reduce(&:merge)
end
```

rboulton:
Test added, and @evilstreak's advice about putting the batching code into a separate method followed. (The comment made perfect sense when I tried implementing it, and it's much nicer that way.)

evilstreak:
:+1: Looks good to me, though since you've used the code I suggested in a comment I'll wait for someone else to merge this.

PR #348, opened 2015-02-04
Update countryside stewardship grant schema

evilstreak:
These changes reflect those made in specialist publisher. No documents of this type have been published yet, so there are no concerns about migrating data.

See https://github.com/alphagov/specialist-publisher/pull/403


jackscotti:
:+1:

PR #351, opened 2015-02-09
Add policies attr to core doctype

tommyp:
As we want content to be tagged to Policies, this commit adds the policies attribute to the core doctype.


rboulton:
@tommyp This field needs to be added to the schema for the government index in `config/schema/elasticsearch_schema.yml` too.  That index hasn't been migrated over to the json-format schema files yet.

PR #353, opened 2015-02-09
Remove duplicate code for registry classes and tests

KushalP:
There are a number of `FooRegistry` classes like `TopicRegistry`, and others which duplicate a lot of code. To make it easier for others to read and understand all of these classes, move them all into a single module `Registry`, where they can inherit from a common `BaseRegistry` class.

That way, the differences and similarities between all of these classes can be made clearer.

Also reduce duplication of tests now that this change has been made.


rboulton:
Lovely code cleanup. :heart:

lib/registry.rb
```diff
@@ -0,0 +1,119 @@
...
+          organisation.organisation_type = MINISTERIAL_DEPARTMENT_TYPE
+        end
+      end
+
+      organisation
+    end
+  end
+
+  class SpecialistSector < BaseRegistry
+    CACHE_LIFETIME_SECONDS = 300
```
alext, line 78: This constant has a different name to the one on the base class, meaning it won;t have any effect.  Additionally, I don't think it would apply even if the name did match due to scoping.  I'll have a look at fixing this.

PR #354, opened 2015-02-09
Add policies to allowed filters

tommyp:
With the policy finder, we need to be able to filter by the policies attr. This commit adds it to the `ALLOWED_FILTER_FIELDS` array so we can use it.

### DIFF
--- a/lib/search_parameter_parser.rb
+++ b/lib/search_parameter_parser.rb
@@ -19,6 +19,7 @@ class BaseParameterParser
     mainstream_browse_pages
     manual
     organisations
+    policies
     section
     specialist_sectors
   )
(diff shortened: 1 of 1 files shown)


rboulton:
This looks fine - but I'd advocate adding it to ALLOWED_RETURN_FIELDS too, so you can actually determine what policies a document is tagged to in the results of searches.

It would be useful to add it to ALLOWED_FACET_FIELDS, also, so that you can easily investigate which policies are included in the results of a search.  (We could then add it to the fields in the tags in https://govuk-content-explorer.herokuapp.com/ too, making it even easier to see this.)

tommyp:
@rboulton Done. Also I didn't know about that App. Handy to know about.

rboulton:
Great :star:

PR #357, opened 2015-02-10
Reduce the duplication of similar methods

KushalP:
All of these methods are doing very similar operations. Simplify them
into a single method so that we can understand where we're doing similar
operations on data structures.

### DIFF
--- a/lib/result_set_presenter.rb
+++ b/lib/result_set_presenter.rb
@@ -29,32 +29,32 @@ private
 
     if result['document_series'] && should_expand_document_series?
       result['document_series'] = result['document_series'].map do |slug|
-        document_series_by_slug(slug)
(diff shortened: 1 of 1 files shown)


lib/result_set_presenter.rb
```diff
@@ -108,55 +108,9 @@ def spelling_suggestions
...
-      {"slug" => slug}
-    end
-  end
-
-  def sector_by_slug(slug)
-    sector = specialist_sector_registry && specialist_sector_registry[slug]
-    if sector
-      sector
+  def structure_by_slug(structure, slug)
+    if items = structure && structure[slug]
```
alext: In this case I think `items` is always actually a single item.  Might be better named in the singular.

PR #366, opened 2015-03-10
Add a doctype for policies

tommyp:
This commit adds a doctype file for policies. It contains no additional properties and is published by Policy Publisher. [Ticket](https://trello.com/c/TvQngvWL/70-add-policies-to-rummager).


rboulton:
Note that this should be deployed to production before https://github.com/alphagov/policy-publisher/pull/29 to ensure that the schema for policies is present.

PR #368, opened 2015-03-16
Update the schema for the CSF with new filters

issyl0:
- A corresponding change for https://github.com/alphagov/specialist-publisher/pull/446.

PR #369, opened 2015-03-16
Rework schema definitions

rboulton:
Reviewers will probably want to read the following description, and then go through the changes commit-by-commit.

Schemas were previously defined for each document type by essentially
writing the elasticsearch mapping for the field, and adding a few additional
properties to it using a "details" hash.

This had several shortcomings:
- it was possible for fields with the same name to be defined in
  different ways in different document types.  This makes it awkward to
  implement, eg, faceting across all document types.
- it was awkward to change the elasticsearch configuration for a
  particular type of field, eg, searchable text fields, since we need
  to change the configuration in all places.  This makes it hard to implement
  better support for synonyms, for example.
- any non-elasticsearch configuration, such as the allowed values for a
  field, needed to be separated out from the elasticsearch
  configuration in order to set the schema.
- elasticsearch configuration was being cut-and-pasted when defining
  new fields, resulting in invalid configurations being created.  For
  example, we had several date fields defined with a `index` property
  with a value of `analyzed` - but this value is only valid for string
  fields.  Keeping the elasticsearch configuration in one place makes
  would make it easier to audit that it is being defined correctly.

Instead the schema definitions now split the elasticsearch configuration
from other metadata, and define the type of each field in only one
place.  This is done by splitting schemas into three pieces (each
defined in JSON files):
- A `field_types` file which defines some named types to be used when
  defining fields.  These types contain all the elasticsearch
  configuration to be used for the field as a property (`es_config`),
  but also contain information on other application-level things such
  as whether the field is allowed to have multiple values.  They also
  contain a `description` property, which is a comment field,
  describing what the type is indended to be used for.
- A `field_definitions` file which defines the actual named fields
  being used in the search index - tying field names to the types
  defined in the `field_types` file.  This ensures that there is only
  one definition for each field name.
- JSON files for each format (and `core.json` for fields which exist in
  all formats) which define which fields are used by each document
  type.  These files can also define the values which are allowed to be
  taken by a named field in documents of a particular type.
- JSON files for each index, defining which field types are allowed in that index.

Changing the schema in this way has been discussed several times, but
the motivation for doing it now is:
- To improve our synonym support, we'll need to modify the index-time
  configuration of searchable text fields, and this will be easier if
  the schema definitions allow us to change how all searchable text
  fields are indexed in a single place.
- Migrating to elasticsearch 1.x requires us to define in some way
  which fields should return array values, in order to retain API
  compatibility in return values (and in order to provide reasonably
  easy to work with return values).  Adding this information to the
  schemas as they were defined would have made it more awkward to
  rework the schemas in future.  In this new scheme, it is simply
  necessary to define a field type with the multivalued property set to
  true, and use that field type for the appropriately named field.

In order for this rework to be useful, the "government" index is now
included in the "JSON" style schemas.  Previously, it had been left
hard-coded in the elasticsearch_schema.yml file.

One other change which this schema change requires is to index all
date fields for sorting and filtering.  This change is simply for
convenience; indexing all the date fields for searching will use slightly
more index space, but very few extra resources unless they're being
used.

No changes should happen to the public API of rummager as a result of
these schema changes.

There's lots of refactoring to areas of the code like `index_group.rb` and
`index.rb` that can easily be done following this change, but I've resisted
doing this for now.


alext:
Needs a rebase (probably because of #368)

config/schema/base_document_type.json
```diff
@@ -0,0 +1,22 @@
+{
+  "fields": [
+    "format",
+    "title",
+    "title",
```
rboulton: @dsingleton pointed out to me that there are dupes in this list - there aren't meant to be - will fix in a bit.

config/schema/README.md
```diff
@@ -0,0 +1,105 @@
+# Schema definitions
+
+This directory tree holds the schema definitions for our search indexes.
+The purpose of the schema is to define:
+
+ - which fields are allowed to be present in documents passed to the search
+   indexer
+ - how these fields should be processed both at search indexing time
```
alext: This doesn't read quite correctly...  is 'both' left over from a previous wording of this?

config/schema/README.md
```diff
@@ -0,0 +1,105 @@
...
+   for (this is essentially a comment field).
+
+ - `es_config`: the elasticsearch configuration to be applied to fields using
+   this type.
+
+ - `multivalued`: (optional, default of `false`) - if true, the field is
+   allowed to contain multiple values, and will be represented as an array when
+   documents are being returned.
+
+ - `children`: (optional) - if present, the field is contains child fields (ie,
```
alext: "is contains" seems wrong.

alext:
:+1: Looks good to me.

PR #373, opened 2015-03-16
Correct regexes used to identify schema files

rboulton:
The service-manual configuration file wasn't matching the regex used to
identify `index_schema` JSON files.  This commit fixes this regex, and
also tidies both it and the `document_type` regex up:
- indexes must start with a letter, followed by letters and hyphens.
- document_types must start with a letter, followed by letters and
  underscores.


evilstreak:
:+1:

PR #375, opened 2015-03-19
Replace the FieldDefinitions object with a hash

rboulton:
The FieldDefinitions object is just a wrapper around a hash, with a
`.get` method to look up values, and a `.es_config` method which was
only used during parsing.  It's note helping much with structuring anything,
and when making use of the definitions it would be easier if we just had a hash,
so this commit replaces the FieldDefinitions object with a hash.

### DIFF
--- a/lib/schema/document_types.rb
+++ b/lib/schema/document_types.rb
@@ -37,7 +37,10 @@ class DocumentTypeParser
     end
 
     fields = Hash[field_names.map { |field_name|
-      [field_name, field_definitions.get(field_name)]
+      field_definition = field_definitions.fetch(field_name) do
+        raise_error(%{Undefined field "#{field_name}"})
+      end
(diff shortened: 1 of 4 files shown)


lib/schema/document_types.rb
```diff
@@ -37,7 +37,11 @@ def parse
     end
 
     fields = Hash[field_names.map { |field_name|
-      [field_name, field_definitions.get(field_name)]
+      field_definition = field_definitions[field_name]
```
evilstreak: I think it would be better to use `fetch` for this:

``` ruby
field_definition = field_definitions.fetch(field_name) {
  raise_error(%{Undefined field "#{field_name}"})
}
```

rboulton: Agreed - I'd forgotten about that form of `fetch`. rebased.

PR #377, opened 2015-03-24
Add content_id to schema

tekin:
This will enable us to reference and search documents based on their content_id. Specifically, we're going to reference whitehall document's new-world "policies" using their content_ids.

Story: https://trello.com/c/h2t4VlGE/31-allow-documents-in-whitehall-to-be-tagged-to-new-policy-areas


bradwright:
Are these going to leak into the main API?

PR #380, opened 2015-03-25
Populate 'public_timestamp' for all documents

tekin:
specialist-publisher sets 'last_update' as the timestamp signifying the freshness of a document. This is semantically identical to 'public_timestamp', which is the field that other publishing applications use. To ease the transition to using 'public_timestamp' across all applications (and removing 'last_update'), 'public_timestamp' will be set on any documents that are indexed with 'last_update'. A re-index will update all existing documents.

This change will allow us to update finder-frontend to order rummager queries by 'public_timestamp', which will mean results can be fetched for both specialist content and whitehall content (whitehall content doesn't set 'last_update' so ordering by 'last_update' excludes them from results).

### DIFF
--- a/lib/elasticsearch/index.rb
+++ b/lib/elasticsearch/index.rb
@@ -473,6 +473,7 @@ module Elasticsearch
         doc_hash = prepare_mainstream_browse_page_field(doc_hash)
         doc_hash = prepare_tag_field(doc_hash)
         doc_hash = prepare_format_field(doc_hash)
+        doc_hash = prepare_public_timestamp_field(doc_hash)
       end
 
       doc_hash = prepare_if_best_bet(doc_hash)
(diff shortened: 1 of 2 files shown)


tekin:
Not entirely sure how to test this. Guidance would be much appreciated.

rboulton:
I think some unit tests of this would be good.  There's no neat way to do this, presently; the ones I added for a similar thing to populate the `mainstream_browse_pages` field are at https://github.com/alphagov/rummager/blob/master/test/unit/elasticsearch_index_test.rb#L293 - it would be reasonable (though ugly) to just make another test like that.  Lots of stubbing, though.

In terms of testing in preview/production - the best approach is probably to run the migration, and then use elasticsearch `head` plugin to check examples of documents which are meant to have had the public_timestamp added.

tekin:
Now with added unit test.

rboulton:
:up: :banana:

PR #384, opened 2015-03-26
Remove content_id from default field

tekin:
We are no longer using content_id to map to policies, which makes the content_id redundant in the field definitions.


rboulton:
Thanks for keeping things clean.

PR #385, opened 2015-03-26
Use schemas in documents

rboulton:
Now that we have well-defined schema objects, which include information in particular about which fields are multivalued or not, we want to use these schemas in the Document object to enforce that non-multivalued fields aren't passed arrays, and to return only multivalued fields as arrays.  This is part of making rummager work correctly with both elasticsearch 0.90.x and 1.x releases.  (It may even be the last part of this, though that is probably too much to hope for.)

As a result of this, rummager will now raise an error at indexing time if it is passed a field which is multivalued when it shouldn't be, or vice-versa.  This would show up in errbit for rummager, and we'll then fix either the publishing application or the schema to match.  Unfortunately, there's no way to report such errors to the publishing application at present, since indexing happens in a queue, but I think we'll just have to live with that.

The actual change to use the multivalued information is very clean and simple: https://github.com/alphagov/rummager/commit/534462ca9199991f3aa97d8e03cb73a80b72752d, however, there's a huge amount of plumbing that needed to be updated to get the correct schema information into the Document, which is what the rest of this PR is for.  It's probably best to review the code changes commit-by-commit, since I've grouped the changes into commits which should make some logical sense, but note that I've rebased it around a lot so many intermediate commits don't produce actual running code.

In order to reduce the stubbing in our test suites, and to avoid having to add a whole load more, I've converted our test suites to use index names based on the schema (but with a "_test" suffix) almost everywhere, instead of made up names like "test-index". To go with this, I've introduced a `test/sample_config.rb` helper which tests use to get at the parts of the schema they need.  There are a unfortunately large number of test changes needed simply to start using this mechanism.

In addition to the tests, I've manually checked that it is still possible to run the `rummager:migrate_index` task on all indexes, since this could easily have been broken by the schema update work, and by being more fussy about the multivalued status of fields.


KushalP:
:+1: overall and you already know about the minor stuff I told you in person.

PR #391, opened 2015-03-31
Rename some field types

rboulton:
Renames the "searchable_date" field type to just "date", and the
"stored_float" field type to just "float".

There are no other types of date or float fields at present, so it's not
particularly helpful to try and be more specific about these names.  If
we need to add separate configuration for only some date fields in
future (probably for performance reasons), we should change the names to
indicate the ways in which the type can be used at that point, but
currently there's no need to add this level of detail.

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -21,7 +21,7 @@
   },
 
   "popularity": {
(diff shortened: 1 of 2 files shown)


evilstreak:
:+1:

PR #394, opened 2015-04-07
Add support for "reject"-style filters

rboulton:
For example:

  /unified_search.json?reject_organisations=hm-revenue-customs

would return all documents, except for those tagged with the
`hm-revenue-customs` organisation.  This is wanted for the new search
interface for manuals, where the results filtered to the manual are
to be shown in one area, and all the other results in a separate area.

This is implemented by adding a "reject" parameter to the Filter object,
which tracks whether the filter is a rejection filter or not.

If a user supplies both "reject_organisations" and
"filter_organisations", both the filters need to match.  This will
usually not be intended behaviour, but it might be convenient to write a
frontend application this way, so I haven't made it into a validation
failure.

### DIFF
--- a/lib/search_parameter_parser.rb
+++ b/lib/search_parameter_parser.rb
@@ -1,4 +1,5 @@
 require "ostruct"
+require "unf"
 
 class BaseParameterParser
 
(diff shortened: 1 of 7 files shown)


tadast:
An alternative syntax to this could be reusing `filter_organisations` with a `-` prefixed to the values e.g. `/unified_search.json?filter_organisations=-hm-revenue-customs`, this would make it less likely for the user to submit unmatching filters.

rboulton:
I'm not sure what you mean by "unmatching" filters.  I considered this kind of syntax, but for some of the filter fields it's possible that arbitrary punctuation could be valid in the field, so I'd prefer to keep special syntax out of the parameter value.  Otherwise, we'll have to start introducing ways to escape values, etc, and it will rapidly become unworkable.

tadast:
I was referring to the first paragraph in your description. Yeah that makes sense, LGTM :+1:

PR #403, opened 2015-04-20
Downweight historic WH content

dsingleton:
We want to downweight historic content so users are less likely
to be shown historic content when using a broad search term. When
searching specifically for historic content it should still rank
highly. Eg, when searching for something like "tax" historic
content should be less visible, but when searching for the title
of a specific historic tax document it should still be top.

(Historic means that the content is political and published under
a prior government, eg, not reflecting the political stance of
the current government)

Non political content will be unaffected.

This is an initial weighting, which needs some testing and
analysis before deciding.

This PR for discussion while we agree an initial value and analysis approach, it probably shouldn't be merged yet.
- [x] Agree initial down weighting value

### DIFF
--- a/lib/unified_search_builder.rb
+++ b/lib/unified_search_builder.rb
@@ -80,7 +80,7 @@ class UnifiedSearchBuilder
   end
 
   def boost_filters
-    (format_boosts + [time_boost, closed_org_boost, devolved_org_boost]).compact
+    (format_boosts + [time_boost, closed_org_boost, devolved_org_boost, historic_edition_boost]).compact
   end
 
   def best_bets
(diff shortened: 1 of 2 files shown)


rboulton:
Tests need fixing, and as indicated a value needs agreeing, but this code looks good to me.

dsingleton:
We've agreed 50% as an initial down-weighting. We'll be updating it based on analytics.

dsingleton:
(Sorry, didn't mean to merge, forgot this wasn't whitehall)

rboulton:
For the record, I'd have been happy to merge this if David hadn't got there first!

PR #404, opened 2015-04-22
Add organisations and people to policy schema.

elliotcm:
Both fields are already defined in `field_definitions.json` as they're used by the edition schema.

### DIFF
--- a/config/schema/document_types/policy.json
+++ b/config/schema/document_types/policy.json
@@ -1,5 +1,7 @@
 {
   "fields": [
-    "slug"
+    "slug",
+    "organisations",
+    "people"
   ]
 }


config/schema/document_types/policy.json
```diff
@@ -1,5 +1,7 @@
 {
   "fields": [
-    "slug"
+    "slug",
+    "organisations",
```
rboulton, line 4: This isn't needed, since "organisations" is already in the `base_document_type.json` schema (on directory level up), which is merged with this one.

Adding "people" is needed, though, since that one isn't in the base.

elliotcm, line 4: OK, in that case I'll close this PR since I've just discovered people aren't yet taggable to new policies.  We've another ticket for handling people on policies so this can be handled then, since it isn't a "while I'm here" task anymore.

elliotcm, line 4: Correction, I forgot how to Rails.

elliotcm:
Superseded by https://github.com/alphagov/rummager/pull/405

PR #407, opened 2015-04-24
Make title sorting case insensitive

russellthorn:
Ruby orders by ascii value by default meaning capital letters come
before lowercase letters i.e. Z = 90 comes before a = 97

I've added .downcase to sorting for titles to fix this and added a
test

https://www.pivotaltracker.com/story/show/89405790


evilstreak:
:+1:

rboulton:
Nice, thanks. :+1:

PR #410, opened 2015-05-05
Add health check for spelling suggestions

tijmenb:
We are currently working on improving the spelling suggestions for search (a PR for that will come later today). This commit adds "health checks" for suggestions.

It works by downloading a list of search terms and their expected results [from a public Google Doc](https://docs.google.com/spreadsheet/pub?key=0AmD7K4ab1dYrdDR5c2tITTNHRUZqajFTTU8wODAzZ1E). The terms are then compared to the actual suggestions returned from gov.uk search. The new health check reports the percentage of correctly corrected search terms in CI, as well as the terms that pass/fail:

screen shot 2015-05-05 at 09 57 49

As far the commits, 4959a30, ec08411, e70b0a2, a8ad9c8 and f9b3071 do some refactoring to make the change easier. 

1cd3600 adds the new Google Doc sheet to the downloader, 94f2b92 adds the actual functionality of running the health checks.

Trello: https://trello.com/c/GibHi6Aa/68-term-suggester-spelling-corrections-using-elasticsearch


lib/health_check/suggestion_checker.rb
```diff
@@ -0,0 +1,41 @@
...
+      @search_client = options[:search_client]
+    end
+
+    def run!
+      Logging.logger[self].info("Connecting to #{@search_client.to_s}")
+
+      calculator = Calculator.new
+
+      parsed_checks.each do |search_term, expected_result|
+        suggested_queries = @search_client.search(search_term)[:suggested_queries]
```
rboulton: It would be good to pass `count: 0` to the client to avoid making rummager return a list of results (which should reduce network transfer and load on the server).

PR #411, opened 2015-05-05
Add a `spelling_text` field for suggestions

tijmenb:
This field contains a copy of all the text that's used for searching, tokenised only by splitting it into words and groups of words ("shingles") and lowercasing.

Stemming, stopwording, etc, are not applied. This is to support spelling correction, which otherwise will try to correct words to their stemmed forms.  eg "prisioner" would be corrected to "prison" if corrected using the `_all` field.  Using the `spelling_text` field, it is corrected to "prisoner".

Lowercasing is needed so that words which are consistently capitalised in our documents (such as "PAYE"), but are not consistently capitalised in searches, are still corrected to useful variants.  It would be nice to avoid having to lowercase, so that we could, for example, suggest proper nouns with the appropriate capitalisation, but that is less important than making basic corrections work at present.

The shingles filter adds "shingles" (token n-grams) to the field to allow corrections like "driving liecence" to "driving licence". [From the docs](http://www.elastic.co/guide/en/elasticsearch/reference/1.5/analysis-shingle-tokenfilter.html):

> 'For example, the sentence "please divide this sentence into shingles" might be tokenized into shingles "please divide", "divide this", "this sentence", "sentence into", and "into shingles".'

The field is generated by elasticsearch at index time by copying any `searchable_text` or `searchable_identifier` fields.  It therefore won't appear in the _source field when fetching documents, or be shown.

However, it is possible to search on it, and use it as the target field for a TermSuggester for spelling corrections, etc.

Original commit by @rboulton
## Testing

Reindex rummager by running:

```
bundle exec rake rummager:migrate_index RUMMAGER_INDEX=all
```

### DIFF
--- a/config/schema/base_document_type.json
+++ b/config/schema/base_document_type.json
@@ -14,6 +14,7 @@
     "updated_at",
     "last_update",
     "public_timestamp",
-    "latest_change_note"
+    "latest_change_note",
+    "spelling_text"
   ]
 }
--- a/config/schema/elasticsearch_schema.yml
+++ b/config/schema/elasticsearch_schema.yml
@@ -22,6 +22,10 @@ index:
           type: custom
           tokenizer: standard
           filter: [standard, lowercase, stemmer_override, stemmer_english]
+        spelling_analyzer:
+          type: custom
+          tokenizer: standard
+          filter: [standard, lowercase, shingle]
       filter:
         stemmer_english:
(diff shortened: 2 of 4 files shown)


rboulton:
This looks good to me, but I did half of it.  @KushalP could you take a look and merge if happy.

PR #412, opened 2015-05-06
Refactor UnifiedSearchPresenter

tijmenb:
These commits refactor the UnifiedSearchPresenter class and associated tests. By consolidating arguments it brings down the arity of `UnifiedSearchPresenter.new` from 10 to 6 and improves readability of the tests. 

Immediate win is that it's no longer necessary to add new parameters for an upcoming change relating to the spelling suggester.

Best reviewed per-commit in chronological order (start at 97ec7e6).

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -190,11 +190,11 @@ class Rummager < Sinatra::Application
     result_streams = searcher.search(@query, organisation, params["sort"])
 
     result_context = {
-      organisation_registry: organisation_registry,
-      topic_registry: topic_registry,
-      document_series_registry: document_series_registry,
-      document_collection_registry: document_collection_registry,
-      world_location_registry: world_location_registry
+      organisations: organisation_registry,
+      topics: topic_registry,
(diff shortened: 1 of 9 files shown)


evilstreak:
This looks good :+1: 

To summarise an IRL conversation: Rather than removing `registry_by_fields`, we could remove `registries` and make `ResultSetPresenter` use that instead, which would be both simple to do as it has a lookup method for each registry, and would enable some neater refactoring as that class is effectively doing a field -> registry mapping in `#build_result`

One other minor point of feedback: you've dropped trailing commas in a few places when refactoring parameter lists. That's probably unintentional?

rboulton:
It's good to see this being tidied up.  I agree with @evilstreak's comment, but otherwise :+1:.

tijmenb:
Rebased & pushed with the amended commits. `ResultSetPresenter` now uses the plural keys from the hash.

Also added back some missing trailing commas. 

Thanks @evilstreak !

PR #413, opened 2015-05-06
Add vehicle_recalls_and_faults_alert document schema

tadast:
To be published from specialist publisher https://github.com/alphagov/specialist-publisher/pull/480

### DIFF
--- a/config/schema/document_types/vehicle_recalls_and_faults_alert.json
+++ b/config/schema/document_types/vehicle_recalls_and_faults_alert.json
@@ -1,0 +1,111 @@
+{
+  "fields": [
+    "fault_type",
+    "alert_issue_date",
+    "faulty_item_type",
+    "manufacturer",
+    "faulty_item_model",
+    "serial_number",
+    "build_start_date",
+    "build_end_date"
+  ],
+
+  "allowed_values": {
+    "fault_type": [
+      {
+        "label": "Recall",
+        "value": "recall"
+      },
+      {
+        "label": "Non-urgent fault",
+        "value": "non_urgent_fault"
+      }
+    ],
+
+    "faulty_item_type": [
+      {
+        "label": "Vehicle",
+        "value": "vehicle"
+      },
+      {
+        "label": "Baby seat",
+        "value": "baby-seat"
+      },
+      {
+        "label": "Tyres",
+        "value": "tyres"
+      },
+      {
+        "label": "Parts",
+        "value": "parts"
+      },
+      {
+        "label": "Agricultural equipment",
+        "value": "agricultural-equipment"
+      },
+      {
+        "label": "Other accessories",
+        "value": "other-accessories"
+      }
+    ],
+
+    "manufacturer": [
+      {
+        "label": "Alfa Romeo",
+        "value": "alfa-romeo"
+      },
+      {
+        "label": "Audi",
+        "value": "audi"
+      },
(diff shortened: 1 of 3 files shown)


config/schema/field_definitions.json
```diff
@@ -345,6 +345,38 @@
     "type": "float"
   },
 
+  "fault_type": {
+    "type": "identifiers"
+  },
+
+  "item_type": {
```
evilstreak: `item_type` and `item_model` are pretty broad terms. Is it worth finding a more specific term? /cc @rboulton

rboulton: Thanks for spotting, @evilstreak, yes; something a bit more specific to the context of vehicles would reduce the probability of confusion or conflict in future.

config/schema/document_types/vehicle_recalls_and_faults_alert.json
```diff
@@ -0,0 +1,111 @@
...
+      {
+        "label": "Parts",
+        "value": "parts"
+      },
+      {
+        "label": "Agricultural equipment",
+        "value": "agricultural-equipment"
+      },
+      {
+        "label": "Other accessories",
```
rboulton, line 47: Having "other" options makes me wonder if this would be better without "allowed values", allowing any value to be added.  The list feels very non-comprehensive and likely to results in support requests for adding more allowed values in future.

Is this actually only using allowed values because finders require fields with filters to use it?  If so, we should fix that, rather than artificially restrict the list.

The same applies to the list of "manufacturer"s, I think.

rboulton:
This looks plausible, but I'd definitely like the field names discussed above to be altered.  The "allowed_values" comment I leave to your judgement.

config/schema/field_definitions.json
tadast: Thanks for pointing this out. It is not always a vehicle. I could rename to `recalled_item_*`, `faulty_item_*`, any other suggestions?

config/schema/document_types/vehicle_recalls_and_faults_alert.json
tadast, line 47: For item types the list should serve the purpose well and not change often if at all.

However, we will definitely need to add more manufacturers to the list (you can see on http://www.dft.gov.uk/vosa/apps/recalls/ there are more manufacturers). We could allow any value for the manufacturers field, but we still need:

1) to have a dropdown with predefined values in specialist-publisher. This will ensure there is no data duplication, typos etc.
2) to have a facet for filtering and subscribing to a specific manufacturer email alerts. 

I don't know where the finder frontend fetches facets from, is it rummager? In any case we would like to test-drive this asap and once we're convinced finders is the right solution to this problem, we can solve it long-term.

I also don't know if changing the schema in rummager is difficult. If so, let me know.

rboulton, line 47: Changing the schema to add allowed values isn't very difficult, but does require a deploy so is a bit fiddly.

Based on your comments, this will be fine for now.

config/schema/field_definitions.json
rboulton: I think `faulty_item_*` sounds ideal.

tadast:
Renamed, rebased and [force-pushed](http://tclhost.com/GBwrwr2.gif)

PR #416, opened 2015-05-07
Refactor unified searcher tests

tijmenb:
While refactoring `unified_searcher_test.rb`, it became increasingly apparent that almost all tests there duplicate other tests.

The first commits in this PR strip away tests and code from `unified_searcher_test.rb`, the last commit (e8f916e) finally replaces the tests with a simpler version. 

UnifiedSearcher is a very simple class. It receives an arguments object, sends that to a query builder, executes the query in elasticsearch and returns the results in a presenter. A test for this class should test that the right methods are being called with the correct arguments.

With these changes, code coverage (as reported by simplecov) increases slightly from 96.63% to 96.64%.

### DIFF
--- a/lib/facet_example_fetcher.rb
+++ b/lib/facet_example_fetcher.rb
@@ -5,7 +5,6 @@ class FacetExampleFetcher
     @response_facets = es_response["facets"]
     @params = params
     @search_builder = search_builder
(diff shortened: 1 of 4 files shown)


KushalP:
:+1: :heart_eyes:

PR #417, opened 2015-05-07
Upgrade Webmock gem

tijmenb:
Specifically for [WebMock.disable_net_connect! now accepts array of regexps as allow param](https://github.com/bblimke/webmock/blob/master/CHANGELOG.md#1120).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,7 @@ group :test do
   gem "minitest", "4.6.1"
   gem "rack-test"
   gem "mocha", :require => false
-  gem "webmock", "1.9.3", :require => false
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -29,7 +29,7 @@ group :test do
   gem "minitest", "4.6.1"
   gem "rack-test"
   gem "mocha", :require => false
-  gem "webmock", "1.9.3", :require => false
+  gem "webmock", "~> 1.21.0"
```
KushalP: What was the reason for removing `:require => false` here?

tijmenb: Woopsie. Amended!

PR #418, opened 2015-05-07
Add elasticsearch spelling suggester

tijmenb:
Adds spell checking using elasticsearch's [Phrase Suggester](http://www.elastic.co/guide/en/elasticsearch/reference/1.4/search-suggesters-phrase.html) feature. Suggestions made by elasticsearch are more accurate, mainly because the suggestions are drawn from the content on GOV.UK itself.
### Before

before
### After

after
### Notes
- When user searches we now do an extra request to elasticsearch to fetch suggestions for the search term. This has to be done in a separate request because of limitations in working with multiple indices (more in `spell_check_fetcher.rb`).
- The "suggest" part of the elasticsearch query is built in `QueryComponent::Suggest`. The pattern of using small builder classes to build the query should be used to refactor `UnifiedSearchBuilder`.
- `SuggestionBlacklist` makes sure we don't suggest corrections for search terms we deem correct.

Trello: https://trello.com/c/GibHi6Aa/68-term-suggester-spelling-corrections-using-elasticsearch
### Commits

The first commit removes everything related to the old spell checker. The second implements the new spell checker.


rboulton:
This is great work.  My only merge-blocking concern is the change to the integration test helper which removes the restriction to only access "_test" indexes.

tijmenb:
Great, thanks. 

Fixed the integration helper & indentation.

PR #421, opened 2015-05-13
Refactor UnifiedSearchBuilder

tijmenb:
`UnifiedSearchBuilder` is responsible for taking a set of parameters (like search string, filters and facets) and producing a [search request body](http://www.elastic.co/guide/en/elasticsearch/reference/1.4/search-request-body.html) that is sent to elasticsearch.

This PR refactors `UnifiedSearchBuilder` to improve readability and maintainability. This is done by splitting it up into `QueryComponents` that are responsible for a specific piece of the search query.

There are two types of components:
- Components that take the `params` hash and return a `payload` for use in the query. For example, the `QueryComponents::Sort` returns the correct `sort` clause for a set of params.
- Components that `#wrap` a query to "enhance" it. For example, `QueryComponents::Booster` takes an existing query and returns a query that uses `boosts`to promote certain types of content.

The tests for UnifiedSearchBuilder have also been split up and refactored.

TODO: rethink the naming of `search_term`, `UnifiedSearchBuilder`, `QueryComponents`.

Didn't rebase to make reviewing easier. Will probably squash all commits into one to keep a clean history.


rboulton:
I'm basically very happy with this.  I think "search_term" might be better named as "query_string", but I'm happy enough with "search_term" if you don't want to go through and change it.

I think UnifiedSearchBuilder is a good name for now (until we remove all other kinds of search builder, and then it should be renamed SearchBuilder).

I'm pretty happy with the name QueryComponents.

If you'd like to rebase and get this ready for merge, I'm happy to do that.

tijmenb:
Thanks! Squashed with appropriate commit message.

PR #424, opened 2015-05-14
Expand 'people' fields into person objects

dsingleton:
Results with a 'people' field were returning an array of string
slugs, compared to fields like 'organisation' where they are
expanded into full objects, with the title/link/etc of the
organisation included.

This adds the appropriate code to expand people fields, so that
the full object representation can be used to display details
about people, eg, in metadata or filter facets in a Finder.

I'm not 100% confident on the `person_registry_index` change,
it's based on what other registries are doing. I think I understand
it, but i'd appreciate someone giving that an extra eye.

### DIFF
--- a/elasticsearch.yml
+++ b/elasticsearch.yml
@@ -6,6 +6,7 @@ topic_registry_index: "government"
 document_series_registry_index: "government"
 document_collection_registry_index: "government"
 world_location_registry_index: "government"
+people_registry_index: "government"
 # These indices are passed in this order to GovukSearcher
 govuk_index_names: ["mainstream", "detailed", "government", "service-manual"]
 metasearch_index_name: "metasearch"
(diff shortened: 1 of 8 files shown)


rboulton:
Will review fully later, but I'd like you to be consistent (regardless of gramattical correctness) about using the name of the field here.  ie, always use the word "people" rather than person.  This will help searchability of the code base in future, and help us refactoring this to be generic across fields.

dsingleton:
Agreed, it's a bit weird and a hard to follow. 

I did it to follow the existing convention of pluralisation for other expanded fields, eg; `specialist_sector_registry` vs `specialist_sectors` vs `Registry::SpecialistSector` vs `result['specialist_sectors']`\- though `people` vs `person` is more awkward :(

Happy to update to consistent term, which would you favour? `people` to match the field mapping in `ResultSetPresenter`?

rboulton:
I'd prefer the other registries to be more systematically named, too :)

I'd go for `people` everywhere to match the field name.

dsingleton:
Updated to `people` everywhere.

PR #425, opened 2015-05-14
Add title to sort fields

tommyp:
For the Policy finder (https://www.gov.uk/government/policies) sorting
by `public_timestamp` is meaningless as content changes don't bubble up
to the Policy. Until we can do that, it makes more sense to sort the
Policies by title. This commit adds title to the `ALLOWED_SORT_FIELDS`.

### DIFF
--- a/lib/search_parameter_parser.rb
+++ b/lib/search_parameter_parser.rb
@@ -11,6 +11,7 @@ class BaseParameterParser
   ALLOWED_SORT_FIELDS = %w(
     last_update
     public_timestamp
+    title
   )
 
   # The fields listed here are the only ones which can be used to filter by.


rboulton:
I have some performance worries about this; sorting by an analyzed text field which isn't stored is going to require fetching and parsing the full stored "source" of every matching document.  We can configure elasticsearch to index the field better for allowing sorting.  It would be good to do some timings of a search for all documents with this sort enabled, compared to the default sort order; and I'd like to look into updating the title field's configuration before this is merged.

tommyp:
There are ~230 Policies and this isn't likely to grow anytime soon. I can see a case where certain Finders other than Policies would need this though (Contacts is another example) so we should definitely get this sorted.

rboulton:
Merging this for now, but please keep an eye on performance in preview, and we can tweak elasticsearch configuration if it turns out to be a problem.

PR #426, opened 2015-05-15
Make sorting by title work.

rboulton:
This is implemented by also indexing "title" to a non-analyzed sub-field
called "title.sort", and making a sort on title use this.

(Depends on elasticsearch 1.0+'s "fields" feature.)

This probably actually wants to have the title lowercased, etc, so that sorting is case insensitive.

The index will need to be migrated first, for this to work.

FAO: @tommyp

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -4,7 +4,7 @@
   },
 
   "title": {
-    "type": "searchable_text"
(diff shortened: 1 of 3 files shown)


tommyp:
Does this mean then that we need to order by `title.sort` or will ordering by `title` be enough?

PR #431, opened 2015-05-19
Remove use of mappings in parameter parser

rboulton:
This is a code cleanup, but also allows slightly more flexibility in the API.

Specifically, this PR removes the `ALLOWED_FILTER_FIELDS` constant in SearchParameterParser, in favour of defining in the schema which fields can be filtered on.  To do this, a new `filter_type` value is supported in the `field_types.json` file.  This may be either set to "text" or "date", and if not present means the field may not be filtered on.

The extra API flexibility is that it is no longer required to filter on document type to make it possible to filter on fields which are defined only for a particular document type.  Performing such filtering will implicitly filter to the document type, anyway (since the field won't be present for any other documents).

This is best reviewed commit by commit.

It would be good to remove the other constants defined in SearchParameterParser in future PRs, but I'd like to merge this one now since it makes a large improvement (by removing the hacks which allowed filtering extra fields when a document_type filter was used), and is self-contained.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -280,7 +280,7 @@ class Rummager < Sinatra::Application
 
     parser = SearchParameterParser.new(
       parse_query_string(request.query_string),
-      current_index.mappings,
+      unified_index_schema,
     )
 
     unless parser.valid?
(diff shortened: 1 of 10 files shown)


KushalP:
I really like the structure and ordering of your commits. It's easy to understand your thinking and the commits make it explicit why the refactoring (simplification) happens.

Apart from the small `assert_equal` ordering nit-pick, I think this is great. Good work :+1:

PR #441, opened 2015-05-26
New ranking system

rboulton:
This pull request reviews the ranking process we're using, and rebuilds it roughly from scratch.  The idea is to build our best guess at how the ranking ought to work, including which features it should use, compare how this performs to what we're currently using, and then switch over.

Nothing in this pull request should change the ordering of search results on the site, unless a new debug parameter is supplied.  If `&debug=new_weighting` is added to a search query string, the new weighting will be used.  The idea is that this PR should be fine to merge without affecting ranking, so that we can review it (and refine the synonym lists for it) over at least a few days without blocking deploys to production.

The individual commits should be reviewed in order, but the changes they make are intended to do the following:
- Change the way in which we implement synonyms, to make synonym groups be replaced at index time with special "marker" terms, and to look for these terms at search time; rather than trying to do the full synonym matching at search time.  This should give better ranking (by treating a synonym as a single term, so its frequency makes sense), and also interact properly with the `minimum_should_match` setting.  The synonym list used by the new ranking is separated from that used by the old ranking, so that the new list can be reworked independently.
- Review the way in which the weighted text query is built, to use per-field boosts (and use different boosts for different fields, such that, for example, matches in title fields are considered more important than `indexable_text` fields).
- Stop using stopwords, since they don't really help us at all.
- Increase the `minimum_should_match` settings, to require more terms to be matched.  It was artificially low to compensate for problems with synonyms and stopwords.
- Use "shingles" (word pairs) to boost the score of partial phrase searches.


KushalP:
First off, this is a really nicely structured PR. I especially appreciated the comments on the code in 0ec9322, 0101af6, and 0d38986. They'll be useful in the future.

In 0101af6 you pointed out that the indexes must be migrated when this PR is deployed, to ensure
that the new `query_with_old_synonyms` analyzer exists. Is this using the existing `rummager:migrate_index` task or something else?

KushalP:
Once the question I have above about how this is deployed is answered, I'm happy for this PR to be merged :+1:

rboulton:
Yes, this just needs to be deployed using "deploy:migrations", which will run the migrate index. (and all preview deploys do this, so it will happen in preview automatically)

tarastockford:
I'm not sure about requiring more terms to match. Although some users expect that adding more words should narrow down their results, other users are already missing what would have been relevant results because their query contains 'extra' words that don't match the content.

Since this will only be on a debug parameter, it's fine for now and we can review examples of longer queries to see whether some relevant content is no longer returned.

tijmenb:
:rocket: awesome PR, should improve the search a lot!

rboulton:
@tijmenb I've rebased to address your comments - which was probably foolish as now it's hard to see the changes I made to do this.  I've pushed the old version to the "old-better-synonym-matching", but can't manage to construct a correct comparison URL in github.  If you do git fetch, you should be able to see the diff by running `git diff old-better-synonym-matching..better-synonym-matching`

PR #443, opened 2015-05-31
Search specially for "ID codes"

rboulton:
# Problem

Often, people will search for ID codes (eg, form or report names, etc)
on the site.  A particular problem with such searches is that these
codes are often split into pieces, but are inconsistently written.

Some examples:
- `P60` vs `P 60`
- MAA Regulatory articles: often written as "(RA) 1000", but searched
  for as "RA1000".
- Visa tiers: often written "tier 1", but searched for as "tier1".

Currently, we have large lists of these manually added to the synonyms
file to handle the most important of these cases, but there are always
going to be cases that we've missed.
# Approach

To stop needing to maintain these lists, this commit adds a special
`with_id_codes` analyzer to pick out codes like these, and uses this to
index them to a special ".id_codes" sub-field.  This is applied to all
searchable text fields.  A match in this field of a single code is
sufficient to make a document be returned from a search (but usually not
with a very high weight).  Per-field boosting is also applied (so a
match in a title field is much more significant than a match in the body
text).

The change is only made when the `new_weighting` debug flag is enabled.

The analyzer is complex, but entirely built of standard elasticsearch
components.  The codes we're looking for have the following
characteristics:
- They contain at least one digit, and are at least 2 characters long.
- They might be split into two words, or separated by characters such
  as `.`, `:`, `/`.
- They're not split across sentences, or quotes, etc.

The data is first split into sentences using a `pattern` tokenizer, and
then progressively broken down and filtered by a sequence of
`pattern_capture` and `pattern_replace` token filters.

The analyzer is a bit over-eager; it will often return matches for
things which aren't really codes.  Eg: "March 2014" will produce the
terms "march2014" and "2014".  However, these matches are pretty
harmless; if a user has entered things like this in their query and they
get matches back the matches will usually be reasonably useful.

Having made these changes, a second commit removes lots of
synonyms (eg "RA 1002, RA1002") which should no longer be
needed.


tijmenb:
Looks good! We might want to split up `elasticsearch_schema.yml` at one point, it growing rather large.

alext:
Is it safe to remove the synonyms while the new weighting is still feature-flagged?

rboulton:
@alext Yes - the non-feature-flagged synonyms are now in "old_synonyms.yml"

rboulton:
@tijmenb Yes - I'd also be interested in making it non-YAML too.  I think we can live with it like that for a while more, though.

PR #446, opened 2015-06-05
Allow sorting of facets by value

tommyp:
This commit modifies the `SearchParameterParser` and the `FacetOption`
to allow facet options to be sorted by the value returned if the value
is just a string. If the value is an object it will be fall back to
sorting it by title.

### DIFF
--- a/test/unit/facet_option_test.rb
+++ b/test/unit/facet_option_test.rb
@@ -105,6 +105,22 @@ class FacetOptionTest  < MiniTest::Unit::TestCase
     )
   end
 
+  def test_compare_by_value
+    orderings = [[:value, 1]]
+    assert(
+      FacetOption.new("a", 0, false, orderings) <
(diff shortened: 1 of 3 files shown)


test/unit/facet_option_test.rb
```diff
@@ -105,6 +105,14 @@ def test_compare_by_link_descending
...
   end
 
+  def test_compare_by_value
+    orderings = [[:value, 1]]
+    assert(
+      FacetOption.new("a", 0, false, orderings) <
+      FacetOption.new("b", 0, false, orderings)
+    )
+  end
+
```
boffbowsh, line 115: Maybe add a test for `{"title" => "a"}` etc too?

rboulton:
I think that suggested extra test would be a good idea, but other than that, :+1:

tommyp:
Added another test for sorting by value when the `FacetOption` has a hash with `title` as a key.

PR #448, opened 2015-06-09
Add small regression test for stemming

tijmenb:
Since PR #447 we're using the `porter2` stemming algorithm. This adds a small regression test for that.


rboulton:
Top work filling in for my omission there. :shipit:

PR #449, opened 2015-06-09
Add detailed_format to edition

tommyp:
As we need to get back a slug version of `display_type` I added a new
field to edition and gave it all the possible values of display type to
the `allowed_values` field in the edition doctype.

The need for this comes from the fact that `display_type` is sent to
rummager as a title case string `News story` instead of `news-story`.

Ticket:
https://trello.com/c/TAuAS4wp/174-8-add-more-filter-options-to-policy-pages

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -98,6 +98,11 @@
     }
   },
 
+  "detailed_format": {
+    "description": "A slugified version of the display_type field",
+    "type": "identifier"
+  },
+
   "display_type": {
     "type": "identifier"
   },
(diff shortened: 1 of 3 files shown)


config/schema/field_definitions.json
```diff
@@ -98,6 +98,10 @@
     }
   },
 
+  "detailed_format": {
+    "type": "identifier"
```
rboulton: Could you add a "description" entry in this field definition, please?  They're optional, and for documentation only, but it's very helpful to have one indicating what sort of data is expected in the field, and where it comes from.  I'm trying to make sure any new fields have one, and then add it for existing fields.

PR #451, opened 2015-06-10
Remove unused code

rboulton:
In app.rb, govuk_indices and the `before` helper for the old search
endpoints haven't been used since those endpoints were removed.

The old "combined" endpoint was also the only thing that combined
ResultSets together, so when this was removed methods in ResultSet for
merging or subtracting result sets became unused.  In turn, these were
the only thing that called `Document.weighted`.


tijmenb:
:tada:

PR #452, opened 2015-06-11
Fix lookup of people for expansion

rboulton:
The format of a document about a person is `person`, not `people`.
Change the lookup performed by the Person registry to look for documents
with this format, so that expansion works.

### DIFF
--- a/lib/registry.rb
+++ b/lib/registry.rb
@@ -122,7 +122,7 @@ module Registry
 
   class Person < BaseRegistry
     def initialize(index, field_definitions)
-      super(index, field_definitions, "people")
+      super(index, field_definitions, "person")
     end
   end
 end


alext:
Would it be possible to write a test to expose this bug?

rboulton:
It would be possible, but I think it would require fixture data in an integration test to do it convincingly (ie, I could write a unit or integration test, but they'd just be using stubbed data expressing the same change).  I don't want to add fixtures like that in haste - I think the appropriate thing to do is to add good fixture data (or better, generated example equivalents) when we split the search and indexing side of this app apart.

If anyone has a good idea for a test of this, though, I'd love to hear it. :)

dsingleton:
Urgh, sorry :(

PR #453, opened 2015-06-11
Parallelise bulk loading of a new index.

alext:
This may be easier to review commit-by-commit.

We can get much more performance while bulk loading a new index (either
migrating an indes, or loading an index from an external dump), by
processing batches of records in parallel.  This is because it allows
elastisearch to use all the resources available to it.  See 
https://www.elastic.co/blog/performance-considerations-elasticsearch-indexing
for more details.

I've picked a default concurrency level of 12 to match our
infrastructure.  Elasticsearch defaults to using an indexing thread pool
size that matches the number of CPU cores available.  Our
api-elasticsearch machines have 4 cores, and we run a 3-node cluster,
hence a value of 12 makes sense. This may need to be tuned after
benchmarking etc.

This change halved the indexing time on my dev VM.

### DIFF
--- a/bin/bulk_load
+++ b/bin/bulk_load
@@ -26,7 +26,7 @@ load the alias is switched over.
 }
   run do |opts, args|
     if args.size == 1
-      BulkLoader.new(search_config, args.first, logger: Logger.new(STDERR)).load_from($stdin)
+      BulkLoader.new(search_config, args.first).load_from($stdin)
     else
       puts self
     end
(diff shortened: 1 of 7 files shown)


rboulton:
Looks pretty great Alex.  The logical separation into commits was very helpful with reviewing.

Made a minor comment on 0bbf43d, but other than that this looks good to go.

alext:
@rboulton Updated and rebased.

PR #454, opened 2015-06-12
Add important to editions

tommyp:
This commit adds an important flag to the edition doctype. It will be
used by the Whitehall editor to set documents as being important.

### DIFF
--- a/config/schema/document_types/edition.json
+++ b/config/schema/document_types/edition.json
@@ -31,7 +31,8 @@
     "has_act_paper",
     "is_political",
     "is_historic",
-    "government_name"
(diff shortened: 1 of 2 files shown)


rboulton:
I thought the plan was to use a less-generic name than `important` - and to only set this for policy content.

tommyp:
Oh yeah. I totally forgot we had that chat. What should I name this then? Didn't we agree on `important_to_policy`?

rboulton:
That seems a good name, to me.

tommyp:
@rboulton Updated it to use `important_to_policy`.

PR #462, opened 2015-06-29
Fix expansion of specialist_sectors.

alext:
They've been moved to the /topic namespace, so the code that inferred
the slug from the link field was no longer working. Instead we now send
the slug to rummager (alphagov/collections-publisher#112), so this
inferrence is no longer necessary.


rboulton:
Looks good to me, pending tests passing.

Issue #162, opened 2013-09-03
Naming of exact_field_boosts and exact_match_boost


matthewford:
Can we rename these fields, on search_query_builder as they dont seem to actually do any boosting?

`exact_field_boosts`

`exact_match_boost`

PR #463, opened 2015-07-02
Bump GDS API adapters to 20.1.1

elliotcm:
This changes the user agent to include the app
name in API requests.

https://trello.com/c/qkHdtob4/247-bump-gds-api-adapters-everywhere-it-s-used-to-include-user-agent-information

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gem "raindrops", "0.11.0"
 gem "sinatra", "1.3.4"
 gem "rake", "0.9.2", :require => false
 gem "rack", "~> 1.6"
-gem "rest-client", "1.6.7"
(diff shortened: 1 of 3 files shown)


alext:
In light of the changes to the router-api client (#162), we should probably remove the `:skip_commit => true` param from the router calls in `lib/tasks/router.rake`.

elliotcm:
@alext Is that the PR you meant to reference?

alext:
Doh, no, that's the PR number on gds-api-adapters.  I meant https://github.com/alphagov/gds-api-adapters/pull/162

elliotcm:
@alext thanks for the note, updated.

PR #471, opened 2015-07-23
Hide mainstream browse pages

tijmenb:
Mainstream browse pages are currently not indexed, but soon will be to allow rummager to deliver content.

We're still unsure if we want to display mainstream browse pages in the search results, and how. By boosting them with zero, they can be hidden from the results.
## Before

screen shot 2015-07-23 at 17 18 59
## After

screen shot 2015-07-23 at 17 19 14

Trello: https://trello.com/c/6vXKBd5d

PR to index browse pages: https://github.com/alphagov/collections-publisher/pull/127

### DIFF
--- a/lib/query_components/booster.rb
+++ b/lib/query_components/booster.rb
@@ -36,6 +36,9 @@ module QueryComponents
         "document_collection" => 1.3,
         "operational_field" => 1.5,
         "contact"           => 0.3,
+
+        # Hide mainstream browse pages for now.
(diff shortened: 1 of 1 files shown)


rboulton:
I've not tested this, but I'd expect that code to just move these to the end of the list, rather than remove the results from search.  If a mainstream browse page is the only match, they'll still come back.  This is probably fine for now, though.

PR #472, opened 2015-07-23
Remove search from command line

tijmenb:
There's currently a `bin/search` executable in this project that could be used to get data on search results (introduced here: https://github.com/alphagov/rummager/commit/9ab1de77aeeb9655c1b05898ef066b431c73a11b).

Originally it used pretty much the same code as normal search, but it has diverged significantly in the last few years, accumulating code only used for this feature. This also means it doesn't actually return the same search results as the website does, making it less useful. 

I'm pretty sure we can delete all this, as I'm assuming nobody has used this feature in a while.

@alext


alext:
This looks good to me, however I don't have enough context around this feature and what it was used for to be confident merging this.

tijmenb:
Closing this until @rboulton is back from holiday and can take a look at it. I've [a reminder story in the prep Trello](https://trello.com/b/459ffE04/icebox-and-story-preparation).

rboulton:
We should totally remove this.

tijmenb:
:+1:

PR #474, opened 2015-08-03
More downweighting for closed organisations

tijmenb:
Closed organisations can currently rank higher than normal organisations. By decreasing the boost ever further we prevent that.

Example: search for "uk border" should not bring up the closed UK Border Agency as high as it does now.

https://www.gov.uk/search?q=uk+border 
## Before

screen shot 2015-08-03 at 10 56 12
## After

screen shot 2015-08-03 at 10 56 24

(Descriptions removed to fit it all on one page).


jackscotti:
PM is happy with this. Merging.

PR #475, opened 2015-08-05
Add /content endpoints

tijmenb:
This PR adds two new API endpoints.

`GET /content?link=/vehicle-tax`

Will return the search result from our content indexes with the specified link.

`DELETE /content?link=/vehicle-tax`

Will delete the search result from the index.

These endpoints are meant to be used by [search-admin](https://github.com/alphagov/search-admin) first, to: 
1. Display the data for a link in elasticsearch, usefull to debug what is in `indexable_content` 
2. Manually delete search results from the indices. Removal of search results is currently done manually, which is taking up developer time.

Notes:
- There is already a `get "/?:index?/documents/*"` route to fetch the document from elasticsearch. However, this endpoint does not support multiple indexes. The difference between our four "content" indexes (mainstream, government, detailed, service-manual) is no longer relevant and users may not even know in which index the document lives. The new `GET` route works on all content indexes.
- `get "/?:index?/documents/*"` doesn't work for some document formats because of the way it calculates the document type when searching (see [this](https://github.com/alphagov/rummager/blob/9b2a1e1748808bdedbc7cbf974c32c1159582e1a/lib/elasticsearch/index.rb#L334) for details). The new endpoint works for all documents that can turn up in the results.
- `delete "/?:index?/documents/*"` needs the `type` and `index` of the result to delete. This is difficult to find, especially since `get "/?:index?/documents/*"` doesn't work. Our new endpoint doesn't need the type.
- A third endpoint `delete "/?:index?/documents"` can also be deprecated.

Notes about implementation:
- We chose to add a new file in `lib/routes` for these endpoints, to keep `app.rb` from growing even bigger.
- The URL would be nicer in the form of `GET /content/:id`. But this would be confusing because of the double slash the URL would contain (a problem with the endpoints mentioned above).
- Added documentation to `/docs/api.md`.
## Trello

https://trello.com/c/dvzAnQ80/272-remove-passport-extension-page-from-search
https://trello.com/c/0Jybv3yh/276-remove-item-from-search-index

Paired with @jackscotti on this one :+1:

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -22,6 +22,8 @@ require "schema/combined_index_schema"
 require_relative "config"
 require_relative "helpers"
 
+require "routes/content"
+
 class Rummager < Sinatra::Application
   def search_server
     settings.search_config.search_server
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,0 +1,55 @@
+## Rummager API
+
+### `GET /content/?link=/a-link`
+
+Returns information about the search result with the specified link.
+
+### Example reponse
+
+```
+curl -XGET http://rummager.dev.gov.uk/content?link=/vehicle-tax
+```
+
+Currently returns a hash with one element: `raw_source`, which contains the raw elasticsearch document (`_source`).
+
+```json
+{  
+   "raw_source":{  
+      "section":"driving",
+      "subsection":"car-tax-discs",
+      "organisations":[  
+         "department-for-transport",
+         "driver-and-vehicle-licensing-agency"
+      ],
+      "popularity":0.08333333333333333,
+      "public_timestamp":"2014-12-09T16:21:03+00:00",
+      "format":"transaction",
+      "title":"Renew vehicle tax",
+      "description":"Renew your vehicle tax, apply online, by phone or at the Post Office",
+      "link":"/vehicle-tax",
+      "indexable_content":"[..snip..]",
+      "mainstream_browse_pages":[  
+         "driving/car-tax-discs"
+      ],
+      "tags":[  
+         "organisation:department-for-transport",
+         "organisation:driver-and-vehicle-licensing-agency"
+      ],
+      "_type":"edition",
+      "_id":"/vehicle-tax"
+   }
+}
+```
+
+### `DELETE /content/?link=/a-link`
+
+Deletes the search result with the specified link.
+
+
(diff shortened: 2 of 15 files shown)


KushalP:
I've only read the description so far, but I'm wondering if using `/content?link=/vehicle-tax` will conflict with the [similar usage](https://github.com/alphagov/content-store#reading-content-from-the-content-store) that exists in Content Store. We should make the two of these consistent or choose a path other than `/content`.

bradwright:
/content is the same as the content store public URI. Does that matter?

tijmenb:
Won't conflict, but still a good point to avoid confusion.

I meant for the `/content` namespace only to operate on rummager-documents that can turn up in search (mainstream, government, detailed, service-manual indexes, not page-traffic and metasearch). How about `/search-results?link=/vehicle-tax`?

tijmenb:
Closed until we have more clarity about when indexes are locked.

rboulton:
I've not looked at the code, but I like the addition of these endpoints - they make semantic sense.  I'd suggest using `/document` instead of `/content`, since that's more the terminology used in elasticsearch and other search systems, but I don't think `/content` is actually problematic.

Though the `?link=` form of the URL is a bit ugly, it has the advantages mentioned in the PR description, and would also support changing the unique identifier in search without breaking things (eg, we could support `?id=` if identifiers became distinct from the link field, such as if we started using content-IDs as identifiers).

I'm not sure what the issue about indexes being locked is.  If you describe in more detail I may be able to help.  Indexes in elasticsearch should only be locked in two cases:
- when a migration is in progress.
- when a bulk import is in progress.

tijmenb:
Slack discussion w/ richard: index locks are temporary. I'll amend the commit to handle locking errors gracefully.

As for the naming, I would prefer it to have `document` refer to _any_ elasticsearch document. A thing from `/content` or `/search-results` then refers only to things that can be returned as a search result, and not metadata that is also kept in ES, such as the popularity data and best bets. A possible future endpoint `/document` would also return those. Does that make sense?

rboulton:
The logic for calling it `/content` makes sense, particularly since we have a "content-index-names" config variable to refer to those indexes.

tijmenb:
Updated PR with the lock handling.

PR #476, opened 2015-08-06
Add Upper Tribunal Immigration and Asylum Chamber decisions

robmckinnon:
- Add `utiac_decision` document type and field definitions for Upper Tribunal Immigration and Asylum Chamber decisions.
- Add new `searchable_identifiers` field type that allows identifiers that occur multiple times in a single document to be considered in non-field-specific text searches.

### DIFF
--- a/config/schema/document_types/utiac_decision.json
+++ b/config/schema/document_types/utiac_decision.json
@@ -1,0 +1,9 @@
+{
+  "fields": [
+    "country",
+    "country_guidance",
+    "decision_reported",
+    "judges",
+    "promulgation_date"
+  ]
+}
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -399,5 +399,21 @@
   "important_to_policy": {
     "description": "A flag set by editors to mark a document as being important to policy",
     "type": "boolean"
+  },
+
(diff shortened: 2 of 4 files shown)


config/schema/document_types/utiac_decision.json
```diff
@@ -0,0 +1,9 @@
+{
+  "fields": [
+    "country",
+    "country_guidance",
+    "decision_reported",
+    "judges",
+    "promulgation_date"
+  ]
+}
```
tommyp, line 9: Missing newline.

config/schema/field_definitions.json
```diff
@@ -399,5 +399,21 @@
   "important_to_policy": {
     "description": "A flag set by editors to mark a document as being important to policy",
     "type": "boolean"
+  },
+
+  "country": {
+    "type": "identifier"
```
rboulton, line 405: Could you add a "description" field for each of these fields, indicating the type of content expected in it, and any special considerations.  This is our way of commenting the file, so that we have some idea what the existing fields are used for.

The comments should explain any special terminology used, too (eg, "promulgation").

robmckinnon:
Closing for now.

PR #477, opened 2015-08-10
Extract and update documentation of search API

rboulton:
There was a 122 line comment in app.rb describing the syntax of the
parameters expected by the unified search endpoint.  This was
- hard to find and read
- slightly out of date

This commit extracts it into a markdown-formatted file in the `docs/`
directory, and updates it to be up-to-date.


tijmenb:
This is nice! :+1:

PR #479, opened 2015-08-10
Temporarily fix leading slashes in links

tijmenb:
The link field should contain either a full URL or a relative URL with leading slash for the frontend to work. Manuals and specialist documents currently send links without leading slash to rummager,
causing problems in search and collections.

This temporarily adds the leading slash. This functionality should be removed once all documents are indexed correctly. For this reason the tests have been isolated in a separate file.

### DIFF
--- a/lib/result_presenter.rb
+++ b/lib/result_presenter.rb
@@ -21,6 +21,7 @@ class ResultPresenter
     result = expand_entities(result)
     result = add_calculated_fields(result)
     result = add_debug_values(result)
+    result = temporarily_fix_link_field(result)
     result
   end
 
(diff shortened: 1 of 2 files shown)


rboulton:
Are you _certain_ that only manuals and specialist documents don't send the leading slash?  Once this has been deployed and a migration run, it will be difficult to work out if other cases of this problem exist - and if we miss them, we'll start having duplicate documents in the search index (some with the slash, and some without).

Perhaps it's worth running a query to dump out all the link fields, together with the document format, to check that there are no more formats with this problem that have been missed?  Or have you checked this in some other way?

tijmenb:
No, not certain, but this is only changing the output of the document, not changing data in elasticsearch. It should be safe to merge.

rboulton:
Ah, sorry - I didn't realise this was in the result presenter.  I'd thought it was an attempt to change the value stored for the field in the index.

PR #480, opened 2015-08-10
Publish special routes via rake task.

elliotcm:
- /sitemap.xml
- /sitemaps

https://trello.com/c/blLdEZN5/292-make-apps-register-special-routes-on-deploy


rboulton:
This code seems fine.

I think it's a bit weird for rummager to be registering these routes, though, because rummager really isn't a publishing app. The `/search` routes aren't something that rummager cares about at all - it just provides a generic API for other apps, which use the search API in all sorts of ways.

Would it be better for frontend to be the publishing app for the /search routes? That's also a bit weird, but at least frontend is an app that has something to do with those routes. Rummager really doesn't have anything to do with how the search results are accessed and presented (and they're presented in lots of ways other than /search, such as the topic page latest feeds).

In future, I expect the functionality of serving search results to be moved out of frontend into its own app.  At that point, it would seem natural for that app to register the routes for its pages.  It's also possible in future that rummager will move into the "central' set of apps that serve content (ie, become more tightly integrated with the content-store, etc), and we wouldn't want special-case routes hard-coded into those apps.

Registering the /sitemap routes from rummager makes more sense, though, since rummager is actually responsible directly for producing the content which is served on those routes.

elliotcm:
@rboulton I've removed the search routes and made use of the gem publisher.  Could you recheck?

PR #484, opened 2015-08-11
Add script to cleanup old sitemaps

edds:
Currently we are keeping all sitemaps ever created. It would be good to
clean then up after a couple of days so that we don't have hundreds of
them saved on disk.

This lets you set a number of days to clean up and then deletes all
sitemaps which aren't the number of days most recent. If the sitemaps
are not created on a particular day this will still maintain at least
the right number of days worth of sitemaps not just the last number of
calendar days worth.

https://www.pivotaltracker.com/story/show/90087288

### DIFF
--- a/lib/elasticsearch/sitemap.rb
+++ b/lib/elasticsearch/sitemap.rb
@@ -41,6 +41,11 @@ class Sitemap
     index_full_path
   end
 
+  def cleanup
+    sitemap_cleanup = SitemapCleanup.new(@output_path)
+    sitemap_cleanup.delete_excess_sitemaps
+  end
+
 private
   def base_url
     Plek.current.website_root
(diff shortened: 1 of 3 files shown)


rboulton:
This code looks fine, and I like that it's well tested.

Did you consider instead using something external like logrotate, or just a shell script to delete the oldest files in the directory?  This is a fair chunk of code, that I'd rather avoid having to maintain if we could use something already written.

As an example of the kind of shell script I'm thinking of

```
ls -1tr *.xml | head -n -4 | xargs -d '\n' rm -f
```

will delete all but the newest 4 files matching `*.xml`.  I'm not sure if the sitemap files are organised into directories - if so, this command could be pointed at the directories instead of directly at the *.xml files

edds:
Due the the fact that we split the sitemap into chunks of 50,000 urls the number of files per day is not a constant, and could grow in the future (and has grown in the past). For example in the past there were 4 files (one index and 3 files full of URLs). Now that we have more than 150,000 urls there are 5 files (one index and 4 files full or urls). This will probably grown in the future to be more files once we have more than 200,000 unique pages on the site. So something naive like just deleting all but the _n_ most recent files won't scale with the site.

I have no experience using anything which would be able to do it based on creation dates of the files, something might exist but I just have no knowledge. This just seemed like an easy way of achieving it reliably.

rboulton:
Ah, so the files are all in a single directory, but there are varying numbers of them per day.  That does make things more awkward - not intractable, but the shell scripting required would get impenetrable quickly.

In the interests of this being robust, tested, and in a language that all our developers understand, this is probably the best approach then.  Thanks for the explanation.

bradwright:
The main reason we didn't want to use logrotate was that the files are produced by the app, so it's in a better position to know what's still useful to it or not. The best we could do with logrotate was probably delete any files older than 30 days.

PR #485, opened 2015-08-11
Use fully qualified time zones offset

edds:
Previously the version of ruby we were using in production didn't
support it. We now use Ruby 2.1.6 in prod so we can use fully qualified
time offsets.


tijmenb:
:+1: thanks!

PR #488, opened 2015-08-12
Clean up Registries code

tijmenb:
This PR makes the following changes:
- Specify one `registry_index`, not for each type (special cases like `specialist_sectors` can be handled in the `Registries` class.
- By using `Registry::BaseRegistry` directly we can save a lot of code and limit inheritance.
- Only allow hash-style access to registries to cut down on public methods on that class.
- Replaces the `govuk_index_names` setting with `content_index_names`, which is identical.
- Refactors `search_config` to be cleaner.

No change in functionality intended.

### DIFF
--- a/lib/search_config.rb
+++ b/lib/search_config.rb
@@ -4,6 +4,19 @@ require "schema_config"
 require "plek"
 
 class SearchConfig
+  %w[
+    registry_index
+    metasearch_index_name
+    popularity_rank_offset
+    auxiliary_index_names
+    content_index_names
+    spelling_index_names
+  ].each do |config_method|
+    define_method config_method do
+      elasticsearch.fetch(config_method)
+    end
+  end
+
   def search_server
     @server ||= Elasticsearch::SearchServer.new(
       elasticsearch["base_uri"],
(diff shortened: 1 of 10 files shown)


lib/search_config.rb
```diff
@@ -22,63 +35,16 @@ def index_names
     content_index_names + auxiliary_index_names
   end
 
-  def content_index_names
-    elasticsearch["content_index_names"] || []
-  end
-
-  def auxiliary_index_names
-    elasticsearch["auxiliary_index_names"] || []
```
jackscotti, line 30: By dinamically creating these methods, you lost the "`|| []`" would this affect something?

KushalP:
A nitpick: the bullet points you've provided with the commit are really clear. It would be really nice to have each bullet point exist as a separate commit. But it's a nitpick and not a blocker on this PR.

lib/search_config.rb
tijmenb, line 30: Should have mentioned that. There is no valid use case for `content_index_names` and `auxiliary_index_names` being nil/empty. They are either set by elasticsearch.yml or the test helpers.

tijmenb:
Fair point @KushalP. I worked on it as a whole and it was too hard to untangle the refactors into separate commits in this case, but some of them definitely could've been split out.

PR #489, opened 2015-08-12
Move schedule file to alphagov-deployment.

danielroseman:
This is to allow it to be different per environment.

Do not deploy until https://github.gds/gds/alphagov-deployment/pull/1028 is merged.


alext:
Instead of deleting the file, it's helpful to leave a comment in there indicating that the schedules are managed at deploy time.

danielroseman:
Done.

PR #490, opened 2015-08-13
Fix bug in facet example fetchers

tijmenb:
In some cases, elasticsearch's `fields` will be empty. That causes exceptions.

It seems very hard to reproduce this in the tests, so we had to extract a method and test it in isolation. This is not nice, but better than no tests.

cc @rboulton

### DIFF
--- a/test/unit/facet_example_fetcher_test.rb
+++ b/test/unit/facet_example_fetcher_test.rb
@@ -43,6 +43,26 @@ class FacetExampleFetcherTest < ShouldaUnitTestCase
     index
   end
 
+  context "#prepare_response" do
+    should "map an empty response" do
+      fetcher = FacetExampleFetcher.new(@index, {}, {}, @builder)
+
+      response = fetcher.send(:prepare_response, [], [])
+
+      assert_equal response, {}
+    end
+
+    should "map a response to facets without fields" do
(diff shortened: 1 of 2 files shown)


test/unit/facet_example_fetcher_test.rb
```diff
@@ -43,6 +43,26 @@ def stub_index(name)
...
+
+      response = fetcher.send(:prepare_response, [], [])
+
+      assert_equal response, {}
+    end
+
+    should "map a response to facets without fields" do
+      fetcher = FacetExampleFetcher.new(@index, {}, {}, @builder)
+      slugs = ['yolo']
+      response_list = [{ 'hits' => { 'total' => 1, 'hits' => [ { 'slug' => 'yolo' }]}}]
```
rboulton: Do the responses from elasticsearch actually look like this?  Specifically, I wouldn't expect the 'slug' key to exist.  I'd expect it to look more like:

`[{ 'hits' => { 'total' => 1, 'hits' => [ { '_id' => 'yolo' }]}}]`

rboulton: This doesn't actually affect the test, I think, but if I'm right about the response format it's a bit misleading.

tijmenb: Good point - updated.

PR #491, opened 2015-08-13
Encapsulate search parameters in value object

tijmenb:
`SearchParameters` is a value object tasked with holding the search parameters produced by the `SearchParameterParser`.

Currently the parsed parameters are passed down through the system as a hash. This causes defensive coding lower down in the system, where we have to check if some value such as `params[:debug]` is nil or an empty hash. It also causes the same variables to have many different names, `applied_filters` is an alias for `params[:filters]` sometimes.

This can be done cleaner by having methods on the object that will never return nil when a array or hash is expected.

The upside to using this value object over a hash is that we can be sure about correctly calling its interface. Previously, a call to `params[:qeury]` returns nil. `search_query.qeury` obviously does not.

A downside to this approach is that we need to type a little more, especially in tests. An upside of the downside is that its super clear that we're handling search parameters there, and not some other hash. The readability of the tests has slightly improved, I believe. 

To signal to the reader that something is a `SearchParameters` object, the variable name `search_params` is consistently used, instead of just `params`.

This will be especially useful to cleanly implement the search highlighter (https://trello.com/c/z0yi0Qit).

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -18,6 +18,7 @@ require "redis"
 require "matcher_set"
 require "parameter_parser/search_parameter_parser"
 require "parameter_parser/facet_parameter_parser"
+require "search_parameters"
 require "registries"
 require "schema/combined_index_schema"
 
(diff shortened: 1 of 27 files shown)


KushalP:
:+1:

PR #492, opened 2015-08-14
Add highlighted title and description

tijmenb:
This commit allows users to request two virtual fields in the search results: `title_with_highlighting` and `description_with_highlighting`.

These contain the title and description with `<em>` tags around the words the user was looking for. Elasticsearch automatically highlights stemmed words and synonyms, so when the user searches for "oversea", the words "overseas" and "abroad" will also be highlighted.

About the configuration of the highlight query: `number_of_fragments` specifies how many "fragments" elasticsearch will return. We set `number_of_fragments: 0` for titles to make ES return the entire string. For descriptions we use `number_of_fragments: 1` to enable truncation. This is preferable to doing the truncation ourselves, because ES can do truncation on both sides of the description by "centering" on the highlighted word. This does mean we need some voodoo to append/prepend ellipses the text.

Screenshot of [frontend once that uses this feature](https://github.com/alphagov/frontend/pull/858):

screen shot 2015-08-14 at 14 42 23

### DIFF
--- a/lib/unified_search_builder.rb
+++ b/lib/unified_search_builder.rb
@@ -8,6 +8,7 @@ require "query_components/popularity"
 require "query_components/best_bets"
 require "query_components/query"
 require "query_components/filter"
+require "query_components/highlight"
 require "query_components/facets"
 
 # Builds a query for a search across all GOV.UK indices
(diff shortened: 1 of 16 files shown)


lib/unified_search_builder.rb
```diff
@@ -22,11 +23,14 @@ def payload
     hash_without_blank_values(
       from: search_params.start,
       size: search_params.count,
-      fields: search_params.return_fields,
+      # `title` and `description` always needed to potentially populate virtual
+      # fields. If not explicitly requested they will not be sent to the user.
+      fields: search_params.return_fields + %w[title description],
```
rboulton, line 28: I feel that these extra fields ought only to be requested if the corresponding virtual fields are requested.

lib/unified_search_builder.rb
```diff
@@ -22,11 +23,14 @@ def payload
...
       query: query,
       filter: filter,
       sort: sort,
       facets: facets,
+      highlight: highlight,
```
rboulton, line 33: highlighting is definitely not a cheap operation.  Could we request it only if the user has asked for the highlighted fields in the response?

PR #495, opened 2015-08-18
Stop using `Document` for Registries

tijmenb:
`Index#documents_by_format` and `IndexForSearch#documents_by_format` use the `Document` class to wrap results. This is not necessary - the only task it performs is converting the elasticsearch response, which return all fields as arrays.

We add a MultivalueConverter class to perform this task.

This cleans up the tests and removes the use of the `Document` from the search-side.

Trello: https://trello.com/c/oCWEBqER


tijmenb:
For posterity, this change was inadvertently responsible for a massive improvement in Rummager's performance:



Average response time per day for the last 30 days.

PR #496, opened 2015-08-19
Remove GET /documents/:id endpoint

tijmenb:
The `GET /documents/:id` can be used to request a document from the rummager index. The endpoint is currently only used by our tests. It is no longer required and can safely be removed because no external services depend on it.

Kibana shows no requests to the endpoint in the last 30 days:

screen shot 2015-08-19 at 10 19 02

Additionally, [`GdsApi::Rummager` doesn't support the endpoint](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/rummager.rb) and [neither does Rummageable](https://github.com/alphagov/rummageable/blob/master/lib/rummageable.rb).

Some indexing-related integration tests currently use the `GET /documents/:id` endpoint to test if they have correctly indexed the data. They have been updated to verify their data with elasticsearch directly.

Trello: https://trello.com/c/oCWEBqER

Note that functionality to fetch an individual document was added in https://github.com/alphagov/rummager/pull/475.


KushalP:
Verified using the following Kibana query for those that are interested:

```
application:rummager AND @fields.path:"/*/documents*" AND @fields.method:"GET"
```

KushalP:
:+1:

PR #497, opened 2015-08-19
Use Rummager as rendering app for sitemap files

bradwright:
In production these routes were originally pointing at Rummager:

```
#<Route _id: 5277d004fa5ee55bbe000015, incoming_path: /sitemap.xml, route_type: exact, handler: backend, backend_id: rummager, redirect_to: nil, redirect_type: nil>
```

`frontend` fails to render this correctly.


bradwright:
Closed via #498

PR #505, opened 2015-08-26
Only enable highlighting when requested by the client

tijmenb:
Currently we request highlighting from elasticsearch for all queries, even when the client hasn't asked for either `title_with_highlighting` or `description_with_highlighting`. This means elasticsearch has to do work that's not needed, and we're throwing away the result.

This commit makes it so that we highlight if the client requests _either_ title or description. We could make it so that it doesn't highlight description when only asking for the title and not description, but that would require more code and is a use case not likely to happen.

For review by @rboulton

### DIFF
--- a/lib/query_components/highlight.rb
+++ b/lib/query_components/highlight.rb
@@ -1,6 +1,8 @@
 module QueryComponents
   class Highlight < BaseComponent
     def payload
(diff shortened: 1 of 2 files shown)


rboulton:
:sparkles:

So, I wonder if this will improve performance further...

PR #508, opened 2015-09-02
Store asylum support facet labels directly in rummager

robmckinnon:
This approach means we no longer need to set `allowed_values` in the schema configuration. 

In `finder-frontend` metadata labels now come directly from `rummager` as separate `_name` fields.

**Required for this specialist-publisher PR**: https://github.com/alphagov/specialist-publisher/pull/552


rboulton:
Had an offline conversation about this - summary: this looks good, but we're going to make all the `_name` fields be "searchable_identifier" (or searchable_identifiers, if appropriate), and all the corresponding not-name fields be just "identifier(s)".  This means that keyword searches for things like "Section 95" will find documents in the appropriate section.

robmckinnon:
Also agreed in offline conversation to alphabetically sort the field names in the document schema file.

rboulton:
I'm happy to merge this change.

One field I'm still concerned about is the "tribunal_decision_sub_category" field - does this include the parent category as well?  ie, is it something like "parent-category/sub-category".  If not, I think it probably should be (but might be missing information about this).  Happy to discuss in person sometime.

PR #509, opened 2015-09-02
Prevent use of section fields in searches

rboulton:
Ban the `section`, `subsection` and `subsubsection` field from use in
searches (in facets or as return fields).  We believe we have removed
all use of these fields by apps (see
https://trello.com/c/N0dU4k3G/296-remove-section-subsection-and-subsubsection-from-rummager),
but will test this carefully on preview to be sure.

Note that the data will be removed from the database automatically after this is deployed by the nightly migration run.

### DIFF
--- a/config/schema/document_types/edition.json
+++ b/config/schema/document_types/edition.json
@@ -1,8 +1,5 @@
 {
   "fields": [
-    "section",
-    "subsection",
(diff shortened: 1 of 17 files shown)


tijmenb:
:fireworks:

PR #517, opened 2015-09-09
Reformat health check

malcolmbaig:
Potentially lots more here that could be improved/cleaned up/extracted, but wanted to get feedback on this sooner rather than later.

Ticket: https://trello.com/c/sia4T2Ip/210-reformat-rummager-s-search-health-check-output-so-it-s-easier-to-analyse

Essentially:
- clean up the logging output of the health check, removing the sentences makes it easier to scan
- write the results of the health check to a CSV file in the root of the project

Sample output:

screen shot 2015-09-09 at 13 14 48

### DIFF
--- a/lib/health_check/search_check.rb
+++ b/lib/health_check/search_check.rb
@@ -1,9 +1,7 @@
 require "health_check/result"
 
 module HealthCheck
-  SearchCheck = Struct.new(:search_term, :imperative, :path, :minimum_rank, :weight)
-
-  class SearchCheck
+  class SearchCheck < Struct.new(:search_term, :imperative, :path, :minimum_rank, :weight)
     def valid_imperative?
       ["should", "should not"].include?(imperative)
     end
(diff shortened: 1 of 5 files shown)


lib/health_check/search_check.rb
```diff
@@ -29,28 +27,26 @@ def positive_check?
...
-      expectation = positive_check? ? "<= #{minimum_rank}" : "> #{minimum_rank}"
-      if found_index
-        message = "#{marker} Found '#{path}' for '#{search_term}' in position #{found_index + 1} (expected #{expectation})"
-      else
-        message = "#{marker} Didn't find '#{path}' for '#{search_term}' in any position (expected #{expectation})"
-      end
+      position_found = found_index ? found_index + 1 : 0
+      found_label    = found_index ? "FOUND" : "NOT FOUND"
+      expectation    = positive_check? ? "<= #{minimum_rank}" : "> #{minimum_rank}"
+      success_label  = success ? "PASS" : "FAIL"
```
alext: It's better not to align the `=`.  It makes for noisy diffs.

lib/health_check/search_checker.rb
```diff
@@ -1,14 +1,16 @@
...
   class SearchChecker
     attr_reader :search_client
 
-    def initialize(options = {})
-      @test_data_file = options[:test_data]
-      @search_client = options[:search_client]
+    def initialize(search_client:, test_data:, produce_report: true)
+      @test_data_file = test_data
+      @search_client  = search_client
+      @file_output    = produce_report ? SearchCheckReport.new : File.open(File::NULL, 'w')
```
alext: Instead of writing to null, can we not write at all when the report hasn't been requested.

malcolmbaig: Could do, the null object spares us an if statement though and is arguably cleaner in `run!`.

malcolmbaig:
@tijmenb yep, totally agree. Was looking for the minimal set of changes that would provide the functionality required, but extracting some of that result processing is a logical progression. Fine to add in this PR?

rboulton:
I'd prefer tidying up the very many arguments thing in a separate PR, keeping the functional changes to this one.  So - merged. :+1:

PR #518, opened 2015-09-10
Extract Result creation for search checks into its own class

malcolmbaig:
https://trello.com/c/sia4T2Ip/210-reformat-rummager-s-search-health-check-output-so-it-s-easier-to-analyse

Why?:

The existing implementation has a `result` method on SearchCheck which
is pretty long and complex. It also returns a struct that needs
populating  with lots of attributes. Extracting this out into its own
class makes SearchCheck less busy (single responsiblity), makes result
evaluation easier to understand, and removes the need for an overly
clunky Result struct.

How?:
- Add a new SearchCheckResult object, creating a separate
  SuggestionCheckResult so that SuggestionCheck logic remains
  largely unaffected.
- Convert all the struct attributes on Result into proper methods.
- Migrate result-related tests from SearchCheck to SearchCheckResult.


lib/health_check/search_check.rb
```diff
@@ -22,37 +26,8 @@ def valid?
...
-      else
-        logger.fail logging_output
-      end
-
-      score = success ? weight : 0
-      Result.new(
-        success, score, weight, success_label, found_label,
-        path, search_term, position_found, expectation
-      )
+      SearchCheckResult.new(check: self, search_results: search_results).build
```
KushalP: Given that `build` doesn't _build_ anything here, it would be cleaner to instantiate the object, call `log_the_result` directly and then return the instantiated object rather than hiding the details here.

malcolmbaig: On second thought -
- agree that build is a lousy name (hangover from a prior version of this method that did a bit more)
- reckon that a sensible place to trigger the logging is back in the SearchChecker - where we also (optionally) generate the CSV. Keeps IO in one place.

Will push something in a bit.

rboulton:
Is there any reason not to just delete test/unit/health_check/search_check_test.rb entirely?

malcolmbaig:
@rboulton it has behaviour, so I'm going to put a few simple tests in. Follow-up PR.

rboulton:
Fair enough.  Merging.

PR #521, opened 2015-10-06
Extract code from Elasticsearch::Index

tijmenb:
`Elasticsearch::Index` is a very large class with too many responsibilities. By delegating things to extracted classes we can start to pull apart the unit tests, which are testing (or stubbing) too much at the moment, making it difficult to add functionality.

There's no change in functionality from this PR.

Ground work for https://trello.com/c/xQtX24kW

### DIFF
--- a/lib/elasticsearch/advanced_search.rb
+++ b/lib/elasticsearch/advanced_search.rb
@@ -1,0 +1,47 @@
+require "elasticsearch/advanced_search_query_builder"
+
+module Elasticsearch
+  class AdvancedSearch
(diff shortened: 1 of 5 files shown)


KushalP:
This looks good, but can you explain why you haven't added tests for the new classes you've created?

tijmenb:
@KushalP Because the code is still covered by the existing tests and I don't want to introduce new tests without removing older (problematic) ones. I've got another PR in progress that cleans up some of the test classes. After that it might be possible to move some tests in `index_test.rb` down to a lower level.

PR #522, opened 2015-10-06
Simplify test helpers and insert data into search directly

tijmenb:
This PR simplifies the test helpers by letting all integration tests inherit from the same `IntegrationTest` class and inlining methods and modules. This makes the code more easy to follow as we don't have to think about the inheritance of `setup` and `teardown` blocks, which have been inlined. 

All this enables us to use consistently insert testing data directly into elasticsearch. Currently this is done via the app itself (by calling `post "/documents"` for example). This tightly couples the indexing of the application to the search. It caused particular pain in the implementation of https://trello.com/c/xQtX24kW, which is now significantly easier.

It would've been better if this was two PRs, but it was hard to separate the two goals here. I can walk provide more explanation of the refactors if needed.


test/unit/elasticsearch/elasticsearch_index_test.rb
```diff
@@ -7,80 +7,28 @@ class ElasticsearchIndexTest < MiniTest::Unit::TestCase
...
-  end
-
-  def successful_response
-    <<-eos
-{"took":5,"items":[
-  { "index": { "_index":"mainstream_test", "_type":"edition", "_id":"/foo/bar", "ok":true } }
-]}
-    eos
-  end
+  # Elasticsearch::Index#real_name
```
rboulton, line 13: Is this comment meant to be here?

rboulton, line 13: Ah - I think this is meant to indicate the start of tests for the "real_name" method of Index.  I've not seen that convention before - perhaps if it said `# Tests for Elasticsearch::Index#real_name` that would be clearer?  Or maybe this indicates that this file should be split up further?

Everything else in this PR looks :sparkles: so I'm going to merge, and if you'd like to split up the file that could be done later.  Or not.

PR #527, opened 2015-10-08
Remove functional testing

tijmenb:
Rummager currently has three types of tests: unit, integration and "functional". Over the last couple of years integration and functional tests have started to look and act the same. Most functional tests even inherit from `IntegrationTest`. The only actual difference in practice seems to be that the functional tests stub out elasticsearch, while integration tests don't. But our unit tests _also_ stub out elasticsearch, making the difference between functional and unit tests quite fuzzy.

To make the test setup easier this PR removes the concept of "functional" tests.

Individual commits tell the entire story, but generally this happened:
- Functional tests that duplicated unit or integration tests have been removed.
- Functional tests that didn't add much value have been removed.
- Functional tests that serve a real purpose were converted to either integration tests or unit tests. I've rewritten some of the integration tests to make them more readable and test more broadly.

This PR should make working on Rummager a bit easier. There's less stubbing, making it more robust and less likely to need change. It's also a bit faster now. Code coverage has stayed about the same.

Before:

```
Total lines: 6059
LoC: 2780
Total coverage: 98.12%
Code coverage: 95.90%
```

After:

```
Total lines: 6042
LoC: 2774
Total coverage: 98.16%
Code coverage: 96.00%
```


KushalP:
:boom: :boom: :boom:

This PR is excellent. The commits have easy to understand messages and small enough that I didn't need too long to understand why you did what you did.

:+1: :+1: :+1:

PR #536, opened 2015-10-20
Allow faceting on all filterable fields

tijmenb:
Rummager restricts the fields we use to facet. Removing that restriction will improve Rummager by having less schema/data knowledge leak into application code, making it easier to work with.

The allowed facets fields is already a subset of allowed_filter_fields, so making them equivalent is safe[1]

[1] _I assume_. To be confirmed by @rboulton.

### DIFF
--- a/lib/parameter_parser/base_parameter_parser.rb
+++ b/lib/parameter_parser/base_parameter_parser.rb
@@ -23,21 +23,6 @@ class BaseParameterParser
     "document_type" => "_type",
   }
 
-  # The fields listed here are the only ones which can be used to calculated
-  # facets for.  This should be a subset of allowed_filter_fields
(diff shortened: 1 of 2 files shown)


rboulton:
The main reason for restricting the set of facet fields is because calculating facets can use up a lot of memory, and calculating them on fields with "very high cardinality" (ie, lots of different terms in the field after the analysis process) can even crash elasticsearch.  Newer versions of elasticsearch have considerably more protection against this being a problem, but I'm a bit nervous of removing the restriction with elasticsearch 1.4.4.  After we've upgraded to at least 1.6 things are a lot safer, so I'd rather upgrade before applying this change.

tijmenb:
Good to know!

PR #540, opened 2015-11-05
TagLookup: add tests & refactor

tijmenb:
This PR refactors the `TagLookup` class and adds tests for behaviours where there are duplicated tags for a document.

The immediate motivation for this is that we tried understanding the transformation process during an incident-related investigation. The complexity of the code made this more difficult than necessary.

Commit messages contain more detail.


malcolmbaig:
lgtm

PR #542, opened 2015-11-06
Implement Exact Phrase Matching

ghost:
If the search term is entirely enclosed in double quotes, this is an exact phrase match.

The search term "London Transport" will return documents that have the phrase "London Transport" in them, but not those that have "London" and "Transport" with other words in between.

Synonyms are not used when doing an exact phrase match.  Therefore searching for "Driving abroad" will not return documents that have "Driving overseas" in them.

Stemming is NOT turned off: so "London Transport" will return "London's Transport"

https://trello.com/c/npQ3A5SO/408-exact-phrase-matching-for-search-queries

### DIFF
--- a/lib/query_components/core_query.rb
+++ b/lib/query_components/core_query.rb
@@ -39,6 +39,15 @@ module QueryComponents
     MINIMUM_SHOULD_MATCH = "2<2 3<3 7<50%"
 
     def payload
+      if @search_params.quoted_search_phrase?
+        payload_for_quoted_phrase
+      else
+        payload_for_unquoted_phrase
+      end
+    end
+
+  private
+    def payload_for_unquoted_phrase
       {
         bool: {
           must: must_conditions,
@@ -47,7 +56,10 @@ module QueryComponents
       }
     end
 
-    private
+    def payload_for_quoted_phrase
+      groups = [ field_boosts_phrase ]
+      groups.map { |queries| dis_max_query(queries) }
+    end
 
     def must_conditions
       [query_string_query]
@@ -82,6 +94,14 @@ module QueryComponents
       end
     end
 
+    def field_boosts_phrase
+      # Return the highest weight found by looking for a phrase match in
+      # individual fields
+      MATCH_FIELDS.map { |field_name, boost|
+        match_query("#{field_name}.no_stop", search_term, type: :phrase, boost: boost)
+      }
+    end
+
     def exact_match_boost
       {
         multi_match: {
@@ -109,6 +129,36 @@ module QueryComponents
         DEFAULT_QUERY_ANALYZER_WITHOUT_SYNONYMS
       else
         DEFAULT_QUERY_ANALYZER
+      end
+    end
+
+    def match_query(field_name, query, type: :boolean, boost: 1.0, minimum_should_match: MINIMUM_SHOULD_MATCH, operator: :or)
+      {
+        match: {
+          field_name => {
+            type: type,
+            boost: boost,
+            query: query,
+            minimum_should_match: minimum_should_match,
+            operator: operator,
+          }
+        }
+      }
+    end
+
+    def dis_max_query(queries, tie_breaker: 0.0, boost: 1.0)
+      # Calculates a score by running all the queries, and taking the maximum.
+      # Adds in the scores for the other queries multiplied by `tie_breaker`.
+      if queries.size == 1
+        queries.first
+      else
+        {
+          dis_max: {
+            queries: queries,
+            tie_breaker: tie_breaker,
+            boost: boost,
+          }
+        }
       end
     end
   end
--- a/lib/query_components/text_query.rb
+++ b/lib/query_components/text_query.rb
@@ -22,6 +22,20 @@ module QueryComponents
     MINIMUM_SHOULD_MATCH = "2<-1 5<-2 7<75%"
 
     def payload
+      if @search_params.quoted_search_phrase?
+        payload_for_quoted_phrase
+      else
+        payload_for_unquoted_phrase
+      end
+    end
+
+  private
+    def payload_for_quoted_phrase
+      groups = [ field_boosts_phrase ]
+      dismax_groups(groups)
+    end
+
+    def payload_for_unquoted_phrase
       {
         bool: {
           must: must_conditions,
@@ -29,8 +43,6 @@ module QueryComponents
         }
       }
     end
-
-  private
 
     def must_conditions
       [all_searchable_text_query]
@@ -59,10 +71,11 @@ module QueryComponents
       groups << field_boosts_synonyms unless search_params.disable_synonyms?
       groups << field_boosts_shingles
       groups << field_boosts_id_codes
+      dismax_groups(groups)
+    end
 
-      groups.map { |queries|
-        dis_max_query(queries)
-      }
+    def dismax_groups(groups)
+      groups.map { |queries| dis_max_query(queries) }
     end
 
     def field_boosts_words
--- a/lib/search_parameters.rb
+++ b/lib/search_parameters.rb
@@ -3,13 +3,20 @@
 # Value object that holds the parsed parameters for a search.
 class SearchParameters
   attr_accessor :query, :order, :start, :count, :return_fields, :facets,
-                :filters, :debug, :suggest
+                :filters, :debug, :suggest, :is_quoted_phrase
+
+  QUOTED_STRING_REGEX = /^\s*"[^"]+"\s*$/   # starts and ends with quotes with no quotes in between, with or without leading or trailing whitespace
 
   def initialize(params = {})
     params = { facets: [], filters: {}, debug: {}, return_fields: [] }.merge(params)
     params.each do |k, v|
       public_send("#{k}=", v)
     end
+    determine_if_quoted_phrase
+  end
+
+  def quoted_search_phrase?
+    @is_quoted_phrase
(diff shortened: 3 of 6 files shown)


lib/search_parameters.rb
```diff
@@ -12,6 +12,11 @@ def initialize(params = {})
     end
   end
 
+  def quoted_search_phrase?
```
rboulton: I think abstracting this makes sense.

I think it would be good to check that there are no double quotes in the body of the query, too.  eg, for the query `"a cat" "a dog"`, I don't think we want to activate this special behaviour (because it won't work as I think the users would expect).

I wonder if this could instead return the phrase, if there's a match - that would avoid needing to put anything in BaseComponent, be a bit neater than the 'sub' version, and make sure there's no chance of this matching and the extraction of the phrase from the quotes getting out of sync if one of them was extended.

lib/query_components/text_query.rb
```diff
@@ -59,10 +77,17 @@ def should_conditions
       groups << field_boosts_synonyms unless search_params.disable_synonyms?
       groups << field_boosts_shingles
       groups << field_boosts_id_codes
+      dismax_groups(groups)
+
+      # groups.map { |queries|
```
rboulton: We shouldn't commit commented-out code.

lib/query_components/base_component.rb
```diff
@@ -2,14 +2,13 @@ module QueryComponents
   class BaseComponent
     include Elasticsearch::Escaping
 
-    attr_reader :search_params
+    attr_reader :search_params, :search_term
 
     def initialize(search_params = SearchParameters.new)
       @search_params = search_params
+      @search_term   = @search_params.query.strip.sub(/^"/, "").sub(/"$/, "") unless @search_params.query.nil?
```
rboulton: Is this just here to abstract shared code between `core_query` and `text_query`?  If so, I won't bother abstracting it (ie, I'd just have the same code twice), because we're working towards removing `core_query`, and that's actually easier if it's more self-contained.

Also, I'm uncomfortable about the `unless` clause here - if I read it right, this will mean that `@search_term` isn't even defined if `search_params.query` is nil, which seems likely to result in unexpected errors due to failing to look up the variable.  I'd rather `@search_term` was at least defined as being `nil` in that case.

test/unit/search_parameter_test.rb
```diff
@@ -0,0 +1,43 @@
...
+			params = SearchParameters.new(query: %Q{  \t  "phrase enclosed with quotes"} )
+			assert_equal true, params.quoted_search_phrase?
+		end
+	
+		should "return true if query enclosed with quotes but with trailing whitespace" do
+			params = SearchParameters.new(query: %Q{"phrase enclosed with quotes"  \t  } )
+			assert_equal true, params.quoted_search_phrase?
+		end
+	end
+end
```
jackscotti: Please use soft tabs, here is our styleguide: https://github.com/alphagov/styleguides/blob/master/ruby.md#general

test/integration/quoted_and_unquoted_searches_test.rb
```diff
@@ -0,0 +1,178 @@
...
+    create_meta_indexes
+    reset_content_indexes
+    end
+
+  def teardown
+    clean_test_indexes
+  end
+
+
+  # NEW WEIGHTING TESTS
```
jackscotti, line 19: I don't think we should refer to code as `new` or `old`. What if we will need to introduce additional tests?

tijmenb, line 19: @jackscotti Yes, normally :thumbsup:. But the "new weighting" here refers to the new search weighting that was implemented a few months back but is still under the feature flag `&debug=new_weighting`, so it's okay here.

lib/query_components/base_component.rb
tijmenb: @rboulton instance variables are nil by default, so this code does the correct thing.

@srmoscow are we sure we need to strip the quotes? It may be that elasticsearch strips the quotes from the query anyway, so perhaps this change is not necessary.

lib/query_components/text_query.rb
```diff
@@ -22,6 +22,18 @@ class TextQuery < BaseComponent
...
 
     def payload
+      if @search_params.quoted_search_phrase?
+        payload_for_quoted_phrase
+      else
+        payload_for_unquoted_phrase
+      end
+    end
+
+    def payload_for_quoted_phrase
```
tijmenb: do these methods have to be public?

lib/query_components/base_component.rb
rboulton: Ah, @tijmenb - thanks.  My python showing through there.

lib/query_components/text_query.rb
```diff
@@ -51,6 +63,12 @@ def all_searchable_text_query
       dis_max_query(queries, tie_breaker: 0.1)
     end
 
+    def quoted_phrase_conditions
+      groups = [ field_boosts_phrase ]
+      dismax_groups(groups)
+    end
+
+
```
rboulton: Nitpick, but try not to add extra whitespace lines.

lib/query_components/text_query.rb
```diff
@@ -59,10 +77,17 @@ def should_conditions
...
+      dismax_groups(groups)
+
+      # groups.map { |queries|
+      #   dis_max_query(queries)
+      # }
+    end
 
-      groups.map { |queries|
+    def dismax_groups(groups)
+      groups.map do |queries|
```
rboulton: We try to use {} instead of `do` `end` when we care about the result of the iteration (not explicitly said in the styleguide, but a consequence of the "Never chain do...end" guideline)

test/integration/quoted_and_unquoted_searches_test.rb
```diff
@@ -0,0 +1,178 @@
...
+  #
+  def test_new_weighting_three_matches_found_for_london
+  	commit_london_transport_docs
+    get "/unified_search?q=london&debug=new_weighting"
+    assert_equal 200, last_response.status
+    assert_equal 3, parsed_response["results"].size
+  end
+
+  def test_new_wieghting_three_matches_found_for_transport
```
rboulton: `weighting` rather than `wieghting`

lib/query_components/text_query.rb
```diff
@@ -51,6 +63,12 @@ def all_searchable_text_query
       dis_max_query(queries, tie_breaker: 0.1)
     end
 
+    def quoted_phrase_conditions
+      groups = [ field_boosts_phrase ]
+      dismax_groups(groups)
```
rboulton: If there's only a single item in the list, there's no benefit to putting it through dismax_groups.

rboulton: Sorry - disregard this, I misread the code.  The "field_boosts_phrase" bit consists of multiple items which dismax_groups joins together with a dismax query.

lib/query_components/core_query.rb
```diff
@@ -47,7 +56,9 @@ def payload
       }
     end
 
-    private
+    def payload_for_quoted_phrase
+      exact_field_boosts
```
rboulton: I think this should be changed to produce exactly the same query as the version in `text_query` does in this case.  The query you're producing here uses fields which have had stop wording applied, so won't work quite right.

test/integration/quoted_and_unquoted_searches_test.rb
jackscotti, line 19: oh! Thanks for the clarification.

lib/query_components/core_query.rb
```diff
@@ -47,7 +56,10 @@ def payload
       }
     end
 
-    private
+    def payload_for_quoted_phrase
+      groups = [ field_boosts_phrase ]
+      groups.map { |queries| dis_max_query(queries) }
+    end
```
tijmenb, line 62: Could this be easier written as `[ dis_max_query(field_boosts_phrase) ]`?

test/unit/search_parameter_test.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    should "not strip leading and trailing whitespace if phrase contains embedded quotes" do
+      params = SearchParameters.new(query: %Q{  \t"Enclosing quotes but with "embedded" quotes"  })
+      assert_equal %Q{  \t"Enclosing quotes but with "embedded" quotes"  }, params.query
+    end
+
+  end
+
+
+
```
tijmenb: Some extra whitespace here.

lib/search_parameters.rb
```diff
@@ -35,4 +42,18 @@ def disable_best_bets?
   def suggest_spelling?
     query && suggest.include?('spelling')
   end
+
+  private
+
+  def determine_if_quoted_phrase
+    unless @query.nil?
+      if @query.strip =~ QUOTED_STRING_REGEX
+        @query = $2                             # remove enclosing quotes if present
```
tijmenb: Why is it necessary to strip the quotes?

rboulton:
I'm happy with this general approach now - Tijmen's remaining comments would be good to address, but consider me happy to have it merged after that.

lib/query_components/core_query.rb
rboulton, line 62: This is like this to match the implementation in text_query, which in turn uses a shared function to apply the dis_max.  It might be better written this way, but I'm on the fence, so not going to block merge on this.

PR #547, opened 2015-11-25
Import scripts for alpha taxonomy data

malcolmbaig:
To import alpha taxonomy definitions into Rummager, providing Design with easier access to taxonomy definitions when prototyping.

This PR includes a rake task for downloading taxonomy definitions in the project `data/prototype_taxonomy` directory. The intention is that this will be run via a Jenkins job to make the data available on all Rummager machines. The existing DocumentPreparer class has been updated to inspect this file for taxon mappings and add them to the document if there is a matching link.

Trello: https://trello.com/c/2f1Ct7tq

### DIFF
--- a/config/schema/base_document_type.json
+++ b/config/schema/base_document_type.json
@@ -14,6 +14,7 @@
     "updated_at",
     "public_timestamp",
     "latest_change_note",
-    "spelling_text"
+    "spelling_text",
+    "alpha_taxonomy"
   ]
 }
--- a/config/schema/base_document_type.json
+++ b/config/schema/base_document_type.json
@@ -1,6 +1,7 @@
 {
   "fields": [
     "all_searchable_text",
+    "alpha_taxonomy",
     "format",
     "title",
     "description",
@@ -14,7 +15,6 @@
     "updated_at",
     "public_timestamp",
     "latest_change_note",
-    "spelling_text",
-    "alpha_taxonomy"
+    "spelling_text"
   ]
 }
--- a/config/schema/base_document_type.json
+++ b/config/schema/base_document_type.json
@@ -2,19 +2,19 @@
   "fields": [
     "all_searchable_text",
     "alpha_taxonomy",
-    "format",
-    "title",
     "description",
-    "link",
+    "format",
     "indexable_content",
-    "popularity",
-    "organisations",
-    "specialist_sectors",
-    "policies",
-    "mainstream_browse_pages",
-    "updated_at",
-    "public_timestamp",
     "latest_change_note",
-    "spelling_text"
+    "link",
+    "mainstream_browse_pages",
+    "organisations",
+    "policies",
+    "popularity",
+    "public_timestamp",
+    "specialist_sectors",
+    "spelling_text",
+    "title",
+    "updated_at"
   ]
 }
--- a/lib/indexer/document_preparer.rb
+++ b/lib/indexer/document_preparer.rb
@@ -38,7 +38,7 @@ module Indexer
     end
 
     def prepare_tags_field(doc_hash)
-      Indexer::TagLookup.new.prepare_tags(doc_hash)
+      Indexer::TagLookup.prepare_tags(doc_hash)
     end
 
     def prepare_format_field(doc_hash)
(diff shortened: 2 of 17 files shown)


lib/indexer/document_preparer.rb
```diff
@@ -27,7 +38,7 @@ def prepare_popularity_field(doc_hash, popularities)
     end
 
     def prepare_tags_field(doc_hash)
-      Indexer::TagLookup.new.prepare_tags(doc_hash)
```
tijmenb, line 30: What's the value of changing the interface? How I see it, if we have to add more args to the method, we'd have to change code in four places instead of two.

If the empty initializer os really bothering you I'd prefer changing it into `TagLookup.new(doc_hash).doc_hash_with_tags`

lib/taxonomy_prototype/data_downloader.rb
```diff
@@ -0,0 +1,82 @@
...
+    def download
+      begin
+        File.open(self.class.cache_location, "wb") do |file|
+          SHEETS.each do |sheet|
+            sheet_url = spreadsheet_url(key: sheet[:key], gid: sheet[:gid])
+            logger.info "Attempting download of #{sheet[:name]} (#{sheet_url})"
+            remote_taxonomy_data = open(sheet_url).read
+            logger.info "Downloaded #{sheet[:name]}"
+
+            relevant_columns_in(remote_taxonomy_data).each do |row|
```
tijmenb: I think we do need a test for the parts of this part of the class. Perhaps we could split up this file into a `Downloader` and a `Parser`, and add some tests for the parser?

test/unit/taxonomy_prototype/taxon_finder_test.rb
```diff
@@ -0,0 +1,24 @@
+require "test_helper"
+require "taxonomy_prototype/taxon_finder"
+
+describe TaxonomyPrototype::TaxonFinder do
+  describe "find_by(slug:)" do
+    before do
+      File.stubs(:exist?).returns(true)
```
tijmenb: Is this block only relevant to the first test? If that's the case it might be clearer to inline it.

malcolmbaig: reckon in-lining the File stub might be clearer, the CSV stuff is needed by both existing tests. Will also introduce a third test for the scenario where no file is present.

lib/indexer/document_preparer.rb
malcolmbaig, line 30: Actually chucked that in temporarily to make stubbing the interface easier in some other test, and just left it in because I thought it was fine. Not terribly fussed - will revert this and update the  corresponding test setup.

lib/taxonomy_prototype/data_downloader.rb
malcolmbaig: You're probably right. Will take a look.

config/schema/base_document_type.json
```diff
@@ -14,6 +14,7 @@
     "updated_at",
     "public_timestamp",
     "latest_change_note",
-    "spelling_text"
+    "spelling_text",
+    "alpha_taxonomy"
```
tijmenb: Can you put this in alphabetically?

rboulton:
This code mostly looks fine.  I'm finding it hard to see from this PR how this is intended to work in our production environment: specifically - when is downloading meant to happen, where is the data stored, is it downloaded on all machines or downloaded centrally and then replicated, etc.

I think the current code forces the download to happen every time migrate index is run.  I think it's probably desirable to keep this independent.

Happy to discuss face-to-face when convenient.

tijmenb:
I think we forgot to account for rummager being on multiple machines, to `rake taxonomy_prototype:download_data` will only download the file on one machine. When the re-indexing jobs are run on other machines, the cached file won't be there.

I can see two solutions:
1. Add a shared database/cache or put the data in an index 
2. Download the data if it doesn't exist when indexing the document (https://github.com/alphagov/rummager/pull/547/files?diff=unified#diff-348239dd834c9c22be9df9175e7417d3R11).

The first adds more tech overhead, the second is slower and could interfere with the re-indexing.

lib/indexer/document_preparer.rb
malcolmbaig, line 30: For completeness here - decided to keep the class method interface after all because minitest makes it unreasonably weird to stub out this method call.

rboulton:
This looks good to me.  Would be good to squash the history a bit to remove the revert bit.

PR #551, opened 2015-12-08
Improve TaxonFinder unit test

malcolmbaig:
This now exercises the logic in TaxonFinder using a tempfile that contains
fixture data that matches the form expected of a typical input CSV file.
This approach more completely tests the unit as a whole, including the
config for how a CSV should be read.

### DIFF
--- a/test/unit/taxonomy_prototype/taxon_finder_test.rb
+++ b/test/unit/taxonomy_prototype/taxon_finder_test.rb
@@ -1,13 +1,18 @@
 require "test_helper"
 require "taxonomy_prototype/taxon_finder"
+require "tempfile"
 
 describe TaxonomyPrototype::TaxonFinder do
   describe "find_by(slug:)" do
     before do
-      CSV.stubs(:read).returns([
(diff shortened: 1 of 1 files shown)


test/unit/taxonomy_prototype/taxon_finder_test.rb
```diff
@@ -1,13 +1,19 @@
...
 describe TaxonomyPrototype::TaxonFinder do
   describe "find_by(slug:)" do
     before do
-      CSV.stubs(:read).returns([
-        ["test-taxon-1", "/test/slug/1"],
-        ["test-taxon-2 > test-taxon-3", "/test/slug/2"]
-      ])
+      @temp_taxon_file = Tempfile.new("taxon_finder_test.csv")
+      @temp_taxon_file.write("test-taxon-1\t/test/slug/1\ntest-taxon-2 > test-taxon-3\t/test/slug/2\n")
+      @temp_taxon_file.rewind
```
rboulton: Do you actually need this `rewind`?  I think you're not actually using the open file descriptor again after this point (just using the path).

PR #553, opened 2016-01-13
Use bearer token for publishing API authentication

danielroseman:
https://trello.com/c/tx50hSOx/492-add-bearer-token-value-to-all-apps-using-publishing-api

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -3,7 +3,15 @@ namespace :publishing_api do
   task :publish_special_routes do
     require 'gds_api/publishing_api/special_route_publisher'
 
-    publisher = GdsApi::PublishingApi::SpecialRoutePublisher.new(logger: Logger.new(STDOUT))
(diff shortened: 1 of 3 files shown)


lib/tasks/publishing_api.rake
```diff
@@ -3,7 +3,15 @@ namespace :publishing_api do
...
     require 'gds_api/publishing_api/special_route_publisher'
 
-    publisher = GdsApi::PublishingApi::SpecialRoutePublisher.new(logger: Logger.new(STDOUT))
+    publishing_api = GdsApi::PublishingApiV2.new(
+      Plek.new.find('publishing-api'),
+      bearer_token: ENV['PUBLISHING_API_BEARER_TOKEN'] || 'example'
+    )
+
+    publisher = GdsApi::PublishingApi::SpecialRoutePublisher.new(
+      logger: logger,
```
elliotcm: Where does this logger come from?

danielroseman:
@elliotcm thanks, fixed. Odd that the task wasn't showing any errors when I ran it, though.

PR #554, opened 2016-01-15
Index documents from published_documents message exchange

bishboria:
Currently, changes to topics and other tags in content-tagger don't cause the search index (accessed via rummager) to be updated correspondingly. We want such changes to be reflected as soon as possible in the search index.

Part of: https://trello.com/c/lRnxI2x5/472-make-rummager-index-the-tags-in-the-links-hash

### DIFF
--- a/lib/index_documents.rb
+++ b/lib/index_documents.rb
@@ -1,0 +1,86 @@
+require_relative "../app"
+require 'gds_api/publishing_api_v2'
+
+class IndexDocuments
+  def process(message)
+    raw_links = links_from_payload(message)
+
+    if !raw_links.nil? && !raw_links.empty?
+      links = remap_keys(raw_links, content_tagger_to_rummager_map)
+      links.each do |(link_type, content_ids)|
+        links[link_type] = paths_from_content_ids(content_ids)
+      end
+
+      base_path = document_base_path(message)
+      update_index(base_path, links)
+    end
+
+    message.ack
+  end
+
+private
+  def links_from_payload(message)
+    message.payload["links"].try(:slice, *content_tagger_tags)
+  end
+
+  def paths_from_content_ids(content_ids)
+    content_ids.map { |content_id| base_path content_id }.compact.sort
+  end
+
+  def content_tagger_tags
+    %w{
+      mainstream_browse_page
+      parent
+      topics
+      organisations
+    }
+  end
+
+  def content_tagger_to_rummager_map
+    {
+      "topics" => "specialist_sectors",
+      "organisations" => "lead_organisations",
+      "mainstream_browse_page" => "mainstream_browse_pages",
+      "parent" => "parent",
+    }
+  end
+
+  def remap_keys(hash, keymap)
+    Hash[hash.map {|k, v| [keymap[k], v] }]
+  end
+
+  def document_base_path(message)
+    message.payload["base_path"]
+  end
+
+  def base_path(content_id)
+    GdsApi::PublishingApiV2.new(Plek.current.find('publishing-api')).get_content(content_id)["base_path"]
+  end
+
+  def update_index(base_path, links)
+    indices_with_base_path(base_path).map do |index|
+      index.amend(base_path, links)
+    end
+  end
+
+  def indices_with_base_path(document_base_path)
+    indices_to_search = search_server.index_for_search(search_config.content_index_names)
+    results = indices_to_search.raw_search(query: {term: {link: document_base_path}})
+
+    indices = results["hits"]["hits"].map do |hit|
+      Elasticsearch::Index.strip_alias_from_index_name(hit["_index"])
+    end
+
+    indices.map do |index|
+      search_server.index(index)
+    end
+  end
+
+  def search_server
+    @search_server ||= search_config.search_server
+  end
+
+  def search_config
+    @search_config ||= Rummager.settings.search_config
+  end
+end
--- a/lib/tasks/message_queue.rake
+++ b/lib/tasks/message_queue.rake
@@ -1,0 +1,20 @@
+namespace :message_queue do
+  desc "Index documents that are published to the publishing-api"
+  task :index_documents_from_publishing_api do
+    # The following environment variables need to be set
+    # RABBITMQ_HOSTS
+    # RABBITMQ_VHOST
+    # RABBITMQ_USER
+    # RABBITMQ_PASSWORD
+
+    require 'govuk_message_queue_consumer'
+    require_relative '../index_documents'
+
+    # routing_key is defaulted to '#'
+    GovukMessageQueueConsumer::Consumer.new(
+      queue_name: "rummager_to_be_indexed",
+      exchange_name: "published_documents",
+      processor: IndexDocuments.new,
+    ).run
+  end
+end
--- a/test/integration/index_documents_test.rb
+++ b/test/integration/index_documents_test.rb
@@ -1,0 +1,71 @@
+require "integration_test_helper"
+require 'govuk_message_queue_consumer'
+require 'govuk_message_queue_consumer/test_helpers'
+require './lib/index_documents'
+
+class IndexDocumentsTest < IntegrationTest
+  include IndexDocumentsTestHelpers
+
+  def setup
+    stub_elasticsearch_settings
+    create_test_indexes
+    stub_tagging_lookup
+    stub_calls_for_index_documents_test
+  end
+
+  def teardown
+    clean_test_indexes
+  end
+
+  def test_topics_get_indexed_from_publishing_api
+    Elasticsearch::Amender.any_instance.expects(:amend)
+
+    m = GovukMessageQueueConsumer::MockMessage.new(payload_with_topics)
+    IndexDocuments.new.process(m)
+
+    assert m.acked?
+  end
+
+  def test_empty_topics_get_indexed_from_publishing_api
+    Elasticsearch::Amender.any_instance.expects(:amend)
+
+    m = GovukMessageQueueConsumer::MockMessage.new(payload_empty_topics)
+    IndexDocuments.new.process(m)
+
+    assert m.acked?
+  end
+
+  def test_no_topics_no_update
+    Elasticsearch::Amender.any_instance.expects(:amend).never
+
+    m = GovukMessageQueueConsumer::MockMessage.new(payload_no_topics)
+    IndexDocuments.new.process(m)
+
+    assert m.acked?
+  end
+
+  def test_no_links_no_update
+    Elasticsearch::Amender.any_instance.expects(:amend).never
+
+    m = GovukMessageQueueConsumer::MockMessage.new(payload_no_links)
+    IndexDocuments.new.process(m)
+
+    assert m.acked?
+  end
+
+  def payload_with_topics
+    payload_empty_topics.merge({"links" => {"topics" => ["cc9eb8ab-7701-43a7-a66d-bdc5046224c0"]}})
+  end
+
+  def payload_empty_topics
+    payload_no_topics.merge({"links" => {"topics" => []}})
+  end
+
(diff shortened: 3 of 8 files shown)


lib/tasks/message_queue.rake
```diff
@@ -0,0 +1,19 @@
+namespace :message_queue do
+  desc "Index documents that are published to the publishing-api"
+  task :index_documents_from_publishing_api do
+    ENV["RABBITMQ_HOSTS"] = "localhost"
```
rboulton: I think these should be set in puppet, rather than in the app; puppet is the place that we normally put configuration.

Also, this value for RABBITMQ_HOSTS won't be right in production.

lib/index_documents.rb
```diff
@@ -0,0 +1,48 @@
...
+    topic_content_ids = topic_content_ids(message)
+
+    unless topic_content_ids.nil? # no topics field in links, don't update
+      topic_base_paths = topic_content_ids.map { |content_id| base_path(content_id) }
+      update_index(base_path, topic_base_paths)
+    end
+
+    message.ack
+  rescue Exception => e # bury any errors, retry
+    message.retry
```
rboulton: Why bury the errors?  We should at least send them to errbit.

benlovell: Also, it's generally better to rescue from at the least `StandardError` (thus its derivatives) so you're not rescuing from runtime issues.

config/rabbitmq.yml
```diff
@@ -0,0 +1,7 @@
+defaults: &defaults
+  host: localhost
+  port: 5672
+  vhost: /
+  user: publishing_api
+  pass: publishing_api
+  recover_from_connection_close: true
```
rboulton: Missing newline at the end of this file.

config/rabbitmq.yml
```diff
@@ -0,0 +1,7 @@
+defaults: &defaults
```
rboulton: What uses this config file?  We're passing the configuration in via environment variables, aren't we?

lib/tasks/message_queue.rake
```diff
@@ -0,0 +1,21 @@
+namespace :message_queue do
+  desc "Index documents that are published to the publishing-api"
+  task :index_documents_from_publishing_api do
+    # The following environment variables need to be set
+    # RABBITMQ_HOSTS
+    # RABBITMQ_PORT
```
rboulton: RABBITMQ_PORT doesn't appear to be a thing that exists in the message consumer code.  I think we're just using the default AMQP port always.

lib/index_documents.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  private
+
+  def document_base_path(message)
+    message.payload["base_path"]
+  end
+
+  def topic_content_ids(message)
+    message.payload["links"].try(:[], "topics")
```
rboulton: I'm not familiar with this idiom, and am having difficulty working out what it does.  Is it equivalent to `message.payload["links"].fetch("topics", {})` for the purposes of what we're doing?

benlovell: That would fail if `payload["links"]` is `nil`.

lib/index_documents.rb
```diff
@@ -0,0 +1,54 @@
...
+
+class IndexDocuments
+  def process(message)
+    base_path = document_base_path(message)
+    topic_content_ids = topic_content_ids(message)
+
+    unless topic_content_ids.nil? # no topics field in links, don't update
+      topic_base_paths = topic_content_ids.map { |content_id| base_path(content_id) }
+      update_index(base_path, topic_base_paths)
+    end
```
rboulton: We need to do this for all the tag types that can be be tagged in content-tagger.  So: mainstream_browse_pages, topics, and organisations (at present).

lib/index_documents.rb
```diff
@@ -0,0 +1,54 @@
...
+  def base_path(content_id)
+    GdsApi::PublishingApiV2.new(Plek.current.find('publishing-api')).get_content(content_id)["base_path"]
+  end
+
+  def update_index(base_path, topic_base_paths)
+    current_index.amend(base_path, { topics: topic_base_paths })
+  end
+
+  def current_index
+    @current_index ||= search_server.index(settings.default_index_name)
```
rboulton: This won't work if the documents aren't in the `mainstream` index, I think.

config/rabbitmq.yml
bishboria: I added this when starting the ticket, I'll remove the file

lib/index_documents.rb
rboulton: I think the best approach would be to use `search_server.index_for_search(settings.search_config.content_index_names)` to first to a "get document", to check if the document exists at all and find out which index it's in.  Then use search_server.index(indexname) for the index returned in the data for that document to update the document in the right index.

This will also avoid errors being thrown by the call to `amend` if/when we get updates for documents which are excluded from the search index (for whatever reason).

lib/index_documents.rb
rboulton: Ah.  From the documentation, it looks like that will then return `nil` instead of `[]`, which will then fail when the return value is passed to `map`.

Would `message.payload.fetch("links", {}).fetch("topics", [])` be better?

bishboria: `topic_content_ids(message)` can validly return nil, depending if `links` hash is present, or if there is no `topics` field.

This seemed like the nicest way to avoid `nil["topics"]` at runtime

lib/index_documents.rb
bishboria: This was what I wanted to discuss with you: missing trello ticket information.

Should the ticket be updated and then add extra code changes to this PR?

lib/index_documents.rb
benlovell: Shame we're not using ruby 2.3 and `#dig`

lib/index_documents.rb
bishboria: I'll go with that, you know `rummager` a lot better :)

lib/index_documents.rb
rboulton: With the code as it stands, if `topic_content_ids()` returns `nil`, the call to `topic_content_ids.map` in IndexDocuments::process() will raise a `NoMethodException`.

rboulton: Ah, scratch that - you're explicitly checking for nil.

lib/index_documents.rb
```diff
@@ -0,0 +1,65 @@
+require_relative "../app"
+require 'gds_api/publishing_api_v2'
+
+class IndexDocuments
+  def process(message)
+    links = message.payload["links"].try(:dup)
+
+    if links
+      links.each do |(k, vs)|
```
rboulton: Could you use multi-word names for `k` and `vs` here to make it easier to read the code?  Perhaps `v` is `link_type` and `vs` is `content_ids`?

lib/index_documents.rb
```diff
@@ -0,0 +1,65 @@
...
+        paths = vs.map { |v| base_path(v) }.compact
```
rboulton: It would be good to sort the output for this, since the ordering we receive them in is not significant.  Sorting will ensure we get a consistent order out when we process the same document, and that in turn will ensure we don't send unnecessary changes to elasticsearch, allowing it to avoid doing work when there's no semantic change to the document.

lib/index_documents.rb
```diff
@@ -0,0 +1,65 @@
...
+        links[k] = paths
+      end
+
+      base_path = document_base_path(message)
+      update_index(base_path, links)
+    end
+
+    message.ack
+  rescue StandardError => e
```
rboulton: What would happen if we didn't do this rescue at all?

I think that errors will then be caught by the message queue consumer gem that called this method, the consumer will exit (and report the error) and then upstart will restart it and the message will be reprocessed.  This would be functionally very similar to this code, but would also trigger alerting if a message was failing too frequently (because the process would restart frequently, too).  It would also mean less code to test and maintain.

lib/index_documents.rb
```diff
@@ -0,0 +1,65 @@
...
+    indices_with_base_path(base_path).map do |index|
+      index.amend(base_path, links)
+    end
+  end
+
+  def indices_with_base_path(document_base_path)
+    indices_to_search = search_server.index_for_search(search_config.content_index_names)
+    results = indices_to_search.raw_search(query:{term:{link: document_base_path}})
+
+    indices = results["hits"]["hits"].map do |h|
```
rboulton: Perhaps rename `h` to `hit`?

lib/index_documents.rb
```diff
@@ -0,0 +1,65 @@
...
+    Airbrake.notify_or_ignore(
+      error: e,
+      payload: message.payload,
+      headers: message.headers,
+      delivery_info: message.delivery_info
+    )
+    message.retry
+  end
+
+  private
```
rboulton: Our style guide says to outdent private (ie, don't have leading space for it here).  Also that it shouldn't have a blank line following it.

lib/index_documents.rb
rboulton: We don't want to use all the link types - only the ones which come from content-tagger.  Also, the values for link types don't always match the field names in rummager (eg, the link type of `topics` maps to the `specialist_sectors` field in rummager), so I think we'll need some custom logic for each link type.

test/integration/index_documents_test.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def teardown
+    clean_test_indexes
+  end
+
+  def test_topics_get_indexed_from_publishing_api
+    m = GovukMessageQueueConsumer::MockMessage.new(payload_with_topics)
+    IndexDocuments.new.process(m)
+
+    assert m.acked?
```
rboulton: I think we need to be checking more than that the method acks the message.  Perhaps check that the appropriate payload was sent to elasticsearch?  Or that the topics from the payload were sent to elasticsearch, at least.

test/support/index_document_test_helpers.rb
```diff
@@ -0,0 +1,168 @@
...
+    stub_request(:post, "http://localhost:9200/mainstream_test/_bulk").
+      with(:body => post_to_elasticsearch).
+      to_return(:status => 200, :body => elasticsearch_post_response)
+
+    stub_request(:post, "http://localhost:9200/mainstream_test/_bulk").
+      with(:body => post_no_topics_to_elasticsearch).
+      to_return(:status => 200, :body => no_topics_elasticsearch_post_response)
+  end
+
+  def post_to_elasticsearch
```
rboulton: This method name sounds like the method performs the action of posting.  Perhaps something like `elasticsearch_post_payload`?.  Similar for the next method.

test/support/index_document_test_helpers.rb
```diff
@@ -0,0 +1,168 @@
...
+    "{\"index\":{\"_type\":\"edition\",\"_id\":\"/topic/animal-welfare/pets\"}}\n{\"popularity\":5.6085249579360626e-05,\"slug\":\"animal-welfare/pets\",\"topics\":[\"/ooo/aar\"],\"description\":\"Info about pets.\",\"format\":\"specialist_sector\",\"link\":\"/topic/animal-welfare/pets\",\"title\":\"Pets\",\"_type\":\"edition\",\"_id\":\"/topic/animal-welfare/pets\"}\n"
```
rboulton: It's quite hard to read this; perhaps it would be clearer to write this as a nested hash datastructure and call .to_json on it to produce the sample body.

The topics should be being sent to elasticsearch in the `specialist_sectors` field, not the `topics` field.  (The `topics` field contains the things which whitehall calls topics internally, but which are now called "policy areas" in the UI.)

bishboria:
Based on the number of comments made, I'll close this PR and reopen with different commits

PR #555, opened 2016-02-04
Index selected links for published documents sent to publishing api

bishboria:
Currently, changes to topics and other tags in content-tagger don't cause the search index (accessed via rummager) to be updated correspondingly. We want such changes to be reflected as soon as possible in the search index.

Part of: https://trello.com/c/lRnxI2x5/472-make-rummager-index-the-tags-in-the-links-hash

### DIFF
--- a/lib/index_documents.rb
+++ b/lib/index_documents.rb
@@ -1,0 +1,96 @@
+require_relative "../app"
+require 'gds_api/publishing_api_v2'
+
+class IndexDocuments
+  MIGRATED_TAGGING_APPS = %w{
+    businesssupportfinder
+    calculators
+    calendars
+    hmrc-manuals-api
+    licencefinder
+    policy-publisher
+    smartanswers
+  }
+
+  def process(message)
+    if publishing_app_migrated?(message)
+      raw_links = links_from_payload(message)
+
+      unless raw_links.empty?
+        links = rummager_fields_from_links(raw_links)
+        base_path = document_base_path(message)
+        update_index(base_path, links)
+      end
+    end
+
+    message.ack
+  end
+
+private
+  def publishing_app_migrated?(message)
+    MIGRATED_TAGGING_APPS.include? message.payload["publishing_app"]
+  end
+
+  def links_from_payload(message)
+    message.payload.fetch("links", {})
+  end
+
+  def rummager_fields_from_links(links)
+    results = {}
+    rummager_field_mappers.each { |field_name, mapper|
+      field_values = mapper.call(links)
+      if field_values
+        results[field_name] = field_values
+      end
+    }
+    results
+  end
+
+  def rummager_field_mappers
+    {
+      "mainstream_browse_pages" => lambda { |links| sorted_link_paths(links, "mainstream_browse_pages") },
+      "organisations" => lambda { |links| (sorted_link_paths(links, "lead_organisations") + sorted_link_paths(links, "organisations")).uniq },
+      "specialist_sectors" => lambda { |links| sorted_link_paths(links, "topics") },
+    }
+  end
+
+  def sorted_link_paths(links, link_types)
+    links.fetch(link_types, []).map { |content_id| base_path(content_id) }.compact.sort
+  end
+
+  def document_base_path(message)
+    message.payload.fetch("base_path")
+  end
+
+  def base_path(content_id)
+    publishing_api.get_content(content_id)["base_path"]
+  end
+
+  def update_index(base_path, links)
+    indices_with_base_path(base_path).map do |index|
+      index.amend(base_path, links)
+    end
+  end
+
+  def indices_with_base_path(document_base_path)
+    indices_to_search = search_server.index_for_search(search_config.content_index_names)
+    results = indices_to_search.raw_search(query: {term: {link: document_base_path}})
+
+    indices = results["hits"]["hits"].map do |hit|
+      index = Elasticsearch::Index.strip_alias_from_index_name(hit["_index"])
+      search_server.index(index) if index
+    end
+  end
+
+  def publishing_api
+    @publishing_api ||= GdsApi::PublishingApiV2.new(Plek.current.find('publishing-api'))
+  end
+
+  def search_server
+    @search_server ||= search_config.search_server
+  end
+
+  def search_config
+    @search_config ||= Rummager.settings.search_config
+  end
+end
--- a/test/integration/index_documents_test.rb
+++ b/test/integration/index_documents_test.rb
@@ -1,0 +1,131 @@
+require "integration_test_helper"
+require 'govuk_message_queue_consumer'
+require 'govuk_message_queue_consumer/test_helpers'
+require './lib/index_documents'
+
+class IndexDocumentsTest < IntegrationTest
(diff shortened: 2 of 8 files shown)


lib/index_documents.rb
```diff
@@ -0,0 +1,86 @@
...
+private
+  def links_from_payload(message)
+    message.payload["links"].try(:slice, *content_tagger_tags)
+  end
+
+  def paths_from_content_ids(content_ids)
+    content_ids.map { |content_id| base_path content_id }.compact.sort
+  end
+
+  def content_tagger_tags
```
rboulton: Since the values here are required to be the keys for `content_tagger_to_rummager_map`, perhaps this would be better implemented by calling `.keys` on that, to avoid risk of them getting out of sync.

lib/index_documents.rb
```diff
@@ -0,0 +1,86 @@
...
+  def remap_keys(hash, keymap)
+    Hash[hash.map {|k, v| [keymap[k], v] }]
+  end
+
+  def document_base_path(message)
+    message.payload["base_path"]
+  end
+
+  def base_path(content_id)
+    GdsApi::PublishingApiV2.new(Plek.current.find('publishing-api')).get_content(content_id)["base_path"]
```
rboulton: I think it would be worth extracting the call to Plek to a method and memoizing it; no need to construct a new client for every content ID we want to look up.

lib/index_documents.rb
```diff
@@ -0,0 +1,86 @@
...
+      organisations
+    }
+  end
+
+  def content_tagger_to_rummager_map
+    {
+      "topics" => "specialist_sectors",
+      "organisations" => "lead_organisations",
+      "mainstream_browse_page" => "mainstream_browse_pages",
+      "parent" => "parent",
```
rboulton: There's also no "parent" field in rummager at present - this should just be ignored for the moment.

lib/index_documents.rb
```diff
@@ -0,0 +1,86 @@
...
+      parent
+      topics
+      organisations
+    }
+  end
+
+  def content_tagger_to_rummager_map
+    {
+      "topics" => "specialist_sectors",
+      "organisations" => "lead_organisations",
```
rboulton: I don't think this is quite right; the field name in rummager is "organisations", too.  Elasticsearch does preserve the ordering of field values, though, so it might be worth ensuring that the value stored in the organisations field in elasticsearch has the "lead" organisations first, but are in alphabetical sorted order.  Something like (in pseudo code) "(organisations.sort + lead_organisations.sort).uniq".

[Aside: we might want to index the `lead_organisations` link type to a rummager field, but that would require adding a new field type in rummager, and besides nothing currently uses that, so that's beyond the scope of this PR.]

test/integration/index_documents_test.rb
```diff
@@ -0,0 +1,71 @@
...
+    stub_tagging_lookup
+    stub_calls_for_index_documents_test
+  end
+
+  def teardown
+    clean_test_indexes
+  end
+
+  def test_topics_get_indexed_from_publishing_api
+    Elasticsearch::Amender.any_instance.expects(:amend)
```
rboulton, line 21: I think this test still needs to check a bit more; it just checks that an amend call was made - I'd like it to at least check that it's trying to amend the right document, and that it's adding the appropriate tags (because I think those are more likely things to go wrong than it failing to call `amend` at all.

jamiecobbett:
Naive question: what's to prevent this from picking up placeholders and other content being migrated but not yet switched over?

rboulton:
Answer to @jamiecobbett; nothing, except they don't have anything in their links hashes for these fields.
If they do have things in their links hashes, it ought to be correct despite them being placeholders, so there'd be no problem with picking it up.

jamiecobbett:
@rboulton OK, if I understand correctly, would a more specific title for this PR be "Index _links_ for _published_ documents sent to the publishing api"?

rboulton:
@jamiecobbett Yes, that would be more accurate.  I've changed it.

test/integration/index_documents_test.rb
bishboria, line 21: I was operating under the assumption that other tests make sure that amend is called for the correct document, but fair enough, i'll add extra expectations on `amend`

lib/index_documents.rb
bishboria: just checked content in rummager, and it does seem to be `lead_orgs.sort + other_orgs.sort`

bishboria: Think it'll be best to remove `.sort` from Line 27 in the above code snippet, since the order in lists being sent appears to be correct

lib/index_documents.rb
```diff
@@ -0,0 +1,103 @@
+require_relative "../app"
+require 'gds_api/publishing_api_v2'
+
+class IndexDocuments
+  MIGRATED_TAGGING_APPS = %w{
+    businesssupportfinder
+    calculators
+    calendars
+    content-tagger
```
tijmenb: `content-tagger` doesn't belong here, because there are no content items with `content-tagger` as publishing_app.

bishboria: damn, forgot to remove that before pushing.

lib/index_documents.rb
```diff
@@ -0,0 +1,102 @@
...
+      if field_values
+        results[field_name] = field_values
+      end
+    }
+    results
+  end
+
+  def rummager_field_mappers
+    mapper = {}
+
```
rboulton: Why assign these individually, rather than just writing them as a literal hash?

lib/index_documents.rb
```diff
@@ -0,0 +1,102 @@
...
+    mapper["mainstream_browse_pages"] = lambda { |links| sorted_link_paths(links, "mainstream_browse_page") }
```
rboulton: The field in the links hash is `mainstream_browse_pages` - https://github.com/alphagov/govuk-content-schemas/blob/master/formats/base_links.json#L10

rboulton:
Apart from the comment I made about the name `mainstream_browse_page` not being the one used in the links hash, this looks correct to me.  Needs review from at least one other finding things dev.

tijmenb:
Reviewed and this looks good. I want to smoke test it in development first though, so https://github.com/alphagov/govuk-puppet/pull/4068 needs to be merged & deployed first.

MatMoore:
Given that we renamed the title of the PR, can we do the same for the task name? Currently it is `:index_documents_from_publishing_api`

Also, I think the IndexDocuments class could have a more descriptive name too. It's very similar sounding to the Indexer module and the ElasticSearch::Index class at the moment.

bishboria:
How about `:index_selected_links_from_published_documents` ?

bishboria:
Actually, can we keep a note of changing the class, rake task, etc, until after this and https://github.com/alphagov/govuk-puppet/pull/4068 are merged?

`govuk-puppet` is already setup with this rake task name, and there is an existing PR trying to fix problems connecting to Rabbitmq.

rboulton:
Approve of not doing name changes here.  Happy to merge this, on the basis that we haven't set credentials up in production yet, so even if we deployed rummager, it wouldn't start running there.

PR #556, opened 2016-02-05
Add a withdrawn field to all document types

MatMoore:
This will allow us to include withdrawn documents when we do content audits.
https://trello.com/c/RkrlGszq/473-keep-items-in-search-index-when-withdrawn-but-not-redirected

If `is_withdrawn` is set on a content item, it is not returned by searches, except if a debug flag is passed in.

When we do a basic search, the elasticsearch query is built up in parts. I've refactored the `Filter` component so that it delegates to a `UserFilter` and a `VisibilityFilter`, where `VisibilityFilter` contains a query to exclude the withdrawn content.

`UserFilter` is responsible for building up filters from the search params. There was a bunch of reusable code here that used to be private methods of this component. They all manipulate hashes that are serialised into JSON for the elasticsearch request. I've moved them out into a mixin in the elasticsearch module so I can reuse them in `Filter` and `VisibilityFilter`. If there is a better place for this let me know. I suspect there might be some duplication with the advanced search logic, which has its own builder class.

`#filter_hash` is slightly different, because this operates on another class object representing a filter that is created in the search parameter parser. I've left this as a private method of the UserFilter for now but it may make sense to move this method to the filter object itself.

I still need to have a look at advanced searches as they are handled differently.

### DIFF
--- a/docs/documents.md
+++ b/docs/documents.md
@@ -1,0 +1,38 @@
+## Rummager Documents API
+
+### `POST /:index/documents`
+
+Insert or overwrite a document.
+
+There must be a link attribute in the JSON body.
+
+Any fields which are not part of the schema for the document type you are posting
+will be silently ignored (see config/schema).
+
+#### Example request and response
+
+```json
+{
+    "organisations": [
+      "department-for-transport",
+      "driver-and-vehicle-licensing-agency"
+    ],
+    "public_timestamp": "2014-12-09T16:21:03+00:00",
+    "description": "Renew or tax your vehicle for the first time, apply online, by phone or at the Post Office",
+    "format": "transaction",
+    "link": "/vehicle-tax",
+    "mainstream_browse_pages": [
+      "driving/car-tax-discs"
+    ],
+    "title": "Tax your vehicle",
+    "_type": "edition",
+    "_id": "/vehicle-tax",
+    "specialist_sectors": []
+}
+```
+
+```json
+{
+  "result": "OK"
+}
+```
--- a/lib/elasticsearch/queries.rb
+++ b/lib/elasticsearch/queries.rb
@@ -1,0 +1,45 @@
+module Elasticsearch
+  # Mixin for building elasticsearch queries
+  module Queries
+  private
+
+    # Combine filters using an operator
+    #
+    # `filters` should be a sequence of filters. nil filters are ignored.
+    # `op` should be :and or :or
+    #
+    # If 0 non-nil filters are supplied, returns nil.  Otherwise returns the
+    # elasticsearch query required to match the filters
+    def combine_filters(filters, op)
+      filters = filters.compact
+      if filters.length == 0
+        nil
+      elsif filters.length == 1
+        filters.first
+      else
+        {op => filters}
+      end
+    end
+
+    def terms_filter(field_name, values)
+      return nil if values.size == 0
+
+      {"terms" => { field_name => values } }
+    end
+
+    def term_filter(field_name, value)
+      {"term" => { field_name => value } }
+    end
+
+    def date_filter(field_name, value)
+      {
+        "range" => {
+          field_name => {
+            "from" => value["from"].iso8601,
+            "to" => value["to"].iso8601,
+          }.reject { |_, v| v.nil? }
+        }
+      }
+    end
+  end
+end
--- a/lib/parameter_parser/search_parameter_parser.rb
+++ b/lib/parameter_parser/search_parameter_parser.rb
@@ -299,6 +299,10 @@ private
         options[:new_weighting] = true
       when "explain"
         options[:explain] = true
+      when "include_withdrawn"
+        # Withdrawn content is excluded from regular searches but is useful for
+        # content audits
+        options[:include_withdrawn] = true
       else
         @errors << %{Unknown debug option "#{option}"}
       end
--- a/lib/query_components/filter.rb
+++ b/lib/query_components/filter.rb
@@ -1,102 +1,52 @@
+require "query_components/user_filter"
+require "query_components/visibility_filter"
+require 'elasticsearch/queries'
+
 module QueryComponents
   class Filter < BaseComponent
-    def payload(excluding = [])
-      rejects = []
-      filters = []
+    include Elasticsearch::Queries
 
-      param_filters = search_params.filters.reject do |filter|
-        excluding.include?(filter.field_name)
-      end
+    def payload(excluded_field_names = [])
+      user_filter = QueryComponents::UserFilter.new(search_params)
+      visibility_filter = QueryComponents::VisibilityFilter.new(search_params)
 
-      param_filters.each do |filter|
-        if filter.reject
-          rejects << filter_hash(filter)
-        else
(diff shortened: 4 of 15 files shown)


lib/query_components/filter.rb
```diff
@@ -1,23 +1,27 @@
...
     def payload(excluding = [])
-      rejects = []
-      filters = []
+      user_filter = QueryComponents::UserFilter.new(search_params)
+      visibility_filter = QueryComponents::VisibilityFilter.new(search_params)
 
-      param_filters = search_params.filters.reject do |filter|
-        excluding.include?(filter.field_name)
-      end
+      filters = combine_filters(user_filter.filters_payload(excluding), :and)
```
jackscotti: for consistency, please rewrite this with the same style you've used in the lines below

lib/query_components/user_filter.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      field_name = filter.field_name
+      values = filter.values
+
+      case filter.type
+      when "string"
+        es_filters << terms_filter(field_name, values)
+      when "date"
+        es_filters << date_filter(field_name, values.first)
+
```
jackscotti: extra white line

docs/documents.md
```diff
@@ -0,0 +1,35 @@
...
+    "_id": "/vehicle-tax",
+    "specialist_sectors": []
+}
+```
+
+```json
+{
+  "result": "OK"
+}
+```
```
jackscotti, line 38: Might be worth to mention that if try to put a document with parameters not in declared in `base_document_type.json`, the request would still be valid and the document will be created but those parameters will be dropped.

lib/elasticsearch/queries.rb
```diff
@@ -0,0 +1,45 @@
...
+        nil
+      elsif filters.length == 1
+        filters.first
+      else
+        {op => filters}
+      end
+    end
+
+    def terms_filter(field_name, values)
+      nil if values.size == 0
```
tijmenb: Should this read `return nil if values.size == 0`?

lib/parameter_parser/search_parameter_parser.rb
```diff
@@ -299,6 +299,10 @@ def debug_options
         options[:new_weighting] = true
       when "explain"
         options[:explain] = true
+      when "include_withdrawn"
+        # Withdrawn content is excluded from regular searches but is useful for
+        # content audits
+        options[:withdrawn] = true
```
tijmenb: Perhaps we could use `include_withdrawn` for the options hash too? That way it's consistent between the URL param and debug option.

lib/query_components/visibility_filter.rb
```diff
@@ -0,0 +1,13 @@
+require 'elasticsearch/queries'
+
+module QueryComponents
+  class VisibilityFilter < BaseComponent
+    include Elasticsearch::Queries
+
+    def rejects_payload
+      return [] if search_params.debug[:withdrawn]
+
+      [term_filter("is_withdrawn", true)]
```
tijmenb: Would it be clearer to inline what `term_filter` does here? `[{ term: { is_withdrawn: true } }]` isn't that much longer and prevents some indirection.

test/unit/query_components/facets_test.rb
```diff
@@ -2,11 +2,19 @@
 require "unified_search_builder"
 
 class FacetsTest < ShouldaUnitTestCase
+  def make_search_params(facets:, filters: [])
```
tijmenb, line 5: Is it necessary to test with `withdrawn` here? Or are the tests related?

tijmenb:
:+1: very nice!

It might be worth it to add a integration test. That could be something like:

``` ruby
def test_withdrawn_content
  reset_content_indexes

  commit_document("mainstream_test",
    title: "I am the result",
    description: "This is a test search result",
    link: "/some-nice-link",
    withdrawn: true
  )

  get "/unified_search?q=test"
  assert_equal 0, parsed_response.fetch("total")
end

def test_withdrawn_content_with_flag
  reset_content_indexes

  commit_document("mainstream_test",
    title: "I am the result",
    description: "This is a test search result",
    link: "/some-nice-link",
    withdrawn: true
  )

  get "/unified_search?q=test&debug=include_withdrawn"
  assert_equal 1, parsed_response.fetch("total")
end
```

docs/documents.md
rboulton, line 38: It's not exactly if they're not in `base_document_type.json`; it's if they're not defined in the schema for the document type which has been submitted (and all schemas include the fields defined in `base_document_type` automatically).  The schema is defined by the value of the `_type` field.

As an aside, a good piece of work to do might be to make rummager report an error to errbit whenever this happens, so that we can ensure we're not silently losing fields.

lib/query_components/filter.rb
rboulton: This is a little confusing; I like that you're trying to keep it generic.  I assume that you're not including "visibility_filter" in the combine because you know it will always be empty: however, it would be even nicer to just make and array of "filter_components", and call .map on that to get the input for the combine_filters calls.

Something like:

``` ruby
filter_components = [
QueryComponents::UserFilter.new(search_params),
QueryComponents::VisibilityFilter.new(search_params),
]
filters = combine_filters(filter_components.map { |component| component.filters_payload(excluding) }, :and)
```

test/unit/query_components/facets_test.rb
MatMoore, line 5: It's not absolutely necessary, but it makes things a bit simpler. Because the expectations are against the elastic search query structure in these tests, our change results in all of them changing, unless the debug flag is set, despite it not making a difference to the end search result for most content.

Even though the new behaviour is on by default, I think it's simpler and clearer to keep the old semantics for these searches, and add new tests for the new behaviour (excluding withdrawn).

lib/query_components/filter.rb
MatMoore: @rboulton yep it is a bit confusing, might add some comments to clarify that.

Your assumption is right, I know visibility is just for restricting stuff so I didn't bother implementing an empty filter payload for that one.

I'm not sure what you are getting at in your example though. Do you just want to make the filter/reject code more similar? I could do something like what you wrote with the filter_components array just containing the one item, which would be clearer than adding extra payload methods I think...?

docs/documents.md
MatMoore, line 38: @rboulton is that work for a separate ticket?

lib/elasticsearch/queries.rb
MatMoore: Yep, will fix.

lib/query_components/filter.rb
MatMoore: We cleaned this up a bit by renaming a lot of the variables that were confusing. It's still a bit difficult to follow I think, for example in some places we can add nils that ultimately get compacted away again. But hopefully its a bit clearer now what is a query hash and what is an array of queries.

docs/documents.md
rboulton, line 38: Definitely a separate piece of work (and separate ticket, if we decide it's worth doing).

PR #557, opened 2016-02-09
Copy legacy "topics" to "policy areas"

tijmenb:
Policy areas used to be called "topics", which is an unfortunate naming collision with the concept of "topic" that lives under gov.uk/topic.

We want to rename it to "policy areas". The plan is:
- Always copy incoming topics to policy_areas
- Request policy_areas instead of topics from whitehall/search
- Use policy_areas in the payload sent to rummager
- Remove the copying of topics to policy_areas
- Remove the topics field

This commit makes rummager copy the contents of `topics` in the payload to `policy_areas`.

Trello: https://trello.com/c/bDFlehc5


rboulton:
:+1:  Happy to be merged when green.

PR #559, opened 2016-02-10
Send the bearer token to publishing-api

tijmenb:
The publishing-api expects a bearer token. This wasn't set when initialising this object.

`PUBLISHING_API_BEARER_TOKEN` is already available to rummager.

Trello: https://trello.com/c/lRnxI2x5
Follow up to: https://github.com/alphagov/rummager/pull/555


jackscotti:
:bear:

PR #561, opened 2016-02-15
Expand policy_areas

tijmenb:
The `topics` field will be renamed to `policy_areas` (see https://github.com/alphagov/rummager/pull/557). Currently we copy `topics` to `policy_areas` during the indexing.

To make rummager return the same data for both fields, we need to expand the field as well. We still have to use the `topic` format for this, because Whitehall will keep sending policy area docs with this
format name (https://github.com/alphagov/whitehall/blob/master/app/models/classifica
tion.rb#L7-L12).

https://trello.com/c/bDFlehc5

### DIFF
--- a/lib/registries.rb
+++ b/lib/registries.rb
@@ -10,6 +10,13 @@ class Registries < Struct.new(:search_server, :search_config)
       organisations: organisations,
       specialist_sectors: specialist_sectors,
       topics: registry_for_document_format('topic'),
+
+      # Whitehall has a thing called `topic`, which is being renamed to "policy
+      # area", because there already are seven things called "topic". Until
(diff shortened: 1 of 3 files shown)


lib/registries.rb
```diff
@@ -10,6 +10,8 @@ def as_hash
       organisations: organisations,
       specialist_sectors: specialist_sectors,
       topics: registry_for_document_format('topic'),
+      # Note that the `topic` format hasn't been renamed.
```
malcolmbaig: Can we elaborate here on why that is?

tijmenb: I elaborated elaborately.

PR #562, opened 2016-02-16
Message queue: add better logging and error handling

tijmenb:
See individual commits for details.

Related to: https://trello.com/c/GRt7NSbL

### DIFF
--- a/lib/index_documents.rb
+++ b/lib/index_documents.rb
@@ -69,7 +69,11 @@ private
 
   def update_index(base_path, links)
     indices_with_base_path(base_path).map do |index|
-      index.amend(base_path, links)
+      begin
+        index.amend(base_path, links)
+      rescue Elasticsearch::DocumentNotFound => e
+        Airbrake.notify_or_ignore(e)
+      end
     end
   end
 
--- a/lib/index_documents.rb
+++ b/lib/index_documents.rb
@@ -14,6 +14,7 @@ class IndexDocuments
   }
 
   def process(message)
+    $stdout.puts "Processing message: #{message.payload.inspect}"
     index_links_from_message(message)
     message.ack
   end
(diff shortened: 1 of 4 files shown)


Davidslv:
:+1:

lib/index_documents.rb
```diff
@@ -14,20 +14,28 @@ class IndexDocuments
...
-      end
-    end
-
+    $stdout.puts "Processing message: #{message.payload.inspect}"
+    index_links_from_message(message)
     message.ack
+  rescue GdsApi::HTTPServerError => e
+    $stderr.puts "An error occurred!"
+    $stderr.puts e.inspect
+    message.retry
```
bishboria, line 23: This has been added back? @rboulton told me to remove it and allow the message queue gem to handle this sort of thing.

tijmenb, line 23: Which part are you talking about?

bishboria, line 23: Refering to that entire rescue section: https://github.com/alphagov/rummager/pull/554#discussion-diff-51092802

tijmenb, line 23: Aha, I see. Thanks! Couple of questions:
- The problem is that the worker doesn't seem to restart after it exits. I've had to manually do `sudo service rummager-publishing-api-document-indexer-procfile-worker start`.
- @rboulton writes "but would also trigger alerting if a message was failing too frequently (because the process would restart frequently, too)". This alerting hasn't happened as far as I know.
- My reasoning was that since `HTTPServerError` (5XX) is "expected" because publishing-api is sometimes down, we should handle that explicitly. Wdyt?
- Doesn't relying on upstart cause the exception-raising message to always be in front of the queue? Retry-ing will put it at the back of the queue, so that new messages coming in can still be processed. Is this correct?

Perhaps we should have a quick discussion with @rboulton about failure states, logging and monitoring in the queue. I've had a lot of trouble getting visibility on the problems in the last week.

PR #563, opened 2016-02-17
Ignore non-existing links in links hash

tijmenb:
In certain situations, the content_ids found in the links hash will not exist in the publishing-api. This is correct behaviour. When this happens, we should ignore the item.

One of the valid cases I found is collections-publisher. For any parent topic, this puts the children in the links hash (`children` key):

https://github.com/alphagov/collections-publisher/blob/7846746aa4b5e62139c7e199c841c938515c18e3/app/presenters/topic_presenter.rb#L25

But these may not have been sent to the publishing-api, thus causing this issue.

Trello: https://trello.com/c/lRnxI2x5

### DIFF
--- a/lib/indexer/links_lookup.rb
+++ b/lib/indexer/links_lookup.rb
@@ -28,6 +28,7 @@ module Indexer
     end
 
   private
+
     def rummager_field_mappers
       {
         "mainstream_browse_pages" => lambda { |links| sorted_link_paths(links, "mainstream_browse_pages") },
@@ -41,7 +42,10 @@ module Indexer
     end
 
     def base_path(content_id)
-      publishing_api.get_content(content_id)["base_path"]
+      publishing_api.get_content!(content_id)["base_path"]
+    rescue GdsApi::HTTPNotFound
+      # Content items in the links hash may not exist yet.
+      nil
     end
 
     def publishing_api
--- a/lib/indexer/links_lookup.rb
+++ b/lib/indexer/links_lookup.rb
@@ -1,0 +1,54 @@
+require 'gds_api/publishing_api_v2'
+
+module Indexer
+  class LinksLookup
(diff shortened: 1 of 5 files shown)


lib/indexer/links_lookup.rb
```diff
@@ -0,0 +1,57 @@
...
+        "specialist_sectors" => lambda { |links| sorted_link_paths(links, "topics") },
+      }
+    end
+
+    def sorted_link_paths(links, link_types)
+      links.fetch(link_types, []).map { |content_id| base_path(content_id) }.compact.sort
+    end
+
+    def base_path(content_id)
+      content_item = publishing_api.get_content!(content_id)["base_path"]
```
benlovell: Is making this assignment of any use here?

test/unit/indexer/links_lookup_test.rb
```diff
@@ -0,0 +1,40 @@
...
+      stub_request(:get, "http://publishing-api.dev.gov.uk/v2/content/b6b7d71f-ecf3-4ff0-8fee-f19041cbe6b5").
+        to_return(body: { base_path: "/hela" }.to_json)
+
+      rummager_links = rummager_links_for({
+        "topics" => ["b6b7d71f-ecf3-4ff0-8fee-f19041cbe6b5"]
+      })
+
+      assert_equal rummager_links, {
+        "mainstream_browse_pages" => [],
+        "organisations"=>[],
```
benlovell: Missing space around the hash rocket here.

tijmenb:
Updated, now 100% govuk-lint compliant :)

MatMoore:
This looks good to me. :+1: Will this impact the frontend in any way or is it just a case of having less information in the index to filter on etc?

tijmenb:
This actually won't have any effect - currently the indexer crashes when it encounters a non-existing `content_id`.

PR #564, opened 2016-02-17
Add govuk-lint

tijmenb:
This PR adds govuk-lint to this repo and fixes most violations. Violations that haven't been fixed are in [our local rubocop.yml](https://github.com/alphagov/rummager/blob/95f828ea25114b92f369e5e93d77249850df2e70/.rubocop.yml), where we also specify the reason why we don't want them enabled.


malcolmbaig:
That's one comprehensive list of commits. LGTM! :+1:

PR #566, opened 2016-02-22
Also index tags from contacts-admin

tijmenb:
contacts-admin has been migrated since https://github.com/alphagov/contacts-admin/pull/231.

Trello: https://trello.com/c/afa0Mfa4


Davidslv:
:+1:

PR #567, opened 2016-02-22
Fix which value is indexed from message queue

tijmenb:
Currently `IndexDocuments` indexes the entire base_path of the topic, organisation or browse page. This should be the slug only.

https://trello.com/c/lRnxI2x5


bishboria:
:+1:

PR #568, opened 2016-02-23
Fix removal of taggings

tijmenb:
Currently the indexer will not touch the taggings of documents if the key (eg `organisations`) is not in the message queue payload. This causes a problem when the links are intentionally empty because the
links have been cleared.

The  makes sure that we always clear tags if they're not sent by the publishing-api.

https://trello.com/c/lRnxI2x5


bishboria:
:+1:

PR #571, opened 2016-02-26
[WIP - do not merge] Rake tasks for snapshot restore

MatMoore:
- Start snapshot
- Monitor snapshot
- List snapshots in bucket
- Get latest snapshot
- Restore a named snapshot
- Monitor a restore
- Switch aliases to a restored index

The ongoing snapshot process will be managed from a cron or jenkins job,
and we plan to add a jenkins job for restoring as well.

In addition to this we already have a nightly job that rebuilds the indexes,
which will be extended to clean up old snapshots.

Elasticsearch's APIs don't handle multiple snapshot/restore operations at
the same time. If we are unable to take a snapshot for this reason, we will just
let the job the fail. Snapshot freshness will be monitored by icinga.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ gem "gds-api-adapters", "~> 28.1.0"
 gem "rack-logstasher", "0.0.3"
 gem 'airbrake', '4.0.0'
 gem "unf", "0.1.3"
+gem 'aws-sdk', '~> 2'
 
 group :test do
   gem "shoulda-context"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -14,6 +14,12 @@ GEM
     ast (2.2.0)
     astrolabe (1.3.1)
       parser (~> 2.2)
+    aws-sdk (2.2.19)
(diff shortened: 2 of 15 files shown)


MatMoore:
Pairing with @jackscotti 

We still need to sort out the lint and do some refactoring, but these tasks will form the core of the snapshot/restore process we're working on. @rboulton @tijmenb would appreciate any feedback you have on the work so far.

jackscotti:
As a side note, I would like to mention the fact that we would have wanted to use the official Elasticsearch gem. It would have made code much simpler and easier to read/understand. Sadly we could not use it because of namespace clashes with the internal files of this repository. 

https://github.com/elastic/elasticsearch-ruby

MatMoore:
:point_up: I think if we do a refactor to separate out the indexing code from the search code, that would be a great time to introduce this gem. It would allow us to remove the API client class and a lot of code that interfaces with it.

MatMoore:
Closing as we are going to continue on a new branch, picking up #573

PR #572, opened 2016-02-29
Allow misspelled ID codes to return relevant results

Davidslv:
Often people search for ID codes (eg forms or reports) which may or may not contain spaces or punctuation. They only see the relevant results when they use exactly same spelling as the content, this often frustrates users for not getting the results they are looking for.

This commit aims to solve this problem without changing too much of the existing weight system, by applying a match query on `all_searchable_text.id_codes` fields.

Related with: https://trello.com/c/lrYj5PdF/516-enable-searching-for-id-codes-with-or-without-spaces-punctuation

### DIFF
--- a/lib/parameter_parser/search_parameter_parser.rb
+++ b/lib/parameter_parser/search_parameter_parser.rb
@@ -289,7 +289,8 @@ private
     debug_options = character_separated_param("debug")
 
     options = {}
-    debug_options.each { |option|
+
+    debug_options.each do |option|
       case option
       when ""
       when "disable_best_bets"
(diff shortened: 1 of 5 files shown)


tijmenb:
This feature is most easily tested behind a feature flag (so @tarastockford can compare the results). Can you make it so that this new query is used only in for `debug=use_id_codes`?

Davidslv:
That makes sense @tijmenb, thanks

tijmenb:
This PR needs some rebasing after https://github.com/alphagov/rummager/pull/573.

Davidslv:
@tijmenb just rebased

tijmenb:
I'm gonna merge this to make testing easier. This is fine because the behaviour change is behind a feature flag anyway.

PR #573, opened 2016-02-29
Split elasticsearch/ into parts

MatMoore:
The elasticsearch module is a holdover from the times when we supported solr as
well. Now it is a big catch all namespace for a lot of different things.

This PR separates it out into parts (see further comments below).

@tijmenb @rboulton Does the new structure make sense to you? If this is a good step forward, we can rework the namespaces to match, and rebase our snapshot/restore on top of this for https://trello.com/c/YAJY2zIl/359-implement-snapshot-restore-to-backup-elasticsearch-indexes

The aim of this refactor is to add some structure to the elasticsearch/ code and facilitate moving to the elasticsearch gem, since we plan to make a lot of changes to the indexing workflow in the future and the official gem makes a lot of things simpler to implement. (This came up before in https://trello.com/c/PqLodfBo/47-fix-elasticsearch-load-balancing)


MatMoore:
The test directories also need moving if we go ahead with this.

MatMoore:
Closing for now, going to take another shot at it this afternoon.

MatMoore:
After discussing with @tijmenb yesterday I decided to redo this. We decided that a large part of the former Elasticsearch module should belong to the indexer, and another chunk to the unified search part. This split is something we've started already, but a lot of classes have been left in the root directory/namespace.

Although not strictly necessary to do right now, I'm using the opportunity to
- Organise those classes living directly under `lib`
- Rename `unified_search` -> `search` in the code.
- Group together presenters for search, and workers for the indexer.

I think the split between search/indexer is fairly good as I've tried to avoid moving anything that couples those two areas of the code.

```
gds0264:rummager matmoore$ grep -rl 'Search::' . | egrep -v '^\./(test|lib\/search)/'
./app.rb
./lib/index.rb
./lib/legacy_search/advanced_search.rb
./lib/legacy_search/advanced_search_query_builder.rb
./lib/searcher.rb
gds0264:rummager matmoore$ grep -rl 'Indexer::' . | egrep -v '^\./(test|lib\/indexer)/'
./bin/bulk_load
./lib/index.rb
./lib/tasks/indices.rake
./lib/tasks/message_queue.rake
```

What remains of the Elasticsearch module is what I view as the "model" code: SearchServer, IndexGroup, Index, SearchConfig. This is used all over the place at the moment, although it may change significantly when we introduce the elasticsearch gem.

`legacy_client` can be removed when the elasticsearch gem is introduced. 

`legacy_search` is the old advanced search functionality, which I wanted to separate from the new search namespace.

I moved `query_components` under search as well, but I'm not renamespacing this as I think nesting them lexically hurts readability and I want to keep it simple. If anyone has differing opinions on this I'm happy to rework it.

In this PR I don't intend to make any API changes or structural changes beyond moving and renaming constants.

I plan to move some tests around after this to reflect the new structure. Tests are all passing but I expect to do a lot of manual regression testing on integration, as everything is affected.

@tijmenb @rboulton @mobaig

tijmenb:
Really happy with this. Makes a lot of sense and will make it easier for new developers.

The only thing I'm not a 100% on is the `Search` namespace because the indexing side is called `Indexer` and it may be harder to grep. Would naming it `Searcher` be better?

MatMoore:
I'm fine with either, or something else. I think both are an improvement on unified search.

I guess the only problem with Searcher is there is there is already a Searcher class, which currently lives outside of that namespace (in contrast to indexer.rb, which just requires stuff).

Maybe we could move this to Searcher::IndexSearcher or something like that? Otherwise it could be a bit confusing as Searcher isn't the thing you use to do the searching.

tijmenb:
Yeah, `searcher.rb` should also go into the new namespace. `Searcher::Search` would make sense, but `Search::Searcher` would also be fine :)

MatMoore:
Cool, I've moved this to Search::Searcher.

malcolmbaig:
Nicely done. 

Some suggestions below for further renaming - the intention here being clarity and succinctness. What do you think?

`Search::Searcher` => `Search::Query` (maybe with a renamed instance method e.g. - `Search::Query.new(search_params).run`)
`Search::Queries (module)` => `Search::QueryHelpers` or `Search::ElasticsearchHelpers`
`Search::SearchBuilder` => `Search::QueryBuilder`
`Search::SearchPresenter` => `Search::ResultPresenter`
`Search::FacetExampleFetcher` => `Search::FacetExample`
`Search::SearchParameters` => `Search::Parameters`
`Indexer (module)` => `Index`
`Indexer::IndexQueue` => `Index::Queue`
`Indexer::IndexDocuments` => `Index::Documents`

MatMoore:
Thanks, those are good suggestions.

I've implemented the search ones, but left the indexer ones for now, because we still have some classes like "Index" and "IndexGroup" that could cause confusion. They are not part of what I'd consider "indexer" code at the moment since they are involved in other things like rake tasks, and advanced search. I expect we will rework this in the future so it might be good to revisit the naming another time.

malcolmbaig:
:+1: :ship: :rocket:

tijmenb:
:+1: massive improvement! :balloon:

rboulton:
:clap:

PR #574, opened 2016-03-02
Split integration tests into appropriate directories

tijmenb:
In https://github.com/alphagov/rummager/pull/573 we've reorganised this repo. This is a follow up to move the integration tests into separate directories.

@MatMoore @jackscotti


MatMoore:
:+1:

PR #576, opened 2016-03-02
Use the correct document_id when indexing docs

tijmenb:
Currently this indexer assumes that a document's `_id` and `link` are the same. This causes tag updates for the `contacts` app to fail because they aren't the same (the _id has no leading slash).

Best reviewed per-commit, as some refactoring was necessary to prevent this from happening again.

https://trello.com/c/lRnxI2x5 https://trello.com/c/afa0Mfa4

### DIFF
--- a/lib/indexer/amender.rb
+++ b/lib/indexer/amender.rb
@@ -8,7 +8,7 @@ module Indexer
 
     def amend(document_id, updates)
       if updates.include?("link")
-        raise ArgumentError, "Cannot change document the `link` attribute of a document."
+        raise ArgumentError, "Cannot change the `link` attribute of a document."
       end
 
       document = index.get_document_by_id(document_id)
--- a/lib/indexer/amender.rb
+++ b/lib/indexer/amender.rb
@@ -6,15 +6,16 @@ module Indexer
       @index = index
     end
 
-    def amend(link, updates)
+    def amend(document_id, updates)
       if updates.include?("link")
         raise ArgumentError, "Cannot change document the `link` attribute of a document."
       end
 
-      document = index.get(link)
+      document = index.get_document_by_id(document_id)
 
       unless document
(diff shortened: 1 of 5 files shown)


rboulton:
Would it be better to fix this by making the `contacts` app send data with a consistent value for the link and _id?  (Following the convention is for other apps, which is to make the _id and link both have a leading slash.)

In order to migrate to this, you could make rummager add a leading slash to any supplied document with an "_id" field with no leading slash (unless it started with `http:` or `https:`) when indexing; then the nightly index migration would update all the _id fields to have a leading slash, and you could update contacts accordingly at a later deploy.

It's probably best to try and use the _id field as the unique identifier, since that's the only reason we have it!

tijmenb:
The reason I don't want to fix this in contacts-admin is that I think the migration process might be error prone, and we don't gain that much. If we start indexing from the message queue we will remove the indexing code from contacts-admin anyway, removing the problem entirely. We'll probably use `content_id` for the new `id`s there.

lib/indexer/amender.rb
```diff
@@ -6,15 +6,16 @@ def initialize(index)
       @index = index
     end
 
-    def amend(link, updates)
+    def amend(document_id, updates)
       if updates.include?("link")
         raise ArgumentError, "Cannot change document the `link` attribute of a document."
```
Davidslv: I found this error message confusing
Does it means "Cannot change `link` attribute of a document." ?

Davidslv:
It looks good to me given your concerns and explanation.

lib/indexer/amender.rb
jackscotti: Yes, this needs rewrite.

jackscotti:
Minor comment, but happy to merge.

PR #578, opened 2016-03-04
Introduce elasticsearch gem and rake tasks for snapshot restore

MatMoore:
- Start snapshot
- Monitor snapshot
- List snapshots in bucket
- Get latest snapshot
- Restore a named snapshot
- Monitor a restore
- Switch aliases to a restored index

The ongoing snapshot process will be managed from a cron or jenkins job,
and we plan to add a jenkins job for restoring as well.
We always create repository before snapshotting (this is idempotent), so settings
get updated when keys are rotated.

In addition to this we already have a nightly job that rebuilds the
indexes, which will be extended to clean up old snapshots.

Elasticsearch's APIs don't handle multiple snapshot/restore operations
at the same time. If we are unable to take a snapshot for this reason, we
will just let the job fail.

Ultimately, we will ensure we always have a useable snapshot by using
snapshots to refresh staging/integration and monitoring this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,8 @@ gem "gds-api-adapters", "~> 29.4"
 gem "rack-logstasher", "0.0.3"
 gem 'airbrake', '4.0.0'
 gem "unf", "0.1.3"
+gem 'aws-sdk', '~> 2.2.19'
+gem 'elasticsearch', '~> 1.0.15'
 
 group :test do
   gem "shoulda-context"
--- a/Rakefile
+++ b/Rakefile
@@ -47,6 +47,10 @@ def search_server
   search_config.search_server
 end
 
+def elasticsearch_uri
+  SearchConfig.new.elasticsearch["base_uri"]
+end
+
 def index_names
   case ENV["RUMMAGER_INDEX"]
   when "all"
--- a/docs/snapshot-restore.md
+++ b/docs/snapshot-restore.md
@@ -1,0 +1,144 @@
+# Snapshot and Restore Tasks
+
+Rummager defines several rake tasks to assist in
+
+  * restoring the search indexes to an earlier state
+  * rebuilding indexes with updated popularity data
+  * mirroring the search indexes between environments
+
+This is built on top of elasticsearch's snapshot/restore functionality,
+documented at <https://www.elastic.co/guide/en/elasticsearch/reference/1.4/modules-snapshots.html>
+
+## How snapshot/restore works in elasticsearch
+
+The Elasticsearch API allows us to create snapshots of one or more
+indexes.
+
+The snapshots live in a `repository`, which maps to some remote data store,
+such as a location in the file system, or in our case an AWS S3 bucket.
+
+Each snapshot is incremental, so successive snapshots don't necessarily use
+up extra storage, but there will be redundancy in the stored data, due to things
+like elasticsearch merging shards over time, or us rebuilding an index entirely.
+Because of this, we regularly delete old snapshots.
+
+### Index aliases
+
+Aliases are like symbolic links to one or more indices, and can be changed
+atomically.
+
+Rummager uses aliases to avoid downtime when restoring a snapshot or rebuilding
+an index.
+
+For example, during a restore operation, we follow the following steps:
+
+1. `mainstream` -> `mainstream-2016-02-25t17:07:13z-8efbb505-759a-4a34-ae0e-85aad065932d`
+2. Restore mainstream from a snapshot to a new index, `restored-mainstream-2016-02-25t17:11:04z-6d2a7dc4-a3b7-4572-b388-c59dc2c40be5`
+3. Wait for the restore to complete. Searches will continue to use the old
+index.
+3. Switch the alias: `mainstream` -> `restored-mainstream-2016-02-25t17:11:04z-6d2a7dc4-a3b7-4572-b388-c59dc2c40be5`
+4. Searches now use the new index.
+
+## Repository setup
+
+If you have an S3 bucket set up, you can create a snapshot repository yourself
+by running:
+
+```
+rake rummager:snapshot:create_repository
+```
+
+This assumes you have some environment variables available:
+
+- AWS_ACCESS_KEY_ID
+- AWS_ACCESS_SECRET_ACCESS_KEY
+- AWS_BUCKET_NAME
+- AWS_BUCKET_REGION
+
+The name of the bucket to be used is configured in `elasticsearch.yml`.
+
+If you want to create your own repository (e.g. for development), you might
+want to create a filesystem backed repository instead. This is documented in
+the [elasticsearch documentation for snapshot restore](https://www.elastic.co/guide/en/elasticsearch/reference/1.4/modules-snapshots.html).
+
+You won't be able to use the `rummager:snapshot:list` and `rummager:snapshot:latest`
+tasks without an S3 bucket, but the basic snapshot/restore can work with any
+repository type.
+
+## Start a snapshot
+Normally, snapshots are triggered regularly from a cron job.
+
+You can initiate a snapshot manually by running:
+
+```shell
+rake rummager:snapshot:run[repository_name,snapshot_name]
+```
+
+As above, you will need to ensure that AWS environment variables are set.
+
+If snapshot_name is left out, Rummager generates one based on the current time.
+
+If the repository does not exist, or elasticsearch is unable to process the
+request, the task exits with a non zero code and prints a stack trace.
+
+Otherwise the exit code will be zero.
+
+## Monitoring snapshots
+### Check if a snapshot has completed yet
+```shell
+rake rummager:snapshot:check[snapshot_name]
+```
+This returns the status of a snapshot operation, e.g. "SUCCESS".
+
+### Find the last successful snapshot
+```shell
+# Print the name of the latest snapshot
+rake rummager:snapshot:latest
+
+# Print the last snapshot excluding any after Jan 1st 2016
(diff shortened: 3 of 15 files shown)


MatMoore:
The main rake task here is `rummager:snapshot:run`. This will be called by a cron job - I'll raise a PR against puppet for this shortly.

The other rake tasks demonstrate what we can do with the snapshots. These could be triggered from jenkins or fabric.

We still need to extend the nightly jenkins job to clean up old snapshots.

Open questions:
- Is there a simple way to configure the bucket name per environment in rummager? 
- Is there a process for taking snapshots from the S3 bucket and transferring it to a separate environment? The work so far only really addresses disaster recovery.

Gemfile
```diff
@@ -18,6 +18,8 @@ gem "gds-api-adapters", "~> 28.1.0"
 gem "rack-logstasher", "0.0.3"
 gem 'airbrake', '4.0.0'
 gem "unf", "0.1.3"
+gem 'aws-sdk', '~> 2'
+gem 'elasticsearch', '~> 1.0', '>= 1.0.15'
```
tijmenb: Would it be better to pin these to something `X.X.X`?

Rakefile
```diff
@@ -47,6 +47,14 @@ def search_server
   search_config.search_server
 end
 
+def bucket
```
tijmenb: Can these be moved to `snapshot.rake`?

docs/snapshot-restore.md
```diff
@@ -0,0 +1,134 @@
...
+3. Switch the alias: `mainstream` -> `restored-mainstream-2016-02-25t17:11:04z-6d2a7dc4-a3b7-4572-b388-c59dc2c40be5`
+4. Searches now use the new index.
+
+## Repository setup
+
+If you have an S3 bucket set up, you can create a snapshot repository yourself
+by running:
+
+```
+rake rummager:snapshot:create_repository[repository_name]
```
tijmenb: Where is the `repository_name` defined?

jackscotti:
:+1: reaaally good and great doc! Let's wait for someone that hasn't work on this to review it.

elasticsearch.yml
```diff
@@ -4,6 +4,7 @@ auxiliary_index_names: ["page-traffic", "metasearch"]
 registry_index: "government"
 metasearch_index_name: "metasearch"
 popularity_rank_offset: 10
+snapshot_bucket_name: "govuk-api-elasticsearch-snapshots-integration"
```
tijmenb: Should this also be in the env variables?

lib/index_group.rb
```diff
@@ -21,6 +21,13 @@ def initialize(base_uri, name, schema, search_config)
       @search_config = search_config
     end
 
+    def find_index(real_name)
+      # TODO: validate this against the API to make sure it actually exists
```
tijmenb: How do we validate this?

lib/snapshot/snapshot_repository.rb
```diff
@@ -0,0 +1,134 @@
...
+      snapshots = @client.status(repository: repository_name)["snapshots"]
+      snapshots.reject! { |snapshot| snapshot.fetch("state") == "SUCCESS" }
+      snapshots.map { |snapshot| snapshot.fetch("snapshot") }
+    end
+
+    def last_successful_snapshot(snapshots)
+      (snapshots - in_progress_snapshots).last
+    end
+
+    def restore_indexes(snapshot_name, index_groups)
```
tijmenb: Feels like this method could be nicely extracted into a `Snapshot::Restore` class. WDYT?

docs/snapshot-restore.md
```diff
@@ -0,0 +1,134 @@
...
+```
+Prints "true" or "false", depending on whether the restore is complete.
+This uses the elasticsearch recovery API.
+
+## Switch an index alias to a restored index
+
+Once an index has been successfully restored from a snapshot, you can switch
+over the alias:
+```shell
+INDEX_NAMES=mainstream rake rummager:switch_to_named_index[new_index_name]
```
tijmenb, line 138: Where do we get `new_index_name` from?

lib/snapshot/bucket.rb
```diff
@@ -0,0 +1,35 @@
+require 'time'
```
tijmenb: Is this `require` needed?

lib/snapshot/snapshot_repository.rb
```diff
@@ -0,0 +1,134 @@
...
+        "Snapshot #{snapshot['snapshot']}: #{snapshot['state']}"
+      )
+      logger.debug(snapshot["shards_stats"].to_s)
+
+      snapshot["state"]
+    end
+
+    def in_progress_snapshots
+      snapshots = @client.status(repository: repository_name)["snapshots"]
+      snapshots.reject! { |snapshot| snapshot.fetch("state") == "SUCCESS" }
```
tijmenb: Perhaps it's better to chain these calls, to avoid mixing mutating / non-mutating methods.

test/unit/snapshot/snapshot_repository_test.rb
```diff
@@ -0,0 +1,30 @@
+require "test_helper"
+require "snapshot/snapshot_repository"
+require "minitest/mock"
```
tijmenb: This is not necessary?

Rakefile
MatMoore: The bucket definitely can.

elasticsearch_uri is also used by indices.rake, because I've added a new index related task that uses the elasticsearch client.

I put the implementation in index.rb since it's related to the other functionality there, although maybe there is somewhere else to put this. The existing code always accesses that url indirectly via the SearchServer instance, but for this we don't need to instantiate the whole hierarchy of SearchServer, IndexGroup, Index for anything else, so it was easier to just make it a class method and pass in the url config.

docs/snapshot-restore.md
MatMoore: It's arbitrary. The actual value we use will be defined in puppet (PR to follow...)

I guess this could be made more obvious though. Is there a better way to document these rake calls?

tijmenb: In that case, would it be possible to have a default and make this argument optional?

Rakefile
tijmenb: I'm not sure but could we have `Snapshot::SnapshotRepository` access `SearchConfig.new.elasticsearch["base_uri"]` directly? That would stop us from having to use `elasticsearch_uri` in the rake tasks/initialiser altogether.

docs/snapshot-restore.md
MatMoore: Absolutely. I might come back to this though because I think there is still some uncertainty on whether we are going to create a new repository every day (in which case we may include a timestamp) or just keep one repo and periodically delete snapshots within it. Plan to look at this next.

elasticsearch.yml
MatMoore: Yes. Yes it should. We will have to make a corresponding change to the puppet for this, but we can do that after.

lib/index_group.rb
MatMoore: Oops, I forgot about this. I was thinking of issuing a basic get request, since the API will return 404 if the index doesn't exist. It will do that anyway if you try do actually do anything with this index, but thought the behaviour was a bit surprising.

Will take another look.

lib/snapshot/snapshot_repository.rb
MatMoore: I think so. `#select_index_from_groups` `#guess_restored_indices` and `#restore_indexes` are all part of the restore process so I think we can move these to a helper object.

I was toying with the idea of having a Snapshot as a separate thing as well but don't think that would help here.

MatMoore: I've extracted this to a Restorer class, and I've added tests for the horrific regex (this was based on an existing regex in index.rb but we weren't able to reuse that one easily, due to running regexes outside of ruby).

Think its a bit cleaner now. See latest commit.

docs/snapshot-restore.md
MatMoore, line 138: It comes from the previous restore task. Added a sentence above to clarify this.

lib/snapshot/bucket.rb
```diff
@@ -0,0 +1,33 @@
...
+          bucket: @bucket_name,
+          prefix: prefix
+      }).contents
+
+      all_snapshots.select! { |snapshot| snapshot.last_modified <= before_time }
+
+      all_snapshots.sort_by!(&:last_modified)
+
+      all_snapshots.map do |snapshot|
+        snapshot.key.gsub(prefix, "")
```
Davidslv: use sub instead since you just expecting to find 1 `prefix`

lib/snapshot/snapshot_repository.rb
```diff
@@ -0,0 +1,163 @@
...
+      indices.select do |index|
+        index_groups.any? do |index_group|
+          index.include?(index_group)
+        end
+      end
+    end
+  end
+
+  class Restorer
+    RENAME_PATTERN = '(restored-)*(.+)-\d{4}-\d{2}-\d{2}t\d{2}:\d{2}:\d{2}z-[0-9a-f][-0-9a-f]*\Z'
```
Davidslv: This could be done in a multiline and commented for the next person who comes into the code.
Also freeze the constant, constants in ruby can still be changed.

jackscotti:
@mobaig - @tijmenb the PR is ready to be reviewed :grin:

malcolmbaig:
LGTM. Happy to merge once rebased.

MatMoore:
@mobaig Rebased now

tijmenb:
:rocket:

PR #579, opened 2016-03-07
Correct lint

MatMoore:
Remove the unnecessary level from the inheritance hierarchy. See https://github.com/bbatsov/ruby-style-guide#no-extend-struct-new


jackscotti:
:+1:

PR #580, opened 2016-03-07
Rename `allowed_values`

tijmenb:
`allowed_values` is misnamed because it implies that this performs some kind of validation of input, which it doesn't.

`expanded_search_result_fields` is more descriptive, because its purpose is to replace certain values in the search results.

Paired with @mobaig on the new naming.


jackscotti:
:+1:

PR #581, opened 2016-03-09
Improve error handling in message queue consumer

tijmenb:
When consuming from the message queue we've encountered two new edge-cases. 

The first one is that the document on the queue doesn't exist in the search index (which can happen, not all pages are in search). In this case we handle the error gracefully but send a Airbrake message for us to inspect. Once we've established that the pages are missing from search legitimately we'll remove that error notification.

The second is where a document is tagged to a access-limited document. When we request the item from the publishing-api it will return a `401 Unauthorised`. In this case we'll just ignore the linked document.

https://trello.com/c/lRnxI2x5


bishboria:
:+1:

PR #593, opened 2016-03-18
Don't index "gone" documents

tijmenb:
These documents are generally not in the search index. We can ignore them when indexing the links.

This will help in reducing the load in elasticsearch when indexing a bunch of documents.

https://trello.com/c/0FKSqAkK


MatMoore:
:+1:

PR #595, opened 2016-03-29
Update Rummager gems Part 1

bishboria:
The gems for Rummager are almost all out of date. Upgrading will make the app more secure (and perhaps faster) - https://trello.com/c/1mXjOji6/565-upgrade-rummager-dependencies

The gems updated here are the simplest to do first since they require
the least amount of work to keep tests passing.


jackscotti:
:+1:

tijmenb:
If there's no reason to pin the gem to a specific version, you can use the pessimistic version operator: `~> X.X.X` (https://github.com/alphagov/styleguides/blob/master/ruby.md#gemfile).

PR #597, opened 2016-03-30
Update simplest project gems (without Unicorn)

bishboria:
https://trello.com/c/1mXjOji6/565-upgrade-rummager-dependencies

The gems updated here are the simplest to do first since they require the least amount of work to keep tests passing.

The previous attempt had problems with Unicorn, so that'll be upgraded in the next PR.


Gemfile
```diff
@@ -5,25 +5,25 @@ gem "sinatra", "1.3.4"
...
-gem 'nokogiri', "1.5.5"
+gem "logging", "2.1.0"
+gem 'nokogiri', "1.6.7.2"
 gem 'whenever', require: false
-gem "slop", "3.4.5"
-gem "sidekiq", "2.13.0"
+gem "slop", "4.3.0"
+gem "sidekiq", "4.1.1"
 gem "sidekiq-statsd", "0.1.5"
 # pin to version that includes security vulnerability fix
```
tijmenb, line 14: This comment is probably out-of-date now, can you remove it?

bishboria, line 14: oops, yep, I missed that comment. If it's not needed in the next PR, it'll be gone.

tijmenb, line 14: It refers to `redis-namespace`, which is now updated so it's definitely not needed anymore.

PR #599, opened 2016-03-31
Fix suggestion health check

tijmenb:
The search suggestion health check measures how good rummager is in suggesting spelling corrections to the user. 

A rummager API change months ago (https://github.com/alphagov/rummager/pull/515) caused the health check to stop working.

Fixes: https://trello.com/c/wae6TIdZ


bishboria:
:+1:

PR #600, opened 2016-04-01
Upgrade minor gems and sidekiq < 3

bishboria:
Rummager gems are out of date and require upgrading. Sidekiq requires
2 major version bumps, and has changed its format from the version
we currently run. This change introduces changes suggested in
https://github.com/mperham/sidekiq/blob/master/3.0-Upgrade.md

Part of https://trello.com/c/1mXjOji6/565-upgrade-rummager-dependencies


jackscotti:
After what happened with the other upgrades, have you tried deploying this branch to integration and checked for error on Errbit?

tijmenb:
Perhaps it would be good to take the opportunity to introduce pessimistic versioning for the gems you are updating now.

Gemfile
```diff
@@ -5,20 +5,20 @@ gem "sinatra", "1.3.4"
...
 gem "rest-client", "1.8.0"
-gem "logging", "1.8.1"
-gem 'nokogiri', "1.5.5"
+gem "logging", "2.1.0"
+gem 'nokogiri', "1.6.7.2"
 gem 'whenever', require: false
-gem "slop", "3.4.5"
-gem "sidekiq", "2.13.0"
+gem "slop", "4.3.0"
+gem "sidekiq", "< 3"
```
tijmenb: As you can see in the `Gemfile.lock` (https://github.com/alphagov/rummager/blob/upgrade-gems-part-1/Gemfile.lock#L140) this hasn't actually updated the sidekiq gem because the current version `2.13` still satisfies this new constraint (`< 3`).

Gemfile
```diff
@@ -5,20 +5,20 @@ gem "sinatra", "1.3.4"
...
 gem "rack", "~> 1.6"
 gem "rest-client", "1.8.0"
-gem "logging", "1.8.1"
-gem 'nokogiri', "1.5.5"
+gem "logging", "2.1.0"
+gem 'nokogiri', "1.6.7.2"
 gem 'whenever', require: false
-gem "slop", "3.4.5"
-gem "sidekiq", "2.13.0"
+gem "slop", "4.3.0"
```
tijmenb: Contrary to the PR title you're doing a major upgrade here from `3` to `4`.

Slop is the option parser for things in `bin`. A quick attempt to run something results in an error, so I suspect this is a breaking update.

```
vagrant@development:~/govuk/rummager$ bundle exec bin/health_check --type=suggestions
bin/health_check:34:in `block in run!': undefined method `banner' for #<HealthCheckCLI:0x007f33a9044820> (NoMethodError)
```

Gemfile
bishboria: FFS, `bundle update sidekiq` didn't work...

bishboria:
Tests pass locally, `bin/health_check…` works, unified search via the browser returns sensible returns: no errors on the VM. If this still doesn't work on integration, i give up.

Gemfile
tijmenb: It does work - bundler will resolve things the Gemfile to any "valid" version - there is no guarantee that it will take the latest. If you want a specific version you need to tell it to use that specific version.

The latest 2.X from https://rubygems.org/gems/sidekiq/versions is `2.17.8` so you'll need to put that in the Gemfile.

MatMoore:
I think we should test the rummager:migrate-index rake task as well. I'm not sure how well the tests cover the indexing code at the moment.

bishboria:
`RUMMAGER_INDEX=all bundle exec rake rummager:migrate_index` seems to be working fine

MatMoore:
:+1: I think we can merge this as long as @tijmenb is happy.

tijmenb:
I am always happy!

PR #601, opened 2016-04-01
Don't send `InvalidQuery` exceptions to Errbit

tijmenb:
This exception is triggered by a user error and should not be sent to errbit. Currently it is, causing unnecessary emails to developers and error-blindness.

Ignores errors like:

https://errbit.publishing.service.gov.uk/apps/539849ba0da115be230008b7/problems/56fd109165786342b80e1a00

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -79,7 +79,7 @@ class Rummager < Sinatra::Application
     halt(503, "Redis queue timed out")
   end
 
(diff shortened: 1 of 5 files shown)


tijmenb:
cc @MatMoore

MatMoore:
:+1:

PR #602, opened 2016-04-01
Rescue all exceptions in message queue worker

tijmenb:
During development of the message queue processor we've experienced multiple times that some unknown exception in the code is triggered.

Currently this will crash the worker, after which it is restarted by `upstart`. This sometimes happens very quickly, causing the message queue processor to try the same thing many times, causing other load
troubles.

This commit makes the consumer more resilient by rescuing any errors and waiting one second before trying again. It also reports the error to Airbrake itself so that we have access to the offending payload.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -83,7 +83,7 @@ class Rummager < Sinatra::Application
     halt(422, env['sinatra.error'].message)
   end
 
-  error SearchIndices::BulkIndexFailure do
+  error Indexer::BulkIndexFailure do
     halt(500, env['sinatra.error'].message)
   end
 
--- a/lib/index.rb
+++ b/lib/index.rb
@@ -18,15 +18,6 @@ module SearchIndices
   class DocumentNotFound < RuntimeError; end
   class IndexLocked < RuntimeError; end
 
-  class BulkIndexFailure < RuntimeError
-    attr_reader :failed_keys
-
-    def initialize(failed_items)
-      super "Failed inserts: #{failed_items.map { |id, error| "#{id} (#{error})" }.join(', ')}"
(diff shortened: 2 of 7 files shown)


MatMoore:
Do we have a way of limiting number of retries in govuk_message_queue_consumer? Or could this retry indefinitely?

tijmenb:
No, currently (and after this PR) messages will be retried indefinitely. What's more, `message.retry` will put the message back (or leave it) at the front of the queue, so if an unrecoverable error happens the queue will block because we have a concurrency of 1:

https://github.com/alphagov/govuk_message_queue_consumer/blob/dc0acfdf09a9260ed4728ae7be142b839ebfd545/lib/govuk_message_queue_consumer/consumer.rb#L12

This PR will at least prevent the consumer from overloading elasticsearch in the case of a recurring error. This in combination with the better Airbrake error and queue-monitoring ("Monitor rummager's current queue size from graphite" - https://trello.com/c/MC4qYvTV) should alert us when this is happening.

A longer term solution to this would be a "errored" queue which we can monitor.

MatMoore:
In that case is it better to discard by default and only handle known cases with retries? (e.g. connectivity errors)

tijmenb:
Good point, but

tijmenb:
I'm not sure - until we have the dead letter exchange (https://www.rabbitmq.com/dlx.html) I think it's preferable to block & fix unknown cases (or let them retry until success). I'd rather not drop the messages. WDYT?

malcolmbaig:
LG, particularly as we've planned to follow this up with better reporting/alerts via statsd and Icinga.

PR #603, opened 2016-04-01
Stop logging entire message body in consumer

tijmenb:
The consumer currently logs too much data, which makes monitoring difficult. This commit restricts that to only data we need.

### DIFF
--- a/lib/indexer/index_documents.rb
+++ b/lib/indexer/index_documents.rb
@@ -25,9 +25,10 @@ module Indexer
     }
 
     def process(message)
-      $stdout.puts "Processing message: #{message.payload.inspect}"
-      index_links_from_message(message.payload)
-      message.ack
+      with_logging(message) do
+        index_links_from_message(message.payload)
+        message.ack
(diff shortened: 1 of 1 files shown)


lib/indexer/index_documents.rb
```diff
@@ -25,7 +25,7 @@ class IndexDocuments
     }
 
     def process(message)
-      $stdout.puts "Processing message: #{message.payload.inspect}"
+      $stdout.puts "Processing message: #{message.payload.slice("content_id", "base_path", "document_type", "title", "update_type", "publishing_app")}"
```
malcolmbaig: Can drop the `$stdout`, just extra line noise.

Also maybe multiline this with a heredoc, for readability?

tijmenb:
@mobaig fair point. I've updated the PR to also add a finish log line and refactored a bit. 

Note that we should fix the logging in rummager and stop using `puts` altogether.

malcolmbaig:
:+1:

PR #604, opened 2016-04-01
Message queue: only listen to link changes

tijmenb:
The publishing-api sends a message when the content item changes: on a `publish` call and a `patch links` call.

It sends the message with a "routing key" in the form of "format.update_type". The update type can be major, minor or republish (when the content has changed), or `links` when somebody patches the
links.

Rummager is only interested in the changing links, so we can restrict the consumer to listen to a specific pattern. You can read more about it here:

https://www.rabbitmq.com/tutorials/tutorial-five-python.html

By restricting this rummager will have to do less work, because less messages will come in.


malcolmbaig:
:shipit:

PR #606, opened 2016-04-04
Bump `sinatra` to v1.4.7

bishboria:
Changes in `sinatra` means that, eg, `settings` becomes `Rummager.settings`.

Leave `tilt`, and small bump for `rack-protection`: bumping these two further cause test failures. That will be fixed in the next PR.

The following work on the development vm:
- all tests pass
- `RUMMAGER_INDEX=all bundle exec rake rummager:migrate_index` works
- `bundle exec bin/health_check --type=suggestions` runs
- `http://rummager.dev.gov.uk/unified_search.json?q=child%20benefit` returns sensible output

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,9 @@
 source "https://rubygems.org"
 
 gem "unicorn", "4.6.2"
-gem "sinatra", "1.3.4"
+gem "tilt", "1.3.3"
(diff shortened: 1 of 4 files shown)


MatMoore:
:confetti_ball:

tijmenb:
This is great! :tada: :+1:

PR #608, opened 2016-04-04
Load entire app environment into workers

tijmenb:
Since https://github.com/alphagov/rummager/pull/606 the sidekiq workers have been failing because it doesn't have the `Rummager` class loaded.

This commit makes the message queue workers require the exact same code in the message queue workers as in server-mode. This makes the code a somewhat easier to follow, it replicates the behaviour we are used to in Rails and prevents scenarios like this where tests are passing because different code is loaded in test environment than in production.

Part of: https://trello.com/c/1mXjOji6


tijmenb:
cc @bishboria

tijmenb:
I've deployed this branch to integration:

https://deploy.integration.publishing.service.gov.uk/job/Deploy_App/2671/console

PR #610, opened 2016-04-05
Fix error reporting to Errbit

tijmenb:
Currently the exception messages for `BulkIndexFailure` and `FailedJobException` are constructed with a very specific error message string. This causes Airbrake to think the errors are unique, and it
will email the admins thousands of exceptions if something goes wrong.

This commit makes the errors always have the same `message` so that is nicely grouped in Errbit. Any additional information is sent in the `parameters`.


bishboria:
@tijmenb this needs a rebase…

tijmenb:
Somehow this PR got in a weird state. It should only contain https://github.com/alphagov/rummager/pull/610/commits/5f66dfcd8ae3d18ffdb5d777c334cd6a8eb5b40f, but now contains other history as well after a force-push to rebase by @bishboria.

All relevant commits seem to be on master however.

PR #612, opened 2016-04-05
Bump `sidekiq` to `3.5.4`

bishboria:
_Do not merge straight away_

Ensure `sidekiq` has no work to retry, see https://github.com/mperham/sidekiq/blob/master/3.0-Upgrade.md

Part of: https://trello.com/c/1mXjOji6/565-upgrade-rummager-dependencies


Gemfile
```diff
@@ -9,7 +9,7 @@ gem "logging", "2.1.0"
 gem 'nokogiri', "1.6.7.2"
 gem 'whenever', require: false
 gem "slop", "3.4.5"
-gem "sidekiq", "< 3"
+gem "sidekiq", "< 4"
```
Davidslv: shouldn't we lock it to `3.4.1` instead? what's the chances of someone upgrading to 3.5.4 ?

bishboria: When this PR goes in, another should be raised to upgrade `sidekiq`:
- https://github.com/mperham/sidekiq/blob/master/3.0-Upgrade.md
- https://github.com/mperham/sidekiq/blob/master/4.0-Upgrade.md

The `< X` is recommended in those upgrade documents too. I think when the last sidekiq bump goes in, it should lock the version

tijmenb:
This can't be merged before we've deployed the previous version to production.

tijmenb:
Why `3.4.1` and not `3.5.4`?

bishboria:
Where does this version `3.5.4` come from? Sidekiq doesn't have that version.

tijmenb:
https://rubygems.org/gems/sidekiq/versions

bishboria:
Bundler picks version `3.4.1` itself. Manually setting `sidekiq` to `3.5.4` gets a compatibility error with `redis-namespace`.

tijmenb:
So why is `redis-namespace` still locked?

bishboria:
I was only looking at sidekiq just now :)

tijmenb:
`redis-namespace` is only in there for sidekiq, we don't use it for something else if I'm not mistaking. We could just remove it and let sidekiq handle that dependency.

bishboria:
Yeah, that's what I'm doing now. Making sure it's all running on the dev vm first.

h-lame:
Note that sidekiq 4+ drops the redis-namespace dependency, so if we are using it for sidekiq it might be better to leave it in as an explicit dep to avoid breakages when you go up to sidekiq 4.

tijmenb:
Ah, thanks @h-lame! In that case, setting `"redis-namespace", "~> 1.5.2"` would make sense.

tijmenb:
This is now mergable because we deployed Rummager to production on Friday.

PR #613, opened 2016-04-05
Sync tagging in rummager with content-tagger

Davidslv:
Follows the approach used to blacklist applications that are not yet
migrated, as seen in [content-tagger](https://github.com/alphagov/content-tagger/pull/52/files)

Part of:
https://trello.com/c/tZd23gW5/568-allow-tagging-in-content-tagger-for-apps-that-can-t-be-tagged-yet-s

### DIFF
--- a/lib/indexer/index_documents.rb
+++ b/lib/indexer/index_documents.rb
@@ -3,20 +3,13 @@ require 'indexer/links_lookup'
 
 module Indexer
   class IndexDocuments
-    MIGRATED_TAGGING_APPS = %w{
-      businesssupportfinder
-      calculators
-      calendars
-      collections-publisher
-      contacts
-      contacts-admin
-      hmrc-manuals-api
-      licencefinder
-      policy-publisher
-      smartanswers
-      tariff
-      travel-advice-publisher
-    }
+    NON_MIGRATED_APPS = %w(
+      publisher
+      service-manual-publisher
+      specialist-publisher
+      whitehall
+      non-migrated-app
+    ).freeze
 
     EXCLUDED_FORMATS = %{
       email_alert_signup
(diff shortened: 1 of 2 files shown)


lib/indexer/index_documents.rb
```diff
@@ -3,20 +3,13 @@
...
-      smartanswers
-      tariff
-      travel-advice-publisher
-    }
+    NON_MIGRATED_APPS = %w(
+      publisher
+      service-manual-publisher
+      specialist-publisher
+      whitehall
+      non-migrated-app
```
Davidslv, line 11: I was not sure what `non-migrated-app` is meant to, so I left it there.

tijmenb, line 11: It is used in content-tagger to provide a dummy to use in the tests, so you don't have to specify an _actual_ unmigrated app, like you're doing here: https://github.com/alphagov/rummager/pull/613/files?diff=unified#diff-b9b8b16fea186bffa8dcaec1391a4fafR72.

If you use `non-migrated-app` there, we will be able to remove publisher from the list and have the tests still work.

lib/indexer/index_documents.rb
```diff
@@ -82,7 +75,7 @@ def update_document_in_search_index(document, links)
     end
 
     def should_index_document?(content_item)
-      MIGRATED_TAGGING_APPS.include?(content_item.fetch("publishing_app")) &&
+      !NON_MIGRATED_APPS.include?(content_item.fetch("publishing_app")) &&
```
tijmenb: This is now pretty hard to understand probably. Perhaps we could flip the naming (`should_skip_document?`) and simplify the conditional?

tijmenb:
:+1: very nice!

PR #614, opened 2016-04-05
Update not pinned versions

jackscotti:
Update several gems and pin down their version.

Ticket: https://trello.com/c/1mXjOji6/565-upgrade-rummager-dependencies


tijmenb:
Can we start pinning those versions now?

jackscotti:
@tijmenb very good point! Thanks for raising it.

PR #618, opened 2016-04-07
Add stats for messages from the pubilshing-api message queue.

benhyland:
This gives us stats for messages rummager is processing
from rabbitmq in addition to the existing stats on the
sidekiq workers.

Not appropriate for unit testing, so in the absence of acceptance testing for now we checked stats were arriving as expected with netcat: `nc -l -u 8125`

Paired with @tijmenb

See https://trello.com/c/Ipl9UVsA/578-monitor-rummager-s-current-queue-size-with-statsd-graphite-m


bishboria:
👍

PR #619, opened 2016-04-07
Weigh down HMRC manuals results

jackscotti:
We should reduce HMRC manuals relevancy score, so that they don't appear above more relevant results.
Ticket: https://trello.com/c/fF7yDgRL/585-weighting-of-hmrc-manuals-in-search


jackscotti:
This has been tested by @tarastockford and @hollygarrett. Good to go. @tijmenb

tijmenb:
:+1: excellent!

PR #622, opened 2016-04-13
Upgrade Rummager dependencies

Davidslv:
This is as far as we can go without having problems with dependencies.

Upgrading `active_support` from version `4.0.0` onwards causes problems with `sidekiq-statsd`

```
ERROR -- : There was an error while trying to load the gem 'sidekiq-statsd'. (Bundler::GemRequireError)
```

Upgrading ruby to `2.3.0` seems to have activated something on `active_support` that required me to install `test-unit` gem.

Upgrading `minitest` past version `4.7.5` will be a story of it's own, as the [API changed and many things happen](https://github.com/seattlerb/minitest/blob/master/History.rdoc#500--2013-05-10)

That will also break `minitest-colorize` usage, but that's a good thing.

I've discovered also that to solve `ActiveSupport::Deprecation::MethodWrapper`
we need to add `require "active_support/deprecation"` in `Rakefile`, but it's not needed in this PR.

Part of: https://trello.com/c/1mXjOji6/565-upgrade-rummager-dependencies


tijmenb:
> Upgrading ruby to 2.3.0 seems to have activated something on active_support that required me to install test-unit gem.

Can you explain this? What was the error you were getting?

tijmenb:
> I've discovered also that to solve ActiveSupport::Deprecation::MethodWrapper we need to add require "active_support/deprecation" in Rakefile, but it's not needed in this PR.

This confuses me - what do we do with this info?

tijmenb:
> Upgrading minitest past version 4.7.5 will be a story of it's own, as the API changed and many things happen. That will also break minitest-colorize usage, but that's a good thing.

Fair enough about minitest, but why is a breakage of minitest-colorize "a good thing"? That's very cryptic :)

Gemfile
```diff
@@ -39,12 +39,13 @@ gem "govuk_message_queue_consumer", "~> 2.1.0"
 gem "govuk-lint", "~> 1.0.0"
 
 group :test do
+  gem "minitest", "4.7.5"
+  gem "minitest-colorize"
+  gem 'test-unit'
```
tijmenb: Can you use double quotes here?

Gemfile
```diff
@@ -39,12 +39,13 @@ gem "govuk_message_queue_consumer", "~> 2.1.0"
 gem "govuk-lint", "~> 1.0.0"
 
 group :test do
+  gem "minitest", "4.7.5"
+  gem "minitest-colorize"
```
tijmenb: Can you use pessimistic versioning for these?

https://robots.thoughtbot.com/rubys-pessimistic-operator

Davidslv:
> Upgrading ruby to 2.3.0 seems to have activated something on active_support that required me to install test-unit gem.
> 
> Can you explain this? What was the error you were getting?

When running `bundle exec rake` this happens:

```
/opt/rubies/2.3.0/lib/ruby/gems/2.3.0/gems/activesupport-3.2.22.2/lib/active_support/test_case.rb:2:in `require': Please add test-unit gem to your Gemfile: `gem 'test-unit', '~> 3.0'` (cannot load such file -- test/unit/testcase) (LoadError)
```

> I've discovered also that to solve ActiveSupport::Deprecation::MethodWrapper we need to add require "active_support/deprecation" in Rakefile, but it's not needed in this PR.
> 
> This confuses me - what do we do with this info?

We don't need to do anything now, you can ignore that comment, that was my attempt to keep that information for the next person that tries to update ActiveSupport knows about it (although I've added that on Trello)

> > Upgrading minitest past version 4.7.5 will be a story of it's own, as the API changed and many things happen. That will also break minitest-colorize usage, but that's a good thing.
> 
> Fair enough about minitest, but why is a breakage of minitest-colorize "a good thing"? That's very cryptic :)

That's a good thing because there is [minitest-reporters](https://github.com/kern/minitest-reporters) which is a maintained gem.

h-lame:
> > > Upgrading ruby to 2.3.0 seems to have activated something on active_support that required me to install test-unit gem.
> > 
> > Can you explain this? What was the error you were getting?
> > When running bundle exec rake this happens:
> 
> ```
> /opt/rubies/2.3.0/lib/ruby/gems/2.3.0/gems/activesupport-3.2.22.2/lib/active_support/test_case.rb:2:in `require': Please add test-unit gem to your Gemfile: `gem 'test-unit', '~> 3.0'` (cannot load such file -- test/unit/testcase) (LoadError)
> ```

Ruby 2.0 switched from shipping test-unit in the stdlib to shipping minitest.  To avoid pain for developers though it added some shims that translated the test-unit classes into the minitest ones so that you didn't need to know about it.  Ruby 2.2 stopped shipping minitest and with it the API translation layer.

Rails 3.2 uses the test-unit API for testing, not the minitest one, so now we're on ruby 2.3 we don't have the translation layer and need to install both minitest and test-unit.  There is [a gem version of the translation layer](https://github.com/iGEL/test-unit-minitest) that might be a better idea than having both modern versions of minitest and test-unit in the gemfile as I can't really be sure how they'd interact with each other, but I've never used it so can't really say if it's good or not.  It surprises me that we'd need minitest _and_ test-unit for this stuff to work, but at the same time, rails 3.2 and supporting gems were written a long time before the current situation and won't have anticipated ruby 2.2 dropping a testing library entirely from stdlib.

Davidslv:
Hi @h-lame thanks for great the explanation, I'm going to give it a try to the minitest wrapper.

Thank you 👍

Gemfile
```diff
@@ -39,9 +39,8 @@ gem "govuk_message_queue_consumer", "~> 2.1.0"
 gem "govuk-lint", "~> 1.0.0"
 
 group :test do
-  gem "minitest", "~> 4.7.5"
+  gem 'test-unit-minitest', require: nil
```
tijmenb: Double quotes and no `require`s please :) See https://github.com/alphagov/rummager/pull/620 for the why's.

Davidslv:
Thank you, I've addressed the comments made.

Davidslv:
@tijmenb I've address the comments we discussed offline.

Thank you

Gemfile.lock
```diff
@@ -229,9 +227,12 @@ DEPENDENCIES
...
   statsd-ruby (~> 1.3.0)
-  test-unit
+  test-unit-minitest (~> 0.9.1)
   timecop (= 0.8.0)
   unf (= 0.1.4)
   unicorn (= 5.1.0)
   webmock (~> 1.24)
   whenever (~> 0.9.4)
+
+BUNDLED WITH
```
jackscotti, line 237: You should not commit this.

tijmenb, line 237: I thought we should...

https://github.com/search?q=org%3Aalphagov+%22BUNDLED+WITH%22&type=Code

tijmenb, line 237: Ok, after reading up on this, we really think should keep it:

> Here’s the long explanation: Bundler 1.10 adds the BUNDLED WITH section to the lock, while Bundler 1.9 removes it. When a team includes some developers on 1.9 or older, and some developers on 1.10 or higher, it’s easy to end up in a situation with commits changing the lock back and forth. The simplest way to stop that problem is to upgrade to 1.10 by running gem install bundler. **When Bundler 1.10 or higher sees a BUNDLED WITH section, it will leave it in the file.**

(Emphasis mine)

http://bundler.io/blog/2015/06/24/version-1-10-released.html

Anyhow as we discussed earlier, you should never cherry-pick from `Gemfile.lock`.

What are the reasons you wouldn't want this committed @jackscotti?

Davidslv, line 237: Just in case, I've removed it when I saw @jackscotti comment. so please don't merge it until I put that back

jackscotti, line 237: I said that because last time I committed it someone told me to remove it. Did not know of these new implications, so please, keep it.

Davidslv:
Ok, I've added `BUNDLED WITH` back again on `Gemfile.lock`

tijmenb:
:ship:

PR #629, opened 2016-04-19
Always use Sidekiq in development

tijmenb:
Rummager currently does not run Sidekiq in development, but has special code paths for a synchronous and a queued mode.

This commit changes that behaviour to always use Sidekiq. A corresponding change to the `github.gds/development` (https://github.gds/gds/development/pull/317) will make sure that in development we automatically start the worker. This will keep the development experience the same as it is now.

The main upside is that we can simplify the code and have less configs. It also makes the development environment look more like production, which is good.

The second commit uses the new setup to refactor the way jobs are enqueued. It inlines the code that performs the Sidekiq-queueing. This:
- Removes some indirection and improves readability
- Decreases the coupling between the `Index` class and the `Indexer` module
- Removes some responsibility from the `Index` class, making it less god-like

Commits are best reviewed individually.


jackscotti:
👍

PR #632, opened 2016-04-25
[Do not merge] Rework tagging sync

tijmenb:
This changes the way tags are indexed by Rummager.
### Context

Currently, every time we send a document to elasticsearch, the document is first "prepared" by the [DocumentPreparer](https://github.com/alphagov/rummager/blob/42f30f37a9a81b5a341af549ee3ec9af4544bd1d/lib/indexer/document_preparer.rb) class. That adds some extra information to the document (like popularity) and copies some fields (`format` to `_type`). It also looks up the tags for the document in the [content-api](https://github.com/alphagov/govuk_content_api).

For example, when indexing the payload for

https://www.gov.uk/state-pension-age

it will look up which tags have been added to this artefact:

https://www.gov.uk/api/content/state-pension-age

(At the time of writing, this is the "State Pension" browse page)
### Changes

1) This PR changes the datasource, so that rummager looks up the tags in the publishing-api rather than the contentapi. This has been made possible by work to make publishing-api the accurate source of all publishing-api tags.

2) It replaces way we index things from the message queue. Currently there's an incomplete implementation (which has been switched off on all environments and [currently isn't running](https://github.com/alphagov/govuk-puppet/blob/2fdf64b8e6bbcafa4b44ec6c70111cd36121677a/modules/govuk/manifests/apps/rummager.pp#L57)), which consumes messages from the publishing-api message queue and indexes the payloads.

This message queue consumer has been changed to treat new messages as a "change notification". Whenever the message queue consumer receives a new message, we simply re-add the existing document to the index, forcing a new tag-lookup by the `DocumentPreparer`, as described above.

This gives us a guarantee that we always index the latest version of the links. If we were to index the links from the queue directly (like the current implementation does) we would run the risk of indexing stale data if the delivery of the message had been delayed.

The trade-off here is that we have to make an extra request to the publishing-api for data that has already been sent to rummager in the message queue payload.
### For review
- [TagLookup](https://github.com/alphagov/rummager/blob/rework-message-queue/lib/indexer/tag_lookup.rb) is called by the DocumentPreparer and does the fetching of links from the publishing-api.
- The [MessageProcessor](https://github.com/alphagov/rummager/blob/rework-message-queue/lib/indexer/message_processor.rb) is responsible for receiving the message from the queue, and reporting back to RabbitMQ. The [ChangeNotificationProcessor](https://github.com/alphagov/rummager/blob/rework-message-queue/lib/indexer/change_notification_processor.rb) takes the actual message payload and triggers a re-index of the document.

Trello: https://trello.com/c/RrN7B6xQ
### Deploy
- [ ] We need to be sure that the links in the publishing-api are accurate (https://trello.com/c/qXmMfTOZ)
- [ ] Panopticon changes to remove tagging interface should be deployed (https://trello.com/c/4vU2VrC0)


lib/indexer/tag_lookup.rb
```diff
@@ -11,43 +13,54 @@ def prepare_tags(doc_hash)
...
       doc_hash
     end
+
+    # This lookup works because each link type in `links` contains items with
+    # format specialist sector, mainstream browse page or organisation. We have
+    # ensured that they have a content_id by updating whitehall &
+    # collections-publisher accordingly.
+    def content_ids_to_slugs(content_ids)
+      content_ids.map do |content_id|
+        result = client.raw_search(query: { term: { content_id: content_id } }, size: 1)
```
MatMoore: Is the assumption here that the target of a tag will always be indexed in rummager before the documents tagged to it?

If so, this seems like it could add complexity if we were to ever rebuild an index from scratch. Not that we do this at the moment, but it might be something we need to do later down the line, particularly if we get rid of the separate government/mainstream/etc indexes.

MatMoore:
The general approach here looks good to me. I'm not sure I have a complete understanding of the "change notification" change though: IIRC this minimises race conditions in the case where there are extra publishing api writes between the time a change is first written to publishing api and the time it is written to ES.

I think there is still a (much less likely) race condition within rummager itself since multiple processes consume from the queue and may process messages out of order - @benhyland commented on this on the trello ticket. I assume this will go into a separate PR though?

Also - do we have a separate ticket for sorting out the inconsistency with lead/supporting organisations in the publishing/frontend apps? We should unify this sooner rather than later as it adds extra complexity to everything that touches the tags.

lib/indexer/message_processor.rb
```diff
@@ -0,0 +1,53 @@
...
+
+# The consumer is currently using the message queue as a notification
+# mechanism. When it gets a message that some content has changed, it simply
+# re-indexes a document, which (among other things) triggers a new lookup of
+# the links from the publishing-api.
+module Indexer
+  class MessageProcessor
+    def process(message)
+      with_logging(message) do
+        Indexer::ChangeNotificationProcessor.new(message.payload).trigger
```
benhyland, line 18: Any particular reason to create a new CNP for each message? Seems wasteful in allocations for no significant readability improvement. Admittedly this is a drop in the ocean compared to other areas of the code, and we haven't got perf tests yet to show any difference.

lib/indexer/change_notification_processor.rb
```diff
@@ -0,0 +1,54 @@
...
+    def should_skip_document?
+      EXCLUDED_FORMATS.include?(content_item.fetch("document_type"))
+    end
+
+    def search_server
+      @search_server ||= search_config.search_server
+    end
+
+    def search_config
+      @search_config ||= Rummager.settings.search_config
```
benhyland: Can we avoid coupling the entire config object to low-level code? Just pass in the bits needed. Also, prefer to pass things in rather than depending on a global.

lib/indexer/tag_lookup.rb
```diff
@@ -11,43 +13,54 @@ def prepare_tags(doc_hash)
...
+        result.dig("hits", "hits", 0, "_source", "slug")
+      end
+      # TODO: crash properly if this returns nil
+    end
+
+    def client
+      @client ||= Rummager.settings.search_config.search_server.index_for_search(
```
benhyland: Can we pass in the client for looser coupling and easier testing?

I understand we don't do this with Services.foo because our gem provides a nice-ish dsl for testing already, and the pattern is well established.

lib/indexer/tag_lookup.rb
benhyland: Can we retrieve the content ids from publishing-api instead of from rummager?

benhyland:
Somebody should pair with @tijmenb on the remainder IMO.

lib/indexer/message_processor.rb
tijmenb, line 18: It makes for a cleaner implementation and a nicer separation of concerns (handling a message queue message vs doing actual work). Probably worth an extra allocation indeed.

benhyland, line 18: The separation of concern and implementation can still be achieved without allocating per-message, but no worries.

lib/indexer/tag_lookup.rb
tijmenb: Yeah, that is the assumption. Initially I had this asking the publishing-api like @benhyland suggests, but a naive implementation of that would cause hundreds of thousands extra requests.

I considered using a cached version of the data, but that seemed like extra complexity. Perhaps I'm wrong about that, making the indexing rely on itself is quite odd.

I'll investigate an alternative.

benhyland: We should consider modifying the publishing-api itself if it doesn't provide a sufficiently non-chatty interface. As @tijmenb points out re the cache, other options will tend to increase complexity here.

tijmenb: What is a "sufficiently non-chatty interface"?

benhyland: Chatty vs non-chatty - common but not ubiquitous phrase to describe the granularity of api requests available. A chatty interface is one which provides very small, modular functionality, so is usually more flexible but requires more calls to get something useful done. A non-chatty interface provides fewer, higher level calls which are enough to do something useful individually.
The design tradeoff involves complexity of client, complexity of service, flexibility, and invocation overhead.
Here, aiui, we use
- one api call to get the content id for a slug
- one api call to get the links for a content id
- some undetermined mechanism to get the slugs for the link ids
- all per-content-item

The minimal non-chatty interface would be
- one batch call to get the content id, link ids and link slugs for several item slugs simultaneously

But of course there is a design continuum. Knowing nothing about the api in question, my first suggestion would be to add the link slugs to the payload returned from the get_links lookup.

lib/indexer/change_notification_processor.rb
MatMoore: `search_config` is just a subset of the application settings, but I agree with your larger point.

The [law of demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) seems relevant here - there are multiple places where we do repeated lookups just to get at the real dependencies, so objects tend to have a bit too much knowledge about how the rest of the application is structured. In this case we really just depend on
- An Index to write to
- An alias representing another Index to read from

I'm not a huge fan of the way some of the classes fit together at the moment (index objects are part of a huge hierarchy that belongs to a config object) but we could revisit this when we strip out all the LegacyClient stuff in favour of the official Elasticsearch. We should really do this ASAP as it also allows to simplify our infrastructure by removing redundant load balancing.

lib/indexer/message_processor.rb
tijmenb, line 18: Can you show how you would do this in this case?

benhyland, line 18: Just make trigger a self method

lib/indexer/change_notification_processor.rb
```diff
@@ -0,0 +1,54 @@
...
+    rescue KeyError => e
+      raise Indexer::MalformedMessage, "Content item attribute missing. #{e.message}"
+    end
+
+  private
+
+    attr_reader :content_item
+
+    def document
+      @document ||= begin
```
jackscotti, line 29: Could this logic be wrapped in a method rather than in a `begin` block?

lib/indexer/message_processor.rb
jackscotti, line 18: Agree with Ben, `.trigger` could be a class method. Not so fussed about its current state though.

tijmenb:
Closing to keep our todo-list clear, see Trello for update:

https://trello.com/c/RrN7B6xQ/593-rework-the-way-rummager-indexes-links

lib/indexer/tag_lookup.rb
tijmenb: This part has now been updated to use the new expanded links endpoint.

lib/indexer/change_notification_processor.rb
tijmenb: Yeah, the way this indexe things fits together is horrible. I've extracted the index selecting into a little class to hide the ugliness: https://github.com/alphagov/rummager/pull/632/commits/d71c51aaf171ef58c528bfa54073f89cf04425b6

This is not a good solution, but at least it moves the problem away a bit. A real solution would involve refactoring the config/index in tandem with the elasticsearch-client stuff, like @MatMoore says.

lib/indexer/change_notification_processor.rb
tijmenb, line 29: @jackscotti any reason to prefer that over a `begin` block?

MatMoore, line 29: I think if we extracted this it would basically be the same method as `document` anyway, minus the memoisation. It's not something I care too much about when calling it.

lib/index_finder.rb
```diff
@@ -0,0 +1,20 @@
+# Hides some of the dirtyness in selecting/generating the correct index objects
+class IndexFinder
```
MatMoore, line 2: 👍  Extracting this makes a lot of sense.

General ruby style question: When a class is not meant to be instantiated, is there any advantage to using a `class` rather than a `module` for defining your methods on?

lib/indexer/tag_lookup.rb
```diff
@@ -30,12 +30,16 @@ def find_content_id(doc_hash)
       if doc_hash["content_id"].present?
         doc_hash["content_id"]
       else
-        Services.publishing_api.lookup_content_id(base_path: doc_hash["link"])
+        GdsApi.with_retries do
```
MatMoore: Can we explicitly specify the number of retries here? Otherwise you can't see what it will do by just glancing at the code.

Gemfile
```diff
@@ -28,7 +28,7 @@ gem "redis-namespace", "~> 1.5.2"
 gem "statsd-ruby", "~> 1.3.0"
 
 gem "plek", "~> 1.12"
-gem "gds-api-adapters", "~> 30.0"
+gem "gds-api-adapters", github: "alphagov/gds-api-adapters"
```
Davidslv: This looks like a dangerous move to me, if someone merges something to master that later reveals to be broken (not caught by the tests) we might have problems.
Or even if something goes to master a few seconds before we deploy, we might have some unexpected behaviour.

tijmenb: We'll release a new version before this.

lib/services.rb
```diff
@@ -23,3 +11,31 @@ def self.publishing_api
...
+  def self.with_retries(maximum_number_of_retries = 4)
+    attempts = 0
+    maximum_number_of_attempts = maximum_number_of_retries + 1
+
+    begin
+      attempts += 1
+      yield
+    rescue Timeout::Error, GdsApi::TimedOutException => e
+      raise e if attempts >= maximum_number_of_attempts
+      sleep sleep_time_after_attempt(attempts)
```
Davidslv, line 25: What you think of moving `sleep` into `sleep_time_after_attempt`?

tijmenb, line 25: We were planning to add this to gds-api-adapters, and we had that as a separate method to make it more easily testable.

benhyland, line 25: @tijmenb glad to see this - separating effects from calculations is a good habit to get into.

malcolmbaig:
I wonder if an ADR might be useful? Something which expands on the 'why' of this change. Appreciate the commit is a decent summary, but an ADR would call out this reasoning in a place that's arguably more visible/discoverable than the git log.

Might also be an idea to explain the temporary nature of this change, and how we expect it to be replaced by a 'better' message queue based implementation later on.

lib/services.rb
```diff
@@ -23,3 +11,29 @@ def self.publishing_api
...
+  end
+
+  # If attempt 1 fails, it will wait 0.03 seconds before trying again
+  # If attempt 2 fails, it will wait 0.09 seconds before trying again
+  # If attempt 3 fails, it will wait 0.27 seconds before trying again
+  # If attempt 4 fails, it will wait 0.81 seconds before trying again
+  # If attempt 5 fails, it will wait 2.43 seconds before trying again
+  # If attempt 6 fails, it will wait 7.29 seconds before trying again
+  def self.sleep_time_after_attempt(current_attempt)
+    (3.0**current_attempt) / 100
```
malcolmbaig, line 37: 👍

test/integration/indexer/indexing_test.rb
```diff
@@ -22,11 +21,8 @@ def teardown
   end
 
   def test_adding_a_document_to_the_search_index
-    content_api_has_an_artefact("an-example-answer", {
-      "tags" => [
-        tag_for_slug("bar", "specialist_sector"),
-      ]
-    })
+    stub_tagging_lookup
+    publishing_api_has_expanded_links({})
```
malcolmbaig: Have we got a WIP PR to make this a valid stub in gds-api-adapters?

lib/services.rb
MatMoore, line 37: I think I missed some of the discussion about this on another PR - did we intentionally reject the idea of `with_retries` taking a max time in seconds rather than max no. of retries?

I think it's better to explicitly say how long you are prepared to wait, as it forces you to think about it up front. In some cases just letting it die could be preferable to blocking a worker thread for too long.

tijmenb, line 37: There has been no discussion on this. Perhaps we could have that discussion later / not in this PR? I think it's an interesting point, but if we're going to discuss that here I'd rather remove the retry and create a new PR for that.

lib/index_finder.rb
malcolmbaig, line 2: Not particularly - this could just as easily be a module that uses `extend self` or `module_function`. 

Some reckon that cases like this should always be a module because it's not being instantiated, and it hints to the reader that it's never meant to be. Can't say I'm particularly troubled either way.

lib/indexer/message_processor.rb
```diff
@@ -0,0 +1,53 @@
...
+        message.ack
+      end
+    rescue ProcessingError => e
+      Airbrake.notify_or_ignore(e, parameters: message.payload)
+      message.discard
+    rescue StandardError => e
+      # This is rescue of last resort. If anything goes wrong during the payload
+      # processing, we don't want to retry the message really quickly because
+      # that might overload elasticsearch or other components. This should be
+      # replaced by a retry mechanism with exponential back-off.
```
malcolmbaig, line 28: > This should be replaced by a retry mechanism with exponential back-off.

Like https://github.com/alphagov/rummager/pull/632/files#diff-3f3e9c13dba2683b03865a1c0b5287dbR36?

malcolmbaig:
A few smallish comments from me,  but overall this looks solid. Happy to review again / merge once build's green and remaining questions addressed.

lib/services.rb
MatMoore, line 37: Ok, happy to discuss it later.

PR #634, opened 2016-04-28
Limit requested result set size to 1000 records

malcolmbaig:
Return an error to the client if the count param is higher than 1000.

Trello: https://trello.com/c/LbfXhokO

### DIFF
--- a/test/unit/parameter_parser/search_parameter_parser_test.rb
+++ b/test/unit/parameter_parser/search_parameter_parser_test.rb
@@ -168,6 +168,14 @@ class SearchParameterParserTest < ShouldaUnitTestCase
     assert_equal(expected_params(count: 2), p.parsed_params)
   end
 
+  should "complain about an overly large count parameter" do
+    p = SearchParameterParser.new({ "count" => %w(1001) }, @schema)
+
+    assert_equal(%{Maximum result set size (as specified in 'count') is 1000}, p.error)
+    refute p.valid?
+    assert_equal(expected_params(count: 10), p.parsed_params)
+  end
+
   should "understand the q parameter" do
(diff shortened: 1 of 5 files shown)


test/unit/parameter_parser/search_parameter_parser_test.rb
```diff
@@ -168,6 +168,14 @@ def setup
     assert_equal(expected_params(count: 2), p.parsed_params)
   end
 
+  should "complain about an overly large count parameter" do
+    p = SearchParameterParser.new({ "count" => %w(1001) }, @schema)
+
+    assert_equal(%{Maximum result set size (as specified in 'count') is 1000}, p.error)
+    assert !p.valid?
```
tijmenb: `refute p.valid?`?

malcolmbaig: copy+paste from above, but yes, quite right.

tijmenb:
👍 

Not a fan of line breaks in markdown documents btw, they make it harder to re-use the text elsewhere and make editing harder (if you add a long word you may need to re-linebreak whole paragraphs).

benhyland:
👍 
looks fine, assuming further param validation (e.g. 0 <= start < count) is out of scope
although that example depends on what start is for...

test/unit/parameter_parser/search_parameter_parser_test.rb
malcolmbaig: updated

malcolmbaig:
@tijmenb point taken, but I'm not sure editing is all that easy when lines of text disappear off-screen. Text editors (once appropriately configured) are generally pretty decent at reflowing paragraphs of text to within 79 chars.

Also, regarding reuse - you could always copy/paste from the rendered markdown no? : )

PR #635, opened 2016-04-29
Allow rendering_app & publishing_app to be set

tijmenb:
This is interesting information when we will start migrating rummager to use the message queue. We’ll slowly migrate apps to send this information.

### DIFF
--- a/config/schema/base_document_type.json
+++ b/config/schema/base_document_type.json
@@ -13,6 +13,8 @@
     "policies",
     "popularity",
     "public_timestamp",
+    "publishing_app",
+    "rendering_app",
     "specialist_sectors",
     "spelling_text",
     "title",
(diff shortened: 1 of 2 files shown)


benhyland:
Does rummager consume information about pages or about content items?
Are pages and content items always synonymous?

I think that content items should always have a single publishing app, as a principle of our architecture. But I don't think the same is true for rendering app. Of course a given page will have a single rendering app, since that is implied by the routing; but I don't believe a page and a content item are the same thing.

tijmenb:
I'm not sure what you are proposing. All pages are content items (not all content items are pages) so they have this information, as mandated by the content schemas. 

Having this data copied into rummager (like content_id before - https://github.com/alphagov/rummager/pull/624) will allow easier debugging, and will allow users to link to examples of content published by/rendered by a specific application.

benhyland:
Spoke in person, and confirmed that rummager input is derived only from content items which represent pages. Since we know that pages have a one-to-one mapping with publishing apps, so does the rummager input, so this is fine.
👍

PR #636, opened 2016-05-12
Make DFID documents searchable

rgarner:
- Add a document type and the one field we use:
  - country
- Add the new dfid_research_output document_type to the mainstream index

### DIFF
--- a/config/schema/document_types/dfid_research_output.json
+++ b/config/schema/document_types/dfid_research_output.json
@@ -1,5 +1,805 @@
 {
   "fields": [
     "country"
-  ]
+  ],
+  "expanded_search_result_fields": {
+    "country": [
+      {
+        "value": "VA",
+        "label": "Vatican City"
+      },
+      {
+        "value": "BS",
+        "label": "The Bahamas"
+      },
+      {
+        "value": "GM",
+        "label": "The Gambia"
+      },
+      {
+        "value": "ZW",
+        "label": "Zimbabwe"
+      },
+      {
+        "value": "ZM",
+        "label": "Zambia"
+      },
+      {
+        "value": "YE",
+        "label": "Yemen"
+      },
+      {
+        "value": "VN",
+        "label": "Vietnam"
+      },
+      {
+        "value": "VE",
+        "label": "Venezuela"
+      },
+      {
+        "value": "VU",
+        "label": "Vanuatu"
+      },
+      {
+        "value": "UZ",
+        "label": "Uzbekistan"
+      },
+      {
+        "value": "UY",
+        "label": "Uruguay"
+      },
+      {
+        "value": "US",
+        "label": "United States"
+      },
+      {
+        "value": "AE",
+        "label": "United Arab Emirates"
+      },
+      {
+        "value": "UA",
+        "label": "Ukraine"
+      },
+      {
+        "value": "UG",
+        "label": "Uganda"
+      },
+      {
+        "value": "TV",
+        "label": "Tuvalu"
+      },
+      {
+        "value": "TM",
+        "label": "Turkmenistan"
+      },
+      {
+        "value": "TR",
+        "label": "Turkey"
+      },
+      {
+        "value": "TN",
+        "label": "Tunisia"
+      },
+      {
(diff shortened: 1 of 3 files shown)


tijmenb:
As the field definitions are "global", it's probably best to re-use some existing definitions / make them more difid specific. `expanded_search_result_fields` also might not be necessary. We'll discuss in person!

rgarner:
Ok – I'm not pushing this with the intention of getting it merged (yet) – just to re-start the conversation you've been having with @tahb (who's rolled off this project for now).

rgarner:
@tijmenb I'm not in the office any day except Mon/Tue, so here's what I think I'd propose:
- `country` to remain as-is – that's coming from https://country.register.gov.uk, so I hope it's sufficiently global
- `theme` to become `dfid_research_theme`
- `region` to become `dfid_research_region` (neither of these are in use yet)
- `published_date` I'm not sure about. Looks like it should map to `first_published_at` but that would be a constant fixed date and I'm not sure if the meaning of this field more closely maps to something like `most_recently_published_at`.

tijmenb:
- Perhaps we could add both `country_code` and `country_name` fields, so that it's clear what the value should be. This would also allow you to remove `expanded_search_result_fields` because the name is already indexed.
- `dfid_research_theme` & `dfid_research_region` 👍 
- `published_date` - not sure too. I'll find out what all the different dates in the definitions mean so we can pick an existing one.

Thanks!

rgarner:
I must confess I have no idea what `expanded_search_results_fields` is for, only that @tahb mentioned that you said we might not need it.

Yes, let's add the two – if we can avoid having to use [`DfidTransition::Patch::Rummager::Countries`](https://github.com/alphagov/dfid-transition/blob/master/lib/tasks/patch/countries.rake) and remembering to commit both here and in Specialist Publisher that would be nice. Perhaps we ought to be specific that it's `country_iso2_code` while we're about it.

tahb:
@rgarner I initially started populating `expanded_search_results_fields` because there were other finders using it I inferred that it was required for 'search' to work.

I later discovered that removing it didn't change the behaviour of search and on checking why the [readme](https://github.com/alphagov/rummager/blob/532af2c128ef2471592db2b50d5fdcd6b62f2e51/config/schema/README.md#document-types) has it marked as "deprecated". I [noted this learning](https://github.com/alphagov/dfid-transition/wiki/AddingRummagerFieldTypes#search) and it lead to the [removal](https://github.com/alphagov/dfid-transition/pull/7/commits/f065e6b63bd51b2331c03096918a580b6cf5c035) of this field in our Patcher which **should** now populate a different schema with the `expanded_search_results_fields` removed.

rgarner:
Ah, ok, thanks @tahb – I'd forgotten that, and as I still had a patched rummager schema lying around, thought it was still current.

tahb:
I should also note that after chatting with @tijmenb I spent some time splitting out `country` into `country_code` and `country_name` without success. Given the time required to reindex to debug and given that without this: search, filtering and metadata all appear to work perfectly in `finder-frontend` I abandoned it.

rgarner:
@tijmenb, I've removed `published_date` – we don't need it yet and I'll bring it back if we do – and likewise `region`, `theme` are also premature given that there's no tested support/import mapping for them yet.

I've left `country` as is; @tahb said he had issues separating these out into two fields, and what we have currently works. Would you like to speak on Monday about this?

rgarner:
@tijmenb, I've reverted to our working `expanded_search_result_fields` solution at the moment. I'm still concerned that this is called `country` – I guess this would be a generic enough name to block a merge? Would you prefer it to be called `country_code` in advance of us attempting a two-field solution?

tijmenb:
Allthough `expanded_search_fields` is deprecated, I'm merging this to get this testable on integration.

rgarner:
Thanks @tijmenb

rgarner:
Context for the [two-field solution](https://github.com/alphagov/rummager/tree/two-field-rummager-dfid) on the Pivotal [ticket](https://www.pivotaltracker.com/story/show/119439623).

PR #639, opened 2016-05-24
Adds Alan Turing to ignore.yml

Davidslv:
We don't want to wrongly suggest his name during search.
#### Before

screen shot 2016-05-24 at 13 37 31
#### After

screen shot 2016-05-24 at 13 37 14


jackscotti:
👍

PR #643, opened 2016-05-31
Opt in to gds-api-adapters request tracing middleware

MatMoore:
Pass the two headers on to all downstream requests, including elasticsearch.

We currently run a local nginx on the search hosts that does some load balancing and proxies to elasticsearch.
Long term, we can remove this, since the elasticsearch gem can do its own load balancing.

For now though, it would be useful to be able to trace requests back to the original request. The nginx logs currently don't contain any identifible information about the request, because the query string is part of the GET body rather than the URL.

We will also pass through the request id to downstream requests made through gds-api-adapters,
eg content api when indexing.

### DIFF
--- a/lib/legacy_client/client.rb
+++ b/lib/legacy_client/client.rb
@@ -105,7 +105,7 @@ module LegacyClient
           url: url_for(path),
         }
         args[:payload] = payload if payload
-        args[:headers] = headers if headers
+        args[:headers] = headers
         args[:timeout] = @timeout if @timeout
         args[:open_timeout] = @open_timeout if @open_timeout
         logger.debug(args.reject { |k| k == :payload })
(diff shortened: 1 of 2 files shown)


lib/legacy_client/client.rb
```diff
@@ -97,6 +97,8 @@ def request(method, path, payload = nil, headers = {})
         headers[:content_type] = "application/json"
       end
 
+      headers = headers.merge(GdsApi::GovukHeaders.headers)
```
tijmenb, line 100: https://github.com/alphagov/rummager/blob/23e84e8d4e6240254b56d220f22443b17e49fb92/lib/legacy_client/client.rb#L108 seems to suggest `headers` can be nil. Can we remove that ~~line~~ conditional?

MatMoore, line 100: Removed this in another commit. It's defaulted to {} everywhere and I can't see anywhere where headers is passed anything that could be nil.

PR #644, opened 2016-05-31
Opt in to gds-api-adapters request tracing middleware

MatMoore:
Pass the two headers on to all downstream requests, including elasticsearch.

We currently run a local nginx on the search hosts that does some load balancing and proxies to elasticsearch.
Long term, we can remove this, since the elasticsearch gem can do its own load balancing.

For now though, it would be useful to be able to trace requests back to the original request. The nginx logs currently don't contain any identifible information about the request, because the query string is part of the GET body rather than the URL.

We will also pass through the request id to downstream requests made through gds-api-adapters,
eg content api when indexing.

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -26,6 +26,10 @@ else
     extra_request_headers: { "GOVUK-Request-Id" => "govuk_request_id", "x-varnish" => "varnish_id" }
 end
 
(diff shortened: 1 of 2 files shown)


tijmenb:
👍

PR #645, opened 2016-05-31
Stop magic tags from being indexed

tijmenb:
Content API adds two "magic tags" to artefacts when requesting an artefact over the API. Prevent these tags from ending up in Rummager. These are not tags that the user has chosen, but automatically added things that cause an extra navigation to show up on the page. For reference:

https://github.com/alphagov/govuk_content_api/blob/b74a48dd9c43f56120893e5b2000c1da2c473da0/govuk_content_api.rb#L686-L691

When we initially started indexing links from the content-api, we started inadvertently indexing these tags too. This caused initial problems on the browse pages, which we solved by excluding the format from the browse pages (https://github.com/alphagov/collections/pull/170).

This commit solves the original sin and removes these rogue tags from the search-index at indexing time.

We need this now because we want publishing-api to be in sync with rummager.

https://trello.com/c/UidTGyou

### DIFF
--- a/lib/indexer/tag_lookup.rb
+++ b/lib/indexer/tag_lookup.rb
@@ -47,6 +47,13 @@ module Indexer
       doc_hash["mainstream_browse_pages"].uniq!
       doc_hash["specialist_sectors"].uniq!
 
+      # Content API adds two "magic tags" to artefacts when requesting an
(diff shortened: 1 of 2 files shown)


Davidslv:
Looks good to me. 👍

PR #646, opened 2016-06-06
[Don't merge] Add `rummager_document_type` as alias for `_type`

tijmenb:
The term `document_type` is also used by the publishing-api, but they are not the same. To prevent confusion, we're renaming it here in Rummager.

https://trello.com/c/KlVeVW91

I'm not a 100% convinced about the name. Perhaps people have other suggestions?
- `internal_type`
- `elasticsearch_type`
- `internal_document_type`

### DIFF
--- a/lib/parameter_parser/base_parameter_parser.rb
+++ b/lib/parameter_parser/base_parameter_parser.rb
@@ -21,6 +21,12 @@ class BaseParameterParser
   # Incoming filter fields will have their names transformed according to the
   # following mapping. Fields not listed here will be passed through unchanged.
   FILTER_NAME_MAPPING = {
+    # The term `document_type` is also used by the publishing-api, but they are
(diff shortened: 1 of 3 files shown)


MatMoore:
What about just leaving it as "_type"? I think we return some other ES internals that way.

Prefer "elasticsearch" to "rummager" since rummager makes it sound like some new concept we introduced.

tijmenb:
@MatMoore 

I think the argument for `rummager_document_type` is that it's clear on the outside that this is an internal thing. There's nothing elasticsearch-y to the type.

I'm going to quickly check if `document_type` is used by any client - if not we should actually hide it from the API.

tijmenb:
Offline convo: we'll perhaps try something like `rummager_schema` to make communicate that this type determines the fields that get returned.

PR #647, opened 2016-06-06
[DO NOT MERGE] Proof of concept for including organisation pages in an organisation search

MatMoore:
When filtering by organisation, we expect to see the organisation page itself in the results.
Right now, we achieve this by automatically adding the organisation page to its
own organisations field - i.e. tagging the page to itself.

Publishing api doesn't know about this tagging, and we need to make the two
consistent. This PR is one possible way of doing this without changing the behaviour
in rummager. The alternative is to have whitehall send organisation self-tags into
publishing api.

This isn't intended to be mergeable, I just want to see if this is an approach is worth continuing with. Thoughts?

Further discussion on https://trello.com/c/aOTMr3WJ/640-whitehall-organisations-tagged-to-itself
cc @alphagov/team-finding-things

### DIFF
--- a/lib/parameter_parser/search_parameter_parser.rb
+++ b/lib/parameter_parser/search_parameter_parser.rb
@@ -220,8 +220,6 @@ private
       case field_name
       when "organisations"
         "organisation"
-      else
-        nil
       end
     end
   end
--- a/lib/parameter_parser/search_parameter_parser.rb
+++ b/lib/parameter_parser/search_parameter_parser.rb
@@ -215,6 +215,15 @@ private
     def valid?
       errors.empty?
     end
+
+    def associated_document_type
+      case field_name
+      when "organisations"
+        "organisation"
+      else
+        nil
+      end
+    end
   end
 
   class DateFieldFilter < Filter
--- a/lib/search/query_components/user_filter.rb
+++ b/lib/search/query_components/user_filter.rb
@@ -52,6 +52,18 @@ module QueryComponents
         raise "Filter type not supported"
       end
 
+      associated_document_type = filter.associated_document_type
+      if associated_document_type
+        associated_document_filter = combine_filters(
+          [
+            terms_filter("slug", values),
+            term_filter("format", associated_document_type)
+          ],
+          :and
+        )
+        es_filters << associated_document_filter
+      end
+
       combine_filters(es_filters, :or)
     end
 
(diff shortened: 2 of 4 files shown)


tijmenb:
Yeah, this would work. I wonder if we could make this more explicit and keep the `filter` API clean (= only return things that are physically tagged to the organisation). 

I think there are three options:

1) Explicitly enable this behaviour with a flag (`include_organisation_itself=true` or something). Search query would be:

https://www.gov.uk/api/search.json?filter_organisations[]=hmrc&include_organisation_itself=true

2) Add a special search case for organisation-scoped search and add behaviour based on that:

https://www.gov.uk/api/search.json?mode=organisation_scope&filter_organisations[]=hmrc

In time we could add more special-behaviour for this kind of search.

3) Add a parameter that force-includes a document into the search results:

 https://www.gov.uk/api/search.json?filter_organisations[]=hmrc&force_include_results[]=https://www.gov.uk/government/organisations/hm-revenue-customs

benhyland:
@MatMoore this looks sensible. Certainly no worse than our other option where we mess with the data in publishing-api.
My order of preference for @tijmenb 's options is essentially in order of level of detail of abstraction.
2, then 1. I think we should avoid 3.

tijmenb:
Why not 3?

benhyland:
I think option 3 exposes too much control over the details of rummager to the client, without actually adding much benefit.
With 3, the functionality is not explicitly tied to any particular use case or abstraction, which will make it harder to maintain in future.
If there is a shortish list of urls that the client believes belong in the results, it could just add them in itself.

lib/parameter_parser/search_parameter_parser.rb
```diff
@@ -215,6 +215,13 @@ def ==(other)
     def valid?
       errors.empty?
     end
+
+    def associated_document_type
+      case field_name
```
Davidslv, line 220: if you `require "active_support/core_ext/string"` you can `field_name.singularize if field_name == "organisations"`

MatMoore:
Discussed with @tijmenb offline.

We're going to keep this logic in rummager, but do it at indexing time instead, i.e. add to the organisations array when indexing organisations.

This solves the immediate problem: publishing api and rummager are sent different data, which blocks us from moving indexing responsibility to the publishing api.

This means we will defer dealing with the underlying inconsistency until we have a clearer idea of what we want we want to do with organisation tags. For the migration work we will need to add a whitelist entry for this.

PR #648, opened 2016-06-07
Move self-linking of organisations from whitehall

MatMoore:
This logic was previously in whitehall, so will be lost at the point we start indexing based on the publishing api exchange.

We tag organisations to themselves so that the organisation home page shoes up in search results filtered by organisations.

I'd like to do some further testing on integration to make sure this uses the correct fields.

https://trello.com/c/aOTMr3WJ/640-whitehall-organisations-tagged-to-itself

### DIFF
--- a/lib/indexer/tag_lookup.rb
+++ b/lib/indexer/tag_lookup.rb
@@ -14,6 +14,7 @@ module Indexer
       artefact = find_document_from_content_api(doc_hash["link"])
       return doc_hash unless artefact
       add_tags_from_artefact(artefact, doc_hash)
+      add_self_links(doc_hash)
     end
 
   private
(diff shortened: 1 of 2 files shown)


MatMoore:
Corresponding whitehall change: https://github.com/alphagov/whitehall/pull/2619

Rummager should be deployed first.

lib/indexer/tag_lookup.rb
```diff
@@ -27,6 +28,23 @@ def find_document_from_content_api(link)
...
+      # Consider an organisation page to linked to itself.
+      # This means that when filtering on an organisation,
+      # the organisation page gets included in the search results.
+      #
+      # This deliberately doesn't match up with the canonical representation
+      # of the organisation in the publishing api, since self-linking has
+      # a very fuzzy meaning: ids in links can mean both the thing (HMRC)
+      # and the content representing the thing (the HMRC home page).
+      if doc_hash["format"] == "organisation" && doc_hash["slug"]
+        doc_hash["organisations"] << doc_hash["slug"]
```
benhyland, line 41: Do we know `doc_hash["organisations"]` is non-`nil` at this point?

benhyland:
LGTM other than the minor question above.

lib/indexer/tag_lookup.rb
MatMoore, line 41: Yeah, although its order dependent which is a bit awkward. It's initialised in `add_tags_from_artefact`.

PR #649, opened 2016-06-13
Add test for `is_withdrawn` type

tijmenb:
There's some confusion about the type of `is_withdrawn`:

https://github.com/alphagov/finding-things-migration-checker/pull/57#discussion-diff-66598952

This adds a test that confirms that Rummager returns booleans as booleans, not as strings.


benhyland:
The confusion was when calling via gds-api-adapters. @Davidslv and I thought we had examined its output in the console.

PR #650, opened 2016-06-13
Rework how tags are indexed

tijmenb:
This changes the way tags are indexed by rummager.

Currently, every time we send a document to elasticsearch, the document is first "prepared" by the `DocumentPreparer` class. This adds some fields (for popularity) and copies some fields (format to `_type`). It also looks up the tags for the document in the content-api.

For example, when indexing the payload for

https://www.gov.uk/state-pension-age

it will look up which tags have been added to this artefact:

https://www.gov.uk/api/state-pension-age.json

(At the time of writing, this is the "State Pension" browse page)

This commit changes the datasource, so that it looks up the tags in the publishing-api rather than the contentapi. This has been made possible by work to make publishing-api the accurate source of all publishing-api tags.

It also replaces way we index things from the message queue. Currently there's an incomplete implementation (which has been switched off on all environments and currently isn't running), which consumes messages from the publishing-api message queue and indexes the payloads.

This message queue consumer has been changed to treat new messages as a "change notification". Whenever the message queue consumer receives a new message, we simply re-add the existing document to the index, forcing a new tag-lookup by the `DocumentPreparer`, as described above.

This gives us a guarantee that we always index the latest version of the links. If we were to index the links from the queue directly (like the current implementation does) we would run the risk of indexing stale data.

The trade-off here is that we have to make an extra request to the publishing-api for data that has already been sent to rummager in the message queue payload.

https://trello.com/c/SqCoDFxu


malcolmbaig:
LG

PR #651, opened 2016-06-13
Correctly self-tag organisations

tijmenb:
This fixes a bug that prevented the "self-tagging" introduced in https://github.com/alphagov/rummager/pull/648 to work properly.

Because `organisations` do not have an artefact in content-api, the tags would never be added because the tag lookup bailed at `return doc_hash unless artefact`.


benhyland:
The tag_lookup class is likely to disappear soon, and in any case it would make sense to separate the lookup and enhancement steps. 👍

PR #652, opened 2016-06-14
Don't block search requests based on referrer

MatMoore:
We recently [blogged about the search API](https://gdsdata.blog.gov.uk/2016/05/26/use-the-search-api-to-get-useful-information-about-gov-uk-content/), but none of the links are clickable, because
rummager returns 403 for requests that have a referer header from another domain :(

This is caused by this middleware: https://github.com/sinatra/rack-protection/blob/master/lib/rack/protection/json_csrf.rb

The reason for having this check is that if a GET request was protected by a session, you can trick
a victim's browser into making a request using that session, and extract sensitive
information from the response (eg by combining with a javascript injection)

For more details see: http://stackoverflow.com/questions/10509774/sinatra-and-rack-protection-setting

In our case the only API we expose is search, which returns no sensitive information,
so we can just switch this off.

Trello: https://trello.com/c/jT2dkcMx/643-api-search-json-sometimes-returns-forbidden-waiting-for-bug-to-happen-again

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -27,8 +27,16 @@ require "indexer/workers/delete_worker"
 require "indexer/workers/amend_worker"
 
 class Rummager < Sinatra::Application
-  # Stop double slashes in URLs (even escaped ones) being flattened to single ones
-  set :protection, except: [:path_traversal, :escaped_params, :frame_options]
+  # - Stop double slashes in URLs (even escaped ones) being flattened to single ones
+  #
+  # - Explicitly allow requests that are referred from other domains so we can link
+  #   to the search API.
+  #   JsonCsrf requires the referer to match this domain. This is one way to prevent
+  #   browsers from being tricked into making a request that returns sensitive
+  #   information or performs some dangerous action.
+  #   In our case the only public API is search and there is no sensitive or
+  #   personalised information in the response.
+  set :protection, except: [:path_traversal, :escaped_params, :frame_options, :json_csrf]
 
   def search_server
     settings.search_config.search_server


tijmenb:
👍  from me, but like someone else to check this too.

MatMoore:
cc @alexmuller @jennyd 🔒

rjw1:
You should write a follow up blog post for the tech blog :)

benhyland:
Looks good except that it is unlikely but possible that we might introduce another endpoint in future which needs this protection. If we consider the accompanying comment to be sufficient documentation of the issue for the future then 👍  Otherwise, can we turn it off per-endpoint, or document better elsewhere?

MatMoore:
@rjw1 good idea, it was a tricky one to track down!

@benhyland Correct me if you find anything, but I wasn't able to find a way of limiting it by endpoint. The set method is global to the application, and I didn't see anything obvious in the middleware itself. http://www.sinatrarb.com/configuration.html

I'm not too concerned about it given the nature of the app. Did you have something else in mind for further documentation?

benhyland:
@MatMoore Not really. I can't think of a better place for it than the top of the file that is the entry point for routing (i.e. where it is already).

MatMoore:
In that case I'll leave it as is, app.rb contains all routing right now, there's no separate file for it like in rails.

rboulton:
An alternative approach would be to change the way we pass requests through our frontend nginx server to strip off the referer headers.  This way, the configuration would be specific to the public endpoint, rather than specific to the application, which might be a cleaner way to specify it.  I'm not sure whether this is better or worse, though - opinions welcome.

If you want to look into that, the configuration in question is at https://github.com/alphagov/govuk-puppet/blob/475ca2f4f23a5046458100fb3329e4b6c62c86f9/modules/govuk/templates/publicapi_nginx_extra_config.erb#L27-L35

benhyland:
@rboulton Nice to know. I think it is more sensible to keep the config with the app than have to know in detail about the govuk-puppet repo contents when working on rummager, even if this leads to a slightly less flexible solution.

jackscotti:
This work can be 🚢 ed for me.

PR #656, opened 2016-06-17
Offload reindexing to Sidekiq

tijmenb:
This offloads reindexing to Sidekiq to make the indexing faster. Also bump Sidekiq's concurrency to 16, which seems to be a fine number for integration.


Davidslv:
👍

PR #658, opened 2016-06-21
Don't rely on publishing-api when deploying

tijmenb:
On each deploy, capistrano runs the rake task:

```
rake publishing_api:publish_special_routes
```

When publishing-api is unavailable this causes the deploy to fail. These pages are republished each deploy for convenience, so that we don't have to remember to republish them each time we change them. In most cases they won't send anything new to the publishing pipeline, so we're fine with giving a warning when the request fails.

Fixes: https://deploy.integration.publishing.service.gov.uk/job/Deploy_App/4441/console

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -31,14 +31,21 @@ namespace :publishing_api do
     ]
 
     routes.each do |route|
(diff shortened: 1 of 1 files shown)


malcolmbaig:
👍

PR #659, opened 2016-06-21
Add Airbrake warning if unexpected links are present during indexing 

malcolmbaig:
Rummager now fetches links data directly from the publishing API when
indexing a document. We're in the process of updating Rummager client
apps to not send mainstream browse pages, specialist sectors and
organisation links.

This commit adds an Airbrake message if links of
this type are encountered when preparing a document for indexing.

trello https://trello.com/c/LbZA0i15/678-stop-apps-from-sending-mainstream-browse-pages-topics-and-organisations-to-rummager


lib/indexer/document_preparer.rb
```diff
@@ -20,6 +21,16 @@ def prepared(doc_hash, popularities, is_content_index)
 
   private
 
+    class UnexpectedLinksError < StandardError
+    end
+
+    def warn_if_links_present_in(doc_hash)
+      unexpected_links = %w{ mainstream_browse_pages organisations specialist_sectors }
+      if doc_hash.keys.any? { |key| key.in?(unexpected_links) }
+        Airbrake.notify_or_ignore(UnexpectedLinksError.new, parameters: doc_hash.keys & unexpected_links)
```
tijmenb: We probably need the full `doc_hash` to find the source of the request.

Gemfile
```diff
@@ -59,4 +59,5 @@ end
 group :development do
   # (Intelligent) reloading server in development
   gem "mr-sparkle", "0.3.0"
+  gem "pry-byebug"
```
jackscotti: This might go in the test group too.

mobaig: updated

PR #660, opened 2016-06-23
Remove topics field, keep policy_areas

MatMoore:
This removes the deprecated topics field to avoid confusion with specialist sectors,
which are also called topics.

Before merging, we should ensure that
- [x] The topics field is not being used for filtering or faceting - policy_areas should be used instead (https://github.com/alphagov/whitehall/pull/2629)
- [x] All existing documents with topics contain the same data in the policy_areas field
- [x] Whitehall is using the policy_areas field when indexing (https://github.com/alphagov/whitehall/pull/2630)

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -141,7 +141,7 @@ class Rummager < Sinatra::Application
   # Arbitrary "filter parameters", anything which is defined in the mappings
   # for the index is allowed. Examples:
   #   search_format_types[]        - eg "consultation"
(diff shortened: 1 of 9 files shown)


jackscotti:
LGTM

tijmenb:
LGTMAW!  💃

jackscotti:
Can this be merged?

MatMoore:
Yep, fine by me. I'll do a final end to end test once this is on integration.

PR #661, opened 2016-06-23
Add sigdump gem

MatMoore:
This allows us to generate thread dumps by sending a signal to the process,
which can be useful for debugging.

Require it in app.rb as this is the entry point for the app. This is also
required by the sidekiq worker.


MatMoore:
Actually sidekiq implements this for us, so I don't think there's much need for this gem in rummager.

```
kill -TTIN [worker_pid]
```

https://github.com/mperham/sidekiq/wiki/Signals

PR #663, opened 2016-06-28
Upgrade sidekiq to 4.1.2

benhyland:
There is a bug which can cause the polling thread to die
when redis is unavailable during sidekiq startup.
This means that, while the worker threads are still around,
they will never be given any work to do.

This occurs often in integration, but could potentially affect
other environments because it introduces an ordering dependency
on application startups.

The upgrade to 4.1.2 fixes the issue with the polling thread,
but introduces the explicit ordering dependency we wanted to avoid!
At least it will fail early and in a more obvious way now.

Since it dies cleanly, puppet will be able to detect and restart it every 30mins.

We probably want to use the govuk_sidekiq gem, but we'll do that later.

See https://trello.com/c/YIfeLd7C/691-rummager-sidekiq-workers-regularly-die-in-integration


tijmenb:
The upgrade PR looks good. 

I seem to remember that puppet or upstart retries a process a number of times if it won't start. Would that take care of this? The startup script feels like it's in the wrong place, because we'd have to duplicate it across GOV.UK apps.

Davidslv:
Should we replace it with https://github.com/alphagov/govuk_sidekiq ?

benhyland:
@Davidslv yes but not in this PR :)

benhyland:
@tijmenb talked to infrastructure - upstart will retry but without backoff, and will stop trying if it fails 5 times in 20 secs. All this is configurable, but not worth the effort, since puppet will ensure running every 30mins. With the newer version of sidekiq puppet will notice the process has dies rather than limped on, so it will try to restart the service.

Davidslv:
👍

PR #664, opened 2016-06-29
Improve error handling

benhyland:
Don't send DocumentNotFound errors to errbit
With the current design which parallelises deletes and amends, they are not avoidable.

We have better ways of seeing whether there is an unwanted difference
in content in rummager and publishing-api.

If we are worried about bugs which induce large numbers of these, we
should count them with a stat and monitor with icinga, rather than clutter errbit.

Also, tidy up bulk_index error handling.

This allows us to detect locked indexes on elasticsearch responses
which contain a 'create' field as well as those which contain an 'index' field.

Does any reviewer know whether we should be handling both 'locked index' and 'other' errors, or whether they are mutually exclusive?

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -23,6 +23,9 @@ configure do
   Airbrake.configuration.ignore << "Sinatra::NotFound"
   Airbrake.configuration.ignore << "LegacySearch::InvalidQuery"
 
+  # DocumentNotFound is inevitable since we process deletes and amends in parallel
+  Airbrake.configuration.ignore << "SearchIndices::DocumentNotFound"
+
   # We manually send `Indexer::BulkIndexFailure` to Airbrake with extra
   # parameters for debugging. Ignore it here so that we don't send them twice.
   Airbrake.configuration.ignore << "Indexer::BulkIndexFailure"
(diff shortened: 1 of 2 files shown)


MatMoore:
👍 for tidy up.

> If we are worried about bugs which induce large numbers of these, we should count them with a stat and monitor with icinga, rather than clutter errbit.

Are we worried? I'd be interested to know how frequently this occurs and whether this actually changes what rummager ends up with. Processing amends/deletes out of order seems harmless, but I'm not sure if there could be a similar scenario with creates/amends getting swapped?  

I'll leave it to somebody more knowledgable to answer your question about locked indexes (I would assume mutually exclusive but haven't looked into it).

benhyland:
@MatMoore aiui creates and amends are equivalent in ES. It'll be last write wins - kinda bad if Rummager picks a different 'last' than some other system (e.g. publishing-api). It's worth considering at some point but I think we have more important architectural problems for now.

tijmenb:
Recap of offline discussion:
- If we're ignoring the `DocumentNotFound` exception we should probably remove it entirely.
- Not sure about what the elasticsearch output with `index` and `create` means. I'd like to find out why we were only looking at `index` previously.

tijmenb:
Superseded by https://github.com/alphagov/rummager/pull/669.

PR #665, opened 2016-06-30
Reduce concurrency of migrate_index task

MatMoore:
This job is run every night so that we can recreate our indices to include popularity measures from google analytics. Currently it takes about 30 minutes.
## How it works
- We create a number of consumer threads that pull documents off a single queue.
- This queue is populated from the original index and the consumers
  reindex the documents in parallel
## What changed

We changed indexing to call publishing api to fetch links
data every time a document indexed via a call to get_expanded_links.

Since then, the nightly job has been failing due to the get_expanded_links requests timing out.

This is a workaround until we can improve the performance of the publishing api.

It's not ideal, because the longer this job takes to run, the longer the index is locked for, which impacts publishing (rummager sidekiq jobs will fail until the index becomes available).

We think this is acceptable in this case because it's an overnight job.

### DIFF
--- a/lib/tasks/indices.rake
+++ b/lib/tasks/indices.rake
@@ -36,7 +36,8 @@ You should run this task if the index schema has changed.
     require 'indexer/bulk_loader'
 
     index_names.each do |index_name|
-      Indexer::BulkLoader.new(search_config, index_name).load_from_current_index
+      # Batch concurrency reduced from 12 to 3 until publishing api can handle the load.
+      Indexer::BulkLoader.new(search_config, index_name, batch_concurrency: 3).load_from_current_index
     end
   end
 


benhyland:
How long does the job take now? How long do you expect it to take with 3 ruby threads (under GIL) instead of 12?
If we think it will be quick enough, 👍

tijmenb:
Did we explore any other options such as a retry mechanism?

benhyland:
@tijmenb No, but the nature of the problem means it is unlikely that retries will be much help IMO.

The assumption is that publishing-api is unable to cope with sustained load above a certain level.
(otherwise, we would get random-ish failures in any calls to get_expanded_links regardless of load).

So, there's a limit to the load that publishing-api can reliably handle. If we retry, we have to back off sufficiently to get below this level. The entire job cannot be run faster that the throughput allowed by this load threshold. So, rather than add the additional pressure of retries to a loaded system, and back off to an acceptable rate, just hit the system at a steady rate that is at or below its capacity. By avoiding retries we might actually manage to go faster since part of the stress on publishing-api may be caused by handling the transport layer and message parsing. I think the only reason we might go faster by backing off and retrying is that it might get us closer to the (unknown) load threshold - as it is we are cutting the parallelism by an arbitrary amount which might be much more than we need to.

Another use of retries would be to avoid failure of the job if there is a short spike of bad performance which resolves itself. Not sure if this is what we are seeing?

MatMoore:
@benhyland I still need to test it on integration - if it takes too long with 3 I'll roll it back or raise another PR.

PR #666, opened 2016-07-04
Handle slow publishing api requests a bit better

MatMoore:
Expanded links requests are timing out at the moment and this is breaking our nightly jenkins job.
https://trello.com/c/kAm186hM/687-jenkins-rake-task-to-update-the-index-based-on-google-analytics-is-failing
- Increase timeout for publishing api (this is only used during indexing)
- Add some logging so we can investigate the content involved (if it's still an issue)

This is not worth merging until I've run through the job successfully. I'll deploy the branch to integration as this is a better test than my dev VM.

### DIFF
--- a/lib/indexer/links_lookup.rb
+++ b/lib/indexer/links_lookup.rb
@@ -4,6 +4,10 @@ require 'services'
 # the document. If there aren't any links, the payload will be returned unchanged.
 module Indexer
   class LinksLookup
+    def initialize
+      @logger = Logging.logger[self]
+    end
+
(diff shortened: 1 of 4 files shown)


MatMoore:
I think this can be merged, even though it doesn't solve the problem described on the trello ticket. Nginx has a proxy_read_timeout of 15s, so it's unlikely we'll actually hit the client timeout, but I think it's worth setting explicitly rather than relying on the gds_api_adapters default (4s).

This PR also reverts changes to the thread count which are unnecessary, and adds some logging.

PR #673, opened 2016-07-06
Add dfid_review_status

rgarner:
`dfid_review_status` is a single `identifier` that can have either of the values `unreviewed` or `peer_reviewed`.

It still uses the deprecated `expand_search_result_fields` Rummager feature. Two more facets will use that feature, but we have a [ticket](https://www.pivotaltracker.com/n/projects/1573681) to address this work during our next sprint.

Add a general integration test to smoke test multiple DFID finder query types in the face of automapping uncertainty.
## Related PRs
- https://github.com/alphagov/govuk-content-schemas/pull/341
- https://github.com/alphagov/specialist-publisher-rebuild/pull/818

### DIFF
--- a/test/integration/search/search_test.rb
+++ b/test/integration/search/search_test.rb
@@ -473,6 +473,31 @@ class SearchTest < IntegrationTest
     assert parsed_response.fetch("elasticsearch_query")
   end
 
+  def test_dfid_can_search_by_every_facet
+    reset_content_indexes
+    commit_document("mainstream_test", dfid_research_output_attributes)
+
+    facet_queries = %w(
+      filter_dfid_review_status[]=peer_reviewed
+      filter_country[]=TZ&filter_country[]=AL
+    )
+
+    facet_queries.each do |filter_query|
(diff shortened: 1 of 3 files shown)


test/integration/search/search_test.rb
```diff
@@ -473,6 +473,31 @@ def test_show_the_query
     assert parsed_response.fetch("elasticsearch_query")
   end
 
+  def test_dfid_can_search_by_every_facet
```
tijmenb, line 476: Perhaps this can go in a separate file? Since it technically duplicates existing tests we could remove it once DFID stuff is live and stable.

tijmenb, line 476: Nvm, that would require you to duplicate some other stuff.

rgarner, line 476: Yes, it did occur that this was duplication – you're right, this is more or less just scaffolding to avoid a contemporaneous problem. You want another PR to move it? I'm not sure what "nvm" stands for ;)

rgarner, line 476: Oh, never mind. Got it 😕

PR #681, opened 2016-07-13
Add alpha taxons to rummager schema

MatMoore:
This will help us highlight popular information tagged to a taxon (eg education).

https://trello.com/c/fG3pFItN/11-show-popular-content-from-education-in-search

Paired with @rubenarakelyan

### DIFF
--- a/test/integration/search/search_test.rb
+++ b/test/integration/search/search_test.rb
@@ -498,6 +498,46 @@ class SearchTest < IntegrationTest
     end
   end
 
+  def test_taxonomy_can_be_returned
+    reset_content_indexes
+
+    commit_document("mainstream_test",
+      title: "I am the result",
+      description: "This is a test search result",
+      link: "/some-nice-link",
+      taxons: ["eb2093ef-778c-4105-9f33-9aa03d14bc5c"]
+    )
+
+    get "/unified_search?q=test&fields[]=taxons"
+    assert_equal 1, parsed_response.fetch("total")
+
+    taxons = parsed_response.dig("results", 0, "taxons")
+    assert_equal ["eb2093ef-778c-4105-9f33-9aa03d14bc5c"], taxons
+  end
+
+  def test_taxonomy_can_be_filtered
+    reset_content_indexes
+
+    commit_document("mainstream_test",
+      title: "I am the result",
+      description: "This is a test search result",
+      link: "/some-nice-link",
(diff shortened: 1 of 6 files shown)


tijmenb:
Would it make sense to 1) call it `taxons` and 2) index the `content_id` of the taxon instead of the slug?

test/integration/search/search_test.rb
```diff
@@ -1,4 +1,5 @@
 require "integration_test_helper"
+require "pry-byebug"
```
malcolmbaig: Probably don't want to leave this here - worth sticking in that test helper above though.

MatMoore:
I think actually the tests are a bit unrealistic here - when we do the tag lookup, we index a hash of title, content id, base path, so using the slug in the tests is a bit misleading.

Happy to go with taxons over alpha_taxons. I was going for consistency with the content store. Are we planning to rename that at some point?

test/integration/search/search_test.rb
MatMoore: whoops. Will move it.

tijmenb:
@MatMoore the link type is `taxons` and the format is `taxon`. Where are you seeing `alpha_taxons`?

MatMoore:
@tijmenb I may have just imagined the whole thing. Changing it to taxons now.

MatMoore:
Discussed offline with @tijmenb:

We're not planning on adding taxons to rummager as documents, since there is a distinction between taxons as a thing within the taxonomy and navigation pages for a taxon (which we don't have yet). We might even remove the base path field from the taxon format to make this clearer.

So the taxon field in rummager shouldn't expand to hashes containing titles and base paths, like we do for specialist sectors, etc. It's just an array of content ids.

I've also renamed everything to use `taxon` throughout.

lib/search/presenters/entity_expander.rb
```diff
@@ -29,6 +29,7 @@ def initialize(registries)
       'world_locations' => :world_locations,
       'specialist_sectors' => :specialist_sectors,
       'people' => :people,
+      'taxons' => :taxons,
```
tijmenb: I think this is not necessary anymore.

tijmenb:
👍 apart from https://github.com/alphagov/rummager/pull/681#discussion_r70653352

lib/search/presenters/entity_expander.rb
MatMoore: Nope, committed that by accident

tijmenb:
👍

PR #683, opened 2016-07-14
Document adding new fields to the schema

MatMoore:
We already have some documentation about the schema, but I thought it would be useful to have a step by step guide to adding new fields, so we can do it more quickly next time.

Best reviewed by reading the file:
https://github.com/alphagov/rummager/blob/c2d16031fb8df6df4fe6f5d479b4e0a36bc0cf33/docs/adding-new-fields.md


tijmenb:
❤️

PR #687, opened 2016-07-21
Remove topics field, keep policy_areas (again)

MatMoore:
Add back the original commit to remove topics field.
It was reverted because policy areas weren't being indexed properly when whitehall content
was republished.

This reverts commit 2372f3f66b89104e09ccba59c42ce414c53cd183.

Depends on: https://github.com/alphagov/whitehall/pull/2672


tijmenb:
(Updated the title, so that it's clearer what it does)

PR #694, opened 2016-08-22
Alias /search.json to /unified_search.json

MatMoore:
The public URL of the search api is /api/search.json
Currently, the internal URL is /unified_search.json

"Unified" doesn't actually mean anything anymore, it's terminology
left over from when we used to have separate searches for different
kinds of content.

gds-api-adapters only works with the internal URL because it hard codes
the "unified_search.json" part of the endpoint. Using the same name
for both internal/public URLs would avoid confusion and allow
gds-api-adapters to work with the public endpoints too.

Follow up if this gets merged:
- Change gds-api-adapters to use `search.json`
- Change nginx to point to `search.json`
- Deprecate `unified_search.json`

Let me know if there's anything I've missed.

https://trello.com/c/cj8UX2jX

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -108,22 +108,24 @@ class Rummager < Sinatra::Application
   # Return results for the GOV.UK site search
   #
   # For details, see docs/unified-search-api.md
-  get "/unified_search.?:request_format?" do
-    json_only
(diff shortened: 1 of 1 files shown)


malcolmbaig:
👍

PR #695, opened 2016-09-06
[DO NOT MERGE] Simplify bulk payload generation code

MatMoore:
Currently, several methods accept either an Array, or a String, containing very
different types of data. This makes it difficult to reason about what values
actually are at certain points in the code.

Most of the time, `document_hashes_or_payload` is actually just an array of
document hashes. When we post to rummager and it queues an index or amend
operation, this is what we generate to send to elasticsearch.

We also have a `bin/bulk_load` script.

`bin/bulk_load` calls `Indexer::BulkLoader.load_from` with text from stdin.

The data is expected to be in the format accepted by the elasticsearch bulk
index api:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html

The `BulkLoader` class sends this data on an adventure through the codebase
where it is transformed from a string suitable for the bulk index api into a
string suitable for the bulk index api. Rummager then passes it to the bulk
index api.

The `bulk_load` needs to accept this stupid elasticsearch format so we can
indicate which operation we're using. The API call can handle various
operations besides "index" which is what we normally use when indexing documents.

Is `bulk_load` actually useful? Maybe - it's the documented way of reloading
best bet data from search admin https://github.com/alphagov/search-admin - but
it's not something I would normally expect to run.

Is Rummager actually doing anything useful with the data besides pass it to
elasticsearch? Yes. When indexing a new document via this script, it adds in
popularity data, and prepares the document (eg adding in links) as normal.

This change splits out everything that operates on these arrays of
maybe-command-maybe-document hashes into separate methods.

See https://trello.com/c/NKWpaF77/88-use-official-elasticsearch-client-as-transport-layer-in-rummager


MatMoore:
Closing this PR - it needs rebasing so I'll create a new branch if I finish this. This refactor wasn't actually needed for the ticket.

PR #696, opened 2016-09-07
Swap out RestClient for Elasticsearch::Transport

MatMoore:
https://www.elastic.co/guide/en/elasticsearch/client/ruby-api/current/_the_ruby_client.html

This gem provides a new client for communicating with the elasticsearch
cluster, Elastisearch::Transport::Client. This wraps Faraday by default which
in turn wraps Net::HTTP.

It also provides methods for each of the elasticsearch API endpoints, so we no
longer need to construct paths or serialize/deserialize JSON ourselves.

Note: The client will only parse JSON responses if the response contains a
Content-Type header. We have to make sure all our stubs pass this so that our
unit tests reflect reality.

The client doesn't seem to let you specify the request Content-Type very easily
(?) so we're sending application/json for everything regardless of whether the
request has a body or not. The GdsApi::GovukHeaders headers for request tracing
have been dropped from elasticsearch requests, but we weren't using them.

Note that we no longer use Plek for finding the elasticsearch hostname - we now
need an environment variable to be passed in. This allows us to stop using an
 intermediate load balancer and instead connect directly to the cluster.

Paired with @rubenarakelyan 
Trello: https://trello.com/c/NKWpaF77/88-use-official-elasticsearch-client-as-transport-layer-in-rummager 
Follow up from this spike: https://github.com/alphagov/rummager/pull/692


app.rb
```diff
@@ -89,10 +88,6 @@ def json_only
     content_type :json
   end
 
-  error RestClient::RequestTimeout do
```
MatMoore, line 92: needs replacing

lib/index.rb
```diff
@@ -59,13 +51,9 @@ def real_name
...
-        alias_info = JSON.parse(@client.get("_aliases"))
-      rescue RestClient::ResourceNotFound => e
-        response_body = JSON.parse(e.http_body)
-        if response_body['error'].start_with?("IndexMissingException")
-          return nil
-        end
-        raise
+        alias_info = @client.indices.get_aliases(index: @index_name)
+      rescue Elasticsearch::Transport::Transport::Errors::NotFound
+        return nil
```
MatMoore, line 56: For review: is it possible for `get_aliases` to 404 with an error besides `IndexMissingException`, and if so do we care about it?

lib/index.rb
```diff
@@ -231,12 +212,13 @@ def analyzed_best_bet_query(query)
...
-        @client.delete("#{CGI.escape(type)}/#{CGI.escape(id)}")
-      rescue RestClient::ResourceNotFound
+        @client.delete(index: @index_name, type: type, id: id)
+      rescue Elasticsearch::Transport::Transport::Errors::NotFound
         # We are fine with trying to delete deleted documents.
         true
-      rescue RestClient::Forbidden => e
-        response_body = JSON.parse(e.http_body)
+      rescue Elasticsearch::Transport::Transport::Errors::Forbidden => e
+        # FIXME: this is error-prone as it parses the response body out of the error message
```
MatMoore: Is there a simpler way of handling this?

test/test_helper.rb
```diff
@@ -30,6 +30,9 @@
 
 require "webmock/minitest"
 
+# Silence log output
```
MatMoore, line 33: Is there a better way to do this?

Without it, the minitest output is interrupted by log output from the client. Minitest doesn't have any built in options to suppress stdout from code under test.

test/unit/elasticsearch_index_test.rb
```diff
@@ -63,20 +68,37 @@ def test_add_sends_updates_to_the_bulk_index_endpoint
         "title" => "TITLE ONE",
     })
 
-    # Note that this comes with a trailing newline, which elasticsearch needs
-    payload = <<-eos
-{"index":{"_type":"edition","_id":"/foo/bar"}}
-{"_type":"edition","link":"/foo/bar","title":"TITLE ONE","popularity":0.09090909090909091,"format":"edition"}
-    eos
+    payload = [
```
MatMoore, line 70: `index.add` normally takes an array. The indexing methods try to support both arrays of doc hashes and elasticsearch-formatted strings like the one I removed here. The latter is only used by one caller (the `bulk_load` script) so it's misleading to have all our tests assume this format. 

The code for this can be refactored but I don't want to do it in this PR.

lib/legacy_client/client.rb
```diff
@@ -1,117 +0,0 @@
...
-      logging_exception_body do
-        args = {
-          method: method,
-          url: url_for(path),
-        }
-        args[:payload] = payload if payload
-        args[:headers] = headers
-        args[:timeout] = @timeout if @timeout
-        args[:open_timeout] = @open_timeout if @open_timeout
-        logger.debug(args.reject { |k| k == :payload })
```
MatMoore, line 112: This logging should be roughly equivalent to the Elasticsearch client's logger.

It will log requests and responses at ℹ️  level although we run at ⚠️ so these will be ignored at the moment.

HTTP 4xx and 5xx responses get logged out.

MatMoore:
On second thoughts, we shouldn't need to define the environment variables before merging this, because our nginx proxy also binds to localhost:9200 on integration/staging/production.

https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_elasticsearch/templates/nginx_local_proxy.conf.erb

app.rb
MatMoore: ✅

PR #699, opened 2016-09-14
Remove “unified search” from more files

rubenarakelyan:
Following on from https://github.com/alphagov/rummager/pull/698, this commit removes “unified search” from a couple of places where they were missed last time.

Trello: https://trello.com/c/cj8UX2jX


Davidslv:
👍 💯

PR #700, opened 2016-09-28
Update name for Democratic Republic of the Congo

Davidslv:
**trello**: https://trello.com/c/rxjklHjL/191-country-name-change-democratic-republic-of-the-congo

> The country formerly known as 'Democratic Republic of Congo' is now officially called 'Democratic Republic of **the** Congo'

### DIFF
--- a/config/schema/document_types/international_development_fund.json
+++ b/config/schema/document_types/international_development_fund.json
@@ -33,7 +33,7 @@
         "value": "burma"
       },
       {
-        "label": "Democratic Republic of Congo",
+        "label": "Democratic Republic of the Congo",
         "value": "democratic-republic-of-congo"
       },
       {


config/schema/document_types/international_development_fund.json
```diff
@@ -33,8 +33,8 @@
         "value": "burma"
       },
       {
-        "label": "Democratic Republic of Congo",
-        "value": "democratic-republic-of-congo"
```
tijmenb: Let's keep this value the same for now, so that we don't have to republish everything (if it's changed now, the search won't work).

tijmenb (changes requested):

config/schema/document_types/international_development_fund.json
Davidslv: 👍

tijmenb (approved):

rubenarakelyan:
👍

PR #703, opened 2016-09-28
Add frontmatter to /docs pages

tijmenb:
This will probably cause them to show up in the table of contents on

http://alphagov.github.io/rummager/


rubenarakelyan:
🚢

PR #704, opened 2016-09-28
Add styling to documentation pages

tijmenb:
Currently pages like this don't have a layout:

http://alphagov.github.io/rummager/adding-new-fields.html


rubenarakelyan (approved):
🚢

PR #706, opened 2016-10-06
Add new fields for existing links by content id

carvil:
Currently, we store the `base_path` for most links when indexing a document.
This is problematic when links change.

This is the first step into migrating to use `content_id` instead of `base_path`
when indexing links on documents.

This commit adds the following extra fields to each document:
- `topic_content_ids`
- `mainstream_browse_page_content_ids`
- `organisation_content_ids`

These will include the same links as their equivalent fields
(`specialist_sectors`, `mainstream_browse_pages` and `organisations`), but
referenced by `content_id` instead.

When we deprecate the existing links, we will use these `*_content_ids` links in
link expansion.


config/schema/base_document_type.json
```diff
@@ -10,13 +10,16 @@
...
+    "mainstream_browse_page_content_ids",
     "organisations",
+    "organisation_content_ids",
     "policies",
     "popularity",
     "public_timestamp",
     "publishing_app",
     "rendering_app",
     "specialist_sectors",
+    "topics",
```
tijmenb: Would it be clearer to call this `topics_content_ids` as well?

carvil: That was actually its original name and then I changed it to `topics`. I agree it would be clearer here, but I thought that as nothing else is called `topics` it would be 1 less thing to worry about later. But I'm definitely happy to change it again to look like the other fields.

carvil: From a chat IRL with Tijmen, we decided to change this to `topic_content_ids` (which is clearer as this is a list of `content_ids`) and on expansion (another PR), add an extra field called `topics` with the expanded topics.

carvil:
@tijmenb Updated

PR #707, opened 2016-10-07
Add fields for links expansion

carvil:
The commit 646d64f252a5393d8d30b0319361b8548bf49d12 added extra fields in order
to hold `content_ids` of links for a given document.

The next step in the process is to expand links based on those content IDs. In
order to do that transparently, this commit adds 3 new fields to each document:
- `expanded_topics`
- `expanded_organisations`
- `expanded_mainstream_browse_pages`

These new fields are meant to replace, in the future, the following fields:
- `specialist_sectors`
- `organisations`
- `mainstream_browse_pages`

For now we will be expanding the new fields, which will allow us to migrate
existing applications to use the new behaviour.

When all applications have been moved across, we can deprecate the fields
mentioned above.

### DIFF
--- a/config/schema/base_document_type.json
+++ b/config/schema/base_document_type.json
@@ -11,8 +11,10 @@
     "link",
     "mainstream_browse_pages",
     "mainstream_browse_page_content_ids",
+    "expanded_mainstream_browse_pages",
     "organisations",
     "organisation_content_ids",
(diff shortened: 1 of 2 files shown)


carvil:
This is actually not needed if those fields are included in the default return list. Closing for now.

PR #708, opened 2016-10-07
Perfom link expansion via content ids

carvil:
We are changing the way we are storing and expanding information on links.
Before, there would be one field per link type (e.g. `organisations`), which
would hold a list of `slugs` in ElasticSearch. When that document was being
served, we would expand those `slugs` into a Hash of attributes in the same
`organisations` field.

We are changing to use 2 fields per link type instead. One containing the
`content_ids` and another containing the expanded version of the links.

So, for the example above, we would have both `organisation_content_ids` and
`expanded_organisations`.

This PR expands the new fields `expanded_organsiations` and `expanded_topics` based on their content IDs. It also adds these fields to the search results so we can have access to them.

Trello: https://trello.com/c/12ijjlxN/209-change-the-way-rummager-refers-to-other-documents

### DIFF
--- a/lib/parameter_parser/base_parameter_parser.rb
+++ b/lib/parameter_parser/base_parameter_parser.rb
@@ -94,6 +94,10 @@ class BaseParameterParser
     title
     policy_areas
     world_locations
+    topic_content_ids
+    expanded_topics
+    organisation_content_ids
+    expanded_organisations
   ).freeze
 
   # Default order in which facet results are sorted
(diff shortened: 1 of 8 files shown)


lib/search/mapping.rb
```diff
@@ -0,0 +1,11 @@
+module Search
+  class Mapping
```
tijmenb: Would it be good to inline this into `entity_expander.rb` to indicate that it's only used in that context? Otherwise, an comment for this class would make it clear to the reader what the purpose is.

tijmenb (approved):

lib/search/mapping.rb
carvil: Yes, that makes sense. I will move it in!

tijmenb (approved):

PR #710, opened 2016-10-19
Return all available fields by default

carvil:
Rummager used to do field filtering by default when no fields were
specified.  I found that behaviour a bit confusing when working on a
feature that touched this functionality. The difference in output
between `search-admin` and the search API was also a bit confusing,
making it harder to get up to speed with how rummager filters fields.

This change aims at simplifying things by always returning all fields
from ElasticSearch when no specific fields have been requested.

When specific fields have been requested, we fetch all fields from
ElasticSearch and only keep the fields we are interested in, which is
the current behaviour.

This means we don't need to add extra fields such as title and
description to make sure virtual fields (e.g. highlights) work. Because
each result includes all fields, we are able to enrich the response and
further down the line remove unwanted fields.

What do you think?

### DIFF
--- a/lib/parameter_parser/base_parameter_parser.rb
+++ b/lib/parameter_parser/base_parameter_parser.rb
@@ -80,26 +80,6 @@ class BaseParameterParser
   #            the query and filters
   ALLOWED_EXAMPLE_SCOPES = [:global, :query].freeze
 
-  # The fields which are returned by default for search results.
-  DEFAULT_RETURN_FIELDS = %w(
-    description
-    display_type
-    document_series
-    format
-    link
-    organisations
-    public_timestamp
-    slug
-    specialist_sectors
-    title
-    policy_areas
-    world_locations
-    topic_content_ids
-    expanded_topics
-    organisation_content_ids
-    expanded_organisations
-  ).freeze
-
   # Default order in which facet results are sorted
   DEFAULT_FACET_SORT = [
     [:filtered, 1],
--- a/lib/parameter_parser/search_parameter_parser.rb
+++ b/lib/parameter_parser/search_parameter_parser.rb
@@ -102,18 +102,10 @@ private
     [SORT_MAPPINGS.fetch(field, field), dir]
   end
 
-  # Get a list of the fields to request in results from elasticsearch
   def return_fields
     fields = character_separated_param("fields")
-    if fields.empty?
-      return DEFAULT_RETURN_FIELDS
-    end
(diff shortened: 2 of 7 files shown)


rboulton:
A few brief thoughts:

I'd be concerned about performance impacts of fetching all fields from elasticsearch even when they're not needed.  Some of the fields can be very large (eg, the indexable_text field), so this could be a significant impact; both due to network transfer, and also due to forcing elasticsearch to look up more data from its index.  It would need to be measured, but I think this could be significant.

In terms of API design, I'm happy to see the default set of fields going away.  This default was set to work with existing frontends which expected to be able to get a specific set of fields returned; and it was too hard to remove the default set.

However, I'd advise against making all the fields be returned as a default, even without performance concerns, because I think it would make future modifications to the indexing schema harder to make without risking breaking existing applications.  Currently, if you need to remove a field (other than those in the default fields list), you can look through the codebase, or through requests in kibana, for search requests which ask for that field to be returned.  This will show all requests using that field, so you can ensure that the field isn't being used before removing or renaming it.  With a default of all fields being returned, this wouldn't be possible.  Being able to do this was very useful in a few cases in the past.

I'd actually advocate one of two options:
- If no fields are requested, don't return any fields.  The search may still return useful information (facet information, or counts of results), so this can be a useful state.
- Make a list of fields be a required parameter, so any use of the search API without specifying the fields to return returns an error, explaining what the problem is.  Most of the search API is designed to be fussy about invalid requests, to make it easier to make changes behind the scenes, so this would be consistent with the existing API design - but might be annoying for casual users.

Finally: what's missing from the search API at the moment is a way to list all the existing fields.  This can be inspected in the source code, but there's no way to get it from the API.

carvil:
Thanks for your comments, @rboulton !

Those are all fair points. I didn't consider performance issues in as I wasn't quite sure how to measure them. My assumption was that saving developer time would yield more benefits, but that's not a good way to put this.

I quite like your suggestion of enforcing the `fields` parameter. That would give a much more granular control over what fields are used or not, and in all cases we would only fetch fields that are necessary, which is a good thing. From a consumer perspective, it would be instantly clear what parameters are needed (by seeing an error message), rather than trying to figure out why the search results have no fields whatsoever.

Furthermore, we could solve the issue of listing what fields are possible to fetch by:
- adding a new endpoint that lists what fields are possible to fetch;
- or include them in the search results based on a query param (e.g. `list_fields=true`)

Would you have a recommendations on that?

PR #713, opened 2016-11-02
Rename contact_group to contact_groups

rubenarakelyan:
This commit renames the `contact_group` field in the `contact` schema to `contact_groups` and the type to `identifiers` to reflect the fact that there can be more than one contact group. The value of the field is an array of slugs.

Trello: https://trello.com/c/wD10DEXt/260-migrate-the-hmrc-contacts-page-to-a-finder

### DIFF
--- a/config/schema/document_types/contact.json
+++ b/config/schema/document_types/contact.json
@@ -1,5 +1,5 @@
 {
   "fields": [
-    "contact_group"
+    "contact_groups"
   ]
 }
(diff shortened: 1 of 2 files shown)


Davidslv:
This seems to make sense to me.
👍

tijmenb:
Note that we can rename this so easily because there's no data in the `contact_group` field currently. Otherwise we would've had to arrange a migration.

PR #714, opened 2016-11-09
Allow faceting for world locations, policy areas and taxons

tijmenb:
World locations and policy areas are used in the whitehall finders like https://www.gov.uk/government/announcements. We're building new finders for those, so to keep feature parity we need to be able to facet on those attributes.

The taxons facet we'd like to use to experiment in finders too.


rubenarakelyan:
🚢

PR #715, opened 2016-11-14
Add start and end dates to edition schema

MatMoore:
These are used for topical events to capture their topicalness.

Currently we store this in the placeholder content schema, but not rummager
https://github.com/alphagov/govuk-content-schemas/blob/0bda3ae3e0be97cba8849eb700dfd34d9c4c3d05/dist/formats/placeholder/frontend/schema.json

This will allow us to replace /government/topical-events with a finder
of topical events, and /government/topics with a finder of policy areas,
so we have one less thing being rendered from whitehall data using a custom
template.

Currently whitehall shows any with an end date in the future: https://github.com/alphagov/whitehall/blob/master/app/models/topical_event.rb#L38

Trello: https://trello.com/c/oVfSAO9W/311-migrate-policy-areas-index-to-a-finder

### DIFF
--- a/config/schema/document_types/edition.json
+++ b/config/schema/document_types/edition.json
@@ -6,6 +6,7 @@
     "display_type",
     "document_collections",
     "document_series",
+    "end_date",
     "government_name",
     "has_act_paper",
     "has_command_paper",
@@ -28,6 +29,7 @@
     "search_format_types",
     "slug",
     "specialist_sectors",
+    "start_date",
     "statistics_announcement_state",
     "world_locations"
   ]
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -93,6 +93,16 @@
     "type": "date"
   },
 
+  "start_date": {
+    "type": "date",
+    "description": "Start date for topical content."
+  },
+
+  "end_date": {
+    "type": "date",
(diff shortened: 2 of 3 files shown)


config/schema/document_types/edition.json
```diff
@@ -29,6 +29,8 @@
     "slug",
     "specialist_sectors",
     "statistics_announcement_state",
-    "world_locations"
+    "world_locations",
+    "start_date",
+    "end_date"
```
tijmenb: Can these go in alphabetically?

tijmenb (changes requested):

config/schema/document_types/edition.json
MatMoore: Have made that change.

Not sure if anything else is needed on this PR - I tried writing a simple integration test to check the field is being used but I'm either misunderstanding what the test's doing or I've forgotten something in the config because it's not showing up yet.

tijmenb:
Can you push the integration test? I believe this should work.

tijmenb (approved):

MatMoore:
<strike>@tijmenb I've pushed the WIP</strike>

Nevermind, it's because I used the wrong field name 😞

MatMoore:
Added a basic test that just checks the configuration works.

Next step after this is deployed is to make whitehall send these fields.

PR #718, opened 2016-11-25
Allow indexing of `organisation_type`

tijmenb:
This field type already exists in the field definitions, but was never used in a schema. I suspect this was forgotten when first implementing the schemas. We'll be sending this from Whitehall soon to allow us to build a organisation finder with this type as a facet.

Trello: https://trello.com/c/FRYiVa3U

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -162,6 +162,7 @@
   },
 
   "organisation_type": {
+    "description": "The organisation type identifier, like 'ministerial_department' or 'public_corporation'. Only applies to organisations. Expected value is the `organisation_type_key` from Whitehall, enumerated in https://github.com/alphagov/whitehall/blob/master/app/models/organisation_type.rb.",
     "type": "identifier"
   },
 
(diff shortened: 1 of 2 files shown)


config/schema/field_definitions.json
```diff
@@ -162,6 +162,7 @@
   },
 
   "organisation_type": {
+    "description": "The organisation type identifier, like 'ministerial_department' or 'public_corporation'. Only applies to organisations. Expected value is the `organisation_type_key` from Whitehall, enumerated in https://github.com/alphagov/whitehall/blob/master/app/models/organisation_type.rb.",
```
MatMoore, line 165: Is it worth making this an enum in the schema as well?

tijmenb: I don't think that's possible in rummager, but maybe I'm missing something.

MatMoore: Nope, I thought I was a looking at govuk-content-schemas - ignore me

PR #719, opened 2016-11-29
Remove self-hosted documentation

tijmenb:
In https://github.com/alphagov/rummager/pull/701, https://github.com/alphagov/rummager/pull/703 and https://github.com/alphagov/rummager/pull/704 we configured the /docs directory to be served via GitHub pages.

This didn't quite work as expected. To serve usable pages we need some more styling and a better structure so that navigation is easy.

The idea now is that https://github.com/alphagov/govuk-developers will replace the hosted version of the docs. It will pull in the docs from this repo via the GitHub API and display them nicely formatted. This will bring all our docs to a central place where they are easily discoverable.


rubenarakelyan:
👍

PR #720, opened 2016-11-29
Rename document_type to elasticsearch_type

tijmenb:
This PR aims to avoid confusion about the concept "document type".

- The word `document_type` is used by the content store to describe the type of a content item. See [RFC 41 for the background](https://gov-uk.atlassian.net/wiki/display/GOVUK/RFC+41%3A+Separate+document+type+from+format).

- But "document type" is also used by this application to refer to the [elasticsearch type of the document](https://www.elastic.co/guide/en/elasticsearch/guide/current/mapping.html).

To avoid confusion and allow us to use `document_type` consistently across GOV.UK this commit renames `document_type` to `elasticsearch_type` where applicable.

This should not change anything for clients of this API. There are 2 todo's after this PR:

1. We currently allow clients to filter on `document_type`. This needs to be removed or renamed to `elasticsearch_type`.
2. We currently return `document_type` in the payload by default. This needs to be removed or renamed too.

## How to review

- [ ] Read through the entire diff to see if the change makes sense and I haven't made a typo
- [ ] Check out the branch and see if the remaining occurrences of `document_type` make sense

Trello: https://trello.com/c/C8A9bnuO


MatMoore:
This rename makes sense to me, and the only places I can see where the old name is still being is used are:
- The message queue consumer via lib/indexer/message_processor.rb
   This is only when logging the payload of a message. It doesn't look like we use `document_type` from incoming messages, so maybe we should stop the publishing api from sending it?
- The result presenter (has a TODO)
- Filter params (has a TODO)

I think whatever we do with `document_type` should be consistent across the API. The endpoints for indexing include the document types as well so rummager doesn't hide that detail even if we dropped the filter/result field. I think a rename to elasticsearch_type would be the simplest thing for now.

 I'm assuming that when we index from the publishing api the two things (elasticsearch_type and document_type/format) would converge anyway?

MatMoore (approved):

tijmenb:
> I think whatever we do with document_type should be consistent across the API. The endpoints for indexing include the document types as well so rummager doesn't hide that detail even if we dropped the filter/result field. I think a rename to elasticsearch_type would be the simplest thing for now.

Not sure that I get this, can you elaborate?

> I'm assuming that when we index from the publishing api the two things (elasticsearch_type and document_type/format) would converge anyway?

I'd actually expect `elasticsearch_type` and `schema_name` to converge, since they both describe the data.

MatMoore:
> I'd actually expect elasticsearch_type and schema_name to converge, since they both describe the data.

Oh yeah, that makes more sense.

> Not sure that I get this, can you elaborate?

I just mean that the way that an API user deals with that field should be consistent across endpoints, and the concept is exposed in a few different places at the moment.

`POST or DELETE /:index/documents` supports a `_type` field
`GET /content/?link=/a-link` returns `_type`
`GET /search` has `filter_document_type` and returns `document_type`

tijmenb:
Good point, we'll update those as well!

PR #721, opened 2016-11-30
Add rake task to report inconsistent data

rubenarakelyan:
This commit adds a rake task that looks for and reports via Airbrake any facet values that are non-expanded due to unsynced data between rummager and the publishing-api.

It also adds `document_series` to the list of allowed facet fields so that it can also be tested.

`gds-api-adapters` is upgraded so that `rummager.search` can be used.

Trello: https://trello.com/c/JMFQcEj6/337-finder-frontend-displays-blank-values-in-facets

In Errbit:

screen shot 2016-11-30 at 14 21 44


lib/tasks/report_inconsistent_facet_values.rake
```diff
@@ -0,0 +1,55 @@
...
+
+  rummager_base_url = Plek.new.find("rummager")
+  facet_values_to_report = {}
+
+  facets.each do |facet|
+    puts "Checking #{facet} facet..."
+    facet_values_to_report[facet] = []
+
+    # Return 1000 facet values by default and suppress actual results
+    uri = URI.parse("#{rummager_base_url}/search?facet_#{facet}=1000&count=0")
```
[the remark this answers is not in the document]
rubenarakelyan: Yup, changed this

lib/tasks/report_inconsistent_facet_values.rake
```diff
@@ -0,0 +1,57 @@
...
+
+    if response.code != 200
+      puts "Error getting data for #{facet} facet: code #{response.code}"
+      next
+    end
+
+    facet_values = response.to_hash.dig("facets", facet, "options")
+
+    facet_values.each do |facet_value|
+      if (facet_value["value"].keys - ["slug"]).empty?
```
tijmenb: Could this also be `facet_value["value"].keys == ["slug"]`?

lib/tasks/report_inconsistent_facet_values.rake
```diff
@@ -0,0 +1,57 @@
+require "gds_api/rummager"
+require "json"
+
+class EmptyFacetValueError < StandardError
```
tijmenb: An "empty facet" is perhaps more the symptom than the cause. Maybe call this something like `DataInconsistencyError`?

lib/tasks/report_inconsistent_facet_values.rake
```diff
@@ -0,0 +1,57 @@
...
+end
+
+desc "Check for and report any non-expanded facet values"
+task :report_inconsistent_facet_values do
+  # List of facets to check
+  facets = %w{
```
tijmenb: Perhaps we should check all the allowed facets that use the expansion mechanism:

https://github.com/alphagov/rummager/blob/master/lib/search/registries.rb#L11-L26

tijmenb (changes requested):

lib/tasks/report_inconsistent_facet_values.rake
rubenarakelyan: Yup

lib/tasks/report_inconsistent_facet_values.rake
rubenarakelyan: Yup

lib/tasks/report_inconsistent_facet_values.rake
rubenarakelyan: I've added `document_collections` and `specialist_sectors` but the other ones (`document_series`, `organisation_content_ids` and `topic_content_ids`) are not valid facet fields.

For example, an error from rummager:

`{"error":"\"document_series\" is not a valid facet field"}`

tijmenb: Hmm, could you add them to the `ALLOWED_FACET_THINGS` constant (and reuse that here)? 

They could have the same problem as the others.

tijmenb: I suspect you could use `Rummager.registries. as_hash.keys.map(&:to_s)`. If not then copy paste is fine!

rubenarakelyan: I'm using `ALLOWED_FACET_FIELDS` now with `document_series` added in.

tijmenb (approved):

tijmenb:
Awesome, good to merge!

PR #723, opened 2016-12-02
Add Jenkinsfile for Rummager

surminus:
This adds the Jenkinsfile to Rummager to allow builds on the Jenkins pipeline.

This has now been successfully tested, although build times are slower than the current CI environment at ~13 minutes rather than ~5 minutes. 

The release tag pushing is still disabled in the PR, but if we decide to migrate then the steps should be:

1) Push new commit enabling tag pushes
2) Update build number on new CI to reflect the build number on the current CI environment
3) Update the webhook 
4) Merge the PR

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,101 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'rummager'
+DEFAULT_SCHEMA_BRANCH = 'deployed-to-production'
+
+node {
+  def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
+
+  properties([
+    buildDiscarder(
+      logRotator(
+        numToKeepStr: '50')
+      ),
+    [$class: 'RebuildSettings', autoRebuild: false, rebuildDisabled: false],
+    [$class: 'ThrottleJobProperty',
+      categories: [],
+      limitOneJobWithMatchingParams: true,
+      maxConcurrentPerNode: 1,
+      maxConcurrentTotal: 0,
+      paramsToUseForLimit: 'Rummager',
+      throttleEnabled: true,
+      throttleOption: 'category'],
+    [$class: 'ParametersDefinitionProperty',
+      parameterDefinitions: [
+        [$class: 'BooleanParameterDefinition',
+          name: 'IS_SCHEMA_TEST',
+          defaultValue: false,
+          description: 'Identifies whether this build is being triggered to test a change to the content schemas'],
+        [$class: 'StringParameterDefinition',
+          name: 'SCHEMA_BRANCH',
+          defaultValue: DEFAULT_SCHEMA_BRANCH,
+          description: 'The branch of govuk-content-schemas to test against']]
+    ],
+  ])
+
+  try {
+    govuk.initializeParameters([
+      'IS_SCHEMA_TEST': 'false',
+      'SCHEMA_BRANCH': DEFAULT_SCHEMA_BRANCH,
+    ])
+
+    if (!govuk.isAllowedBranchBuild(env.BRANCH_NAME)) {
+      return
+    }
+
+    stage("Checkout") {
+      checkout scm
+    }
+
+    stage("git merge") {
+      govuk.mergeMasterBranch()
+    }
+
+    stage("bundle install") {
+      govuk.bundleApp()
+    }
+
+    stage("rubylinter") {
+      govuk.rubyLinter('app test lib')
+    }
+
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,73 @@
...
+      govuk.rubyLinter()
+    }
+
+    stage("Run tests") {
+      govuk.setEnvar('USE_SIMPLECOV', 'true')
+      govuk.setEnvar('RACK_ENV', 'test')
+      govuk.runRakeTask('ci:setup:minitest test --trace')
+    }
+
+    // Commented out for testing
```
MatMoore: If it's going to stay commented out I think you should update this comment. It sounds like we're disabling this until we switch over to the new system, is that right?

Jenkinsfile
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    stage("bundle install") {
+      govuk.bundleApp()
+    }
+
+    stage("Delete static symlinks and reconnect") {
+      sh '''for d in images javascripts templates stylesheets; do
+              rm -f public/$d
+              ln -s ../../../Static/workspace/public/$d public/
```
djsd123: Looking at current and previous jenkins agents.  The specified directory does not exist and creates dead symlinks.  Is this expected?

Jenkinsfile
```diff
@@ -0,0 +1,73 @@
...
+            done
+         '''
+    }
+
+    stage("rubylinter") {
+      govuk.rubyLinter()
```
djsd123: in the `jenkins.sh` the last line in the `if` block states "test", where the new ruby linter function states `spec`.

MatMoore:
Do we know why build times are slower? Would be great to speed that up if we can. Waiting 13m to get to a deploy can be quite annoying if you're just making minor improvements on a PR 🙁

djsd123:
Looks good.  One query.

In the original job on the old CI, two post build steps are specified.  Were these omitted deliberately?

surminus:
@MatMoore build times are slow for an unknown reason, but I suspect underlying infrastructure. I will be looking to migrate the CI machines over to more performant hardware next week. We obviously don't want to switch over while build times are almost three times of the current CI.

surminus:
@djsd123 great point, and I overlooked that because I just looked at the "rummager_branches" build job, which means there is some extra stuff still required on this. Thanks!

Jenkinsfile
surminus: @MatMoore do you know if there's any difference between running [this](https://github.com/alphagov/rummager/blob/master/jenkins.sh#L20-L25) and [this]( https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy#L54-L59) in this repo?

Jenkinsfile
surminus: @MatMoore as per the above can this step be removed?

surminus:
@MatMoore we've upgraded the CPUs and the build time took 8 mins 52 seconds. I hope that's more acceptable, although I note it still feels slower than the current CI.

Jenkinsfile
suzannehamilton: I was looking into this because I'm adding a Jenkinsfile to some other projects.

The last line determines which directories are analysed by Rubocop. So "spec" will be appropriate for some repos, and "test" will be appropriate for others.

It might be safe to include both, i.e. make the last line `app test spec lib` to cover most (all?) projects. The linting tool handles missing directories gracefully, but only if you use the `--diff` option, which we do. If there are projects with a different structure, maybe we could also modify the puppet script so that it accepts extra directories to include when linting?

tijmenb: Perhaps we could just omit the directories to lint by default, in most projects we want to lint everything anyway.

surminus: I'll do a PR for the groovy script and see if anyone objects to just linting everything.

surminus: Gone with the following approach instead:
https://github.com/alphagov/govuk-puppet/pull/5272

surminus:
@suzannehamilton I've pushed some changes (different rubylinter options and removed the linking of images as it looks superfluous). Can you let me know if this looks OK?

suzannehamilton:
I've added build parameters so this can be used to test changes to the content schema, and removed the commented-out release tagging, because it's OK for Jenkins 2 to do the release tagging once this has been merged to master.

https://trello.com/c/S4DZbsPa/285-jenkins-2-migration

tijmenb (approved):

PR #729, opened 2016-12-15
Add assessment_date to the allowed sort fields

klssmith:
- This is to allow service standard reports to be sorted by assessment
date.

Trello:
https://trello.com/c/ADPnDXsE/385-add-assessment-date-to-service-standard-specialist-document


tijmenb (approved):
👍

PR #733, opened 2017-01-18
Treat search input keywords as strings

tijmenb:
This morning a bot visited the site using weirdly formatted query strings, presumably some automated SQLI attempt.

The request returned 500 because we expect `keywords` to be a string. This commit casts it to a string by default, so that it won't error and warn us via errbit.


MatMoore:
Could we just reject these requests with a 400 in the parameter parser instead?

Then we could treat all the parameters the same way (if they're not supposed to be an array don't allow it)

PR #736, opened 2017-01-20
Add content_store_document_type field to schema

alecgibson:
Add a `content_store_document_type` field to the search schema. This is to allow us to populate at least Guidance material with the corresponding Content Store Document Type in the appropriate publishing applications.

Trello: https://trello.com/c/M1cXENeB/341-add-content-store-document-type-to-all-guidance-content-in-rummager


tijmenb (approved):

tijmenb:
Can you also add it to the allowed facet fields? That will allow us to fetch all the distinct document types in the database.

PR #741, opened 2017-01-30
Make hidden_indexable_content field searchable

lgarvey:
Each MOJ finder has a field called "hidden_indexable_content" which is
used by HMCTS staff to manually copy/paste PDF text into the document,
purely for search purposes.

Previously content in the hidden_indexable_content field was
searchable, but this seems to have stopped working. Strangely I see there has
never been a hidden_indexable_content field in the rummager
field_definitions schema.

I have limited knowledge of specialist-publisher and rummager, so it may be that further amends are required to allow the field to be searchable.

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -483,6 +483,9 @@
   "tribunal_decision_reference_number": {
     "type": "identifier"
   },
+  "hidden_indexable_content": {
+    "type": "searchable_text"
+  },
   "dfid_authors": {
     "description": "A set of author names, which aren't selected from a predefined list and don't repeat",
     "type": "searchable_identifiers"


tijmenb:
Hey @lgarvey,

`hidden_indexable_content` is a specialist-publisher specific thing. It is not supposed to be sent to rummager separately, but combined with other content and sent as `indexable_content`.

It seems this was forgotten in the rebuild of specialist publisher.

So this line:

https://github.com/alphagov/specialist-publisher/blob/f8e93142dfad6f3971a73c923b01f2e7352bdb54/app/presenters/search_presenter.rb#L25

should be something like:

```ruby
def indexable_content
  Govspeak::Document.new(document.body).to_text + " " + document.hidden_indexable_content
end
```

lgarvey:
Hey @tijmen,

Thanks for the quick response.  OK that makes perfect sense - I initially
thought ti was being combined/processed in the specialist-publisher repo,
but couldn't see where.  Do you want me to go ahead and make this amend in
specialist-publisher?

tijmenb:
Yeah, that'd be great!

PR #742, opened 2017-01-31
Add support for “more like this” queries

carvil:
This commit adds support in rummager for “more like this” queries to be sent to Elasticsearch. “More like this” queries allow searching using an existing indexed document as a starting point, and returning a set of documents that are “similar” to it content-wise.

A “more like this” search can be carried out using the existing rummager search endpoint with the new `similar_to` option, which takes a base path as its value (more correctly, the ID of the document in the Elasticsearch index). These searches cannot be combined with a text query (using the `q` option) and they also cannot be custom sorted (they are already sorted by similarity).

### DIFF
--- a/lib/parameter_parser/search_parameter_parser.rb
+++ b/lib/parameter_parser/search_parameter_parser.rb
@@ -31,7 +31,8 @@ private
     @parsed_params = {
       start: single_integer_param("start", 0),
       count: capped_count,
-      query: normalize_query(single_param("q")),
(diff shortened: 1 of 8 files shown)


carvil:
Closing this as we decided not to pursue this route for now.

tijmenb:
See: https://trello.com/c/8hyrTPXO/344-sidebar-content#comment-5890b6ce622476e6c343ad72

PR #744, opened 2017-02-03
Return an error for facet parameters with empty values

rubenarakelyan:
If a user requests a facet parameter with an empty value (for example /search.json?facet_policies) then rummager currently throws an exception when attempting to split the nil value.

This commit adds a check to change the nil value to an empty string, allowing the rest of the function to complete and display an error message.

Trello: https://trello.com/c/6LLgmAFM/470-return-4xx-from-rummager-when-making-a-facet-request-request-without-value


Davidslv (approved):
👍

PR #745, opened 2017-02-07
Catch publishing-api exceptions

rubenarakelyan:
This commit adds `rescue` blocks to catch timeout and HTTP exceptions raised by gds-api-adapters when communicating with the publishing-api.

Catching these exceptions allows us to log them and send them manually to Airbrake with custom messages such that errors are grouped in a sane manner.

This will prevent Airbrake seeing each message as a separate error to be reported.

Trello: https://trello.com/c/Ku9HEzFc/467-rummager-should-bundle-errors-to-avoid-spamming-from-errbit

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -27,5 +27,11 @@ configure do
   # parameters for debugging. Ignore it here so that we don't send them twice.
   Airbrake.configuration.ignore << "Indexer::BulkIndexFailure"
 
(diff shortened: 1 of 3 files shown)


MatMoore (approved):
I think this workaround is OK for now. Let's hope we retire errbit soon!

PR #746, opened 2017-02-10
Speed up integration tests

carvil:
This commit speeds up the entire test suite from:

```
Finished tests in 457.060166s, 0.3063 tests/s, 0.6564 assertions/s.
140 tests, 300 assertions, 0 failures, 0 errors, 0 skips
```

To:

```
Finished tests in 48.789358s, 2.8695 tests/s, 6.1489 assertions/s.
140 tests, 300 assertions, 0 failures, 0 errors, 0 skips
```

It does so by changing the approach we have been using so far.
Previously, we were deleting and creating the test indexes before every
test. Even though this provides a clean environment for each test to
run, it is also quite expensive in terms of time spent.

The approach I'm proposing only creates the test indexes once, on
startup, and makes sure all data on those indexes is deleted on every
test run (on teardown more specifically).  This also provides a clean
environment for each of the tests to run, since most tests don't deal
specifically with functionality that creates and deletes indexes.

It also makes sure we clean up test indexes after the entire test suite
runs. This isn't strictly necessary, because the test suite cleans them
up on startup, but a good thing to do to avoid leaving unwanted indexes
around.

Finally, this commit also moves test indexes' actions to a helper class.
This means we will be able to use it in other contexts without having to
require the test stack.

**NOTE**

There are exceptions, unfortunately. A few test cases still delete and
create certain indexes because that's exactly what they are testing. I
didn't remove this logic, meaning some tests will still be quite slow.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,7 +13,11 @@ require "config"
 Dir[File.join(PROJECT_ROOT, 'lib/tasks/**/*.rake')].each { |file| load file }
 
 desc "Run all the tests"
-task "test" => ["test:units", "test:integration"]
+task "test" => [
+  "test:units",
+  "test:integration",
+  'test:clean_test_indexes'
+]
 
 namespace "test" do
   desc "Run the unit tests"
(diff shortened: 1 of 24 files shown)


carvil:
The different in CI is still around 10x:

This branch:
Finished tests in 89.866474s, 1.5579 tests/s, 3.3383 assertions/s.


Master:
Finished tests in 854.835015s, 0.1638 tests/s, 0.3509 assertions/s.

tijmenb:
😍

tijmenb (approved):

PR #748, opened 2017-02-13
Index parents of the taxon for each page

klssmith:
Content tools would like to be able to fetch all content related to education.
We added a filter, part_of_taxonomy, that returns everything that is tagged
to the given taxon and any of that taxon's children.

Trello card: https://trello.com/c/dIMlC5JH/481-index-parents-of-the-taxon-for-each-page

Paired with @MatMoore

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -85,6 +85,11 @@
     "type": "identifiers"
   },
 
+  "part_of_taxonomy": {
+    "description": "Any taxon tagged to the document and any of their ancestor taxons",
+    "type": "identifiers"
+  },
+
   "updated_at": {
     "type": "date"
   },
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -85,7 +85,7 @@
     "type": "identifiers"
   },
 
-  "part_of_taxonomy": {
+  "part_of_taxonomy_tree": {
     "description": "Any taxon tagged to the document and any of their ancestor taxons",
     "type": "identifiers"
   },
--- a/test/integration/indexer/indexing_test.rb
+++ b/test/integration/indexer/indexing_test.rb
@@ -57,25 +57,6 @@ class ElasticsearchIndexingTest < IntegrationTest
     })
   end
 
-  def test_parts_of_taxonomy
-    post "/documents", {
-      "title" => "TITLE",
-      "format" => "organisation",
-      "slug" => "my-organisation",
-      "link" => "/an-example-organisation",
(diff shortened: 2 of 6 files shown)


config/schema/field_definitions.json
```diff
@@ -508,5 +508,10 @@
   "content_store_document_type": {
     "description": "The document type as stored in the Content Store",
     "type": "identifier"
+  },
+
+  "part_of_taxonomy": {
+    "description": "Any taxon tagged to the document and any of their ancestor taxons",
```
tijmenb: Does "ancestor" imply that it's the _parents_ that are included?

test/integration/indexer/indexing_test.rb
```diff
@@ -57,6 +57,25 @@ def test_tagging_organisations_to_self
     })
   end
 
+  def test_parts_of_taxonomy
```
tijmenb: Not sure about this integration test. Apps will never not send `part_of_taxonomy` directly via a POST.

Would it be better to skip this test and add something extra to the links lookup test?

test/integration/indexer/links_lookup_test.rb
```diff
@@ -80,6 +80,7 @@ def test_indexes_documents_with_links_from_publishing_api
       "specialist_sectors" => ["my-topic/a", "my-topic/b"],
       "mainstream_browse_pages" => ["my-browse/1"],
       "organisations" => ["my-org/1", "my-court"],
+      "part_of_taxonomy" => ["TAXON-1"],
```
tijmenb: It would be nice to add a full test of the behaviour here. So given a taxon with parents a & b, the  `part_of_taxonomy` of something tagged to taxon c will be a, b & c.

tijmenb (changes requested):

config/schema/field_definitions.json
MatMoore: Yes,

So if you have the taxonomy

- Schools
   - Key stage 1
      - Maths
   - Key stage 2
     - Maths

And you tag a document to key stage 1 maths and key stage 2 maths, then:

The field is set to [ks1 maths, ks2 maths, ks1, ks2, schools]

And the document is returned in any of these searches:
 - ks1 maths
 - ks2 maths
 - ks1
 - ks2
 - schools

tijmenb: Right of course. I confused myself, I think I was thinking of the way you retrieve the pages.

MatMoore: Yeah it's a bit counterintuitive

MatMoore:
Have added a more complete test and rebased this onto master.

tijmenb (approved):

PR #751, opened 2017-02-15
Add more metadata for business finance support scheme

rubenarakelyan:
This commit adds more metadata fields to the new “business finance support scheme” elasticsearch type to rummager for content items migrated from business-support-finder.

Trello: https://trello.com/c/UHsavPI5/485-create-a-new-specialist-document-type-for-business-support-scheme

### DIFF
--- a/config/schema/elasticsearch_types/business_finance_support_scheme.json
+++ b/config/schema/elasticsearch_types/business_finance_support_scheme.json
@@ -1,6 +1,7 @@
 {
   "fields": [
     "additional_information",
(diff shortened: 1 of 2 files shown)


tijmenb:
Does all of this need to be in rummager? Perhaps limiting to what the finder will use would make sense.

rubenarakelyan:
I've removed `area_gss_codes`, `locations` and `purposes` but everything else here will be used for faceting.

tijmenb (approved):

PR #756, opened 2017-02-28
Start adding document types to index

tijmenb:
The "document supertype" is a grouping for document types. It will allow us to segment pages by more types, like "Publication" or "Announcement" (used in Whitehall).

With this we'll be able to search like this:

```
https://www.gov.uk/api/search.json?filter_navigation_document_supertype=guidance
https://www.gov.uk/api/search.json?filter_whitehall_document_supertype=publication
```

See https://github.com/alphagov/govuk_document_types/pull/1.

https://trello.com/c/07lAcDtC

### DIFF
--- a/test/unit/elasticsearch_index_test.rb
+++ b/test/unit/elasticsearch_index_test.rb
@@ -87,28 +87,6 @@ class ElasticsearchIndexTest < MiniTest::Unit::TestCase
     end
   end
 
-  def test_should_bulk_update_documents_with_raw_command_stream
-    stub_tagging_lookup
-    stub_traffic_index
-    stub_popularity_index_requests(["/foo/bar"], 1.0)
-
-    # Note that this comes with a trailing newline, which elasticsearch needs
-    payload = <<-eos
-{"index":{"_type":"edition","_id":"/foo/bar"}}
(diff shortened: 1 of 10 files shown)


test/unit/elasticsearch_index_test.rb
```diff
@@ -137,63 +87,6 @@ def test_should_raise_error_for_failures_in_bulk_update
     end
   end
 
-  def test_should_bulk_update_documents_with_raw_command_stream
```
MatMoore, line 140: How come this was removed?

MatMoore:
I'm fine with the rummager changes, just wondering if we need a gem for this or we could add this into publishing api somehow.

Would it work have a document type of `document_type_group` and make each of these groupings a separate content item? (This may be a really terrible idea)

test/unit/elasticsearch_index_test.rb
tijmenb: Did you see the commit message?

tijmenb:
Good point, @carvil and I talked to the architects about this, but we didn't write down the outcome of that (which was to use this method). I'll write up some of the considerations today.

MatMoore (approved):

PR #758, opened 2017-03-02
Add docs for taxonomy fields

MatMoore:
https://trello.com/c/OaHe7Ylt/514-document-taxon-fields-in-rummager

Rendered markdown: https://github.com/alphagov/rummager/blob/24d5eb339fd56cbce22d05033182cfb3568511ff/docs/searching-the-taxonomy.md

### DIFF
--- a/docs/search-api.md
+++ b/docs/search-api.md
@@ -4,9 +4,8 @@ This API is the main endpoint for performing searches on GOV.UK.  It supports
 keyword searching, ordering by relevance or date fields, filtering and
 faceting.
 
-At the time of writing, there is one other endpoint, the `advanced_search`
-endpoint, which is slowly being replaced by the `search` endpoint.  The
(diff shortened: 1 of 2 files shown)


tijmenb:
Does this duplicate https://docs.publishing.service.gov.uk/guides.html#taxonomy? How could we better cross-link between these?

MatMoore:
There is some overlap, but they have a different focus. I think it's a bit too detailed to incorporate into that page, and not all rummager API users are GOV.UK developers. Agree we should cross link though.

Could we pull these docs into https://docs.publishing.service.gov.uk/apis/search-api.html, and then link to them from https://docs.publishing.service.gov.uk/guides.html#3-accessing-the-taxonomy ?

MatMoore:
Closing - will follow up with an edit to taxonomy docs.

PR #759, opened 2017-03-03
Rescue the correct error from links lookup

MatMoore:
Since 114f78dcbf23ca3f3ab019b6ca4baefdb90cf2f7 we've been raising a different
exception class. Even before that, we were returning nil from the method after
handling the error, so I don't think this has ever worked.

This means that the overnight job to reindex search was aborting early if
publishing api times out, which is happening a lot at the moment.
We don't care if we can't fetch the links while reindexing - we can fall back
to whats already there.

Also, even if the links lookup worked, the document type group method would
have failed, since we only require it when running as a sinatra app.

The job has been failing since Jan 28, but we didn't notice this because of a
separate issue with our jenkins job:
https://github.com/alphagov/search-analytics/pull/11


tijmenb (approved):
When green.

PR #761, opened 2017-03-09
Add Docker support

thomasleese:
So that we can run rummager in our e2e test environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM ruby:2.3.1
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential && apt-get clean
+
+ENV REDIS_HOST redis
+ENV ELASTICSEARCH_URI http://elasticsearch:9200
+ENV PORT 3009
+ENV RACK_ENV development
+
+ENV APP_HOME /app
+RUN mkdir $APP_HOME
+
+WORKDIR $APP_HOME
+ADD Gemfile* $APP_HOME/
+RUN bundle install
+ADD . $APP_HOME
+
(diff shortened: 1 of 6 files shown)


cbaines (approved):
Don't know much about the docker related changes, but the config ones look great :)

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+ENV APP_HOME /app
+RUN mkdir $APP_HOME
+
+WORKDIR $APP_HOME
+ADD Gemfile* $APP_HOME/
+RUN bundle install
+ADD . $APP_HOME
+
+CMD bash -c "bundle exec mr-sparkle --force-polling -- -p $PORT -o '0.0.0.0'"
```
tijmenb: If you're running this in non-development mode you can probably run it like in production. `mr-sparkle` is regularly causing problems.

kevindew (approved):

config.ru
```diff
@@ -11,18 +11,21 @@ require "logger"
...
   enable :logging
-  log = File.new("log/production.log", "a")
+  log = File.new(log_path + "/#{ENV['RACK_ENV']}.log", "a")
   log.sync = true
   STDOUT.reopen(log)
   STDERR.reopen(log)
 
   use Rack::Logstasher::Logger,
-    Logger.new("log/production.json.log"),
+    Logger.new(log_path + "/#{ENV['RACK_ENV']}.json.log"),
```
kevindew: Sorry - I'd expect log_path to be full path to log file. See https://github.com/alphagov/publishing-e2e-tests/blob/master/docker-compose.yml#L58

kevindew (changes requested):

kevindew (approved):

PR #762, opened 2017-03-09
Add and use the govuk_taxonomy_helpers gem

klssmith:
Added the govuk_taxonomy_helpers gem. We were then able to replace some
existing methods with a method from the gem.

Trello card: https://trello.com/c/PBIEgMMB/520-rummager-cleanup-use-new-taxonomy-gem

Paired with @MatMoore

### DIFF
--- a/lib/indexer/links_lookup.rb
+++ b/lib/indexer/links_lookup.rb
@@ -1,4 +1,5 @@
 require 'services'
+require 'govuk_taxonomy_helpers'
 
 # LinksLookup finds the tags (links) from the publishing-api and merges them into
 # the document. If there aren't any links, the payload will be returned unchanged.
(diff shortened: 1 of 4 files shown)


test/integration/indexer/links_lookup_test.rb
```diff
@@ -120,6 +125,9 @@ def test_extracts_parts_of_taxonomy
       "content_id" => grandparent_1_content_id,
       "base_path" => "/grandparent-1",
       "title" => "Grandparent 1",
+      "details" => {
```
tijmenb, line 128: Why does it need the `internal_name` here?

klssmith: We needed to add the internal name to the tests because the method we used from the govuk_taxonomy_helpers gem looks for the internal_name of the content item: https://github.com/alphagov/govuk_taxonomy_helpers/blob/master/lib/govuk_taxonomy_helpers/publishing_api_response.rb#L26-L31

MatMoore: I've raised a separate PR to make the gem more lenient: https://github.com/alphagov/govuk_taxonomy_helpers/pull/7

MatMoore:
Bumped this to v0.1.0 of the gem.

lib/indexer/links_lookup.rb
```diff
@@ -23,12 +24,19 @@ def prepare_tags(doc_hash)
       content_id = find_content_id(doc_hash)
       return doc_hash unless content_id
 
+      content_item = {
```
tijmenb, line 27: Hmm, maybe this is related to https://github.com/alphagov/email-alert-service/pull/67#discussion_r106368769. Could we somehow avoid doing this just for the sake of the gem? Seems we're fighting the API here. @MatMoore

MatMoore: Yeah, I was thinking the same thing. But I'm not sure about what to do about it yet.

At the moment, all the fields are optional, so 
if https://github.com/alphagov/govuk_taxonomy_helpers/pull/10 is merged, we could just call that method. This would give you a `LinkedContentItem` with a lot of null fields on it, but the `taxons` and `taxons_with_ancestors` methods would still work. This API still seems problematic though (having one big class with a lot of optional attributes/behaviour), so it may need revisiting.

tijmenb:
We're going to get to this PR next week. Closing in the meantime.

PR #766, opened 2017-03-20
Allow multiple values for business finance support scheme facets

rubenarakelyan:
This commit allow business finance support schemes to have multiple values for each facet.

Trello: https://trello.com/c/UHsavPI5/1102-create-a-new-specialist-document-type-for-business-support-scheme

Deployment checklist:

- [x] Deploy rummager
- [x] Run the rake task `republish:document_type["business_finance_support_scheme"]` in specialist-publisher

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -529,21 +529,21 @@
 
   "business_sizes": {
     "description": "The sizes of business served by a business finance support scheme",
-    "type": "searchable_identifier"
+    "type": "searchable_identifiers"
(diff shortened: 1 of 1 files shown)


whoojemaflip (approved):

tijmenb:
@rubenarakelyan could we make something so that we would have noticed this in development or test?

rubenarakelyan:
@tijmenb Possibly some kind of test against the content schema examples since those would have arrays for these fields.

PR #770, opened 2017-03-28
Configure Errbit via env vars

kevindew:
This removes the file that is added at deploy time. There are related PRs going into puppet, deployment and alphagov-deployment.


cbaines (approved):

kevindew:
Thanks for the reviews @cbaines 👍

PR #771, opened 2017-04-24
Small refactor of best bets checker

dwhenry:
Simplify the code to try and make it more readable
and easier to understand.

### DIFF
--- a/lib/search/best_bets_checker.rb
+++ b/lib/search/best_bets_checker.rb
@@ -9,16 +9,12 @@ module Search
     end
 
     def best_bets
-      unless @fetched
-        fetch
-      end
+      fetch
       @best_bets
     end
 
     def worst_bets
-      unless @fetched
-        fetch
-      end
+      fetch
       @worst_bets
     end
 
@@ -30,14 +26,15 @@ module Search
 
     # Fetch the best bets, and populate @best_bets and @worst_bets
     def fetch
+      return if @fetched
       if @query.nil?
-        best = []
-        worst = []
+        @best_bets = {}
+        @worst_bets = []
       else
         best, worst = select_bets(fetch_bets)
+        @best_bets = combine_best_bets(best)
+        @worst_bets = combine_worst_bets(worst)
       end
-      @best_bets = combine_best_bets(best)
-      @worst_bets = combine_worst_bets(worst)
       @fetched = true
     end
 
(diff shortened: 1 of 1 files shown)


lib/search/best_bets_checker.rb
```diff
@@ -9,16 +9,12 @@ def initialize(query, index = nil)
     end
 
     def best_bets
-      unless @fetched
-        fetch
-      end
+      @fetched || fetch
```
MatMoore: Since you're not using the value, I think this is slightly clearer with `unless`, but on a single line: `fetch unless @fetched`

The confusing part to me is the attributes are being set by side effects of `fetch`, so the behaviour depends on previous calls.

You could perhaps avoid this by splitting the class up, but up to you whether that's worth doing. I'm thinking of something like
1. a value object for the combined best/worst bets
2. a thing that takes a query, fetches the data and does the combining

lib/search/best_bets_checker.rb
```diff
@@ -31,13 +27,13 @@ def metasearch_index
     # Fetch the best bets, and populate @best_bets and @worst_bets
     def fetch
       if @query.nil?
-        best = []
-        worst = []
+        @best_bets = {}
```
MatMoore: 👍

lib/search/best_bets_checker.rb
```diff
@@ -47,52 +43,35 @@ def fetch
...
-      combined
     end
 
     def combine_worst_bets(bets)
-      combined = Set.new
-      bets.each do |bet|
-        combined.add bet["link"]
-      end
-      combined.to_a
+      bets.map { |bet| bet['link'] }.uniq
```
MatMoore, line 60: does this need a `.to_set` on the end to keep the previous behaviour?

lib/search/best_bets_checker.rb
```diff
@@ -109,23 +88,20 @@ def select_bets(bets)
...
         # The search on the stemmed_query field is overly broad, so here we need
         # to filter out such matches where the query in the bet is not a
         # substring (modulo stemming) of the user's query.
-        unless stemmed_query_as_term.nil?
-          unless analyzed_users_query.include? stemmed_query_as_term
-            next
-          end
-        end
-        result << [bet_query, bet_type, details["best_bets"], details["worst_bets"]]
+        next if stemmed_query_as_term && !analyzed_users_query.include?(stemmed_query_as_term)
```
MatMoore: Might just be me, but I think the use of `compact` isn't obvious here. I'd prefer the `nil` value to be more explicit, like

```ruby
if stemmed_query_as_term && ....
  nil
else
  [bet_type, details["best_bets"], details["worst_bets"]]
end
```

MatMoore (commented):
Looks good, just commented on some minor stuff.

lib/search/best_bets_checker.rb
dwhenry: The `.to_a` was turning it back into an array so this just skips the set altogether.

lib/search/best_bets_checker.rb
dwhenry: It is a bit of a horrible pattern and I think you are right long term splitting the class would be a nicer solution, but was turning super hard to do small cleanup rather than a big refactor.

I have changed it so the `@fetched` check is inside the `fetch` method which I think is nicer

lib/search/best_bets_checker.rb
dwhenry: done

lib/search/best_bets_checker.rb
```diff
@@ -47,52 +44,35 @@ def fetch
     #
     # If a link occurs at multiple positions, it will only occur in the group for
     # the highest position it occurred at.
+    # Where high means the smallest positional value.
```
suzannehamilton, line 47: 👍

suzannehamilton (approved):

PR #772, opened 2017-04-25
Downweight service standard reports

suzannehamilton:
Downweight search results for documents with format `service_standard_reports`.

These reports rank highly in search results because the page title is a user-friendly service names. They also cause duplicate titles in search results when a service has had multiple assessments.

https://trello.com/c/gCeAhOCh/57-make-service-standard-reports-less-prominent-and-misleading-in-search-results

Marked as DO NOT MERGE for now because I'm going to double-check the exact weighting with a performance analyst.


suzannehamilton:
Tara's approved the 0.2 weighting, so this is ready to review.

MatMoore (approved):

PR #774, opened 2017-05-03
Update descriptions for dates

tijmenb:
Part of trying to figure out what each date means and what to send.

@MatMoore do you agree that `public_timestamp` should be the same as publishing-api's `public_updated_at`?

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -103,10 +103,17 @@
   },
 
   "updated_at": {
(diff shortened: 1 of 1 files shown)


MatMoore (approved):
👍 That's what I'd expect.

I have no idea why `updated_at` is needed though. It doesn't have any useful meaning outside of a publishing app that I'm aware of.

It doesn't seem to be used for any of the specialist document finders, and I'd expect everything else to be using `public_timestamp`.

PR #775, opened 2017-05-12
Add support for ab_test configuration to be passed in

dwhenry:
A/B test variants should be controlled by upstream apps to ensure
users receive a consistent experience for the life of the test. This
requires the test configuration to be passed in.

We are using a single field to receive the test configuration to avoid
adding a new fields for each test that we run.

https://trello.com/c/43eBQfwf/80-prepare-the-infrastructure-for-the-minimum-should-match-a-b-test


MatMoore (approved):
👍

PR #777, opened 2017-05-15
Do not insert reserved words in Elasticsearch document

dwhenry:
The _type and _id fields are already included in the command hash so including
them in the document hash is redundant and appears to be causing the _type
field to be indexed (and added to the `_mapping` hash which is seen as a
blocker to upgrading elasticsearch.

# Warning

This change could have unforeseen consequences on the search index and needs to be tested on integration before being merged.

https://trello.com/c/mkn7SAn0/84-fix-reserved-field-name-errors-in-elasticsearch-2-x-upgrade

### DIFF
--- a/lib/indexer/bulk_payload_generator.rb
+++ b/lib/indexer/bulk_payload_generator.rb
@@ -78,7 +78,7 @@ module Indexer
         else
           [
             command_hash,
-            doc_hash,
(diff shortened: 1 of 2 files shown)


dwhenry:
Tested on Integration as build the index without the `_type` mapping

MatMoore (approved):

PR #778, opened 2017-05-17
Add licence-specific fields to Rummager

carvil:
These include:
- `licence_identifier`, which is the GDS_ID of a licence;
- `licence_short_description`, which is a short text description of a
  licence.

These fields will be used by `publisher` to send data to Rummager and
subsequently by `licence-finder` to use in the frontend.

We are replacing calls to the Content API with calls to Content Store and Rummager. In this case, we are particularly interested in removing [licences_for_ids](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/content_api.rb#L90-L93), which is basically a search query.

Trello: https://trello.com/c/vlqSHmTZ/76-make-licence-finder-use-something-else-to-fetch-licenses

### DIFF
--- a/config/schema/base_elasticsearch_type.json
+++ b/config/schema/base_elasticsearch_type.json
@@ -11,6 +11,8 @@
     "navigation_document_supertype",
     "is_withdrawn",
     "latest_change_note",
+    "licence_identifier",
+    "licence_short_description",
     "link",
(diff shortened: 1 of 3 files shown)


alecgibson (approved):

MatMoore:
👍

PR #780, opened 2017-05-18
Add minimum should match ab test query

brenetic:
The first ab test that we are running is a change to the minimum should match query. For variant b queries, minimum should match is `2<2` instead of `2<2 3<3 7<50%`.

https://trello.com/c/J2nUnoMm/99-decide-on-choice-of-change-for-a-b-test


brenetic:
Closed in favour of https://github.com/alphagov/rummager/pull/786

PR #781, opened 2017-05-18
Add health-check summary based on word count of search terms

suzannehamilton:
Summarise the search health-check by grouping terms by word count. This will help us assess how changes to the search API affect search queries of different lengths.

https://trello.com/c/gbhkiTCS/100-segment-healthcheck-output-by-word-count


dwhenry (approved):
I'm not 100% about this approach as long term we would not want to keep adding calculators for each variant that we think of.   

It does quickly solve our current problem.

PR #782, opened 2017-05-19
Move to aggregates

dwhenry:
Initial code to replace facets with aggregates in preparation of upgrading from ES 1.7 to 2.X

This code still needs to be tested with the first hurdle being that the healthcheck is not affected by these changes.

https://trello.com/c/lsUYGUtp/93-change-facets-to-aggregates


lib/tasks/report_inconsistent_facet_values.rake
```diff
@@ -19,43 +19,48 @@ task :report_inconsistent_facet_values do
...
-          facet_values: facet_values_to_report[facet]
+          aggregate: aggregate,
+          aggregate_values: aggregate_values_to_report[aggregate]
         }
       )
     end
   end
 end
+
+desc "Check for and report any non-expanded facet values [DEPRECIATE]"
```
MatMoore: "depreciate" should be "deprecate"

test/integration/search/search_test.rb
```diff
@@ -150,97 +150,121 @@ def test_only_contains_fields_which_are_present
...
         "documents_with_no_value" => 0,
         "total_options" => 2,
         "missing_options" => 0,
         "scope" => "exclude_field_filter",
       }
-    }, facets)
+    }, aggregate)
+  end
+
+  # backwards compatibility
```
MatMoore: Can you elaborate on this comment (eg "we changed facet -> aggregate but still support both")

Also: This method has the name same as the above one. Assuming it's meant to be `test_facet_counting`!

lib/search/presenters/result_set_presenter.rb
```diff
@@ -37,7 +37,7 @@ def present
         results: presented_results,
         total: es_response["hits"]["total"],
         start: search_params.start,
-        facets: presented_facets,
+        aggregates: presented_aggregates,
```
MatMoore: We are preserving backwards compatibility for facets in the request params, but this won't work if we remove `facets` from the response.

Assuming we are going ahead with the rename, I think we should make it return `facets` if you ask for facets and `aggregates` if you ask for aggregates (and just reject requests that mix the two).

MatMoore (changes requested):

MatMoore:
@dwhenry I've added some inline comments, but it seems like there isn't much left to get this mergeable, other than rebasing and squashing the commits. Are there any blockers you're aware of?

If we can simplify things like the facet example fetcher we should raise a separate card to look into this.

lib/search/query_components/aggregates.rb
```diff
@@ -0,0 +1,70 @@
...
+      # included in the result set.
+      if options[:scope] == :exclude_field_filter
+        aggregates_filter = filters_hash([field_name])
+      elsif options[:scope] == :all_filters
+        aggregates_filter = filters_hash([])
+      end
+
+      if aggregates_filter && aggregates_filter.count > 0
+        {
+          filter: aggregates_filter,
```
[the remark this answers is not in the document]
dwhenry: This was the second bit of refactoring I was speaking of that we can do once we are using aggregates. I haven't really looked at what is required just because I wanted to limit scope on this PR. :)

config/schema/field_types.json
```diff
@@ -1,6 +1,6 @@
 {
   "identifier": {
-    "description": "An identifier, can be used for exact (case and whitespace sensitive) lookup and facetting",
+    "description": "An identifier, can be used for exact (case and whitespace sensitive) lookup and aggregating",
```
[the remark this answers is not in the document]
dwhenry: I think we should use aggregations as the term internally as it describes the used technology, which (for the time being) supporting both `facet_` and `aggreagte_` naming of fields. Once we can move a to versioned API we can look to do a full migration and remove `facet_` naming.

lib/parameter_parser/base_parameter_parser.rb
```diff
@@ -29,8 +29,8 @@ class BaseParameterParser
   }.freeze
 
   # The fields listed here are the only ones which can be used to calculated
-  # facets for.  This should be a subset of allowed_filter_fields
-  ALLOWED_FACET_FIELDS = %w(
+  # aggrgates for.  This should be a subset of allowed_filter_fields
```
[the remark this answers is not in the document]
dwhenry: fixed

lib/parameter_parser/search_parameter_parser.rb
```diff
@@ -289,26 +289,27 @@ def filter_name_lookup(name)
...
   end
 
-  def facets
-    facets = {}
+  def aggregates
+    aggregates = {}
     @params.each do |key, _values|
-      if (m = key.match(/\Afacet_(.*)/))
+      # to ensure backwards compatibility we will support both facet_* and aggregate_* style naming
+      if (m = key.match(/\A(?:facet_|aggregate_)(.*)\Z/))
```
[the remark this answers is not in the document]
dwhenry: done

lib/search/query_components/aggregates.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def aggregates_hash_for_aggregate(field_name, options)
+      with_filters(
+        {
+          terms: {
+            field: field_name,
+            order: { _count: "desc" },
+            # We want all the aggregate values so we can return an accurate count of
+            # the number of options.
+            size: 100000,
```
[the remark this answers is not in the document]
dwhenry: I'm not really sure, I have left this as is for the moment add will add to the backlog for review/refactoring

lib/search/query_components/aggregates.rb
```diff
@@ -0,0 +1,70 @@
...
+        {
+          filter: aggregates_filter,
+          aggs: { 'with_applied_filter' => query }
+        }
+      else
+        query
+      end
+    end
+
+    # Possible duplication.
```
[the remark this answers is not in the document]
dwhenry: removed

lib/tasks/report_inconsistent_facet_values.rake
dwhenry: fixed

test/integration/search/search_test.rb
dwhenry: done

lib/search/presenters/result_set_presenter.rb
dwhenry: Made this value based on query criteria.

test/test_helper.rb
```diff
@@ -1,4 +1,5 @@
 ENV['RACK_ENV'] = 'test'
+require 'pry'
```
MatMoore, line 2: Assuming this is here intentionally I think pry_byebug would be better than pry.

MatMoore (dismissed):
Looks good 🚢

dwhenry:
@MatMoore Have tested this on the following pages:

* https://www-origin.integration.publishing.service.gov.uk/government/organisations/hm-revenue-customs/services-information
* https://docs.integration.publishing.service.gov.uk/document-types/announcement.html
* https://www-origin.integration.publishing.service.gov.uk/search?q=organisations&filter_organisations%5B%5D=department-for-international-development

And get identical results between this branch and the current master branch.

test/test_helper.rb
dwhenry: This add either `pry` or `pry-byebug` depending on what's included in the `gemfile`

config/schema/field_types.json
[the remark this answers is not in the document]
tijmenb: I would agree with the keeping faceting as the internal and external feature name. Aggregations seem to be an elasticsearch abstraction, and feel more like a implementation detail to rummager.

As Mat says faceting has an established meaning. Keeping the rummager faceting feature naming nicely separates the use case (faceted search) from the implementation (which uses the `elasticsearch facets` feature now, will start using the `aggregations` feature with this PR, but could use another implementation in the future).

brenetic (approved):
🚀

PR #783, opened 2017-05-19
Add index name as a field

dwhenry:
This has been added to perform two tasks:

* Allow the government index boost to be implemented the same as our other
boosts. It currently required a full copy of the query and uses
`no_match_query`.
* Merging data into a single index.

It should be noted that we are also looking to `rendering_app` and
`publishing_app` fields to the index, once this is done we can potentially
remove this field as it will be redundent.


suzannehamilton:
Is this change still necessary now that we're replacing index boosting with format boosting?

https://trello.com/c/FhscGKf2/127-refactor-code-where-weightings-are-applied

PR #786, opened 2017-05-22
Use minimum should match for weighting in a/b tests

brenetic:
The first ab test that we are running is a change to the minimum should match query. For variant b queries, minimum should match is now added for weighting in the should query params.

https://trello.com/c/J2nUnoMm/99-decide-on-choice-of-change-for-a-b-test

Mobbed with @JonathanHallam and @suzannehamilton.


MatMoore (approved):
🎉

PR #787, opened 2017-05-23
Add duplicate deleter rake task

dwhenry:
We can have duplicates when the type is different, but the content_id
and _id (link) are the same.

Paired with: @MatMoore 


tijmenb (approved):
👍

PR #788, opened 2017-05-23
[DO NOT MERGE] Remove dependence on _type field being in _source

MatMoore:
Elasticsearch documents are uniquely identified by the combination of their `_id`
and `_type` values.

In elasticsearch 1.7, there is a `_type` metadata field, which by default is

- indexed
- not analyzed
- not stored

https://www.elastic.co/guide/en/elasticsearch/reference/1.7/mapping-type-field.html

This means you can search on it, but not retrieve it in the body of the
document.

In our configuration, we add "_type" to our mappings, so that `_type` *is*
stored in the `_source` of the document.

In elasticsearch 2.0+ this is no longer allowed; `_type` is a reserved field
name and you can't customise it anymore.

We attempted to remove this in
https://trello.com/c/mkn7SAn0/84-fix-reserved-field-name-errors-in-elasticsearch-2-x-upgrade
thinking that it wasn't needed, because search results still correctly
displayed `_type` with it missing.

It turns out that lots of rummager relies on the field being in the `_source`.
In particularly, the nightly job which rebuilds the search index, uses it
during the export.  Additionally, if the field is missing, it silently coerces
the type to edition.

The field is also used by the legacy "advanced search" result sets, although
these are only used by whitehall, and as far as I'm aware all whitehall
documents are indexed as editions.

It's also possible that rummager filters for `document_type` rely on this
source field, because they use term filters rather than type filters:
https://www.elastic.co/guide/en/elasticsearch/reference/1.7/query-dsl-type-filter.html

This commit is an attempt to remove code that relies on `_type` being in
`_source`, so that we can again try to remove the configuration from our
mappings. It also removes the dangerous behaviour of defaulting to "edition"
wherever possible. We should always preserve the original type of the document
when doing anything with already indexed documents.


MatMoore:
There are at least 3 processes that are directly affected by this:

1) When you hit `post "/?:index?/documents"`

2) `BulkLoader`

3) When you hit `get "/:index/advanced_search.?:request_format?"`

I don't *think* I've changed what happens when you hit the `search.json` endpoint. I'll investigate this separately.

lib/index.rb
```diff
@@ -136,14 +136,43 @@ def amend(document_id, updates)
...
         nil
       end
     end
 
-    def document_from_hash(hash)
-      Document.from_hash(hash, @elasticsearch_types)
+    def document_from_original_hash(hash)
+      # FIXME: validate this and fail the request if missing
+      type = hash.delete('_type') || "edition"
+      id = hash.delete('_id')
```
MatMoore, line 148: id shouldn't be there

MatMoore:
Closing this for now as https://github.com/alphagov/rummager/pull/791 should hopefully fix the immediate problem.

We'll still need to remove dependence on `['_source']['_type']` to upgrade elasticsearch, but we might want to split into this smaller PRs.

PR #789, opened 2017-05-25
Add rake task to find and delete all duplicate content

suzannehamilton:
The existing `delete:duplicates` task requires a list of content IDs or paths when deleting duplicate entries. This is quite unwieldy when deleting large numbers of entries, such as the ones caused by the recent rummager `_type` bug.

To speed up the duplicate cleanup process, this new rake task finds the duplicate content before deleting it.

### DIFF
--- a/lib/tasks/delete.rake
+++ b/lib/tasks/delete.rake
@@ -1,5 +1,12 @@
 namespace :delete do
-  desc "Delete duplicates from index"
+  CONTENT_SEARCH_INDICES = %w(mainstream detailed government).freeze
+
+  desc "
+  Delete duplicates with the content IDs and/or links provided
+  Usage
+  TYPE_TO_DELETE=edition CONTENT_IDS=id1,id2 rake delete:duplicates
+  TYPE_TO_DELETE=edition LINKS=/path/one,/path/two rake delete:duplicates
+  "
   task :duplicates do
     require 'duplicate_deleter'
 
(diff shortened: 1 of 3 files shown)


lib/tasks/delete.rake
```diff
@@ -11,4 +11,18 @@ namespace :delete do
...
     deleter.call(links, id_type: 'link') if links.any?
   end
+
+  task :all_duplicates do
+    require 'duplicate_deleter'
+    require 'duplicate_links_finder'
+
+    type_to_delete = ENV.fetch("TYPE_TO_DELETE")
+
+    links = DuplicateLinksFinder.new("localhost:9203", ["mainstream", "detailed", "government"]).find
```
suzannehamilton: Just spotted that I've left this pointing to localhost. I'll fix.

lib/tasks/delete.rake
```diff
@@ -11,4 +11,20 @@ namespace :delete do
...
+
+  task :all_duplicates do
+    require 'duplicate_deleter'
+    require 'duplicate_links_finder'
+
+    type_to_delete = ENV.fetch("TYPE_TO_DELETE")
+
+    elasticsearch_config = YAML.load("../../elasticsearch.yml")
+
+    links = DuplicateLinksFinder.new(elasticsearch_config["base_uri"], %w(mainstream detailed government)).find
```
brenetic: `INDEXES` constant?

brenetic (approved):
Small comment. MWA 🚀

lib/tasks/delete.rake
suzannehamilton: Constant extracted. 👍

PR #790, opened 2017-05-25
Allow duplicate content to be deleted if the bad data has no ID

suzannehamilton:
Before, duplicate indexed content would be skipped if the content had different content IDs. This does not take into the account the case where the good data (the item with the correct `_type` value) has a content ID, but the bad data has a *missing* content ID.

### DIFF
--- a/lib/duplicate_deleter.rb
+++ b/lib/duplicate_deleter.rb
@@ -62,8 +62,13 @@ class DuplicateDeleter
 
       content_ids = results[:results].map { |a| a['content_id'] }
       if content_ids.uniq.count != 1
-        io.puts "Skipping #{id_type} #{id} as multiple content_id's detected #{content_ids.uniq.join(', ')}"
-        next
+        if results[:results].any? { |a| a[:elasticsearch_type] != type_to_delete && a['content_id'].nil? }
+          io.puts "Skipping #{id_type} #{id} as there is another document indexed with a valid '_type' but a missing content ID"
+          next
+        elsif results[:results].any? { |a| a[:elasticsearch_type] == type_to_delete && !a['content_id'].nil? }
+          io.puts "Skipping #{id_type} #{id} as multiple non-null content_id's detected #{content_ids.uniq.join(', ')}"
+          next
+        end
       end
 
       index_names = results[:results].map { |a| a[:index] }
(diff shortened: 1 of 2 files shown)


lib/duplicate_deleter.rb
```diff
@@ -62,8 +62,13 @@ def call(ids, id_type: 'content_id')
 
       content_ids = results[:results].map { |a| a['content_id'] }
       if content_ids.uniq.count != 1
-        io.puts "Skipping #{id_type} #{id} as multiple content_id's detected #{content_ids.uniq.join(', ')}"
-        next
+        if results[:results].any? { |a| a[:elasticsearch_type] != type_to_delete && a['content_id'].nil? }
+          io.puts "Skipping #{id_type} #{id} as it is indexed with a valid '_type' but a missing content ID"
```
MatMoore: I think this message is a bit unclear about which result is missing the content ID.

Maybe something like
```
"Skipping #{id_type} #{id} as there is another document indexed with a valid '_type' but a missing content ID"
```

lib/duplicate_deleter.rb
```diff
@@ -62,8 +62,13 @@ def call(ids, id_type: 'content_id')
...
+          next
+        elsif results[:results].any? { |a| a[:elasticsearch_type] == type_to_delete && !a['content_id'].nil? }
```
MatMoore: I think this ok. If anything it will be too strict and still skip too much, but if so we can raise another PR.

These are the different combinations I'm thinking of

1. Deleted type has non-nils, Non-deleted type(s) has all nils = bad, we've lost information
2. Deleted type has non-nils, non-deleted type(s) has all non-nils = bad, because content ids must be different
3. Deleted type has non-nils, non-deleted types has a mix = unknown
4. Deleted type has all nils = should always be ok

😖

MatMoore (dismissed):

lib/duplicate_deleter.rb
MatMoore: On second thoughts, I think this is not quite right... the code as is makes sense.

lib/duplicate_deleter.rb
suzannehamilton: Agreed and updated. 👍

MatMoore (approved):

PR #791, opened 2017-05-26
Ensure the metafields are preserved when amending a document

suzannehamilton:
When amending a document, which happens when the publishing API sends an update, copy the `_type` and `_id` fields directly to the new Elasticsearch item. Currently, they are copied from the nested `_source` data, which has caused problems when the main `_type` and the `_source` `_type` are out of sync.

This will help prevent some of the bad data that has been appearing in search results over the past week: search results with the wrong `_type` value and duplicate results with the same URL but different `_type`s.

Paired with @MatMoore

### DIFF
--- a/lib/index.rb
+++ b/lib/index.rb
@@ -136,7 +136,7 @@ module SearchIndices
     def get_document_by_id(document_id)
       begin
         response = @client.get(index: @index_name, type: "_all", id: document_id)
-        document_from_hash(response["_source"])
+        response
       rescue Elasticsearch::Transport::Transport::Errors::NotFound
         nil
       end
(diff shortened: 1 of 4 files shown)


MatMoore:
LGTM.

A bit more context:
The document type returned in `['_type']` and the document type returned in `['_source']['_type']` are different. We believe `['_source']['_type']` to be missing for most documents as a result of https://trello.com/c/mkn7SAn0/84-fix-reserved-field-name-errors-in-elasticsearch-2-x-upgrade, which we later reverted.

Deploying this will stop us from creating duplicate documents when we should be updating existing ones.

tijmenb (approved):

PR #792, opened 2017-05-30
Copy document type from canonical place when reindexing

suzannehamilton:
- When fetching all documents, look up the document `_type` from the top-level `_type` field, not the copy in `_source`. This is a step towards deleting `_source._type` entirely, which we need to do before upgrading to Elasticsearch 2.
- Move default `_type` assignment to POST request. Add more validation of search content `_type`, and move the fallback to "edition" to the /documents POST endpoint, which is the only place which is permitted to create documents without supplying a specific type. This should help us avoid situations like the recent corrupted search data, in which the `_type` was not being specified properly and was falling back to "edition" even for specialist document types.

Paired with @MatMoore.

Marked as DO NOT MERGE while we add more integration tests.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -176,6 +176,7 @@ class Rummager < Sinatra::Application
   post "/?:index?/documents" do
     request.body.rewind
     documents = [JSON.parse(request.body.read)].flatten.map { |hash|
+      hash["_type"] ||= "edition"
       current_index.document_from_hash(hash)
     }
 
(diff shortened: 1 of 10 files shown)


suzannehamilton:
@tijmenb or @carvil - would one of you be able to review this? It should let us reenable the overnight analytics job (once I've tested it on staging).

carvil (approved):
This looks good to me

suzannehamilton:
Cheers. I'm going to merge and deploy to staging for some more testing.

carvil:
Ok, sounds good to me. You can also deploy the branch to staging if you prefer to do that before merging, I'm happy either way.

suzannehamilton:
Good idea. I'll mark this as DO NOT MERGE while we're testing. 👍

PR #794, opened 2017-05-31
Add "primary_publishing_organisation"

tijmenb:
This field is populated in Whitehall (https://github.com/alphagov/whitehall/pull/3259) and represents the org that published the page. It will be used in the auditing of content.

https://trello.com/c/86sDRf3t


suzannehamilton (approved):
Looks OK to me. 👍

PR #795, opened 2017-05-31
Allow duplicate deleter to handle contacts with different IDs

suzannehamilton:
Main change:

The duplicate deleter was skipping over duplicate HMRC contacts because their `_id` fields were different, even though their content IDs and links were the same. This is because the `edition` duplicates were created with the wrong `_id`: contact IDs do not have a leading `/`.

Add this special case to the duplicate deleter so that it deletes the `edition` version if it also has the wrong `_id`.

Related changes:
- Give integration test helper parameter a more accurate name
- Fix typo in existing test name

This is part of the overall data cleanup arising from the Rummager `_type` bug.

### DIFF
--- a/lib/duplicate_deleter.rb
+++ b/lib/duplicate_deleter.rb
@@ -55,7 +55,7 @@ class DuplicateDeleter
       end
 
       ids = results[:results].map { |a| a[:_id] }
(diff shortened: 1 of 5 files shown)


carvil (approved):
👍

PR #796, opened 2017-06-01
Ensure duplicate deleter uses the correct ID for contacts

suzannehamilton:
The duplicate deleter assumed that duplicated content always had the same `_id` field. This isn't true for HMRC contacts, where the link was the same but the `_id` of the bad version had a leading `/`. This meant that the duplicate deleter sometimes tried to delete non-existent items with the correct `_type` but the wrong `_id`.

Fix this by looking up the correct `_id` to delete rather than picking one of the duplicate `_id`s arbitrarily.

### DIFF
--- a/lib/duplicate_deleter.rb
+++ b/lib/duplicate_deleter.rb
@@ -77,8 +77,15 @@ class DuplicateDeleter
         next
       end
 
-      Indexer::DeleteWorker.new.perform(index_names.first, type_to_delete, ids.first)
-      io.puts "Deleted duplicate for #{id_type} #{id}"
+      item_to_delete = results[:results].detect { |a| a[:elasticsearch_type] == type_to_delete }
+      if item_to_delete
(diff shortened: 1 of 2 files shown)


lib/duplicate_deleter.rb
```diff
@@ -77,7 +77,8 @@ def call(ids, id_type: 'content_id')
         next
       end
 
-      Indexer::DeleteWorker.new.perform(index_names.first, type_to_delete, ids.first)
+      id_to_delete = results[:results].detect { |a| a[:elasticsearch_type] == type_to_delete }[:_id]
```
carvil: Are we sure `detect` will find it? In the case it doesn't, will `Indexer::DeleteWorker` handle a `nil` ID to delete?

suzannehamilton: Line 52 should guard against that case, since it checks whether any of the results have `type_to_delete` type.

I've added an extra nil check anyway, since there's a lot of logic in this class and an error message might help someone track down the problem.

carvil (approved):

PR #802, opened 2017-06-07
Replace government index boost with format boosts

suzannehamilton:
- Simplify unit test of format boosting
- Tiny refactoring of format boost query
- Extract format boosts to config file to make the next change simpler
- **Main change**: Replace government index boost with format boosts:

Remove the search query boost based on index, and replace it with boosts to the individual search formats.

This produces a much simpler Elasticsearch query, because it no longer has to generate one weighted query for the `government` index and a separate unweighted query for everything else.

This should have almost no effect on search results. The only difference will be in the few formats which are shared between `government` and `mainstream`: `edition`, `finder` and `inside-government-link`. However, this just affects 23 mainstream documents and so the overall change will be very small. It didn't have any effect on the search healthcheck results when I tested it locally.

This should make it easier for us to make more changes to format weightings because they won't appear twice in the generated Elasticsearch query.

https://trello.com/c/FhscGKf2/127-refactor-code-where-weightings-are-applied

### DIFF
--- a/config/query/format_boosting.yml
+++ b/config/query/format_boosting.yml
@@ -1,0 +1,22 @@
+format_boosts:
+  # Mainstream formats
+  service_manual_guide: 0.3
+  service_manual_topic: 0.3
+  smart-answer: 1.5
+  transaction: 1.5
+  # Should appear below mainstream content
+  aaib_report: 0.2
+  dfid_research_output: 0.2
+  hmrc_manual_section: 0.2
+  service_standard_report: 0.2
+  # Inside Gov formats
+  contact: 0.3
+  document_collection: 1.3
+  document_series: 1.3
+  minister: 1.7
+  operational_field: 1.5
+  organisation: 2.5
+  topic: 1.5
+  topical_event: 1.5
+  # Hide mainstream browse pages
+  mainstream_browse_page: 0
--- a/config/query/format_boosting.yml
+++ b/config/query/format_boosting.yml
@@ -1,3 +1,37 @@
+government_index:
+  boost: 0.4
+  # Search formats used in the government index. These are defined by the
+  # publishing apps which push pages to the search API, and are different to the
+  # `format` field used in the content store.
+  #
+  # Any new formats used in search should be added to this list. Once all pages
+  # in search have a document_type, the document supertype should be used for
+  # boosting instead of formats, which will remove the need for this list.
+  formats:
+    - case_study
+    - consultation
+    - corporate_information_page
+    - document_collection
+    - edition
+    - fatality_notice
+    - finder
+    - inside-government-link
+    - minister
+    - news_article
+    - operational_field
+    - organisation
+    - person
+    - policy_group
+    - publication
+    - speech
+    - statistical_data_set
+    - statistics_announcement
+    - take_part
+    - topic
+    - topical_event
+    - world_location
+    - world_location_news_article
+    - worldwide_organisation
 format_boosts:
   # Mainstream formats
   service_manual_guide: 0.3
--- a/lib/search/query_components/booster.rb
+++ b/lib/search/query_components/booster.rb
@@ -4,13 +4,13 @@ module QueryComponents
       {
         function_score: {
           boost_mode: :multiply,
+          score_mode: :multiply,
           query: {
             bool: {
               should: [core_query]
             }
           },
           functions: boost_filters,
-          score_mode: "multiply",
         }
       }
     end
--- a/lib/search/query_components/booster.rb
+++ b/lib/search/query_components/booster.rb
@@ -3,6 +3,7 @@ require "yaml"
 module QueryComponents
   class Booster < BaseComponent
     FORMAT_BOOST_CONFIG = YAML.load_file('config/query/format_boosting.yml')
+    DEFAULT_BOOST = 1
 
     def wrap(core_query)
       {
@@ -25,8 +26,25 @@ module QueryComponents
       format_boosts + [time_boost, closed_org_boost, devolved_org_boost, historic_edition_boost]
     end
 
+    def boosted_formats
+      government_index_config = FORMAT_BOOST_CONFIG["government_index"]
+      government_index_boost = government_index_config["boost"]
+      government_formats = government_index_config["formats"]
+
+      individual_format_boosts = FORMAT_BOOST_CONFIG["format_boosts"]
+
+      boosted_formats = (government_formats + individual_format_boosts.keys).uniq
+
+      boosted_formats.each_with_object({}) do |format_name, boosts|
+        format_index_boost = government_formats.include?(format_name) ? government_index_boost : DEFAULT_BOOST
+        individual_format_boost = individual_format_boosts.fetch(format_name, DEFAULT_BOOST)
+
+        boosts[format_name] = format_index_boost * individual_format_boost
+      end
+    end
+
     def format_boosts
-      FORMAT_BOOST_CONFIG["format_boosts"].map do |format, boost|
+      boosted_formats.map do |format, boost|
         {
           filter: { term: { format: format } },
           boost_factor: boost
--- a/lib/search/query_components/booster.rb
+++ b/lib/search/query_components/booster.rb
@@ -1,5 +1,9 @@
+require "yaml"
+
 module QueryComponents
   class Booster < BaseComponent
+    FORMAT_BOOST_CONFIG = YAML.load_file('config/query/format_boosting.yml')
+
     def wrap(core_query)
       {
         function_score: {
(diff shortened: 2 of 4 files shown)


rboulton:
Haven't reviewed the code, but this is a good change to make - we always intended to do this, but never got around to analysing the impact it would have to weightings.

lib/search/query_components/booster.rb
```diff
@@ -22,32 +24,24 @@ def boost_filters
...
+      boosted_formats = (government_formats + individual_format_boosts.keys).uniq
 
-        # Hide mainstream browse pages for now.
-        "mainstream_browse_page" => 0,
-      }
+      boosted_formats.each_with_object({}) do |format_name, boosts|
+        format_index_boost = government_formats.include?(format_name) ? government_index_boost : DEFAULT_BOOST
+        individual_format_boost = individual_format_boosts.fetch(format_name, DEFAULT_BOOST)
+
+        boosts[format_name] = format_index_boost * individual_format_boost
```
dwhenry, line 42: I feel we should either:

A/ treat all format boost the same regardless of if the are government formats or not - this would require each government format to have a boost value - and would allow more flexibility if we want different down weights for different formats.

B/ Treat Government format boosts as a completely different thing to normal format boosts so we don't have all the edge cases in the code (just move it to a different method)

lib/search/query_components/booster.rb
```diff
@@ -22,32 +24,24 @@ def boost_filters
...
-        "operational_field" => 1.5,
-        "contact" => 0.3,
+      individual_format_boosts = format_boost_config["format_boosts"]
 
-        # Should appear below mainstream content
-        "aaib_report" => 0.2,
-        "dfid_research_output" => 0.2,
-        "hmrc_manual_section" => 0.2,
-        "service_standard_report" => 0.2,
+      boosted_formats = (government_formats + individual_format_boosts.keys).uniq
```
dwhenry, line 36: This is a change as currently if the format is present in both places it would be multiplied twice.

test/unit/search/query_components/booster_test.rb
```diff
@@ -2,54 +2,23 @@
...
-          ]
-        }
-      }
+      assert_equal :multiply, result[:function_score][:boost_mode]
+      assert_equal :multiply, result[:function_score][:score_mode]
 
-      assert_equal expected, result
+      assert_boost(result, "organisation", 1.0)
+      assert_boost(result, "service_manual_guide", 0.3)
+      assert_boost(result, "mainstream_browse_page", 0)
```
dwhenry: Have we chosen to only test these three formats for a reason? Would it not make all sense to tests all formats are present? and potentially to check that no unexpected formats are present - in addition, we have lost tests on the organisation_state and search_format_type boosts.

lib/search/query_components/booster.rb
```diff
@@ -22,32 +24,24 @@ def boost_filters
...
+      boosted_formats.each_with_object({}) do |format_name, boosts|
+        format_index_boost = government_formats.include?(format_name) ? government_index_boost : DEFAULT_BOOST
+        individual_format_boost = individual_format_boosts.fetch(format_name, DEFAULT_BOOST)
+
+        boosts[format_name] = format_index_boost * individual_format_boost
+      end
+    end
+
+    def format_boost_config
+      @format_boost_config ||= YAML.load_file('config/query/format_boosting.yml')
```
dwhenry: does this mean we are reading the config data in for the filesystem each time we create a new instance of this class? If so have we checked if this affects performance at all?

config/query/format_boosting.yml
```diff
@@ -0,0 +1,49 @@
+government_index:
```
dwhenry, line 1: It is important to point out that as a result of this change new Government formats will now need to be manually added to this list or they won't receive the boost they require. Given this is unlikely this will be remembered are we sure this is the correct way change to make.

dwhenry (changes requested):
I think use a config file is a good idea, just want to check it won't affect performance.

A few comments on the code, happy to chat if it doesn't make sense.

config/query/format_boosting.yml
suzannehamilton: Good point. The long-term plan is to use document supertypes, which will remove the need for a hard-coded format list. I've added a comment to explain.

lib/search/query_components/booster.rb
suzannehamilton: Can I check what you mean? I was trying not to change the overall boosts, so if I've added an extra factor then perhaps I've misunderstood how the index and format boosts interact. When you say "currently", do you mean the current implementation on master or the current implementation in this PR?

lib/search/query_components/booster.rb
suzannehamilton: You're right, it does read the config file each time. Can you see any issues with extracting it to a constant instead, so that the config is read when the class is loaded?

(For reference, the read doesn't take long (~1 ms elapsed time) but it's much slower than reading it from memory (~5 μs).)

test/unit/search/query_components/booster_test.rb
suzannehamilton: I didn't test every format because we're basically testing config. My aim was to add a few regression tests for the logic without making the tests fragile. (The original tests tripped me up because they were so tied to the exact output of `builder.wrap`.) What do you think?

Good spot on the non-format cases. I've added some more tests to cover those.

lib/search/query_components/booster.rb
suzannehamilton: I'm going to defer this decision until we've decided on the next A/B test: we might end up removing the government weighting entirely.

config/query/format_boosting.yml
MatMoore: I think whatever we do we should assume people will forget/not know to update this file when adding a new format.

So for this work (not necessarily in this PR, but we should have a plan before deploying it) we should make sure that unrecognised formats either get boosted down by default or alert us in some way.

There's a master list here we might be able to use if we opt for the latter: https://github.com/alphagov/govuk-content-schemas/blob/master/lib/govuk_content_schemas/allowed_document_types.yml

lib/search/query_components/booster.rb
MatMoore: My vote is for A, so `format_boosting.yml` values would be inclusive of what is now `format_index_boost`.

There's no reason to treat it as a different thing because the index doesn't correspond to anything meaningful anymore.

suzannehamilton: As discussed in person, I'm going to leave this it as it is because the next A/B test will remove the government format boosting so it's helpful to keep the configuration for the government index separate from the individual weightings.

We'll revisit this if we end up keeping the A version of the test.

config/query/format_boosting.yml
suzannehamilton: As mentioned above, we'll be removing this list of formats in the B version of the next A/B test, so we'll only have to handle this if we keep the A version.

lib/search/query_components/booster.rb
MatMoore: Discussed offline: there are a few odd "formats" like "edition" that are present in both indexes at the moment. They seem to be used for static pages, lists, finders, rather than regularly published content.

There aren't many of them, so I think for now we can decide either way (apply the boost or don't) without it affecting many searches.

MatMoore (dismissed):
🚢

suzannehamilton:
@dwhenry and @MatMoore: I've updated booster_test.rb with a change suggested by David, to split up the tests for format weighting. There are now separate tests for:

* Formats in the government index
* Individually boosted formats
* Formats in the government index with an extra individual weighting
* Formats with no boost at all

MatMoore (dismissed):

test/unit/search/query_components/booster_test.rb
```diff
@@ -2,54 +2,96 @@
...
+  should "combine government index and individual format weightings" do
+    builder = QueryComponents::Booster.new(search_query_params)
+    result = builder.wrap({ some: 'query' })
+
+    assert_format_boost(result, "minister", 0.68)
+    assert_format_boost(result, "organisation", 1.0)
+    assert_format_boost(result, "topic", 0.6)
+  end
+
+  should "not boost mainstream formats with no weighting" do
```
dwhenry: no biggy but might read slightly nicer
```should "not apply a boost to unspecified formats" do```

dwhenry (dismissed):
Just the one comment on the test name.. feel free to ignore

test/unit/search/query_components/booster_test.rb
suzannehamilton: Fixed, thanks. 🙂

MatMoore (approved):

PR #803, opened 2017-06-09
Promote minimum should match to production

dwhenry:
The A/B test around making the minimum should match criteria part of the
`should` calculation instead of having it as a `must` criteria has provide to
give better results. This change makes the B criteria live for all users.

https://trello.com/c/YTXRJKUE/130-deploy-minimum-should-match-a-b-test

### DIFF
--- a/test/unit/search/query_components/core_query_test.rb
+++ b/test/unit/search/query_components/core_query_test.rb
@@ -20,27 +20,13 @@ class CoreQueryTest < ShouldaUnitTestCase
     end
   end
 
-  context "when ab testing minimum should match" do
-    should "use the default value when no variant is passed in" do
+  context "the search query" do
+    should "down-weight results which match fewer words in the search term" do
       builder = QueryComponents::CoreQuery.new(search_query_params)
-
-      query = builder.payload
-      assert_match(/"2<2 3<3 7<50%"/, query[:bool][:must].to_s)
-    end
-
-    should "use variant b when it is passed in" do
-      builder = QueryComponents::CoreQuery.new(search_query_params(ab_tests: { search_match_length: 'B' }))
 
       query = builder.payload
       refute_match(/"2<2 3<3 7<50%"/, query[:bool][:must].to_s)
       assert_match(/"2<2 3<3 7<50%"/, query[:bool][:should].to_s)
-    end
-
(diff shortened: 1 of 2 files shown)


test/unit/search/query_components/core_query_test.rb
```diff
@@ -20,27 +20,13 @@ class CoreQueryTest < ShouldaUnitTestCase
...
-      builder = QueryComponents::CoreQuery.new(search_query_params)
-
-      query = builder.payload
-      assert_match(/"2<2 3<3 7<50%"/, query[:bool][:must].to_s)
-    end
-
-    should "use variant b when it is passed in" do
+  context "minimum should match" do
+    should "is part of the should clause" do
       builder = QueryComponents::CoreQuery.new(search_query_params(ab_tests: { search_match_length: 'B' }))
```
suzannehamilton: Should we get rid of the `ab_tests` parameter from this query?

test/unit/search/query_components/core_query_test.rb
```diff
@@ -20,27 +20,13 @@ class CoreQueryTest < ShouldaUnitTestCase
...
-    should "use the default value when no variant is passed in" do
-      builder = QueryComponents::CoreQuery.new(search_query_params)
-
-      query = builder.payload
-      assert_match(/"2<2 3<3 7<50%"/, query[:bool][:must].to_s)
-    end
-
-    should "use variant b when it is passed in" do
+  context "minimum should match" do
+    should "is part of the should clause" do
```
suzannehamilton: Super nitpicky: the wording "should is" is a bit odd.

I wonder how we would word this if we were writing it from scratch rather than as part of an A/B test. Maybe something like:

```
context "the search query" do
  should "down-weight results which match fewer words in the search term" do
```

suzannehamilton (changes requested):

test/unit/search/query_components/core_query_test.rb
suzannehamilton: I'm going to make these two changes and push a new version so that we can deploy this along with the new A/B test. I've checked that that's OK with @MatMoore.

suzannehamilton (approved):

MatMoore (approved):

PR #804, opened 2017-06-12
[WIP] Add rake task to back-populate metadata

MatMoore:
This is intended as a one off script to add `content_store_document_type` to
existing documents. Most newly published documents should already be sending
this to rummager (but we need to check this).

This field is really useful for debugging and analytics. If it's populated
reliably, it's more useful than `format` for format boosting, because `format`
contains stuff like `publication`, which covers lots of very different document
types.

Also add in `publishing_app` and `rendering_app`, and `content_id` (if missing).

There are about 53,000 documents with missing content store document types:

https://www.gov.uk/api/search.json?filter_content_store_document_type=_MISSING&count=0

Am currently timing how long this takes on my dev environment.


MatMoore:
The script itself took about 30 minutes, but I'm not sure how long the rummager workers take yet. I'll try it on integration tomorrow.

lib/missing_metadata_fetcher.rb
```diff
@@ -0,0 +1,49 @@
...
+      content_id = publishing_api.lookup_content_id(base_path: base_path)
+
+      if content_id.nil?
+        raise MissingDocumentError.new("Failed to look up base path")
+      end
+
+    elsif index_name.nil? || document_id.nil?
+      raise MissingDocumentError.new("Missing index name or id in search results")
+    else
+      response = publishing_api.get_content(content_id)
```
dwhenry: Shouldn't this code be run if the content_id was missing?

lib/tasks/fetch_missing_metadata.rake
```diff
@@ -0,0 +1,41 @@
...
+task :populate_metadata do
+  rummager = GdsApi::Rummager.new(Plek.new.find("rummager"))
+  publishing_api = GdsApi::PublishingApiV2.new(Plek.new.find("publishing-api"))
+  fetcher = MissingMetadataFetcher.new(publishing_api)
+
+  0.upto MAX_PAGES do |page|
+    response = rummager.search({
+      "filter_content_store_document_type" => "_MISSING",
+      "count" => PAGE_SIZE,
+      "start" => page * PAGE_SIZE,
```
dwhenry: This is slightly dangerous as you are requesting page two after updated records on page 1. potentially it would be better to get all the pages first then process or process pages in reverse order.

lib/missing_metadata_fetcher.rb
MatMoore: :man_facepalming:

lib/tasks/fetch_missing_metadata.rake
MatMoore: I went with reading all results into an array first, then doing the indexing.

test/integration/missing_metadata_test.rb
```diff
@@ -0,0 +1,20 @@
...
+      "mainstream_test",
+      "link" => '/path/to_page',
+      "_type" => "edition",
+    )
+
+    runner = MissingMetadata::Runner.new('content_id')
+    # publishing_api_has_content()
+    results = runner.retrieve_records_with_missing_value
+    assert_equal [:a], results
+    runner.update
```
MatMoore: is this test complete?

PR #805, opened 2017-06-13
[REVIEW ONLY] Add A/B test of format boosts

suzannehamilton:
In the B variant, remove the down-weighting of content with formats used in the government index. Replace it with an up-weighting of guidance content.

The government index down-weighting was originally added when there was a clear split between specialist content int the government index and mainstream content in the mainstream index. This distinction is no longer as clear because new specialist content has been added to the mainstream index.

The B variant uses the `navigation_document_supertype` field to boost guidance content. This should push truly mainstream content (as opposed to content in the `mainstream` index) higher in search results, helping mainstream users find content faster.

A boost of 2.5 was chosen because it is the inverse of the 0.4 boost that is applied to government index formats in the A variant.

https://trello.com/c/TjolBPDX/126-have-new-format-weightings-ready-to-use

Marked as REVIEW ONLY until we've run this against the healthcheck and updated the base branch to master.

### DIFF
--- a/lib/search/query_components/booster.rb
+++ b/lib/search/query_components/booster.rb
@@ -22,11 +22,29 @@ module QueryComponents
 
   private
 
+    def format_boost_b_variant?
+      @search_params.ab_tests[:format_boosting] == 'B'
+    end
+
     def boost_filters
-      format_boosts + [time_boost, closed_org_boost, devolved_org_boost, historic_edition_boost]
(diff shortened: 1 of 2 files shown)


dwhenry (approved):
This all looks fine on the condition that we will need to revisit/refactor this code if we do switch the B variant after the test, I assume we would move completely away from format boosts and use all supertype boosts instead?

suzannehamilton:
Apparently GitHub gets confused if you delete a branch that's been used as a base for a PR, so this PR has been replaced by #806. 🙃

PR #806, opened 2017-06-13
Add A/B test of format boosts

suzannehamilton:
In the B variant, remove the down-weighting of content with formats used in the government index. Replace it with an up-weighting of guidance content.

The government index down-weighting was originally added when there was a clear split between specialist content int the government index and mainstream content in the mainstream index. This distinction is no longer as clear because new specialist content has been added to the mainstream index.

The B variant uses the `navigation_document_supertype` field to boost guidance content. This should push truly mainstream content (as opposed to content in the `mainstream` index) higher in search results, helping mainstream users find content faster.

A boost of 2.5 was chosen because it is the inverse of the 0.4 boost that is applied to government index formats in the A variant.

https://trello.com/c/TjolBPDX/126-have-new-format-weightings-ready-to-use

### DIFF
--- a/lib/search/query_components/booster.rb
+++ b/lib/search/query_components/booster.rb
@@ -26,9 +26,9 @@ module QueryComponents
       boosts = format_boosts + [time_boost, closed_org_boost, devolved_org_boost, historic_edition_boost]
 
       if @search_params.format_boosting_b_variant?
-        boosts + [guidance_boost]
+        boosts + [guidance_boost, service_standard_report_boost(0.05)]
       else
-        boosts
+        boosts + [service_standard_report_boost(0.2)]
       end
     end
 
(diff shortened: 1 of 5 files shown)


lib/search/query_components/booster.rb
```diff
@@ -23,10 +23,24 @@ def wrap(core_query)
...
+      if @search_params.format_boosting_b_variant?
+        boosts + [guidance_boost]
+      else
+        boosts
+      end
     end
 
     def boosted_formats
+      if @search_params.format_boosting_b_variant?
+        FORMAT_BOOST_CONFIG["format_boosts"]
```
MatMoore: I think it might be clearer to inline `formats_boosted_by_index` (or restructure it in some other way) - so that it's more obvious that `FORMAT_BOOST_CONFIG["format_boosts"]` is used in both cases, and the A case is just doing something extra.

As is, it's a bit hard to see the difference.

MatMoore (dismissed):
👍 

I think FOIs (and a few other formats) should actually be recency weighted like we do with announcements, so we might want to revisit this in a future test.

For service assessments - if you haven't already done this - it would be good to test searches for each service name and make sure the boost is enough to get the assessment out of the top N results. We could revisit this if we ever change how the titles are displayed.

lib/search/query_components/booster.rb
suzannehamilton: 👍 I've inlined it. Hopefully we'll be able to delete the index boosting at some point anyway.

MatMoore (approved):
👍

PR #808, opened 2017-06-16
Use services publishing-api for missing meta data fetcher

brenetic:
We were missing the bearer token previously.


MatMoore (approved):
:man_facepalming:

PR #810, opened 2017-06-16
Add a dummy param for cache busting

MatMoore:
When testing the API it's sometimes useful to bypass the caching layer.


lib/parameter_parser/search_parameter_parser.rb
```diff
@@ -340,6 +340,9 @@ def debug_options
         options[:use_id_codes] = true
       when "show_query"
         options[:show_query] = true
+      when /^cache_bust:/
```
dwhenry: I super like the idea but feel a shorter param might be easier.

dwhenry (approved):

PR #811, opened 2017-06-20
Add rake task for generating analytics data

suzannehamilton:
- Fix Elasticsearch config in existing rake task
- Move `ScrollEnumerator` out of the `legacy_client` directory. It's being used by site search so it doesn't belong with the legacy advanced search code. And we wanted to reuse it for this task.
- Decrease the search scroll timeout. It used to be 1 hour, which was only necessary in old version of Elasticsearch (<0.9.12).
- **Main change:** add rake task to create a CSV for importing into Google Analytics
- Add data fallbacks for important fields which may be empty

This will allow us to associate search data with Google Analytics events for search without having to send that data with every GA event.

https://trello.com/c/UqfHi27u/135-implement-query-time-data-import-for-enhanced-ecommerce-dimensions

Paired with @brenetic for some of the changes.

Marked as [REVIEW ONLY] until we've tested the file upload. We've tried uploading a small subset of the data but not the whole thing: we may need to split the CSV into chunks.


MatMoore:
👍 

On the script itself, a consequence of driving it from the search index is GA will only have products for the pages we index, so it won't show us gaps in the index (things that are never clicked because we don't actually index them). This isn't something we specifically set out to track but we might want to think about it.

Good spot on the scroll timeout.

The ScrollEnumerator I was hoping we could get rid of. When I introduced the elasticsearch-ruby client I got rid of most of our handwritten elasticsearch client code, and I assumed we could probably refactor this too later. I think it's fine to move it, but maybe search isn't the best module to put it in - it feels lower level than the rest of the classes there.

dwhenry (dismissed):

lib/analytics_data.rb
```diff
@@ -0,0 +1,58 @@
...
+      ga:productCategoryHierarchy
+      ga:dimension72
+      ga:dimension73
+      ga:dimension74
+      ga:dimension75
+      ga:dimension76
+      ga:dimension77
+      ga:dimension78
+      ga:dimension79
+      ga:dimension80
```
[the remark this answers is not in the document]
suzannehamilton: I've refactored the class so that the CSV headers and values are defined together. This makes their purpose clearer (I've added comments on the placeholders) and should stop them from getting out of sync.

suzannehamilton:
I've added an extra commit which handles titles with line breaks. They were causing the CSV import to fail.

There were only around 20 pages like this, and the line breaks didn't look intentional. Is it worth fixing them in the content itself?

suzannehamilton:
@MatMoore As discussed, I've moved `ScrollEnumerator` to `lib` but haven't made any other changes to it. We're already using the client library to perform the actual scroll - other libraries might give us a nicer enumerator wrapper, but we don't think it's worth investigating them at the moment.

lib/analytics_data.rb
```diff
@@ -45,4 +45,11 @@ def columns
       "ga:dimension80" => lambda { |_| nil }, # Placeholder: is page a best bet?
     }
   end
+
```
MatMoore: Maybe worth naming this "for_google_analytics" or adding a comment explaining it. 

Otherwise it looks unnecessary at a glance since you're running it through the CSV module.

MatMoore (dismissed):

lib/analytics_data.rb
suzannehamilton: Done. 👍

lib/analytics_data.rb
```diff
@@ -0,0 +1,55 @@
...
+      "ga:dimension78" => lambda { |item| item["public_timestamp"].nil? ? nil : Date.parse(item["public_timestamp"]).strftime("%Y%m%d") },
+      "ga:dimension79" => lambda { |_| nil }, # Placeholder: purpose not yet decided
+      "ga:dimension80" => lambda { |_| nil }, # Placeholder: is page a best bet?
+    }
+  end
+
+  def sanitise_for_google_analytics(data)
+    data
+      &.gsub("\r\n", " ")
+      &.gsub("\n", " ")
```
dwhenry: ```&.gsub(/[\r\n]*/, " ")```
or

```&.gsub(/\r?\n/, " ")``` 
if you prefer

File: test/unit/elasticsearch_index_test.rb [hunk 4]
```diff
@@ -126,7 +126,7 @@ def test_commit
   end
 
   def test_can_fetch_documents_by_format
-    search_pattern = "http://example.com:9200/mainstream_test/_search?scroll=60m&search_type=scan&size=500"
+    search_pattern = "http://example.com:9200/mainstream_test/_search?scroll=1m&search_type=scan&size=500"
```
dwhenry, line 129: I would generally prefer not including the test domain if possibel.

I think the below would work (needs to be tested)

```
search_pattern = %r{/mainstream_test/_search?scroll=1m&search_type=scan&size=500$}
```

lib/analytics_data.rb
```diff
@@ -0,0 +1,55 @@
...
+    query = {
+      "query": {
+        "match_all": {}
+      }
+    }
+
+    ScrollEnumerator.new(client: client, index_names: @indices, search_body: query) do |hit|
+      item = hit["_source"]
+
+      columns.values.map { |i| i.call(item) }
```
dwhenry, line 23: This will be slower than explicit declarations, have we checked it not a bit slowdown (don't believe it would be)

suzannehamilton: Good point - we are running this 300,000 times. I've timed it and the task still takes around 1 minute.

lib/analytics_data.rb
suzannehamilton: Yep, that's simpler. Fixed. 👍

test/unit/elasticsearch_index_test.rb
suzannehamilton: It's just the domain that's configured in the unit test, so it seems OK to me to include it when the request is stubbed. Is there some problem that you think it might cause in the future?

dwhenry: Personal preference.. :)

dwhenry (approved):

PR #812, opened 2017-06-21
Add rake task to delete duplicate DFID reports from search

suzannehamilton:
Around 130 DFID reports appear twice in search results: one result is a direct link to the page, and one is to a redirect to the correct page.

The duplicates appear because the base paths of these reports were updated in a bulk process which added the new paths to the search index but did not remove the old ones, which have become redirects. This rake task fixes that by deleting the old entries.

The CSV file containing the duplicate redirects can be generated from the publishing API using the rake task created in alphagov/publishing-api#951.

This rake task should be deleted once it has been run on production.

https://trello.com/c/ECoFt3Yq/136-dfid-research-output-finder-issues-to-investigate

### DIFF
--- a/lib/tasks/delete.rake
+++ b/lib/tasks/delete.rake
@@ -39,4 +39,31 @@ namespace :delete do
     deleter = DuplicateDeleter.new(type_to_delete)
     deleter.call(links, id_type: 'link')
   end
+
+  desc "
+  Delete the documents from a search index as specified in the given file. The
+  file should contain a list of page paths separated by new lines.
+  Usage
+  rake 'delete:documents_from_file[/path/to/file, mainstream]'
+  "
+  task :documents_from_file, [:file_path, :index_name] do |_, args|
+    require "csv"
+    require "indexer/workers/delete_worker"
+
+    search_config = SearchConfig.new
+    index = search_config.search_server.index(args[:index_name])
+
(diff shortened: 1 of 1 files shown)


MatMoore:
I think this might be useful for other things. We could make it a general purpose task to delete links specified from a file. 

It could also go in search admin, since that has functionality for removing single links (but either repo is fine)

MatMoore (dismissed):

suzannehamilton:
👍 I've generalised it. A few things to note:

- The task now requires you to specify the index, which was simpler than searching for the documents in all indexes. We can change it if we end up needing to delete things across multiple indexes a lot.
- Is it OK to require people to upload a csv file to the server? It seemed OK for a one-off DFID cleanup task, but can we make it simpler for the more generic task? I couldn't think of a way around it but I might have missed something.
- The output contains noisy `FATAL  Services : [404]` logs if the item doesn't exist. Again, that seemed OK for a one-off task but isn't great in the long term.

MatMoore:
I think that's all fine.

If we find ourselves using this a lot, we can always stick a UI on it and make it part of search admin. But hopefully by integrating with the publishing platform this will become less of a problem in the long term :crossed_fingers:

MatMoore (approved):

PR #813, opened 2017-06-21
Delete old business finance format items

brenetic:
Business finance and support content has been moved from a mainstream format to a finder in specialist publisher.

The old `business_support` format used to be excluded from the search index. `business_finance_support_scheme` is the new format used.

This task can be removed once it has been run on production.

[Trello card](https://trello.com/c/d0cc64W2/137-old-businesssupport-items-appearing-in-search-results)

### DIFF
--- a/lib/tasks/delete.rake
+++ b/lib/tasks/delete.rake
@@ -39,4 +39,32 @@ namespace :delete do
     deleter = DuplicateDeleter.new(type_to_delete)
     deleter.call(links, id_type: 'link')
   end
+
+  # Can be removed once run in production
+  desc "Delete old business support scheme items"
+  task :old_business_support do
+    require "indexer/workers/delete_worker"
+    elasticsearch_host = SearchConfig.new.elasticsearch["base_uri"]
+
+    params = {
+      query: { match_all: {} },
+      filter: {
+        bool: {
+          must: {
+            terms: { format: ["business_support"] }
+          }
+        }
+      },
+      size: 10_000
+    }
+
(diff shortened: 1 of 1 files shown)


lib/tasks/delete.rake
```diff
@@ -39,4 +39,37 @@ namespace :delete do
...
     deleter.call(links, id_type: 'link')
   end
+
+  # Can be removed once run in production
+  desc "
+  Delete old business support scheme items.
+  "
+  task :old_business_support do
+    require "indexer/workers/delete_worker"
+    elasticsearch_config = YAML.load("../../elasticsearch.yml")
```
suzannehamilton: I discovered in another PR that this doesn't actually work - or rather, it only works because the Elasticsearch client falls back to `localhost:9200` if the URL is nil. (See https://github.com/alphagov/rummager/pull/811/commits/0210a1d2600d5bc15e4bc12c1119b6410cfe3b55.)

Fortunately you can use `SearchConfig.new.elasticsearch` instead. :)

lib/tasks/delete.rake
```diff
@@ -39,4 +39,36 @@ namespace :delete do
     deleter = DuplicateDeleter.new(type_to_delete)
     deleter.call(links, id_type: 'link')
   end
+
+  # Can be removed once run in production
+  desc "
+  Delete old business support scheme items.
+  "
```
dwhenry: this doesn't need a multiline comment

dwhenry (approved):
I would have used a count variable, but it's fine as is.

PR #814, opened 2017-06-21
Decouple searching code from Sinatra search controllers

MatMoore:
This finishes up the refactor we started in https://github.com/alphagov/rummager/pull/807

It's best reviewed a commit at a time.

Goals:

- Don't rely on global objects - like `Rummager.settings.search_config` - except in the Sinatra controllers themselves.
- Don't instantiate `Index` and related objects all over the place, use `SearchConfig` methods and dependency injection instead
- Enable reuse of code for performing searches outside of the Sinatra app

`SearchConfig` probably has too many responsibilities, but I think this refactor makes it clearer what's using it.


lib/parameter_parser/base_parameter_parser.rb
```diff
@@ -8,6 +8,10 @@ class ParseError < StandardError
     def initialize(errors)
       @errors = errors
     end
+
+    def error
+      errors.join(". ")
```
dwhenry, line 13: any reason we have changed this from an array to a string?

lib/search_config.rb
```diff
@@ -10,6 +10,8 @@
 require "erb"
 
 class SearchConfig
+  ParseError = BaseParameterParser::ParseError
```
dwhenry: why have we done this? And if we do need this would we be better off use inheritance rather than constant assignment (think it would be easier to debug without breaking anything)

dwhenry (commented):
It all looks pretty good, want to have a bit more or a think and look through the previous commits to understand all the changes before approving. :)

lib/parameter_parser/base_parameter_parser.rb
MatMoore: The API only returns a single string, and I didn't want to change that.

The unit tests for the parser are also (annoyingly) all testing for a string, via a copy of this method on
`search_parameter_parser`. I initially tried moving it but decided it wasn't worth the time to change the tests.

lib/search_config.rb
MatMoore: Nope, we don't need it.

At the time I was just thinking it was weird for app.rb to have to know about the internals of parsing but this isn't a great solution. I think I'll just revert this.

test/support/integration_test.rb
```diff
@@ -16,6 +16,10 @@ def initialize(args)
   end
 
   def setup
```
suzannehamilton, line 18: I've just noticed that there are a few integration tests that don't call `super` in their `setup` methods, e.g. `ElasticsearchMigrationTest`. Should we fix those now?

test/support/integration_test.rb
```diff
@@ -16,6 +16,10 @@ def initialize(args)
...
+    # search_config is a global object that has state, so make sure it's reset
```
suzannehamilton, line 19: It might be nice to split `SearchConfig` up at some point: it's a bit surprising (to me at least 🙂) that it holds state like the registries which contain data from Elasticsearch rather than static config.

lib/search_config.rb
dwhenry: I see you changed where this was used to use `BaseParameterParser::ParseError` but think you might have forgotten to delete this line. :)

MatMoore: Yep my mistake, will fix

test/support/integration_test.rb
MatMoore: Yeah definitely. I'm hoping that by just making the dependencies on it more explicit, it will make further refactoring easier.

For now I think it's useful for *something* to have knowledge about all the data sources (content indexes, metasearch index, spelling index, registries) since before you had to dig into the query building code to understand where data was coming from.

I think we will be able to remove some of the complexity once we start rearchitecting the indexing. :crossed_fingers:

test/support/integration_test.rb
MatMoore: Good point, I've added a new commit addressing these.

suzannehamilton (approved):

PR #817, opened 2017-06-23
Store exported file in a shared location

dwhenry:
This means historical exports are stored in a consistant location


MatMoore:
Can you make the path a setting so it will work on dev environments too?

dwhenry:
@MatMoore done.

It'll now export to the EXPORT_PATH if it's set or <app>/data if it's not.

MatMoore (approved):

PR #819, opened 2017-06-23
Add rake task to delete old export files

dwhenry:
https://trello.com/c/RFFTzN3z/144-automate-ecommerce-csv-generation

### DIFF
--- a/lib/tasks/analytics.rake
+++ b/lib/tasks/analytics.rake
@@ -26,4 +26,16 @@ namespace :analytics do
       end
     end
   end
+
+  desc "Delete old export files (specify the number to keep with EXPORT_FILE_LIMIT)"
+  task :delete_old_files do
+    path = ENV['EXPORT_PATH'] || 'data'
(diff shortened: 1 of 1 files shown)


lib/tasks/analytics.rake
```diff
@@ -26,4 +26,16 @@ namespace :analytics do
       end
     end
   end
+
+  desc "Delete old export files (specify the number to keep with EXPORT_FILE_LIMIT)"
+  task :delete_old_files do
+    path = ENV['EXPORT_PATH'] || 'data'
```
brenetic, line 32: Do we need `/data`? Or is this just for the task already there?

brenetic (approved):
👍

brenetic:
Can we have a Trello card link please

lib/tasks/analytics.rake
dwhenry: This is so that you don't need to specify a path when running locally.. it will just put the generated file in you `<app>/data` folder and then in prod we specify the path in the jenkins job.. :)

PR #820, opened 2017-06-26
Document what publishing apps send to rummager

MatMoore:
Document the content in our indexes and where it comes from.

[Rendered version](https://github.com/alphagov/rummager/blob/dfca1777660ad239f6b59545493c35b7d5b6e0b4/docs/content_overview.md)

Link through to the code that sends stuff to rummager. We'll be changing how content gets indexed and I think it will be useful to be able to quickly find the source when comparing across 10 applications.

It's fine for this to be removed when it becomes out of date.


MatMoore:
Note that the numbers on the pie chart don't add up; there's a bunch of specialist content with missing `publishing_app` data (so it gets added to the "government" slice) and there's likely to be other content I haven't accounted for.

This is just intended as a rough illustration of what we have - so I'm going to leave it as it is until we have accurate information about publishing app and document type for documents we index.

dwhenry (approved):
I didn't realise it was that complicated.

PR #821, opened 2017-06-26
Remove search format boosting a/b test code

suzannehamilton:
The test has reached statistical significance, so it can be removed. The version that is being kept was the b variant which boosts guidance content rather than non government index content and reduces the boost for service standard report.

Also fix missing ampersand in README docs.

[Trello card](https://trello.com/c/Z5KURpnB/147-deploy-format-boosting-%F0%9F%9A%80)

paired with @brenetic

### DIFF
--- a/lib/search/query_components/booster.rb
+++ b/lib/search/query_components/booster.rb
@@ -23,53 +23,23 @@ module QueryComponents
   private
 
     def boost_filters
-      boosts = format_boosts + [time_boost, closed_org_boost, devolved_org_boost, historic_edition_boost]
-
-      if @search_params.format_boosting_b_variant?
-        boosts + [guidance_boost, foi_boost, service_standard_report_boost(0.05)]
-      else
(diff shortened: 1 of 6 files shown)


lib/search/query_components/booster.rb
```diff
@@ -23,55 +23,25 @@ def wrap(core_query)
...
-
-        boosted_formats = (government_formats + individual_format_boosts.keys).uniq
-
-        boosted_formats.each_with_object({}) do |format_name, boosts|
-          format_index_boost = government_formats.include?(format_name) ? government_index_boost : DEFAULT_BOOST
-          individual_format_boost = individual_format_boosts.fetch(format_name, DEFAULT_BOOST)
-
-          boosts[format_name] = format_index_boost * individual_format_boost
-        end
-      end
```
dwhenry, line 53: 👍

dwhenry (approved):
nice

PR #823, opened 2017-06-27
Return 400 response when invalid ElasticSearch query

dwhenry:
We are catching error responses from elasticsearch instead of validating the
value before they are passed in. This has the advantage of catching all errors
for the types we are interested in and ensuring we don't block requests which
would be valid.

https://trello.com/c/C6WRzJGx/139-early-validation-of-rummager-fields

### DIFF
--- a/lib/search/query.rb
+++ b/lib/search/query.rb
@@ -8,6 +8,10 @@ require_relative 'suggestion_blacklist'
 
 module Search
   class Query
+    class Error < StandardError; end
+    class NumberOutOfRange < Error; end
+    class QueryTooLong < Error; end
+
     attr_reader :index, :registries, :spelling_index, :suggestion_blacklist
 
     def initialize(registries:, content_index:, metasearch_index:, spelling_index:)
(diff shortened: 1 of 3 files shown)


lib/search/query.rb
```diff
@@ -81,5 +85,22 @@ def content_index_names
...
+    def fetch_spell_checks(search_params)
+      SpellCheckFetcher.new(search_params, registries).es_response
+    end
+
+    def process_elasticsearch_errors
+      yield
+    rescue Elasticsearch::Transport::Transport::Errors::BadRequest => e
+      case e.message
+      when /Numeric value \([0-9]*\) out of range of/
+        raise(NumberOutOfRange, 'Integer value exceeds maximum allowed')
```
suzannehamilton: Maybe we could include the out-of-range number in the error message? It looks like we can extract that easily even if it would take a lot of effort to parse the field name.

dwhenry: Done.. but it's not pretty

MatMoore (approved):

PR #824, opened 2017-06-28
Remove _type and _id from the document hash

MatMoore:
This is a blocker to upgrading Elasticsearch. When the Elasticsearch migration plugin is run, all the indexes return an error like this:

> Reserved field names ([info](https://www.elastic.co/guide/en/elasticsearch/reference/2.0/breaking_20_mapping_changes.html#migration-meta-fields))
The `_uid`, `_id`, `_type`, `_source`, `_all`, `_parent`, `_field_names`, `_routing`, `_index`, `_size`, `_timestamp`, and `_ttl` field names are reserved and can no longer be used in the document `_source`., in type: `edition`.

We tried removing these fields in https://github.com/alphagov/rummager/pull/777, and it didn't work out so well, but it's now safe to remove from the document hash, because we've changed the codebase to reference the top level
"_type" and "_id" from elasticsearch responses. :crossed_fingers: 

See
- https://github.com/alphagov/rummager/pull/791
- https://github.com/alphagov/rummager/pull/792

Trello: https://trello.com/c/2aAKWCwE/149-remove-reserved-id-and-type-fields-from-source-again

### DIFF
--- a/lib/indexer/bulk_payload_generator.rb
+++ b/lib/indexer/bulk_payload_generator.rb
@@ -70,7 +70,6 @@ module Indexer
       popularities = lookup_popularities(links.compact)
       actions.flat_map { |command_hash, doc_hash|
         if command_hash.keys == ["index"]
-          doc_hash["_type"] = command_hash["index"]["_type"]
           [
             command_hash,
             index_doc(doc_hash, popularities),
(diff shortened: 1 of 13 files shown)


dwhenry (commented):
This all lookss pretty good to me. 

My only comment would be that it might have been nice for the helper methods to use `'_type' => ...` instead of `type: ...` but that relates more to previous decisions.

brenetic (approved):
🚀

PR #825, opened 2017-06-29
[DO NOT MERGE] Validate reindexing

dwhenry:
This is an attempt at building a way of comparing two indexes and getting some simple data related to addition/deletions and changes.

https://trello.com/c/d3s6Q8XV/148-safety-checks-for-switching-to-a-newly-rebuilt-index


lib/indexer/comparer.rb
```diff
@@ -0,0 +1,122 @@
...
+#   'government',
+#   'government-old-2017-06-21t11:04:56z-00000000-0000-0000-0000-000000000000',
+#   'government-2017-06-21t11:55:51z-00000000-0000-0000-0000-000000000000'
+# ).run
+
+module Indexer
+  class Comparer
+    # Source fields to ignore when comparing old and new objects for equality.
+    # These are ones which we know have changed for most documents.
+    FIELDS_TO_IGNORE = ["_id", "_type", "popularity"]
```
MatMoore: Will we remove _id and _type after removing them from _source?

lib/indexer/comparer.rb
```diff
@@ -0,0 +1,122 @@
...
+      @rejected_fields = rejected_fields || ['popularity']
+    end
+
+    def log(msg)
+      puts "#{Time.now}: #{msg}"
+    end
+
+    def read_index(index_name, filter = nil)
+      log "Access index: #{index_name}"
+      index = search_config.search_server.index_group(@base_name).index_for_name(index_name)
```
MatMoore: I don't think this should use the `Index` class, since the old and new index name are passed in we can just instantiate the elasticsearch client directly.

lib/indexer/comparer.rb
```diff
@@ -0,0 +1,122 @@
...
+    def reject_fields(hash)
+      hash.reject{ |k, _| @rejected_fields.include?(k) }
+    end
+
+    def run
+      outcomes = Hash.new { |h, k| h[k] = 0 }
+
+      filters.each do |filter|
+        old_data = read_index(@old_index_name, filter)
+        new_data = read_index(@new_index_name, filter)
```
MatMoore, line 71: I think the only part of `new_data` you need to keep in memory is the keys (for the `removed_keys` calculation below). So if memory usage is still an issue you might be able to do a second pass with the ScrollEnumerator instead.

lib/indexer/comparer.rb
```diff
@@ -0,0 +1,122 @@
...
+            fields = changed_fields(old_item, new_item)
+            if fields.any?
+              outcomes[:changed] += 1
+              outcomes[:"changes: #{fields.join(',')}"] += 1
+            else
+              outcomes[:unchanged] += 1
+            end
+          end
+
+          outcomes[:inconsistent_id] += 1 if new_item['_id'] != new_item['_root_id']
```
MatMoore: I think we can remove these two lines, since we're about to remove the non root `_id` and `_type`.

lib/indexer/comparer.rb
```diff
@@ -0,0 +1,122 @@
...
+          outcomes[:inconsistent_type] += 1 if new_item['_type'] != new_item['_root_type']
+        end
+      end
+
+      outcomes
+    end
+
+    # break the index down into managable chunks to reduce memory overhead
+    def filters
```
MatMoore, line 97: 👍

MatMoore (commented):
Looks good to me, I don't see any problems with the filter based approach.

If the filter value itself changes, it will show up as a bunch of removals and additions, but as long as we catch it somehow it shouldn't matter.

You could maybe slice it using the public_timestamp instead, but I think there are a lot of documents without that field so format is probably better.

lib/indexer/comparer.rb
dwhenry: yes

lib/indexer/comparer.rb
dwhenry: done

lib/indexer/comparer.rb
dwhenry: So I am actually checking that the values for all the keys have remained consistent.

Hense the reason for loading the whole dataset into the index.

lib/indexer/comparer.rb
dwhenry: done

PR #827, opened 2017-06-29
[DO NOT MERGE] Validation using hashed

dwhenry:
This is an alternative to https://github.com/alphagov/rummager/pull/825 which uses significantly less ram by hashing the `_source` data and then comparing the hash. This will increase processing time due to the hashing but save a heap of ram.

A middle ground when key fields (i.e. format, etc) are stored and the remaining fields are hashed and then compared may get the best of bost worlds

https://trello.com/c/d3s6Q8XV/148-safety-checks-for-switching-to-a-newly-rebuilt-index


lib/indexer/comparer.rb
```diff
@@ -0,0 +1,131 @@
...
+    end
+
+    def reject_fields(hash)
+      hash.reject{ |k, _| @rejected_fields.include?(k) }
+    end
+
+    def consistent_hash(hash)
+      # sort the object as reasonible then convert to a string and final hash teh result
+      # this allow us to check if to objects match with needing to store the full object
+      hash_str = consistentify(hash).inspect
```
MatMoore, line 65: Can you just call `#hash` on the document, or are there edge cases you're avoiding here?

This does what I'd expect:

```ruby
irb(main):008:0> {a:1,b:2}.hash
=> 3918977153981370324
irb(main):009:0> {b:2,a:1}.hash
=> 3918977153981370324
```

MatMoore (approved):
Looks good to me, I think we could go with this approach (after rebasing and checks are passing). 👍

PR #829, opened 2017-07-03
[DO NOT MERGE] Rspit test

dwhenry:
This should not be merged into production. It is a play to see what would be required to use rspec to run tests defined using unit-test syntax.


dwhenry:
Maybe not the best approach, but fun to try

PR #830, opened 2017-07-03
Validation elasticsearch migration using streams

dwhenry:
__Compare Elasticsearch indices using streams__

This approach is taken to avoid the need to load
the full datset into memory.

By retrieving order data we can simple to a side by side
comparison incrementing the list position of the smallest
(string comparison) value when the record keys don't match.

This process will consume more CPU/Network/ES CPU time while
providing a stable memory usage, this should not be an issue
so long as it is run on a multiple CPU processor.


speed_test.rb
```diff
@@ -0,0 +1,56 @@
+require_relative 'app'
```
MatMoore: Do we need to check `speed_test.rb` in?

test/unit/indexer/compare_enumerator_test.rb
```diff
@@ -0,0 +1,133 @@
...
+require 'indexer/compare_enumerator'
+
+class CompareEnumeratorTest < MiniTest::Unit::TestCase
+  def test_when_matching_keys_exist
+    data_left = { '_root_id' => 'abc', '_root_type' => 'stuff', 'custom' => 'data_left' }
+    data_right = { '_root_id' => 'abc', '_root_type' => 'stuff', 'custom' => 'data_right' }
+
+    stub_scroll_enumerator(left_request: [data_left], right_request: [data_right])
+
+    results = Indexer::CompareEnumerator.new('index_a', 'index_b').map { |a| a }
```
suzannehamilton: IIRC, you can call `.to_a` instead of `map { |a| a }`.

test/unit/indexer/compare_enumerator_test.rb
```diff
@@ -0,0 +1,133 @@
...
+    stub_client_for_scroll_enumerator(return_values: [[data], []], search_body: search_body)
+
+    enum = Indexer::CompareEnumerator.new('index_a', 'index_b').get_enum('index_name', search_body)
+
+    assert_equal enum.map { |a| a }, [
+      { '_root_id' => 'abc', '_root_type' => 'stuff', 'custom' => 'data' }
+    ]
+  end
+
+  def test_scroll_enumerator_mappings_wthout_sorting
```
suzannehamilton: Typo: wthout

test/unit/indexer/compare_enumerator_test.rb
```diff
@@ -0,0 +1,133 @@
...
+
+  def test_with_matching_ids_but_different_types
+    data_left = { '_root_id' => 'abc', '_root_type' => 'stuff', 'custom' => 'data_left' }
+    data_right = { '_root_id' => 'abc', '_root_type' => 'other_stuff', 'custom' => 'data_right' }
+
+    stub_scroll_enumerator(left_request: [data_left], right_request: [data_right])
+
+    results = Indexer::CompareEnumerator.new('index_a', 'index_b').map { |a| a }
+    assert_equal results, [
+      [Indexer::CompareEnumerator::NO_VALUE, data_right],
```
suzannehamilton: Just to check: is this the most useful way of representing this? I'd have thought it would be easier to interpret the results if they were grouped by ID and we checked the type directly, rather than including them in the added/removed count, since adding and removing documents is expected but changing types is not.

speed_test.rb
dwhenry: no.. removed

test/unit/indexer/compare_enumerator_test.rb
```diff
@@ -0,0 +1,133 @@
+require "test_helper"
+require 'indexer/compare_enumerator'
+
+class CompareEnumeratorTest < MiniTest::Unit::TestCase
+  def test_when_matching_keys_exist
```
suzannehamilton: I think it'd also be handy to have a test with a bit more data in each enumerator so that we test what happens _after_ the enumerator encounters a missing value, i.e. does it correctly increment just one of the indexes?

suzannehamilton (dismissed):

test/unit/indexer/compare_enumerator_test.rb
dwhenry: I'm not 100% sure on this one as it is a result of the implementation. This test is here to clearly show the limitations of the current implementation. We could potentially remove this limitation by removing the grouping functionality which was to help solve the memory issue that we killed by using streaming. I am open to discussion, but wanted to get the code to a state it could be review and iterated (hense this PR)

test/unit/indexer/compare_enumerator_test.rb
dwhenry: fixed

test/unit/indexer/compare_enumerator_test.rb
dwhenry: done

test/unit/indexer/compare_enumerator_test.rb
dwhenry: Add a test for multiple entries on each side.

test/unit/indexer/compare_enumerator_test.rb
```diff
@@ -0,0 +1,151 @@
...
+    stub_scroll_enumerator(left_request: [data_left], right_request: [data_right])
+
+    results = Indexer::CompareEnumerator.new('index_a', 'index_b').to_a
+    assert_equal results, [
+      [data_left, Indexer::CompareEnumerator::NO_VALUE],
+      [Indexer::CompareEnumerator::NO_VALUE, data_right],
+    ]
+  end
+
+  def test_correct_aligns_records_with_matching_keys
```
suzannehamilton, line 61: 👍

test/unit/indexer/comparer_test.rb
```diff
@@ -0,0 +1,86 @@
...
+    comparer = Indexer::Comparer.new(
+      'index_a',
+      'index_b',
+      io: StringIO.new
+    )
+    outcome = comparer.run
+    assert_equal outcome, { added_items: 1 }
+  end
+
+  def test_can_detect_when_a_record_is_remove
```
brenetic: `d`

dwhenry: fixed

lib/indexer/compare_enumerator.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      ScrollEnumerator.new(
+        client: client,
+        index_names: index_name,
+        search_body: search_body,
+        batch_size: BATCH_SIZE
+      ) do |document|
+        result = document['_source']
+        root_elements = %w(_id _type).each_with_object({}) do |key, h|
+          h["_root#{key}"] = document[key]
```
MatMoore: can you rename `h` to something more descriptive?

MatMoore (approved):
👍 Nice job, let's get this deployed

lib/indexer/compare_enumerator.rb
dwhenry: refactored to remove the need for this variable.

PR #831, opened 2017-07-04
Add user_journey_document_supertype

tijmenb:
This is the only supertype that is not indexed in search. We use it to differentiate between content that is the destination (the thing) and the part of the journey (the finding). In the short term I need this field to filter the results to get all documents that should be tagged. In the long term we might use it to create a finder (in which you would only expect "things").


brenetic (approved):
🚀

PR #833, opened 2017-07-12
Create new govuk index

suzannehamilton:
- Also add a rake task to allow this to be done in prod.

We need to create a new index in order to replace the existing
mainstream, government and detailed indices. All queries will therefore
be done against a single index. This is the first step in the migration
process.

Mobbed with @dwhenry and @brenetic

https://trello.com/c/2U7jTkC3/158-create-new-elasticsearch-index

### DIFF
--- a/config/schema/indexes/govuk.json
+++ b/config/schema/indexes/govuk.json
@@ -1,0 +1,31 @@
+{
+  "elasticsearch_types": [
+    "aaib_report",
+    "asylum_support_decision",
+    "business_finance_support_scheme",
+    "cma_case",
+    "contact",
+    "countryside_stewardship_grant",
+    "drug_safety_update",
+    "dfid_research_output",
+    "edition",
+    "employment_appeal_tribunal_decision",
+    "employment_tribunal_decision",
+    "european_structural_investment_fund",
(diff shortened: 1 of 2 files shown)


config/schema/indexes/govuk.json
```diff
@@ -0,0 +1,5 @@
+{
+  "elasticsearch_types": [
+    "aaib_report"
```
MatMoore: I think this could include all the document types at this stage, because

- We want to test we can populate the whole index asap, and measure the performance of that
- We're not planning to actually change the document types

MatMoore (dismissed):
👍 Looks good! Just think we should add in the other document types.

Something that just occurred to me is when/how we could switch to the new index for
- registries
- spelling corrections

Which is defined in https://github.com/alphagov/rummager/blob/master/elasticsearch.yml

I think this is a question for next sprint though.

config/schema/indexes/govuk.json
suzannehamilton: 👍 I've copied all the document types from the mainstream index.

binaryberry (approved):

PR #834, opened 2017-07-17
Insert data from Rabbit MQ into Elastic Search

brenetic:
This is MVP, we're not dealing with most errors, or too worried about the shape of the data going in. This PR is dependent on the queue already being created in Rabbit MQ, which requires a separate PR that we haven't done yet. 

https://trello.com/c/dmBzBVLG/160-insert-data-into-govuk-index-from-rabbitmq-message

mobbed with @suzannehamilton @binaryberry @dwhenry @Rosa-Fox


lib/govuk_index/elasticsearch_presenter.rb
```diff
@@ -0,0 +1,23 @@
+module GovukIndex
+  class ElasticsearchPresenter
+    def initialize(payload)
+      @payload = payload
+    end
+
+    def call
+      raise Indexer::ValidationError.new("base_path missing from payload") if payload["base_path"].nil?
```
dwhenry: Just a thought on increasing the validations on this class. It might be worth looking at using the ActiveModel validations code when we extend the checks we are doing (I am not suggesting changing this now)

lib/govuk_index/publishing_event_processor.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def initialize(index_name)
+      @index_name = index_name
+    end
+
+    def process(message)
+      document = ElasticsearchPresenter.new(message.payload).call
+      save_to_elastic_search(document)
+      message.ack
+    rescue Indexer::ValidationError => e
```
dwhenry: I thought the plan was to avoid reuse of the `Indexer` namespaced code? Any reason we have declared the error class there?

test/integration/fixtures/publishing_api_payload.json
```diff
@@ -0,0 +1,216 @@
+{
+	"title": "Pitts S-2B, G-SKYD, 21 June 1996",
```
dwhenry: any reason we have used 4 space indentation? the existing JSON file is using 2 space indentation.

test/integration/publishing_event_processor_test.rb
```diff
@@ -0,0 +1,64 @@
...
+      rabbitmq_connection: bunny_mock
+    )
+
+    @queue = channel.queue("bigwig.test")
+    consumer.run
+  end
+
+  def test_should_save_new_document_to_elasticsearch
+    fixture_file = File.expand_path("../fixtures/publishing_api_payload.json", __FILE__)
+    payload = JSON.parse(File.read(fixture_file))
```
dwhenry: The consumer is setup to do JSON parsing internally, as discuss previously I believe that we should be passing in JSON data so that we match the production implementation as closely as possible. I think this can be done using:

```
@queue.publish(json_payload, headers: { content_type: "application/json" })
```

test/integration/publishing_event_processor_test.rb
```diff
@@ -0,0 +1,64 @@
...
+      "document_type" => "aaib_report"
+    }
+
+    Airbrake.expects(:notify_or_ignore).with(instance_of(Indexer::ValidationError), error_message: "base_path missing from payload")
+    @queue.publish(invalid_payload)
+
+    assert_equal 0, @queue.message_count
+  end
+
+  def test_should_requeue_message_when_processing_fails
```
dwhenry: test name does not match test action.

test/integration/search/search_test.rb
```diff
@@ -102,7 +102,7 @@ def test_filter_by_field
 
     get "/search?filter_mainstream_browse_pages=browse/page/1"
 
-    assert_equal ["/government-1", "/mainstream-1"],
+    assert_equal ["/government-1", "/govuk-1", "/mainstream-1"],
```
dwhenry: Would like to discuss this as we would want to exclude any data from the `govuk` index from the results for the time being. Is this here intentionally or because it was the easiest way to set everything up?

test/support/test_index_helpers.rb
```diff
@@ -1,6 +1,6 @@
 class TestIndexHelpers
   AUXILIARY_INDEX_NAMES = ["page-traffic_test", "metasearch_test"].freeze
-  INDEX_NAMES = %w(mainstream_test government_test).freeze
+  INDEX_NAMES = %w(mainstream_test government_test govuk_test).freeze
```
dwhenry: Again, we should be considering `govuk` as a separate thing from the existing index list - I think.

test/integration/publishing_event_processor_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    @queue.publish(payload)
+
+    document = fetch_document_from_rummager(id: "/aaib-reports/pitts-s-2b-g-skyd-21-june-1996", index: "govuk_test")
+
+    assert_equal "/aaib-reports/pitts-s-2b-g-skyd-21-june-1996", document["_source"]["link"]
+    assert_equal "/aaib-reports/pitts-s-2b-g-skyd-21-june-1996", document["_id"]
+    assert_equal "aaib_report", document["_type"]
+  end
+
+  def test_should_raise_error_when_invalid_payload
```
dwhenry: I would prefer to see the validation checking in the unit tests. I don't feel this is necessarily worth an integration test.

dwhenry (commented):
All looks pretty good, just a couple of comments/questions. Think it is worth discussing at least some of the points to ensure we are all moving in the same direction. :)

lib/govuk_index/elasticsearch_presenter.rb
binaryberry: Sure - we'll do this once we get to adding more validations

lib/govuk_index/publishing_event_processor.rb
binaryberry: done

test/integration/fixtures/publishing_api_payload.json
binaryberry: fixed

test/integration/publishing_event_processor_test.rb
binaryberry: fixed!

test/integration/publishing_event_processor_test.rb
binaryberry: fixed

lib/govuk_index/publishing_event_processor.rb
```diff
@@ -0,0 +1,46 @@
+require 'govuk_index/elasticsearch_presenter'
+
+module GovukIndex
+  class ValidationError < StandardError; end
+
+  class PublishingEventProcessor
+    TIMEOUT_SECONDS = 5.0
+
+    def initialize(index_name)
```
MatMoore: Is `index_name` the long index name or the alias name (I think it should be the latter)

test/integration/search/search_test.rb
binaryberry: Because it was the easiest way to set everything up - amended now!

test/support/test_index_helpers.rb
binaryberry: sure, amended now, we assigned govuk_test to a different constant

test/integration/fixtures/publishing_api_payload.json
```diff
@@ -0,0 +1,216 @@
+{
```
tijmenb: I would avoid using fixtures, as they will get out of date (and you won't know). You can generate a random, valid, payload using the `govuk_schemas` gem:

```
GovukSchemas::RandomExample.for_schema(notification_schema: "specialist_document")
```

http://rubydoc.info/gems/govuk_schemas

Procfile
```diff
@@ -1,2 +1,3 @@
 worker: bundle exec sidekiq -C ./config/sidekiq.yml
 publishing-queue-listener: bundle exec rake message_queue:listen_to_publishing_queue
+govuk-index-queue-worker: bundle exec rake message_queue:insert_data_into_govuk
```
tijmenb: I think I chose the name `-listener` rather that worker to differentiate it from a internal "worker". Might be worth keeping that convention (but I don't feel strongly about it).

lib/govuk_index/elasticsearch_presenter.rb
```diff
@@ -0,0 +1,27 @@
...
+        _source: {
+          link: payload["base_path"],
+          title: payload["title"]
+        }
+      }
+    end
+
+    def valid!
+      return if payload["base_path"]
+      raise(ValidationError, "base_path missing from payload")
```
tijmenb, line 23: There's legitimate cases where valid content won't have base paths (world locations I think) - perhaps you want to just skip those instead of raising?

test/support/test_index_helpers.rb
```diff
@@ -1,8 +1,9 @@
 class TestIndexHelpers
   AUXILIARY_INDEX_NAMES = ["page-traffic_test", "metasearch_test"].freeze
   INDEX_NAMES = %w(mainstream_test government_test).freeze
+  GOVUK_INDEX_NAME = "govuk_test".freeze
   DEFAULT_INDEX_NAME = INDEX_NAMES.first
-  ALL_TEST_INDEXES = (AUXILIARY_INDEX_NAMES + INDEX_NAMES).freeze
+  ALL_TEST_INDEXES = ([GOVUK_INDEX_NAME] + AUXILIARY_INDEX_NAMES + INDEX_NAMES).freeze
```
tijmenb, line 6: This might have changed in the meantime, but adding a new index here used to be super expensive and slow down the tests a lot.

lib/govuk_index/publishing_event_processor.rb
```diff
@@ -0,0 +1,46 @@
+require 'govuk_index/elasticsearch_presenter'
+
+module GovukIndex
```
tijmenb: Perhaps it would be good to separate out the indexing from the rabbitmq processing? This class mixes concerns a bit. Separating them would make the tests cleaner too probably.

tijmenb (commented):
See comments below. 

One broader thing: the way it's set up now assumes that we either handle the message successfully or unsuccessfully, in which case we `discard`. This doesn't account for when elasticsearch times out or there's a transient network error. 

It might be an idea to put the incoming messages directly onto a sidekiq queue. That will give you automatic retries with exponential back-off, automatic instrumentation and (important later) more concurrency.

test/integration/fixtures/publishing_api_payload.json
tijmenb: Related in email-alert-service:

https://github.com/alphagov/email-alert-service/blob/master/spec/integration/random_input_spec.rb#L10-L23

lib/govuk_index/elasticsearch_presenter.rb
dwhenry: We have chosen to leave this as is and will deal with valid cases that fail this validation in a future PR (we will be significantly reworking the validation model)

test/support/test_index_helpers.rb
dwhenry: The index will be created as a once off task before the tests run, and in the long term we are hoping to remove the indices defined in `INDEX_NAMES`

brenetic:
@tijmenb We have addressed all the comments. Could you have another gander please?

lib/govuk_index/elasticsearch_saver.rb
```diff
@@ -0,0 +1,32 @@
...
+      client.bulk(
+        index: index_name,
+        body: [
+          { index: presenter.identifier },
+          presenter.document
+        ]
+      )
+    end
+
+    def client(options = {})
```
suzannehamilton: Should this be a private method? I can't see why it would want to share the client with other objects.

lib/govuk_index/publishing_event_processor.rb
```diff
@@ -0,0 +1,10 @@
+require 'govuk_index/publishing_event_worker'
+
+module GovukIndex
+  class PublishingEventProcessor
+    def process(message)
+      PublishingEventWorker.perform_async(message.payload)
+      message.ack
```
suzannehamilton, line 7: Do we have to worry about the case where we fail to add the message to the Sidekiq queue?

lib/govuk_index/elasticsearch_presenter.rb
MatMoore: I think there are also contact content items that don't have base paths.

lib/govuk_index/publishing_event_processor.rb
brenetic: We will cover this in future tasks

dwhenry (approved):
By mob rule..

PR #835, opened 2017-07-20
Update test dependencies

suzannehamilton:
- Re-enable and fix a test which wasn't being run
- Replace test-unit-minitest with test-unit because test-unit-minitest is not compatible with Minitest 5
- Replace minitest-colorize with minitest/pride because minitest-colorize is not compatible with Minitest 5
- Upgrade minitest from version 4 to 5

The immediate reason for doing this is that I want to add some more gems while doing some exploration of the Elasticsearch data and they aren't compatible with old versions of minitest. But updating the gems generally seems like a good idea.

I've added a tech debt ticket to our backlog about upgrading more of the old gems.


cbaines (approved):
Looks good :)

PR #836, opened 2017-07-21
[MERGING 11 Sep] Elasticsearch 2 4

dwhenry:
This is an initial implementation of rummager to work with ES 2.4 with all test passing.

Please refer to: https://docs.google.com/document/d/1eWyUXbub1p2D-BPVcqS2zPnh5PnjeNzlbhf4D53brtQ/edit for issues that will need to be resolved prior to the code going live.

https://trello.com/c/3JnIm4rM/161-load-existing-data-into-es-24-and-make-a-list-of-issues

### DIFF
--- a/lib/indexer/compare_enumerator.rb
+++ b/lib/indexer/compare_enumerator.rb
@@ -74,7 +74,7 @@ module Indexer
     end
 
     def search_config
-      @search_config ||= SearchConfig.new
+      @search_config ||= SearchConfig.instance
     end
   end
 end
--- a/lib/search/query.rb
+++ b/lib/search/query.rb
@@ -85,11 +85,11 @@ module Search
 
     def process_elasticsearch_errors
       yield
-    rescue Elasticsearch::Transport::Transport::Errors::BadRequest => e
+    rescue Elasticsearch::Transport::Transport::Errors::InternalServerError => e
       case e.message
       when /Numeric value \(([0-9]*)\) out of range of/
         raise(NumberOutOfRange, "Integer value of #{$1} exceeds maximum allowed")
-      when /TooManyClauses\[maxClauseCount is set to/
+      when /maxClauseCount is set to/
         raise(QueryTooLong, 'Query must be less than 1024 words')
       else
         raise
--- a/lib/search/query_builder.rb
+++ b/lib/search/query_builder.rb
@@ -128,7 +128,10 @@ module Search
       end
 
       {
-        more_like_this: { docs: docs }
+        more_like_this: {
+          docs: docs,
+          min_doc_freq: 0, # This overrides the new ES 2.4 default with the ES 1.7 default so that we can complete the migration
+        }
       }
     end
   end
--- a/lib/search/query_builder.rb
+++ b/lib/search/query_builder.rb
@@ -130,7 +130,7 @@ module Search
       {
         more_like_this: {
           docs: docs,
-          min_doc_freq: 0, # This overrides the new ES 2.4 default with the ES 1.7 default so that we can complete the migration
+          min_doc_freq: 0, # Revert to the ES 1.7 default
         }
       }
     end
--- a/lib/search_config.rb
+++ b/lib/search_config.rb
@@ -1,6 +1,5 @@
 class SearchConfig
   %w[
-    base_uri
     registry_index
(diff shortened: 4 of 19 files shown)


lib/indexer/compare_enumerator.rb
```diff
@@ -74,7 +74,7 @@ def client
     end
 
     def search_config
-      @search_config ||= SearchConfig.new
+      @search_config ||= Rummager.search_config
```
brenetic: Would this need to revert considering how we got stuff working for the `govuk_index` PR?

dwhenry: yep, rebase and update.. :)

lib/search_config.rb
```diff
@@ -1,5 +1,6 @@
 class SearchConfig
   %w[
+    base_uri
```
suzannehamilton: This looks like a duplicate - it's already in the array on line 10.

test/support/integration_test.rb
```diff
@@ -73,8 +78,9 @@ def clean_index_content(index)
   end
 
   def commit_document(index_name, attributes, id: attributes["link"], type: "edition")
-    insert_document(index_name, attributes, id: id, type: type)
-    commit_index(index_name)
+    insert_document(index_name, attributes, id: id, type: type).tap do
```
suzannehamilton: Is `tap` being used to ensure that the return value is the ID returned by `insert_document`? If so, I think it would be much clearer to just assign a temporary variable - I had to read this a few times to work out why it was written like this (and I'm still not totally sure 🙂).

suzannehamilton (approved):
A couple of minor comments, but otherwise this looks good. 👍

lib/search_config.rb
dwhenry: Removed

test/support/integration_test.rb
dwhenry: Ya, not sure why I went with that approach.

lib/indexer/compare_enumerator.rb
```diff
@@ -74,7 +74,7 @@ def client
     end
 
     def search_config
-      @search_config ||= SearchConfig.new
+      @search_config ||= SearchConfig.instance
```
brenetic, line 77: What does this give `SearchConfig` that is different from `new`?

dwhenry: I think I did this to help with testing as it gets the stubs on the instance by default in integration tests. But I am not 100% sure. The `instance` method just returns a memorised copy of `new`

lib/search/query.rb
```diff
@@ -85,11 +85,11 @@ def fetch_spell_checks(search_params)
 
     def process_elasticsearch_errors
       yield
-    rescue Elasticsearch::Transport::Transport::Errors::BadRequest => e
+    rescue Elasticsearch::Transport::Transport::Errors::InternalServerError => e
```
MatMoore, line 88: What's the API change behind this? It feels odd for it not to raise BadRequest here...

lib/search/query_builder.rb
```diff
@@ -128,7 +128,10 @@ def more_like_this_query_hash
       end
 
       {
-        more_like_this: { docs: docs }
+        more_like_this: {
+          docs: docs,
+          min_doc_freq: 0, # This overrides the new ES 2.4 default with the ES 1.7 default so that we can complete the migration
```
MatMoore: Do we have a card to revisit this?

MatMoore (approved):

lib/search/query_builder.rb
dwhenry: Not at this stage. It's an odd choice they have made with this as it means how the cluster is balanced can have an impact on `more_like_this` results for low volume terms.

Having the default as 0 feels like the right approach, I never investigated the reasons for the change, as such I might change the message text to explain what it does better and remove the reference to getting tests to pass.

Thoughts?

lib/search/query.rb
dwhenry: This means that if either of the below two errors occurs ES is now returning a `500` response, while ES 1.7 returned a `400` response. 

I don't have any further details on why this change was made. I can investigate further if you think it is important?

PR #838, opened 2017-07-27
Fix integration tests

Rosa-Fox:
- Tests have been intermittently failing recently.
- This PR fixes some of those failures.
- We don't fully understand why this has only just become a problem or why we are only seeing this on CI and not in development. 
- We are continuing to look into this to fix the remaining failures.
- See commit messages for further explanation.
- See #746 for more context on why the integration tests set up and delete indexes before and after the whole test suite rather than for each individual test.

Paired with @suzannehamilton

### DIFF
--- a/test/integration/indexer/index_group_test.rb
+++ b/test/integration/indexer/index_group_test.rb
@@ -11,8 +11,11 @@ class ElasticsearchIndexGroupTest < IntegrationTest
   end
 
   def teardown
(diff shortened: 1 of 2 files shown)


dwhenry (approved):
If this help gets stuff passing, as you say it would be good have an idea of the underlying cause as this feel like we are addressing the problem rather than the cause.

PR #839, opened 2017-07-27
Update docs path in READMEs

brenetic:
~~docs/~~ => `doc`


suzannehamilton (approved):
I think we'll also need to fix the developer docs, which pull in some of these files: https://docs.publishing.service.gov.uk/apis/search/search-api.html

PR #840, opened 2017-07-31
Add page timestamps to sitemap

suzannehamilton:
- Refactor method which fetches all documents. This gives the sitemap generator access to more fields, such as the last modified date (which is used in the next commit) and the format (which will be used to modify page priority in a later PR).
- Add a `lastmod` field to entries in the sitemap, based on the `public_timestamp` field in Elasticsearch. This will help search engines rank GOV.UK pages against each other by boosting more recent content. This should help users to find more recent, relevant pages when searching in an external search engine.

https://trello.com/c/DJx7rxiy/190-add-timestamps-to-sitemap

### DIFF
--- a/lib/sitemap/sitemap_generator.rb
+++ b/lib/sitemap/sitemap_generator.rb
@@ -13,13 +13,11 @@ class SitemapGenerator
   def get_all_documents
     Enumerator.new do |yielder|
       # Hard-code the site root, as it isn't listed in any search index
-      yielder << "/"
+      yielder << homepage_document
 
       @sitemap_indices.each do |index|
-        index.all_document_links(EXCLUDED_FORMATS).each do |document|
-          if document
-            yielder << document
-          end
+        index.all_documents(exclude_formats: EXCLUDED_FORMATS).each do |document|
+          yielder << document
         end
       end
     end
@@ -34,8 +32,9 @@ class SitemapGenerator
   def generate_xml(chunk)
     builder = Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
       xml.urlset(xmlns: "http://www.sitemaps.org/schemas/sitemap/0.9") do
-        chunk.each do |url|
-          url = Array(url).first
+        chunk.each do |document|
+          url = document.link
+
           url = URI.join(base_url, url) unless url.start_with?("http")
           xml.url {
             xml.loc url
@@ -50,5 +49,11 @@ private
 
   def base_url
     Plek.current.website_root
+  end
+
+  StaticDocument = Struct.new(:link, :public_timestamp)
+
+  def homepage_document
+    StaticDocument.new("/", nil)
   end
 end
--- a/lib/sitemap/sitemap_generator.rb
+++ b/lib/sitemap/sitemap_generator.rb
@@ -4,6 +4,7 @@ class SitemapGenerator
   def initialize(sitemap_indices)
     @sitemap_indices = sitemap_indices
     @all_documents = get_all_documents
+    @logger = Logging.logger[self]
   end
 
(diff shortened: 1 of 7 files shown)


lib/sitemap/sitemap_generator.rb
```diff
@@ -34,15 +35,26 @@ def sitemaps
   def generate_xml(chunk)
     builder = Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
       xml.urlset(xmlns: "http://www.sitemaps.org/schemas/sitemap/0.9") do
-        chunk.each do |url|
-          url = Array(url).first
-          url = URI.join(base_url, url) unless url.start_with?("http")
-          xml.url {
-            xml.loc url
-          }
+        chunk.each do |document|
```
dwhenry: do we still need the chunking?

lib/sitemap/sitemap_generator.rb
```diff
@@ -34,15 +35,26 @@ def sitemaps
...
+          if document == "/"
```
dwhenry: can we replace this root node placeholder with something better? we could avoid this if condition if it responded to link with '\'

dwhenry (changes requested):
just a small change.

lib/sitemap/sitemap_generator.rb
suzannehamilton: Yes, I think so. The sitemap is split into separate files because the [sitemap specification](https://www.sitemaps.org/protocol.html) limits each file to 50,000 URLs.

lib/sitemap/sitemap_generator.rb
suzannehamilton: 👍 I've added a struct which responds to `link` and `public_timestamp`. What do you think?

lib/sitemap/sitemap_generator.rb
```diff
@@ -13,10 +14,10 @@ def self.sitemap_limit
   def get_all_documents
     Enumerator.new do |yielder|
       # Hard-code the site root, as it isn't listed in any search index
-      yielder << "/"
+      yielder << homepage_document
 
       @sitemap_indices.each do |index|
-        index.all_document_links(EXCLUDED_FORMATS).each do |document|
+        index.all_documents(exclude_formats: EXCLUDED_FORMATS).each do |document|
           if document
```
dwhenry: so this catch was previously guarding against instances of the link field being falsy, so we still need this? can the link field be falsy?

dwhenry (approved):
Just had that one last question, but otherwise looks good.

lib/sitemap/sitemap_generator.rb
suzannehamilton: Good question. In principle this is possible because it's not a required field. In practice, every document in ES does have a `link` field. (I've just checked on production.)

Do you think it's worth checking during sitemap generation?

dwhenry: I feel like we should be able to remove this check. We should add something to the new govuk index work to ensure the field is never blank.

@MatMoore do you agree?

suzannehamilton:
Marking as DO NOT MERGE because the plan is to release all the sitemap changes at the same time.

lib/sitemap/sitemap_generator.rb
MatMoore: Fine with me. Worst case scenario we have some invalid entries in the sitemap, which will just be ignored.

suzannehamilton: I've got rid of the `if document`. 👍

suzannehamilton:
I'm merging this into the feature branch. We'll merge everything into master together when we're ready to deploy.

PR #841, opened 2017-07-31
Down-weight withdrawn pages in the sitemap

suzannehamilton:
Set a priority for each page in the sitemap:
- For regular, non-withdrawn pages, set the priority to 1
- For all withdrawn pages, set the priority to 0.25

The priority is a value from 0 to 1 which influences the relative ranking of pages in search engines.

The values of 1 and 0.25 are fairly arbitrary because we have not been able to find advice on how to pick relative priorities. The aim is to push relevant non-withdrawn content above withdrawn content so that users are not confused by old guidance or publications, but still keep the withdrawn content findable in external search if it really is the most relevant information.

Probably easiest to review each commit separately, because the second one is just refactoring.

https://trello.com/c/iR68Sm6R/192-downweight-withdrawn-content-in-the-sitemap

Marked as DO NOT MERGE until all the sitemap changes are ready to be deployed.

### DIFF
--- a/test/integration/sitemap/sitemap_test.rb
+++ b/test/integration/sitemap/sitemap_test.rb
@@ -101,6 +101,17 @@ class SitemapTest < IntegrationTest
     assert_equal "2017-07-01T12:41:34+00:00", pages[0].css("lastmod").text
   end
 
+  def test_links_should_include_priorities
+    generator = SitemapGenerator.new(search_server.content_indices)
+
+    sitemap_xml = generator.sitemaps
+
+    priorities = Nokogiri::XML(sitemap_xml[0])
+      .css("url > priority")
+
+    assert_equal 6, priorities.count
+  end
+
 private
 
   def add_sample_documents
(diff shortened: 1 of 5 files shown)


test/integration/sitemap/sitemap_test.rb
```diff
@@ -101,6 +101,17 @@ def test_links_should_include_timestamps
...
 
+  def test_links_should_include_priorities
+    generator = SitemapGenerator.new(search_server.content_indices)
+
+    sitemap_xml = generator.sitemaps
+
+    priorities = Nokogiri::XML(sitemap_xml[0])
+      .css("url > priority")
+
+    assert_equal 6, priorities.count
```
suzannehamilton, line 112: I've deliberately left this as a weak assertion because the exact values might change, especially if we add format weighting. The unit tests make some stricter assertions.

Thoughts?

brenetic:
🚀

suzannehamilton:
Accidentally closed this by deleting the base branch. I've now switched the base to be the new feature branch. 👍

PR #842, opened 2017-08-01
Add format weighting to sitemap

suzannehamilton:
Probably easiest to review commit-by-commit:

- Small fix of assertion parameters in existing test
- Split unit tests for the sitemap generator and sitemap presenter to make the next change easier
- Set sitemap priorities based on format

  Take the format boosting configured for internal search and apply it to sitemap priorities for external search.

  Search engines use these values to rank GOV.UK pages against each other, so this should help increase the position of more generally relevant pages in external search results.

I'll raise a separate PR for boosting pages based on document supertype, which will boost all guidance content. It would be good to get some feedback on this approach first, though.

https://trello.com/c/wTcs0JaG/194-add-format-weightings-to-sitemap

### DIFF
--- a/lib/sitemap/format_boost_calculator.rb
+++ b/lib/sitemap/format_boost_calculator.rb
@@ -1,0 +1,19 @@
+class FormatBoostCalculator
+  def initialize
+    format_boost_config = YAML.load_file('config/query/format_boosting.yml')
+    @format_boosts = format_boost_config["format_boosts"]
+  end
+
+  def boost(format)
+    page_boost = @format_boosts[format] || 1
+
+    # Normalise format boost to always give a value between 0 and 1
+    page_boost.to_f / max_boost
+  end
+
+private
+
+  def max_boost
+    @_max_boost ||= @format_boosts.values.max
+  end
+end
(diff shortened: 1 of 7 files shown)


MatMoore (approved):

lib/sitemap/format_boost_calculator.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def initialize
+    @format_boosts = FORMAT_BOOST_CONFIG["format_boosts"]
+  end
+
+  def boost(format)
+    page_boost = @format_boosts[format] || 1
+
+    # Normalise format boost to always give a value between 0 and 1
+    page_boost / max_boost
```
dwhenry: there is a small risk here that both values could be integers resulting in an incorrect result.

```page_boost.to_f / max_boost```

dwhenry (approved):
looks good.

lib/sitemap/format_boost_calculator.rb
suzannehamilton: 👍 I've updated it and added a test.

I've also added the rounding that we discussed in person: all the priorities are now rounded to two decimal places.

PR #845, opened 2017-08-01
Add message payload to ValidationError

brenetic:
ValidationErrors sent to Airbrake now contain the message payload

Mobbed with ❤️ along with @binaryberry and @dwhenry


dwhenry (approved):
mob approval

PR #846, opened 2017-08-02
Apply more weightings to page priorities in the sitemap

suzannehamilton:
Probably easiest to review commit-by-commit:

- Some refactoring commits to make it easier to add other types of boosting to sitemap priorities
- Apply more boosts to page priorities in the sitemap

  Internal search results are upweighted or downweighted based on several
properties of a document, such as whether it is guidance or whether it
is marked as historic (i.e. it is a publication by a previous
government).

  This commit applies the same boosts to *external* search by using them
as a multiplier in the sitemap priorities. This should improve the
relative rankings of GOV.UK pages in external search results, helping
users to find the information that they're more likely to be searching
for.

https://trello.com/c/wTcs0JaG/194-add-format-weightings-to-sitemap


dwhenry (changes requested):
Looks really good, but you have left a couple of reference to `format_boost_calculator` when doing the refactor.

suzannehamilton:
Good spot! Now fixed. 👍

lib/search/query_components/booster.rb
```diff
@@ -23,39 +23,20 @@ def wrap(core_query)
...
     def boost_filters
-      format_boosts + [
-        time_boost,
-        closed_org_boost,
-        devolved_org_boost,
-        historic_edition_boost,
-        guidance_boost,
-        foi_boost,
-      ]
+      format_boosts + [time_boost]
```
dwhenry: Might be worth renaming this variable as well :(

dwhenry (approved):
And just the one more. Happy for you to ignore and update later. :)

lib/search/query_components/booster.rb
suzannehamilton: Done. 👍

PR #847, opened 2017-08-02
Delete ES record when unpublishing message is received

dwhenry:
This code deletes the record for all unpublished messages, and
Lays some groundwork for supporting batch processing at a later
stage.

https://trello.com/c/pxp1XtNt/198-unpublish-message-should-remove-record-from-es

### DIFF
--- a/lib/govuk_index/elasticsearch_processor.rb
+++ b/lib/govuk_index/elasticsearch_processor.rb
@@ -2,16 +2,27 @@ require 'services'
 require 'search_config'
 
 module GovukIndex
-  class ElasticsearchSaver
+  class ElasticsearchProcessor
     TIMEOUT_SECONDS = 5.0
 
+    def initialize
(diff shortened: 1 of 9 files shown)


lib/govuk_index/elasticsearch_actions.rb
```diff
@@ -2,16 +2,27 @@
 require 'search_config'
 
 module GovukIndex
-  class ElasticsearchSaver
+  class ElasticsearchActions
```
brenetic: ElasticsearchActioner or ElasticsearchProcessor ?

brenetic (approved):
Naming aside 🚀

lib/govuk_index/elasticsearch_actions.rb
dwhenry: gone with `ElasticsearchProcessor`

PR #848, opened 2017-08-03
Increase the ElasticSearch request timeout for index creation

suzannehamilton:
This fixes the intermittently failing integration tests. These were failing because index creation sometimes takes 6 or 7 seconds, but the request timeout was set to 5 seconds for all requests other than bulk reloading.

Confusingly, the error seen in the integration tests was not a timeout, but an error like this:

```
Elasticsearch::Transport::Transport::Errors::BadRequest: [400]
{"error":"IndexAlreadyExistsException[[mainstream_test-2017-07-31t09:41:05z-e5229cbf-6da5-4eb2-af27-7f9ad623ef53]
already exists]","status":400}
```

This is because the elasticsearch-transport library uses Faraday, which uses Net::HTTP under the hood. Net::HTTP automatically retries any idempotent requests which time out, including the PUT request to create an index. This request _is_ idempotent, but the second request returns a 400 because the index has already been created.

https://trello.com/c/vbOLcr7b/208-fix-intermittent-indexalreadyexistsexception-error-in-tests

### DIFF
--- a/lib/index_group.rb
+++ b/lib/index_group.rb
@@ -15,6 +15,9 @@ module SearchIndices
     def initialize(base_uri, name, schema, search_config)
       @base_uri = base_uri
       @client = Services::elasticsearch(hosts: base_uri)
+      # Index creation can take longer than other requests, so create a separate
+      # client with a longer timeout
+      @index_creation_client = Services::elasticsearch(hosts: base_uri, timeout: 10)
       @name = name
       @schema = schema
       @search_config = search_config
@@ -32,7 +35,7 @@ module SearchIndices
         "settings" => settings,
         "mappings" => mappings,
       }
-      @client.indices.create(
+      @index_creation_client.indices.create(
         index: index_name,
         body: index_payload
       )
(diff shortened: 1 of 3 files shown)


lib/index_group.rb
```diff
@@ -15,6 +15,9 @@ class IndexGroup
     def initialize(base_uri, name, schema, search_config)
       @base_uri = base_uri
       @client = Services::elasticsearch(hosts: base_uri)
+      # Index creation can take longer than other requests, so create a separate
+      # client with a longer timeout
+      @index_creation_client = Services::elasticsearch(hosts: base_uri, timeout: 10)
```
dwhenry, line 20: Are we happy to do this everywhere, or are we confident enough this is a testing issue (as that is the only place we regularly do multiple slow calls together that we could make the timeout environment specific?

dwhenry: Also, would it not be easier to just change the default to 10 seconds?

suzannehamilton: @MatMoore and I had a chat about this. We think that it's safe to do this everywhere because we don't create indexes very often (I think it's just the overnight bulk import) and it's OK if that takes longer than a typical fast request which fetches or updates data.

MatMoore (approved):
🎉

lib/index_group.rb
suzannehamilton: Sorry, I missed your second question. I think we want to keep the default timeout low(ish) because the other types of requests should be fast. We'd want to know if they became much slower, even if the first indication was request timeouts.

PR #849, opened 2017-08-03
Set withdrawn flag for withdrawn content

brenetic:
When documents are withdrawn, we receive a message with the unpublished routing key but with the full payload for that format.

In rummager, we don't display withdrawn documents in search results (yet) but we still index them.

Trello: https://trello.com/c/nKQIS5Gu/209-unpublished-withdrawn-should-flag-the-document-instead-of-removing-it
Mobbed with @Rosa-Fox @MatMoore @dwhenry


dwhenry (approved):
mob merge

PR #850, opened 2017-08-07
Compact requires

dwhenry:
This is something I have spoken about previously and as I had a small amount of free time the other day while waiting for other tasks to complete I did a quick spike.

This is now a real thing and passes all the tests. It is my belief that this makes the dependency tree easier to understand and ensures our tests run in a manner which is more consistent with production. The downside is that load times will be slightly slower, however, I have not seen this in practice.

Anyway have a look and let me know if you agree this is a simplification. :)

https://trello.com/c/zBx4635B/216-load-app-without-loading-sinatra


brenetic:
Very neat!

Are we entering 'this should have been built as a Rails API app' territory? I thought that one of the reasons for using Sinatra was that it didn't load the entire app in a Rails way but rather only when things were required? For example is there really a benefit to loading the entire app in a rake task as opposed to just the bits we need?

dwhenry:
@brenetic 

This is to help simplify how we load the environment for background tasks, tests and from the console.

In term of loading the full environment for all the rails applications, no it isn't completely necessary for all our tasks, but having a consistent approach is easier to reason, and means we don;t have odd bits of code that are only included in special circumstances (i.e. health check).

Maybe converting this to a rails app would make maintenance easier, it would simplify the learning curve for new developers, as rails apps have a set layout/location which I find most Sinatra app lack.

brenetic (approved):
🚀

PR #851, opened 2017-08-07
Simplify query builder

MatMoore:
Extract query logic from QueryComponents::CoreQuery and remove QueryComponents::Query.

It's hard to see the structure of the query when its split over so many files, and it's confusing to have so many classes called "Query".

`QueryBuilder#query` should show how the query is structured and what fields are being used:

```ruby
    def query
      return more_like_this_query_hash unless search_params.similar_to.nil?
      return { match_all: {} } if search_params.query.nil?

      core_query = QueryComponents::CoreQuery.new(search_params)

      best_bets.wrap(
        popularity_boost.wrap(
          format_boost.wrap(
            if search_params.enable_new_weighting?
              new_weighting_query
            elsif search_params.quoted_search_phrase?
              core_query.quoted_phrase_query
            else
              {
                bool: {
                  must: [core_query.optional_id_code_query].compact,
                  should: [
                    core_query.match_phrase("title"),
                    core_query.match_phrase("acronym"),
                    core_query.match_phrase("description"),
                    core_query.match_phrase("indexable_content"),
                    core_query.match_all_terms(%w(title acronym description indexable_content)),
                    core_query.match_bigrams(%w(title acronym description indexable_content)),
                    core_query.minimum_should_match("_all")
                  ],
                }
              }
            end
          )
        )
      )
    end
```

There is a bunch of stuff that's only used when debug params are passed: we should look at removing some of this.

I think this is worth doing now because we'll need to make the query more complex to allow switching back and forth between the old and new search indexes. I'd also like to add in functionality for versioning the API.


dwhenry (approved):
I think this is a great idea and will help simplify the code.

I have some thoughts on additional changes but think we should merge this now as it is an improvement and schedule time to do further work on simplification of this part of the codebase rather than get bogged down.

PR #852, opened 2017-08-08
Add 'elasticsearch' Jenkins tag

dwhenry:
This is required so that we can support elasticsearch 2.4 builds for
our upgrade branch without causing other branches to fail (as they
still require elasticsearch 1.7)

https://trello.com/c/7UxzIqNo/205-update-es-version-for-ci


brenetic (approved):
🚀

PR #853, opened 2017-08-08
Add versioning

brenetic:
**Add external versioning to documents in Elasticsearch:**
Use the `payload_version` field from the publishing API to add a document version. This will prevent old versions of documents overwriting new versions if the publishing events are processed out of
order.

**Handle messages published out of order:**
If messages from the publishing-api arrive out of order and cause a versioning error in Elasticsearch, discard the error and increment the statsd counter.

**Bump govuk_schemas gem to 2.2.0:**
This enables us to exclude fields when using the RandomExample generator in integration tests

**Fix intermitent failures of unpublishing tests**
RandomExampleGenerator was generating a `withdrawn_notice` field 50% of the time, even when we were trying to test unpublishing messages which were not withdrawn. We changed the tests to use `customise_and_merge` which now allows us to exclude the `withdrawn_notice`.

[Trello](https://trello.com/c/LatZVRg8/213-make-versioning-work-for-all-publishing-events)

Mobbed with @suzannehamilton, @dwhenry and @Rosa-Fox.

### DIFF
--- a/lib/govuk_index/elasticsearch_presenter.rb
+++ b/lib/govuk_index/elasticsearch_presenter.rb
@@ -7,7 +7,9 @@ module GovukIndex
     def identifier
       {
         _type: payload["document_type"],
-        _id: payload["base_path"]
+        _id: payload["base_path"],
+        version: payload["payload_version"],
+        version_type: "external",
       }
     end
 
@@ -15,7 +17,7 @@ module GovukIndex
       {
         link: payload["base_path"],
         title: payload["title"],
-        is_withdrawn: withdrawn?
+        is_withdrawn: withdrawn?,
       }
     end
 
--- a/test/integration/govuk_index/versioning_test.rb
+++ b/test/integration/govuk_index/versioning_test.rb
@@ -1,0 +1,42 @@
+require 'govuk_schemas'
+require 'integration_test_helper'
+require 'bunny-mock'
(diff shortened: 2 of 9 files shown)


lib/govuk_index/elasticsearch_presenter.rb
```diff
@@ -7,7 +7,9 @@ def initialize(payload)
     def identifier
       {
         _type: payload["document_type"],
-        _id: payload["base_path"]
+        _id: payload["base_path"],
+        version: payload["payload_version"],
+        version_type: "external"
```
dwhenry: can we have a trailing comma

test/integration/govuk_index/versioning_test.rb
```diff
@@ -0,0 +1,115 @@
...
+    assert_equal 2, document["_version"]
+
+    version2 = version1.merge(payload_version: 1)
+
+    @processor.process(stub_message_payload(version2, unpublishing: true))
+
+    document = fetch_document_from_rummager(id: base_path, index: "govuk_test")
+    assert_equal 2, document["_version"]
+  end
+end
```
dwhenry: the top 5 lines of each methods are almost identical, might be nice to pull these out to a setup method or similar

dwhenry (approved):
Looks good

dwhenry (approved):

PR #854, opened 2017-08-08
Only clean the index is the RUMMAGER_VERSION matches

dwhenry:
We only want to clean the indices for the versions that 
matches from the overnight copy task

If we didn't check this we could potentially attempt to 
delete one of the new indices as we are importing data 
into it.

https://trello.com/c/3vligOJD/219-resolve-the-write-index-lock-for-mainstream-copied-to-integration


brenetic (approved):
🚀

PR #855, opened 2017-08-08
Add architecture record on versioning in the govuk index

suzannehamilton:
Explain why we are adding external versioning to the new Elasticsearch indexing and which edge cases it will handle.

### DIFF
--- a/doc/arch/adr-002-handle-race-conditions-in-govuk-index.md
+++ b/doc/arch/adr-002-handle-race-conditions-in-govuk-index.md
@@ -1,0 +1,105 @@
+# Decision record: use versioning to handle race conditions when updating the govuk index
+
+## Context
+
+We are in the process of creating a new 'govuk' index to replace the existing
+mainstream, government and detailed indexes. The new index will be populated by
+rummager using publishing and unpublishing event messages pulled from a queue
+controlled by the publishing API.
+
+Rummager cannot rely on messages from the queue being processed in the order
+that they were generated. This means that either rummager should handle edge
+cases such as these or we will need to accept the consequences:
+
+- A document is published twice in quick succession. The second publishing event
+  arrives in rummager first. The later, stale message should not overwrite the
+  more recent data in the index.
+- A document is unpublished and then immediately republished. The republishing
+  message is processed first. The later, stale unpublishing message should be
+  ignored and the published document should remain in the search index.
+- A document is published and then immediately unpublished. The unpublishing
+  event is processed first. The later, stale publishing message should not
+  re-add the document to the search index. It should remain unpublished.
+
+## Decision
+
+### Use Elasticsearch's external versioning
+
+Elasticsearch supports [external versioning][es-versioning] for documents. This
+means that each document includes a version number from an external system,
+which in our case is the publishing event number from the publishing API.
+
+When the version number is set, Elasticsearch uses optimistic locking to ensure
+that a document is never updated or deleted if the new version number is less
+than or equal to the existing version number.
+
+This will take care of some of the edge cases:
+
+- If two publishing events arrive out of order, the second stale event will be
+  ignored because the version number is lower. Result: the document in the index
+  ends up in the expected state, which is the most recent version.
+- If a republishing message arrives before an earlier unpublishing message, the
+  unpublishing will be ignored because its version number is lower. Result: the
+  document remains in the index as expected.
+
+[es-versioning]: https://www.elastic.co/blog/elasticsearch-versioning-support
+
+### Delete unpublished documents
+
+The simplest way to handle an unpublishing event is to delete the document from
+the index. Unfortunately, this could lead to a race condition which will leave
+the index in a bad state:
+
+- If a document is published and then immediately unpublished but the
+  unpublished message arrives first, it will initially be deleted from the
+  index. But when the stale republishing message arrives, it will incorrectly
+  recreate the document. Result: the document remains in the search index
+  despite being unpublished.
+
+We considered handling this by storing unpublished documents as tombstone
+records. These would be temporary: they would be cleaned up by reindexing tasks,
+but would be present for long enough to prevent this race condition:
+
+- The unpublishing message arrives first and saves the document with an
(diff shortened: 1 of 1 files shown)


doc/arch/adr-002-delete-unpublished-documents-from-index.md
```diff
@@ -0,0 +1,90 @@
+# Decision record: delete documents from the govuk search index when an unpublishing message is received
```
dwhenry: I feel that something like "Handling unpublishing edge case" is more in line with what we are discussing and would allow us to update this when we handle the current delete case in a nicer fashion.

doc/arch/adr-002-delete-unpublished-documents-from-index.md
```diff
@@ -0,0 +1,90 @@
...
+- If two publishing events arrive out of order, the second stale event will be
+  ignored because the version number is lower. Result: the document in the index
+  ends up in the expected state, which is the most recent version.
+- If a republishing message arrives before an earlier unpublishing message, the
+  unpublishing will be ignored because its version number is lower. Result: the
+  document remains in the index as expected.
+
+[es-versioning]: https://www.elastic.co/blog/elasticsearch-versioning-support
+
+### Delete unpublished documents
```
dwhenry: These seem like options for the last edge case rather than `decisions` which they are nested under.

doc/arch/adr-002-delete-unpublished-documents-from-index.md
```diff
@@ -0,0 +1,90 @@
...
+- The unpublishing message arrives first and saves the document with an
+  'unpublished' flag. The stale republishing message then arrives, but the
+  message is ignored because its version is lower than the unpublishing message.
+  Result: the tombstone record remains in the search index as expected.
+
+We have decided to postpone the implementation of tombstone records until we
+have time to work on it. In the meantime, this race condition may occur, but the
+sequence of events leading to it are expected to be rare.
+
+### Rely on reindexing to fix inconsistencies
```
dwhenry: this feels like a statement rather than a decision and should potentially be included in the delete option as a note on it.

dwhenry (commented):
Look pretty good and quite clear. Just a feel thoughts on the ordering/structure of the document.

doc/arch/adr-002-delete-unpublished-documents-from-index.md
```diff
@@ -0,0 +1,90 @@
...
+
+[es-versioning]: https://www.elastic.co/blog/elasticsearch-versioning-support
+
+### Delete unpublished documents
+
+The simplest way to handle an unpublishing event is to delete the document from
+the index. Unfortunately, this could lead to a race condition which will leave
+the index in a bad state:
+
+- If a document is published and then immediately unpublished by the unpublished
```
brenetic: ~~by~~ => `but`

brenetic (approved):
One little typo, otherwise LGTM MWA 🚀

rboulton:
This looks good, and is in line with what rough discussions about this I was involved in a couple of years ago.

One point about the case of items being potentially left in the index after being unpublished due to the race condition; I think it would be worth including a note on considering the potential impact of leaving a document in the search index in this situation.

In particular, can you include consideration of what a support team should do if this happens to a high profile sensitive document, that the search index is still returning despite it having been removed? I experienced situations like this (though with different causes) in my time on the team. I believe currently there is a way to manually delete a document from the index using the search-admin application - is this mechanism or something going to continue to exist?  Or is re-indexing expected to be quick enough that that could be kicked off to resolve the situation within a few minutes if needed?

rboulton:
Just to say, also, it's really good to see clear documentation of decisions like this being made for the future.

doc/arch/adr-002-delete-unpublished-documents-from-index.md
suzannehamilton: Yeah, when I wrote the title I was thinking of the ADR as just covering the decision to delete unpublished docs rather than keeping them in the index, but it actually covers all of the versioning decisions. Is the new name clearer?

doc/arch/adr-002-delete-unpublished-documents-from-index.md
suzannehamilton: I've removed the heading from the tombstone records section to make it clearer that it's an alternative to deleting unpublished documents.

All the subheaders under Decisions are now actual decisions that we've chosen at this point. Does that help?

doc/arch/adr-002-delete-unpublished-documents-from-index.md
suzannehamilton: Fixed. 👍

suzannehamilton:
@rboulton I've expanded the Consequences section explaining what a user would see and how the support team could fix it. They can continue using Search Admin to remove links, but we think we should also add a publishing API rake task to resend the unpublishing message to rummager.

We don't yet know how long reindexing would take, but I'm guessing it could be an hour or so (once every page has been migrated to the new index). That's how long it takes to run the current reindexing task which updates the popularity data.

doc/arch/adr-002-handle-race-conditions-in-govuk-index.md
```diff
@@ -0,0 +1,105 @@
+# Decision record: use versioning to handle race conditions when updating the govuk index
```
dwhenry, line 1: @suzannehamilton I think this is a better title. 

@MatMoore wanted to clarify your understanding of the best way to write the title for ADR's.

I like what @suzannehamilton has here but would have leaned towards "Handling Race conditions when updating govuk index" as that describes the problem we are making a decision about while adding "use versioning" describes the solution as well. 

Which way should we lean? Are both valid depending on personal preference?

MatMoore: Both seem valid to me.

I'd lean towards including the solution in the title because if you're interested in understanding why something is the way it is, you should be able to scan the list of ADRs and see if there's anything relevant, but you may not know the original problem beforehand.

MatMoore (approved):
This is a really good write up 👍

PR #856, opened 2017-08-08
Allow sitemap format boosts to override internal search boosts

suzannehamilton:
Before, the weightings applied to documents in the sitemap were the same as those used to rank documents in internal search. The assumption was that search results that rank highly in internal search should also rank highly in external search because users are searching for the same content.

But this isn't always the case: navigation pages are inappropriate results in internal search because the user is already on the site, but may help to orient a user coming from external search. So a boost of 0 for mainstream browse pages should be increased in the external search ranking.

https://trello.com/c/8JQ38H0K/217-override-boost-for-navigation-pages

### DIFF
--- a/lib/sitemap/property_boost_calculator.rb
+++ b/lib/sitemap/property_boost_calculator.rb
@@ -1,34 +1,34 @@
 class PropertyBoostCalculator
   def initialize
     @boost_config = YAML.load_file('config/query/boosting.yml')
-    @max_boosts = calculate_max_boosts
   end
 
   def boost(document)
-    boost_values = @boost_config.map do |property, boosts|
+    raw_boosts = @boost_config.map do |property, boosts|
       if document.has_field?(property) && boosts[document.get(property)]
-        page_boost = boosts[document.get(property)]
+        boosts[document.get(property)]
       else
-        page_boost = 1
+        1
       end
-
-      # Normalise format boost to always give a value between 0 and 1
-      page_boost.to_f / @max_boosts[property]
     end
 
-    boost_values.inject(:*).round(2)
+    overall_boost = raw_boosts.inject(:*)
+
+    map_boost_to_priority(overall_boost).round(2)
   end
 
 private
 
-  def calculate_max_boosts
-    @boost_config.keys.each_with_object({}) do |property, max_boosts|
-      boosts = @boost_config[property]
-
-      max_configured_boost = boosts.values.max
-      max_boost = max_configured_boost >= 1 ? max_configured_boost : 1
-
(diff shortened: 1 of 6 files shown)


MatMoore:
We should also consider prioritising topic pages (specialist_sector), policy pages, and policy area pages, since these could be good starting points for a variety of search terms and are kept up to date.

Notably missing are the new `taxon` topic pages. This is because we never actually indexed them in rummager. I think now would be a good time to do this.

suzannehamilton:
@MatMoore I've added `topic` and `policy` to the boost overrides. As far as I can tell, the `topic` format also includes policy areas.

I've added a [card about taxon pages to Trello](https://trello.com/c/frcfJuOq/195-add-taxon-pages-to-the-sitemap-to-prioritise-or-deprioritise-them). They are not currently in the search index so we're not planning to add them to the sitemap as part of this task.

I've also change the algorithm in the priority calculator, as we discussed in person. It uses an exponential function to calculate a priority between 0 and 1.

test/unit/sitemap/property_boost_calculator_test.rb
```diff
@@ -2,129 +2,117 @@
 require "sitemap/sitemap"
 
 class PropertyBoostCalculatorTest < Minitest::Test
-  def test_documents_are_boosted_relative_to_default_for_property
+  def test_boosts_are_between_0_and_1
```
suzannehamilton, line 5: I would love to use property-based tests here because the values in these unit tests seem so over-specific. But as far as I know there's no good PBT library for Ruby.

MatMoore: Yeah I don't know of any either 😞

lib/sitemap/property_boost_calculator.rb
```diff
@@ -1,34 +1,36 @@
...
-    end
+  # Convert a boost (which may be any positive or zero value) to a sitemap
+  # priority.
+  #
+  # The conversion must produce a number between zero and one to match the
+  # sitemap format (https://www.sitemaps.org/protocol.html), and boost ordering
+  # must be preserved (i.e. if one page has a higher combined boost than
+  # another one, its priority must also be higher).
+  #
+  # https://www.wolframalpha.com/input/?i=plot+1+-+2%5E(-x),+x+%3D+0+to+5
```
MatMoore, line 32: 📈

MatMoore (approved):
Looks good to me.

For future reference, `topic` in rummager (and various other places) refers to a policy area. Yes, it makes no sense.

Other teams are currently evaluating what to do with taxon pages so it's fine to leave this alone for now.

PR #858, opened 2017-08-09
Fix statsd elasticsearch logging

brenetic:
This was previously not logging sidekiq events.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -26,6 +26,7 @@ require "indexer/workers/delete_worker"
 require "indexer/workers/amend_worker"
 require "indexer/comparer"
 require "govuk_index/publishing_event_processor"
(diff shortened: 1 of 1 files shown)


app.rb
```diff
@@ -26,6 +26,7 @@
 require "indexer/workers/amend_worker"
 require "indexer/comparer"
 require "govuk_index/publishing_event_processor"
+require "govuk_index/publishing_event_worker"
```
dwhenry, line 29: ??? https://github.com/alphagov/rummager/blob/master/lib/govuk_index/publishing_event_processor.rb#L1

dwhenry (commented):
That file is already included so don't understand the change?

PR #862, opened 2017-08-14
Fix rummager clean task

dwhenry:
Move the index version definition to a location that is loaded inside the rake
task.


brenetic (approved):
🚀

PR #863, opened 2017-08-14
Fix requires for health check

dwhenry:
This could also be fixed by change the order of requires to:

```
require "rummager"
require 'health_check/logging_config'
```

However this would result in the following warning:

```
warning: already initialized constant Logging::MAX_LEVEL_LENGTH
```

due to how logging is initialized.


search_check_results.csv
```diff
@@ -0,0 +1 @@
+Pass or Fail,Found or not,Page,Search Term,Position found,Position wanted
```
MatMoore: Is this supposed to be checked in?

MatMoore (approved):

PR #864, opened 2017-08-14
Parse publishing API json payload correctly

brenetic:
We were not parsing the publishing API payload before putting it onto the sidekiq queue. This resulted in the `payload_version` being read as a string instead of its actual value.

Mobbed with @Rosa-Fox @MatMoore

### DIFF
--- a/lib/govuk_index/publishing_event_processor.rb
+++ b/lib/govuk_index/publishing_event_processor.rb
@@ -2,7 +2,7 @@ module GovukIndex
   class PublishingEventProcessor
     def process(message)
       Services.statsd_client.increment('govuk_index.rabbit-mq-consumed')
-      PublishingEventWorker.perform_async(message.delivery_info[:routing_key], message.payload)
+      PublishingEventWorker.perform_async(message.delivery_info[:routing_key], JSON.parse(message.payload))
       message.ack
     end
   end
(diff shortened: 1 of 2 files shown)


lib/govuk_index/publishing_event_processor.rb
```diff
@@ -2,7 +2,7 @@ module GovukIndex
   class PublishingEventProcessor
     def process(message)
       Services.statsd_client.increment('govuk_index.rabbit-mq-consumed')
-      PublishingEventWorker.perform_async(message.delivery_info[:routing_key], message.payload)
+      PublishingEventWorker.perform_async(message.delivery_info[:routing_key], JSON.parse(message.payload))
```
dwhenry, line 5: This should NOT be required.

If the message is correctly formatted (i.e. has the JSON_FORMAT content_type) then it has already been parsed from here: https://github.com/alphagov/govuk_message_queue_consumer/blob/master/lib/govuk_message_queue_consumer/json_processor.rb

We should be using the existing functionality if at all possible, as this change could potentially result in trying to parse a message multiple times.

dwhenry (changes requested):
Can we find out why the messages do not have the expected header?

brenetic:
These are not the droids we are looking for. Closing to investigate further.

PR #865, opened 2017-08-14
Map content store document_types to elasticsearch document_types

brenetic:
- Previously sent content_store document_types. We need to send the
correct elasticsearch types.
- Currently we only care about the `edition` type as we are adding the
`mainstream` formats. This will expand to encompass all elasticsearch
document_types as we proceed to add each format.

Mobbed with @MatMoore & @Rosa-Fox

### DIFF
--- a/lib/govuk_index/document_type_inferer.rb
+++ b/lib/govuk_index/document_type_inferer.rb
@@ -11,7 +11,8 @@ module GovukIndex
         raise NotFoundError if existing_document.nil?
         existing_document['_type']
       else
-        payload['document_type']
+        raise UnknownDocumentTypeError if elasticsearch_document_type.nil?
+        elasticsearch_document_type
       end
     end
 
@@ -23,8 +24,18 @@ module GovukIndex
 
     attr_reader :payload
 
+    def mapped_document_types
+      @_document_types ||= begin
+        YAML.load_file(File.join(__dir__, '../../config/govuk_index/mapped_document_types.yaml'))
+      end
+    end
+
     def existing_document
       @_existing_document ||= Client.get(type: '_all', id: payload['base_path'])
+    end
+
+    def elasticsearch_document_type
+      @_elasticsearch_document_type ||= mapped_document_types[payload['document_type']]
     end
   end
 end
--- a/lib/govuk_index/publishing_event_worker.rb
+++ b/lib/govuk_index/publishing_event_worker.rb
@@ -25,8 +25,10 @@ module GovukIndex
     # processed out of order so we don't want to notify errbit but just allow
     # the process to continue
     rescue NotFoundError
+      logger.info("#{payload['base_path']} could not be found.")
       Services.statsd_client.increment('govuk_index.not-found-error')
     rescue UnknownDocumentTypeError
+      logger.info("#{payload['document_type']} document type is not known.")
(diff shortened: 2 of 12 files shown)


dwhenry:
I not total sure I understand what this code does or why it is required now given that we are using a placeholder. Is this a change that is needed to get versioning working?

MatMoore:
@dwhenry this is because there are two document type fields. We were passing along the publishing api one but that's not what we should be using for the elasticsearch document type (if we want to keep things backwards compatible)

It's a placeholder because a lot of formats - including all the mainstream ones - map to "edition" in elasticsearch, so we can deal with the mapping later on when we tackle other formats.

lib/govuk_index/document_type_inferer.rb
```diff
@@ -21,10 +23,20 @@ def unpublishing_type?
...
+        YAML.load_file(File.join(__dir__, '../../config/govuk_index/mapped_document_types.yaml'))
+      end
+    end
 
     def existing_document
       @_existing_document ||= Client.get(type: '_all', id: payload['base_path'])
     end
+
+    def elasticsearch_document_type
+      @_elasticsearch_document_type ||= mapped_document_types[payload_document_type]
```
dwhenry: `payload_document_type` is only used the once so may be worthwhile just inlining `payload['document_type']`

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -25,6 +26,8 @@ def perform(routing_key, payload)
     # the process to continue
     rescue NotFoundError
       Services.statsd_client.increment('govuk_index.not-found-error')
+    rescue UnknownDocumentTypeError
+      Services.statsd_client.increment('govuk_index.unknown-document-type')
```
dwhenry: I'm not sure on this, but it might be helpful when we are deciding what to do next:
```
rescue UnknownDocumentTypeError
  Services.statsd_client.increment("govuk_index.unknown-document-type.#{e.message}")
```

where the message is the document_type from the payload. 〰️

test/unit/govuk_index/elasticsearch_processor_test.rb
```diff
@@ -20,4 +20,31 @@ def test_should_save_valid_document
...
+        {
+          delete: presenter.identifier
+        }
+      ]
+    )
+
+    actions = GovukIndex::ElasticsearchProcessor.new
+    actions.delete(presenter)
+    actions.commit
+  end
```
dwhenry, line 49: do we want a test for should not delete a document with an invalid document type or do we already have.

dwhenry (approved):

test/unit/govuk_index/elasticsearch_processor_test.rb
brenetic: We currently raise/rescue the `UnknownDocumentTypeError` from the `DocumentTypeInferer` so it shouldn't ever try to delete a document if the payload document type is invalid. That is currently tested

lib/govuk_index/publishing_event_worker.rb
brenetic: We have added logging with payload details instead

dwhenry (approved):
🚢

PR #868, opened 2017-08-15
Logging is a good thing

MatMoore:
The logging we have is actually useful but it's suppressed most of the time.

No data that gets sent to rummager is sensitive, so there is no need for logs to be much less detailed on production.

This PR changes the following:
- Log at INFO level in integration/staging/production instead of WARN
- Log at DEBUG level in development instead of INFO (we can change this if it's annoying)
- Remove unwanted output from tests.
- Add logging for the new index so we can see what the event processor is doing.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -36,6 +36,9 @@ namespace "test" do
 
   desc 'Clean all test indexes'
   task :clean_test_indexes do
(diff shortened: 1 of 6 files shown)


brenetic (approved):
🚀

PR #869, opened 2017-08-15
Move to rspec

dwhenry:
Convert our tests to run using RSpec.

The migration process is broken down into steps:

* Update the setup/support code to run with RSpec
* Migrate the tests to RSpec syntax - this is done via a script: https://gist.github.com/dwhenry/9a58a537b0aad40c061bd23ce20949c6
* fixes and tidy up after the migration process

This is a straight migration to run using rspec and still relies on existing tests expectations and matchers.

Future PR will try and:

* enable `disable_monkey_patching!`
* clean up integration tests implementation to simplify the implementation
* remove shoulda matches
* remove mocha matches


spec/integration/analytics_data_spec.rb
```diff
@@ -1,12 +1,11 @@
-require 'integration_test_helper'
+require 'spec_helper'
 
-class AnalyticsDataTest < IntegrationTest
-  def setup
-    super
+RSpec.describe 'AnalyticsDataTest', tags: ['integration'] do
```
MatMoore, line 3: It would be good to change these to constants if we can (`RSpec.describe AnalyticsData` instead of AnalyticsDataTest)

spec/integration/analytics_data_spec.rb
```diff
@@ -41,7 +40,7 @@ def test_fetches_rows_of_analytics_dimensions
     assert_equal [expected_row], rows.to_a
   end
 
-  def test_missing_data_is_nil
+  it "missing_data_is_nil" do
```
MatMoore, line 43: It seems a shame to leave the test names with broken english....

Maybe we could dump all the names into spreadsheet, challenge people to fill in better names, then do a big search and replace?

MatMoore (approved):

PR #871, opened 2017-08-16
Strip html from indexable content

brenetic:
We tried several approaches when implementing this code including:

* Elasticsearch `html_strip` filter - according to the documentation this would copy the full content to merged fields rather than the stripped content.

* Storing the value in a separate field in ES - This would cause inconsistencies in the scoring due to data being split across multiple fields resulting in documents with field not being calculated correctly in some edge cases.

* Use of the `Rails::Html::Sanitizer` gem - This did not correctly insert spaces when removing tags, nor did it separate block tags onto separate lines.

Indexable content passed to elasticsearch should not have html tags in it otherwise it can potentially effect search indexing.

The publishing api sends a few different types of payloads for notifications it puts onto the queue therefore it is necessary to deal with strings as well as `text/html` but to ignore `text/govspeak`.

https://trello.com/c/SoyBhfU8/226-change-indexablecontent-analyser-to-work-with-html

Mobbed with @Rosa-Fox and @dwhenry


MatMoore:
From talking to David offline I assume merged fields ==  `_all`?

It's a bit of a shame this blocks us from using `html_strip` but looks like this stripping the tags in ruby will do the job.

Are there any edge cases where the publishing api sends a payload without `text/html` but with some other content type?

MatMoore (approved):

dwhenry:
@MatMoore our assumption is that a single "text/html" element will always be present. As it would be difficult and time-consuming to try and reason any situations where it doesn't, we are using airbrake to notify us if this does occur - we will monitor this closely for the first weeks after deployment.

This felt like a good compromise in order to get this feature out the door ASAP.

lib/govuk_index/indexable_content_sanitiser.rb
```diff
@@ -0,0 +1,50 @@
+require 'loofah'
```
suzannehamilton: Should this be in rummager.rb to fit with the new `require` system?

test/unit/govuk_index/indexable_content_sanitiser_test.rb
```diff
@@ -0,0 +1,97 @@
...
+    presenter = elasticsearch_presenter(payload)
+
+    assert_equal "\nhello\n marmaduke", presenter.document[:indexable_content]
+  end
+
+  def test_strips_html_tags_from_single_string_item
+    payload = {
+      "base_path" => "/some/path",
+      "document_type" => "aaib_report",
+      "details" => { "body" =>  "hello marmaduke"}
```
suzannehamilton: Should this have some HTML tags in it? Or is this a test of a body without HTML? In which case the test name is a bit confusing.

suzannehamilton: Ditto for the test on line 38.

suzannehamilton:
Just a couple of minor comments, but otherwise this looks fine to me. 👍

suzannehamilton (approved):

lib/govuk_index/indexable_content_sanitiser.rb
```diff
@@ -0,0 +1,48 @@
...
+      return nil if cleaned_content.empty?
+      cleaned_content.join("\n")
+    end
+
+  private
+
+    attr_reader :payload
+
+    def details
+      @_details ||= payload.dig('details')
```
dwhenry: can we move this into the initializer as we don't actually need the payload in this class (we could actually just pass in the details object).

Also we don't  need to use dig here as we are only getting one level of data.

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -47,7 +48,11 @@ def process_action(actions, routing_key, payload)
       Services.statsd_client.increment('govuk_index.sidekiq-consumed')
 
       document_type_inferer = DocumentTypeInferer.new(payload)
-      presenter = ElasticsearchPresenter.new(payload, document_type_inferer.type)
+      presenter = ElasticsearchPresenter.new(
+        payload: payload,
+        type: document_type_inferer.type,
+        sanitiser: IndexableContentSanitiser.new(payload)
```
dwhenry: as discussed can we pass the payload into the clean method instead of the initializer

dwhenry (approved):
Looks good

PR #873, opened 2017-08-17
Update popularity nightly

binaryberry:
Update popularity data inplace using Sidekiq Workers. This means we can avoid the manual lock and unlocking of indices and the need to switch indices each night. We are able to achieve this using versioning.

This means we can leverage Sidekiqs concurrency model to provide the worker threads.

paired with @dwhenry 

https://trello.com/c/WzmXihZa/225-update-the-new-index-with-popularity-data-during-the-nightly-rebuild


suzannehamilton:
Just for context, what's the reason for avoiding the bulk_loader?

lib/govuk_index/popularity_updater.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+  end
+
+  class PopularityWorker < Indexer::BaseWorker
+    def perform(records, destination_index)
+      actions = ElasticsearchProcessor.new(client: GovukIndex::Client.new(timeout: 30, index_name: destination_index))
+
+      popularities = retrieve_popularities_for(records)
+      records.each do |record|
+        actions.save(
```
MatMoore: Instead of `save`, we could add another method to the processor to support the `update` type, which would mean we don't have to send all the data for every document back to elasticsearch.

```
{ "update" : {"_id" : "1", "_type" : "type1", "_index" : "index1"} }
{ "doc" : {"field2" : "value2"} }
```
https://www.elastic.co/guide/en/elasticsearch/reference/1.7/docs-bulk.html

The semantics are documented a bit more here
https://www.elastic.co/guide/en/elasticsearch/reference/1.7/docs-update.html#_updates_with_a_partial_document

> Note, this operation still means full reindex of the document, it just removes some network roundtrips and reduces chances of version conflicts between the get and the index. The _source field need to be enabled for this feature to work.

We'd still have to query ES for the links but we'd be passing much less data around so I'd expect this to help with performance.

dwhenry: @MatMoore  It inserts into a different index, so unless we want to change the process here and have it update the existing index instead of creating a new one....

dwhenry:
@suzannehamilton The main reason for avoiding the Bulk loader code is that its complexity. It's a multiple purpose class that also implements threaded workers. As we can get this essentially for free by using sidekiq instead, and that makes the code easier to understand, this feels like an improvement over the current process. It also means we will hopefully be able to retire some of the functionality from the bulk loader.

lib/govuk_index/popularity_updater.rb
MatMoore: Ah I didn't realise that. I'm happy to leave it as it is in that case.

lib/govuk_index/popularity_updater.rb
```diff
@@ -0,0 +1,93 @@
...
+      end
+    end
+
+  private
+
+    def wait_for_sidekiq
+      wait_start = Time.now
+      while Sidekiq::Queue.new.size > MAX_SIDEKIQ_QUEUE_SIZE
+        # alert us that we are waiting.. StatsD?
+        print '.'
```
MatMoore: can you replace this with logging?

lib/govuk_index/popularity_updater.rb
```diff
@@ -0,0 +1,93 @@
...
+    def run
+      scroll_enumerator.each_slice(PROCESSOR_BATCH_SIZE) do |documents|
+        wait_for_sidekiq
+        PopularityWorker.perform_async(documents, @destination_index)
+      end
+    end
+
+  private
+
+    def wait_for_sidekiq
```
MatMoore: How does sidekiq behave when we publish 1000 of these messages without this throttling?

Could you leave a comment explaining the rationale for it?

lib/search_server.rb
```diff
@@ -41,6 +41,14 @@ def content_indices
       end
     end
 
+    def index_from_real_name(real_name)
```
MatMoore: Is this used?

MatMoore (approved):
I think this generally makes sense, but I'm not sure I understand the reasoning behind all the limits in `PopularityUpdater`. Can you make sure this is commented before merging?

lib/govuk_index/popularity_updater.rb
```diff
@@ -0,0 +1,93 @@
...
+        )
+      end
+
+      actions.commit
+    end
+
+    def process_record(record, popularities)
+      base_path = record['identifier']['_id']
+      OpenStruct.new(
+        identifier: record['identifier'],
```
suzannehamilton: Does the identifier include the version by default? You need to add `"version": true` to a query when making a GET request to Elasticsearch, so I'm wondering if we need to do something extra here in order to copy the version to the new record.

lib/govuk_index/popularity_updater.rb
dwhenry: We tried with mainstream locally and it didn't seem to cause a heap of issues (1300 message in the queue).

My thinking is more around broken state and not flooding the queue in that event. However, I would also be putting this data into a separate queue (not done in the spike) potentially with its own workers which might be enough isolation for this.

lib/govuk_index/popularity_updater.rb
dwhenry: good spot, I totally forgot about versioning.

lib/search_server.rb
dwhenry: no.. :(

test/integration/govuk_index/updating_popularity_data_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    setup_page_traffic_data(document_count: document_count)
+
+    popularity = 1.0 / (document_count + SearchConfig.instance.popularity_rank_offset)
+
+    GovukIndex::PopularityUpdater.update('govuk_test')
+
+    assert_document_is_in_rummager({ 'link' => '/test', 'popularity' => popularity }, type: 'edition', index: 'govuk_test')
+  end
+
+  def test_correct_copies_version_information
```
suzannehamilton: The name of this test is a bit clunky. Maybe just drop the "correct"?

Also, do you think it's worth testing the race condition where the version has been increased by a concurrent publishing update?

lib/govuk_index/popularity_worker.rb
```diff
@@ -0,0 +1,34 @@
...
+    def process_record(record, popularities)
+      base_path = record['identifier']['_id']
+      OpenStruct.new(
+        identifier: record['identifier'].merge('_version_type' => 'external_gte'),
+        document: record['document'].merge('popularity' => popularities[base_path]),
+      )
+    end
+
+    def retrieve_popularities_for(records)
+      # The first parameter is only used to check that it is not running for the `page-traffic` index.
```
suzannehamilton: Might be less confusing to pass in the actual index name? It's a bit confusing even with the comment, and I could imagine this causing a bug in the future because the PopularityCode assumes that the index name is correct.

lib/govuk_index/popularity_updater.rb
```diff
@@ -0,0 +1,62 @@
...
+    def self.update(index_name)
+      new(
+        source_index: index_name,
+        destination_index: index_name,
+      ).run
+    end
+
+    # This task is designed to migrate data when the schema changes, It waits for the
+    # queued jobs to be processed, this is done by monitoring the sidekiq queue and workers.
+    # This process should be closely monitored as it is not gauranteed to work.
```
suzannehamilton: Typo: guaranteed.

How could it fail, and should this warning be on the rake task as well?

lib/govuk_index/popularity_updater.rb
```diff
@@ -0,0 +1,62 @@
...
+    def self.migrate(index_name)
+      raise('Please set `CONFIRM_INDEX_MIGRATION_START` to continue') unless ENV['CONFIRM_INDEX_MIGRATION_START']
```
suzannehamilton: What's the purpose of this check, and should it be documented in the rake task too?

lib/govuk_index/popularity_updater.rb
```diff
@@ -0,0 +1,62 @@
...
+      index_group = SearchConfig.instance.search_server.index_group(index_name)
+      new_index = index_group.create_index
+      index_group.current.with_lock do
+        new(
+          source_index: index_group.current.real_name,
+          destination_index: new_index.real_name,
+        ).run
+        # wait for queued tasks to be started
+        sleep 1 while(Sidekiq::Queue.new(GovukIndex::PopularityWorker::QUEUE_NAME).size > 0) # rubocop: disable Style/ZeroLengthPredicate
+        # wait for started tasks to be finished
```
suzannehamilton, line 27: If we've locked the index (line 22), will this ever complete?

suzannehamilton (commented):
The approach looks sensible. 👍 Just a few questions and suggestions.

lib/govuk_index/popularity_updater.rb
dwhenry: Yes as it is inserting the data into a newly created, unlocked index.

lib/govuk_index/popularity_updater.rb
dwhenry: moved to rake file

lib/govuk_index/popularity_updater.rb
dwhenry: fixed the typo. I'm not sure how it could fail, however it feels unstable to me. I think having the warning here is enough, as running this method should be the exception and not the rule.

lib/govuk_index/popularity_worker.rb
dwhenry: done

test/integration/govuk_index/updating_popularity_data_test.rb
dwhenry: changed and added

lib/govuk_index/popularity_updater.rb
suzannehamilton: I'm probably misreading the code, but it looks like it would do this:

1. Create a new index, but don't switch to it
2. Lock the current, pre-existing index
3. Wait for queued tasks to insert data into the current index (which is locked, so I'd expect insertions to fail)

Have I misunderstood step one of these steps? What causes the data to be inserted into the new index when the "current" index is still the old one?

suzannehamilton (approved):

PR #878, opened 2017-08-18
Ignore very old publishing dates in sitemap

suzannehamilton:
Some documents have been backdated with very old `public_timestamp` dates, e.g.
https://www.gov.uk/government/publications/ukho-1914-archives-catalogue

The Google Search Console (understandably) raises an error when these dates are used as the `lastmod` date in the sitemap.

Ideally we would use a date which corresponds to the last time the document was updated on GOV.UK, but we do not have that information in the search index at the moment.

As a workaround, limit the `lastmod` date to recent dates, and use the launch of GOV.UK as a cut-off point because documents could not have been on the site before this date. This affects around 76,000 documents.

https://trello.com/c/O2FcYxgQ/241-fix-old-dates-in-sitemap

### DIFF
--- a/lib/sitemap/sitemap_presenter.rb
+++ b/lib/sitemap/sitemap_presenter.rb
@@ -1,4 +1,6 @@
 class SitemapPresenter
+  GOVUK_LAUNCH_DATE = DateTime.new(2012, 10, 17).freeze
+
   def initialize(document, property_boost_calculator)
     @document = document
     @property_boost_calculator = property_boost_calculator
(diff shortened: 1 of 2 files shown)


MatMoore (approved):

lib/sitemap/sitemap_presenter.rb
```diff
@@ -17,7 +19,11 @@ def last_updated
     return nil unless document.public_timestamp
 
     # Attempt to parse timestamp to validate it
-    DateTime.iso8601(document.public_timestamp)
+    parsed_date = DateTime.iso8601(document.public_timestamp)
+
+    # Ignore timestamps for documents backdated to before GOV.UK was launched
+    return nil if parsed_date < GOVUK_LAUNCH_DATE
```
dwhenry: would there be any difference if we used the launch date?

suzannehamilton: Yeah, Mat suggested this as well because it would stop search engines from making assumptions about the recency of the pages.

I've made the change. 👍

dwhenry (approved):
As discuss looks good, just point out that `GOVUK_LAUNCH_DATE` is a `DateTime` but don't think that matters too much.

PR #879, opened 2017-08-18
ADR explaining our changes to the nightly reindex process

dwhenry:
paired with @binaryberry 

https://trello.com/c/WzmXihZa/225-update-the-new-index-with-popularity-data-during-the-nightly-rebuild

### DIFF
--- a/doc/arch/adr-003-popularity-updating-without-index-locks.md
+++ b/doc/arch/adr-003-popularity-updating-without-index-locks.md
@@ -26,14 +26,13 @@ The code is quite complicated as it has to manage a threaded workload, which
 is difficult to test and reason about.
 
 As part of building the new `govuk` index, we want to simplify this process and 
-ideally avoid the index locking.
+ideally avoid the index locking. This means we can use sidekiq's concurrency
+implementation instead of having our own, as we no longer need to be able to 
+unlock the index at the end of the process.
 
-## Decision
+## Options
 
-### Leverage external versioning to avoid locking
-
-We have previously documented the decision to implement external versioning. As 
-a result of this functionality, we can rethink how we do our overnight processing.
+#### Using external versioning
 
 If we don't use index locking and switching and instead just overwrite the existing
 document with a copy of itself with the updated popularity figure, the following 
@@ -55,3 +54,31 @@ today, let's call this `pop-today`.
 > the search index
 > * `job-A` occurs before `update-A` - in this case both updates will occur, 
 > leaving the values from `update-A` in the search index
+
+This process would fail with version_type set to `external` but would succeed with
+version_type set to `external_gte`. The elasticsearch documentation does state that
+when using `external_gte`:
+ 
+> "If used incorrectly, it can result in loss of data."
+
+#### External versioning with multiplier
+
+By multiplying the external version fields by a multiplier when it is inserted into
+elasticsearch, it is then possible to increment the version field each time you
+do a transient data update. This can then be used as normal with the version_type
+of `external`, and would skip the transient data update if the content had been 
+updated. 
+
+This would mean that with a multiplier of 10,000 we could do a daily update for approx 
+27 years and with a multiplier of 100,000 we could do a daily update for approx 273 years.
+
+## Decision
+
+The two options have similar implementation with only minor differences, as a result we
+will be going with the easier of the two (using `external_gte`), as we will only be
+updating a single field (popularity) the risk of data loss is quite low.
+
+An additional advantage of choosing this approach is that we can use it with indices
+that aren't currently using `external` versioning, as the inplace edit with the current
+version - taken at the start of the reindex process - will be ignored if the content 
+has changed and the version has been automatically incremented.
--- a/doc/arch/adr-003-popularity-updating-without-index-locks.md
+++ b/doc/arch/adr-003-popularity-updating-without-index-locks.md
@@ -1,0 +1,57 @@
+# Decision record: perform popularity updating without using an index lock
+
+## Context
+
+The current nightly job works by:
+ 
+1. locking the current index
+1. queuing all records from the index
+1. starting threaded works to insert the records into the new index with the updated popularity
+1. waiting for the works to finish
+1. switch to the new index
+1. release the index lock
+
+The fact that this process requires the index to be locked while the process 
+runs means that we prefer to run it out of hours so that it doesn't block
+the insertion of new content.
+
+The index locking is required so that document edits during the reindex process 
+are blocked from writing to the old index, as they would then be discarded. The
+lock blocks the write until the process has completed and switched to the new index.
+
+This in turn means we need to create a new index to write the updated popularity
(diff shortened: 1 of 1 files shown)


MatMoore (approved):
Looks good 👍 

Is it also worth mentioning that this also lets us use sidekiq for concurrency, since we don't have special tasks to execute at the beggining/end?

doc/arch/adr-003-popularity-updating-without-index-locks.md
```diff
@@ -0,0 +1,84 @@
...
+lock blocks the write until the process has completed and switched to the new index.
+
+This in turn means we need to create a new index to write the updated popularity
+data to, as the old index is currently locked. 
+
+The code is quite complicated as it has to manage a threaded workload, which
+is difficult to test and reason about.
+
+As part of building the new `govuk` index, we want to simplify this process and 
+ideally avoid the index locking. This means us can use sidekiq's concurrency
```
suzannehamilton: Typo: us -> we

doc/arch/adr-003-popularity-updating-without-index-locks.md
```diff
@@ -0,0 +1,84 @@
...
+>
+> If we get an update on `DocumentA` while the popularity update process is running,
+> let's call this `update-A`, `update-A` will be generated with a popularity figure
+> equal to `pop-today`, the same as what is in `job-A`.
+> 
+> This can result in one of two things occurring:
+>
+> * `update-A` occurs before `job-A` - in this case `job-A` will be ignored as it
+> is for an earlier version of `DocumentA`, leaving the values from `update-A` in
+> the search index
```
suzannehamilton, line 54: Are we OK with this?

I suspect it's fine because the worst case scenario is that the popularity data is one day out of date - the stale data will be replaced 24 hours later by the next popularity update job. If that's the case, then we should state it explicitly.

doc/arch/adr-003-popularity-updating-without-index-locks.md
```diff
@@ -0,0 +1,84 @@
...
+> * `job-A` occurs before `update-A` - in this case both updates will occur, 
+> leaving the values from `update-A` in the search index
+
+This process would fail with version_type set to `external` but would succeed with
+version_type set to `external_gte`. The elasticsearch documentation does state that
+when using `external_gte`:
+ 
+> "If used incorrectly, it can result in loss of data."
```
suzannehamilton, line 62: Is it worth explaining that we think this is just a suggestion to carefully consider the race conditions? (Which we've done.)

doc/arch/adr-003-popularity-updating-without-index-locks.md
```diff
@@ -0,0 +1,84 @@
...
+updated. 
+
+This would mean that with a multiplier of 10,000 we could do a daily update for approx 
+27 years and with a multiplier of 100,000 we could do a daily update for approx 273 years.
+
+## Decision
+
+The two options have similar implementation with only minor differences, as a result we
+will be going with the easier of the two (using `external_gte`), as we will only be
+updating a single field (popularity) the risk of data lose is quite low.
```
suzannehamilton: Typo: lose -> loss

doc/arch/adr-003-popularity-updating-without-index-locks.md
```diff
@@ -0,0 +1,84 @@
...
+
+An addition advantage of choosing this approach is that we can use it with indicies
```
suzannehamilton: Typos:

addition -> additional
indicies -> indices (or indexes - we're a bit inconsistent on this)

doc/arch/adr-003-popularity-updating-without-index-locks.md
```diff
@@ -0,0 +1,84 @@
...
+of `external`, and would skip the transient data update if the content had been 
+updated. 
+
+This would mean that with a multiplier of 10,000 we could do a daily update for approx 
+27 years and with a multiplier of 100,000 we could do a daily update for approx 273 years.
+
+## Decision
+
+The two options have similar implementation with only minor differences, as a result we
+will be going with the easier of the two (using `external_gte`), as we will only be
```
suzannehamilton, line 78: 👍 I agree that this will be simpler. People can refer to the Elasticsearch docs on `external_gte` and we won't have to explain and maintain the multiplication factor code.

suzannehamilton (approved):
Spotted a few typos, but this all sounds good. 👍

doc/arch/adr-003-popularity-updating-without-index-locks.md
dwhenry: That is not the case.. as pointed out above if `job-A` arrives before `Document-A` it will arrive with the correct popularity figure, as it hit the system after the migration started (i.e. after the page-traffic data has been populated). As result you get the same popularity eitehr way. This is really had to explain in words. :(

doc/arch/adr-003-popularity-updating-without-index-locks.md
dwhenry: I would prefer not to put a guess like that in here in case people then read it as more than an assumption. I did mention below that as we are only updating the popularity field we think that data loss should not be a cause for concern and feel that is a strong enough statement.

doc/arch/adr-003-popularity-updating-without-index-locks.md
dwhenry: we should try and use `indices`.. as that is what ES use.

doc/arch/adr-003-popularity-updating-without-index-locks.md
dwhenry: done

doc/arch/adr-003-popularity-updating-without-index-locks.md
dwhenry: done

doc/arch/adr-003-popularity-updating-without-index-locks.md
suzannehamilton: You're right, I just missed the fact that it would have the correct popularity despite the popularity update being ignored. Might be worth adding something like "with the most recent popularity value" to the end of both cases, so that it's clear that the data is always up to date regardless of which process happens first?

doc/arch/adr-003-popularity-updating-without-index-locks.md
suzannehamilton: My concern is that if we leave that sentence as it is, it sounds like we haven't thought about it at all! How about something like this, which makes it clear that we're making an assumption?

> It's not clear what the documentation means by this, but we're assuming that it means that you have to carefully consider the consequences of allowing a document to overwrite another with exactly the same version. We think that this is OK in our case, provided we only use `external_gte` when updating popularity. Regular publishing updates will use the stricter `external` type. We will revisit this assumption if it turns out to be wrong.

PR #880, opened 2017-08-21
Add document supertypes and remaining fields for help_pages

Rosa-Fox:
This covers all the fields for help pages now. 

Add `help_page` to migrated formats. Search results for help pages will now be returned from `govuk` index.

Mega-mobbed with @brenetic and @binaryberry 

https://trello.com/c/t0TCT7DX/231-implement-the-helppages-format


lib/govuk_index/elasticsearch_presenter.rb
```diff
@@ -16,28 +16,42 @@ def identifier
     end
 
     def document
+      {}.merge(common_fields)
+        .merge(fields_from_expanded_links)
```
dwhenry: I know we seem to like manually creating the JSON output at GDS, have we looked at/thought about using a gem to do the combinational work, this leaves more easily testable methods, rather than only being able to test the JSON response?

i.e. something like ActiveModel::Serializer

dwhenry (commented):
I have seen this approach of manually building hashes in code and it is always difficult to test. While alternatives that leverage gems can make the testing - and more importantly - later modification, easier. IMO.

lib/govuk_index/elasticsearch_presenter.rb
```diff
@@ -63,23 +73,37 @@ def withdrawn?
...
+
+    def values_from_collection(collection, key)
+      collection.map { |item| item[key] }
     end
 
-    def organisations
-      expanded_links["organisations"] || {}
+    def organisation_slugs(type)
+      expanded_links_item(type).map do |content_item|
+        content_item["base_path"].sub('/government/organisations/', '').sub('/courts-tribunals/', '')
```
suzannehamilton: This might be too much work, but is there any chance that we could push this logic into the publishing API? It seems a bit fragile - rummager shouldn't have to know how to parse URLs for different formats.

Alternatively, could rummager store them with their full paths? The applications which use them would have to search by path rather than by slug.

Rosa-Fox: Our preference was to use the full path but @MatMoore requested we keep it the same as it is currently done - https://github.com/alphagov/rummager/blob/c780376f4e7a0add2e3cf3af48eda5ab01724753/lib/indexer/links_lookup.rb#L108

Hopefully this will keep consistency with results from both indexes

MatMoore: Yup, this is rummager weirdness IMO. Really there's no need to have it without the leading slash but we're leaving it like that to keep API compatibility.

If it's cleaner we could also move this to the presentation layer: https://github.com/alphagov/rummager/blob/master/lib/search/presenters/result_presenter.rb

lib/govuk_index/elasticsearch_presenter.rb
MatMoore: @dwhenry Could you give an example? Do you mean creating intermediate objects for different parts of the response, and then serialising those?

test/unit/govuk_index/elasticsearch_presenter_test.rb
```diff
@@ -10,29 +10,8 @@ def setup
...
-      },
-      "payload_version" => 1,
-      "public_updated_at" => "2016-02-29T09:24:10Z",
-      "publishing_app" => "rails_for_the_win",
-      "rendering_app" => "react_rules_ok",
-      "title" => "This is a help page",
-    }
-
-    presenter = elasticsearch_presenter(payload, "help_page")
+    payload = File.read(File.join(__dir__, 'fixtures/publishing_api_payload.json'))
```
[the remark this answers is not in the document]
MatMoore: We should definitely validate against the notification schema here as it could easily become invalid over time.

I don't think we need one test per field but my preference would be to split to up a bit so its a bit easier to see whats failing.

Maybe we could use the schema helpers gem to and then override a specific set of fields we want to test?

lib/govuk_index/elasticsearch_presenter.rb
dwhenry: @MatMoore exactly, that way it makes sense to test a field at a time - as each field is a method on the object. This hopefully makes the objects cleaner and easier to test.

binaryberry: We are leaving this for now as we feel that this should be considered a little later on perhaps when moving a format with parts or moving licences

lib/govuk_index/elasticsearch_presenter.rb
binaryberry: We chose to keep it the same as the current mainstream index, but happy to discuss further.

suzannehamilton (approved):

lib/rummager.rb
```diff
@@ -56,8 +56,10 @@
 
 require 'govuk_index/client'
 require "govuk_index/document_type_inferer"
-require 'govuk_index/elasticsearch_presenter'
 require 'govuk_index/elasticsearch_processor'
+require 'govuk_index/presenters/elasticsearch_presenter'
+require 'govuk_index/presenters/common_fields_presenter'
+require 'govuk_index/presenters/expanded_links_presenter'
```
dwhenry: alphabetical

test/unit/govuk_index/presenters/common_fields_presenter_test.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    Indexer::PopularityLookup.expects(:new).with('govuk_index', SearchConfig.instance).returns(@popularity_lookup)
+    @popularity_lookup.expects(:lookup_popularities).with([payload['base_path']]).returns({})
+
+    presenter = common_fields_presenter(payload)
+
+    assert_equal nil, presenter.document['popularity']
+  end
+
+  def assert_directly_mapped_fields(presented_fields, generated_example)
```
dwhenry: Why have we made this a method considering it is only called once?

dwhenry (approved):
Just the two comments. feel free to ignore

PR #881, opened 2017-08-22
[WIP] Use Sentry for error reporting

tijmenb:
This is a proof of concept to see how Sentry behaves in this application. Let me know if there's anything weird looking.

https://trello.com/c/2AP3O9dn


suzannehamilton:
👍 This looks good to me.

We might also be relying on Airbrake logging in the govuk_message_queue_consumer library, which checks if `Airbrake` is defined and then uses it to log exceptions: https://github.com/alphagov/govuk_message_queue_consumer/search?utf8=%E2%9C%93&q=airbrake&type=

Should we look into adding Sentry to that gem as well? (Or some other way of passing the logging configuration.)

PR #882, opened 2017-08-22
[WIP] Proof of concept for reindexing from a separate queue

MatMoore:
At the moment, this reuses the existing sidekiq queue to process the messages.
I think we should add another sidekiq queue as well so that we can priorise
normal indexing messages, but for the moment I just want to make sure that
reindexing does not affect other parts of the system if the consumer isn't
working properly.

- If no consumers are running RabbitMQ should drop the messages (auto delete)
- If RabbitMQ falls over, the queues can be lost and the consumers can just die (transient)
- We are expecting a high throughput so the assumptions the message queue makes about prefetching are wrong. I've picked a fairly arbitrary number (10) instead.

Trello: https://trello.com/c/k2L67c4a/272-cleanup-puppet-module-for-declaring-message-queue-consumers

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,4 @@
 worker: bundle exec sidekiq -C ./config/sidekiq.yml
 publishing-queue-listener: bundle exec rake message_queue:listen_to_publishing_queue
 govuk-index-queue-listener: bundle exec rake message_queue:insert_data_into_govuk
+govuk-index-queue-reindex-listener: bundle exec rake message_queue:listen_to_reindexing_messages
--- a/lib/rummager.rb
+++ b/lib/rummager.rb
@@ -30,12 +30,14 @@ require 'yaml'
 initializers_path = File.expand_path("../config/initializers/*.rb", __dir__)
 Dir[initializers_path].each { |f| require f }
 
+require 'govuk_message_queue_consumer'
 require 'analytics_data'
 require File.expand_path('../config/logging_setup', __dir__)
 require 'document'
 require 'duplicate_deleter'
 require 'duplicate_links_finder'
 require 'govuk_document_types'
+require 'bulk_consumer'
 
 require 'indexer'
 require 'indexer/amender'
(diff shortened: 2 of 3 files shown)


lib/bulk_consumer.rb
```diff
@@ -0,0 +1,19 @@
...
+  class BulkConsumer < Consumer
+    NUMBER_OF_MESSAGES_TO_PREFETCH = 10
+
+  private
+
+    def queue
+      @queue ||= begin
+        channel.prefetch(NUMBER_OF_MESSAGES_TO_PREFETCH)
+
+        channel.queue(@queue_name, auto_delete: true, durable: false).tap do |queue|
```
suzannehamilton: I'm just trying to wrap my head around `auto_delete` from the docs, so this question might be a bit confused. 🤔 

Does `auto_delete: true` mean that the queue would be deleted if the consumer goes away, e.g. if rummager was redeployed in the middle of a bulk indexing?

Is that the intended behaviour, and does the queue get recreated if a new message with the queue's binding key arrives at the exchange?

lib/bulk_consumer.rb
```diff
@@ -0,0 +1,19 @@
+module GovukMessageQueueConsumer
+  # FIXME: this is a proof of concept for a consumer that uses self-declared,
+  # transient queues. We should adapt the consumer in the gem if this works.
```
suzannehamilton: 👍 

There's a comment in the gem saying "We're not expecting a high throughput on this queue" which seems like a bit of an assumption. When we update the gem it might be nice to do something with it, e.g. making it more configurable or adding a warning in the readme.

https://github.com/alphagov/govuk_message_queue_consumer/blob/389275f8a7dfb919e95d6f739cc4b4e0e84caff7/lib/govuk_message_queue_consumer/consumer.rb#L7

MatMoore:
After talking to Paul we've decided not to use `auto_delete` and non-durable queues, so I'm going to redo this PR.

PR #883, opened 2017-08-23
Revert migrating help page format

brenetic:
Searching for help page results when not all the content from the mainstream index has been added to the govuk index has resulted in the relevancy score being different to that in production. Therefore any help page results appear lower than their previous positions.

We are reverting this for now until we are able to add all other mainstream content to the govuk index or we find an alternative solution.


suzannehamilton (approved):
😢

PR #884, opened 2017-08-23
Delete stale comment

suzannehamilton:
The scroll timeout was dropped from 60 minutes to 1 minute in PR #811, so this comment is no longer necessary.


dwhenry (approved):
Ship

PR #885, opened 2017-08-23
Cater for documents with parts

brenetic:
Formats such as `guides` have parts in their details hash. We are now able to index these as well.

Also refactor the presenters. A little bit.

https://trello.com/c/kIyIPdEn/257-add-mainstream-guide-format-to-govuk-index

Paired with @dwhenry


lib/govuk_index/presenters/indexable_content_presenter.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    attr_reader :details, :sanitiser
+
+    def elements
+      if details['parts']
+        parts(details)
+      else
+        details.values
+      end
+    end
```
dwhenry: Just thinking if we want to support parts and anything else at the same time the following alternative might be nicer:
```
def elements
   details.map do |key, value|
      key == 'parts' ? parts(value) : value
  end
end
```

It would require a small change to the parts method as well

lib/govuk_index/presenters/expanded_links_presenter.rb
```diff
@@ -4,60 +4,68 @@ def initialize(expanded_links)
...
-    def expanded_links_item(key)
-      expanded_links[key] || {}
+    def organisation_content_ids
+      content_ids("organisations")
     end
 
-    def values_from_collection(collection, key)
-      collection.map { |item| item[key] }
+    def primary_publishing_organisation
+      organisation_slugs("primary_publishing_organisation")
```
suzannehamilton, line 20: This looks like it's going to return array for a single organisation. Is that OK?

lib/govuk_index/presenters/indexable_content_presenter.rb
```diff
@@ -0,0 +1,31 @@
...
+      sanitiser.clean(elements)
+    end
+
+  private
+
+    attr_reader :details, :sanitiser
+
+    def elements
+      if details['parts']
+        parts(details)
```
suzannehamilton: Just to clarify: if a guide has parts, the indexable content just comes from the parts? We don't include any details from the top-level page?

suzannehamilton (approved):
A couple of questions, but the refactoring and the parts mapping all look good. 👍 

Is it at all possible to split this into two separate commits to make the parts mapping clearer? I realise that might be quite fiddly, so don't worry about it if it's going to take too long.

lib/govuk_index/presenters/expanded_links_presenter.rb
brenetic: Do you mean it should just be the content id as a string?

lib/govuk_index/presenters/indexable_content_presenter.rb
brenetic: The top level page for something with parts will start with one of the parts. This seems to be the overview from all the examples I have found. Each part is always in the `details` `parts` payload

brenetic:
@suzannehamilton In hind sight we should have indeed done it in two commits. Kinda difficult now to split it out. Sorry! 😞

lib/govuk_index/presenters/expanded_links_presenter.rb
suzannehamilton: The name implies it's a single string, but I didn't know if we stored it as an array for backwards compatibility. Just wanted to make sure it was intentional. :)

PR #887, opened 2017-08-24
Enable help pages

dwhenry:
We need to fully populate data into govuk (and keep it updated) for it to correctly
calculate document scores. This is a set of changes to support this while we are
migrating formats.


lib/govuk_index/popularity_updater.rb
```diff
@@ -30,31 +31,29 @@ def self.migrate(index_name)
...
-        client: Services.elasticsearch(hosts: SearchConfig.instance.base_uri, timeout: TIMEOUT_SECONDS),
-        index_names: @source_index,
-        search_body: { query: { match_all: {} } },
-        batch_size: SCROLL_BATCH_SIZE
-      ) do |record|
-        {
-          identifier: record.slice(*%w{_id _type _version}),
-          document: record.fetch('_source'),
+    def search_body
+      # only sync migrated formats as the reset will be updated via the sync job.
```
brenetic: ~~reset~~ => `rest`

MatMoore:
This looks sensible based on what we've discussed (I'll raise an ADR to document this)

It's annoying we can't use the [reindex api](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/docs-reindex.html#docs-reindex) yet!

brenetic (approved):

PR #888, opened 2017-08-24
Licenses implementation

brenetic:
This adds the ability for licences content to be indexable as well as the identifier and short description fields to the main document sent to elasticsearch

https://trello.com/c/d5chF1sF/260-add-licences


brenetic:
@MatMoore Are we ok with sending empty fields for documents that are not licences?

lib/govuk_index/presenters/indexable_content_presenter.rb
```diff
@@ -15,9 +22,10 @@ def indexable_content
     attr_reader :details, :sanitiser
 
     def elements
-      details.flat_map do |key, value|
+      details.flat_map { |key, value|
```
dwhenry: I am sure rubocop normally complains about multiline blocks with {} syntax.... I agree with this rule.

lib/govuk_index/presenters/elasticsearch_presenter.rb
```diff
@@ -40,7 +40,9 @@ def document
         specialist_sectors:                 expanded_links.specialist_sectors,
         taxons:                             expanded_links.taxons,
         topic_content_ids:                  expanded_links.topic_content_ids,
-        indexable_content:                  indexable_content.indexable_content
+        indexable_content:                  indexable_content.indexable_content,
+        licence_identifier:                 licences.identifier,
+        licence_short_description:          licences.short_description,
```
dwhenry: I would like to combine the indexable_content and licences presenters into a single `DetailsPresenter` as we don't want to end up with 100's of presenters, instead I think we should aim to have as few as we can get away with and still have each dealing with related data.

Ignore if you disagree

dwhenry (approved):
Just a couple of small style things. feel free to ignore

lib/govuk_index/presenters/elasticsearch_presenter.rb
brenetic: Sounds good

lib/govuk_index/presenters/indexable_content_presenter.rb
brenetic: It complains with do; end, hence the {} 🙈

MatMoore:
@brenetic Yup, AFAIK it's the same as not passing it at all. Some info about it here: https://www.elastic.co/guide/en/elasticsearch/guide/current/_dealing_with_null_values.html

lib/govuk_index/presenters/details_presenter.rb
```diff
@@ -10,14 +12,24 @@ def indexable_content
...
+    end
+
   private
 
     attr_reader :details, :sanitiser
 
     def elements
-      details.flat_map do |key, value|
+      content = details.flat_map do |key, value|
+        next unless ALLOWED_FIELDS.include?(key)
```
dwhenry, line 29: not totally happy with combination of guard clause `if` and inline `if` but couldn't find a nicer way to write it.

dwhenry (approved):
Just the one comment, I wish it was better

PR #889, opened 2017-08-24
[DO NOT MEREGE] add support for deep pagination

dwhenry:
Proof of concept - please review and leave thoughts around including this into our codebase


dwhenry:
We decided to increase the `max_result_window` setting in ES rather than trying to code around the issue.

PR #891, opened 2017-08-24
Document how we plan to transition to a publishing API derived index

MatMoore:
[Rendered version](https://github.com/alphagov/rummager/blob/2374b65c568d907dca442c6d0e530faae8981de8/doc/arch/adr-004-transition-mainstream-to-publishing-api-index.md)

This describes some of the decisions we made while implementing https://github.com/alphagov/rummager/pull/887

### DIFF
--- a/doc/arch/adr-004-transition-mainstream-to-publishing-api-index.md
+++ b/doc/arch/adr-004-transition-mainstream-to-publishing-api-index.md
@@ -76,7 +76,7 @@ At search time, Rummager reads untransitioned formats from the old indexes. Docu
 
 At index time, Rummager will ignore publishing API messages affecting untransitioned formats.
 
-The nightly update job will update the popularity field of untransitioned formats in the `mainstream` index. Then it will copy untransitioned format documents from the `mainstream` index into the `govuk` index.
+The nightly update job will update the popularity field of untransitioned formats in the `mainstream` index. It will also copy untransitioned format documents from the `mainstream` index into the `govuk` index (it doesn't matter which order these two things happen).
 
 Net effect: untransitioned formats are considered in the TF-IDF statistics of transitioned formats, but are not ready to be returned from the `govuk` index themselves.
 
--- a/doc/arch/adr-004-transition-mainstream-to-publishing-api-index.md
+++ b/doc/arch/adr-004-transition-mainstream-to-publishing-api-index.md
@@ -1,0 +1,107 @@
+# Decision record: Transition mainstream formats to a Publishing API derived search index
+
+Date: 2017-08-24
+
+## Definitions
+
+Throughout this document, `format` refers to the Rummager field `format`, not the document type provided by Publishing API (`content_store_document_type`).
+
+Every `content_store_document_type` is mapped to a single `format`.
+
+`mainstream` is used as an example of a search index we want to retire. This document should also apply to the `government` and `detailed` indexes.
+
+## Context
+### We're building a new search index using publishing API data
+We're currently replacing the existing search indexes (`mainstream`, `government`, `detailed`) with a single index (`govuk`), which is derived from Publishing API.
+
+The existing indexes are populated from [at least 10 different applications](https://docs.publishing.service.gov.uk/apps.html).
+
+This change should improve the quality of the data our search system relies on:
+
+- all formats are handled in a consistent manner
+- we can update documents after any publishing event (publishing, withdrawing, unpublishing, tagging)
+- it should be easier to rebuild if we need to revert to an old backup
+
+### Lifecycle of our search indexes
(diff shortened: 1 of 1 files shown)


doc/arch/adr-004-transition-mainstream-to-publishing-api-index.md
```diff
@@ -0,0 +1,107 @@
...
+As of [ADR 003: Popularity updating without index locks](adr-003-popularity-updating-without-index-locks.md),
+this uses updates, rather than rebuilding the search index every time.
+
+There is a separate task in place to reindex the content currently in the search indexes with [zero downtime](https://www.elastic.co/guide/en/elasticsearch/guide/current/index-aliases.html).
+This is something we need to do after adding new fields or otherwise changing
+the Elasticsearch mappings, for those fields to work properly.
+
+Note that neither of these mechanisms add or remove documents from the search index.
+
+- If an edition is ever published, and the search index doesn't get updated, it stays out of date until the next time the document is updated
```
dwhenry: this could use a heading like `Issues with the current implementation`

doc/arch/adr-004-transition-mainstream-to-publishing-api-index.md
```diff
@@ -0,0 +1,107 @@
...
+from publishing API, to avoid these kind of problems, but we aren't planning to do this right now.
+
+### Immediate needs
+
+We need to be able to populate the `govuk` index with all the documents that
+already exist, so we can start using this index in the Search API.
+
+We also want to be able to reindex everything that has been published within a
+short period of time (1 or 2 days), so that we can easily recover the index
+from backups.
```
dwhenry, line 52: without using backups? we will be rebuilding from scratch after all. Also, ES will take approx 1-2hrs to reinsert the data so maybe we should be more optimistic with the time estimate

doc/arch/adr-004-transition-mainstream-to-publishing-api-index.md
```diff
@@ -0,0 +1,107 @@
...
+- Reindexing by date range lets us bring the `govuk` index up to date when restoring from a backup.
+
+Secondly, when we change the indexing process for a format to use the new index, the format will go through the following phases:
+
+### Phase 1: Untransitioned
+At search time, Rummager reads untransitioned formats from the old indexes. Documents belonging to untransitioned formats that are stored in the `govuk` index will be filtered out.
+
+At index time, Rummager will ignore publishing API messages affecting untransitioned formats.
+
+The nightly update job will update the popularity field of untransitioned formats in the `mainstream` index. Then it will copy untransitioned format documents from the `mainstream` index into the `govuk` index.
```
dwhenry: This step is not required as we won't be using the popularity field for untransitioned documents. The popularity is a boost after the TF-IDF calculation. :)

dwhenry (changes requested):
Look pretty good, just a couple of small comments.

doc/arch/adr-004-transition-mainstream-to-publishing-api-index.md
MatMoore: I'm assuming that we won't do the "from scratch" part. This simplifies the implementation a bit as we can be less concerned about performance, and don't have to coordinate all the index management stuff (working out when to actually switch aliases).

I think we don't actually need to do a full rebuild from scratch now though, and the big win is just replaying the last N days.

doc/arch/adr-004-transition-mainstream-to-publishing-api-index.md
MatMoore: I'm not sure what you mean here - this is talking about docs that get returned from the mainstream index, so they still get their popularity field updated there.

suzannehamilton (approved):
This is great. Every question that occurred to me while reading it was answered somewhere in the document. 👍

doc/arch/adr-004-transition-mainstream-to-publishing-api-index.md
MatMoore: I've pushed a new commit with a separate section for "problems we're not solving", so hopefully it's a bit clearer now.

doc/arch/adr-004-transition-mainstream-to-publishing-api-index.md
dwhenry: What I mean is that they don't need to have a valid popularity. This is because they will never be scored as they will never be returned. So we could easily do the copy to govuk before/during the mainstream popularity update as after it is finished.

MatMoore: Ah I see. I'll clarify this

dwhenry (approved):

brenetic (approved):
🚀

PR #892, opened 2017-08-24
Add mainstream transaction format to govuk index

brenetic:
his adds the necessary additional fields for the transaction format. The fields are also used by other formats in the publishing api payload but not necessarily required in elasticsearch so we need to specify certain fields by format

https://trello.com/c/k0gpEuQi/264-add-transactions

### DIFF
--- a/lib/govuk_index/presenters/elasticsearch_presenter.rb
+++ b/lib/govuk_index/presenters/elasticsearch_presenter.rb
@@ -64,7 +64,12 @@ module GovukIndex
     end
 
     def details
-      @_details ||= DetailsPresenter.new(payload["details"])
+      @_details ||=
+        DetailsPresenter.new(
+          details: payload["details"],
+          format: common_fields.format,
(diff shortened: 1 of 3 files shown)


MatMoore (approved):

lib/govuk_index/presenters/elasticsearch_presenter.rb
```diff
@@ -64,7 +64,13 @@ def common_fields
...
 
     def details
-      @_details ||= DetailsPresenter.new(payload["details"])
+      @_details ||= begin
+        DetailsPresenter.new(
+          details: payload["details"],
+          format: common_fields.format,
+          sanitiser: IndexableContentSanitiser.new
+        )
+      end
```
dwhenry: why is this wrapped in a begin/end block

brenetic: multi line memoization

PR #897, opened 2017-08-25
Make simple smart answer format indexable

brenetic:
https://trello.com/c/ONNnYqJ3/267-make-simple-smart-answer-format-indexable


suzannehamilton (approved):
⛵️ 🚢 🚀 

Might be worth putting future changes to this file in the same PR if we're planning to deploy them all together?

PR #899, opened 2017-08-25
Make place format indexable

brenetic:
https://trello.com/c/iX375XO9/261-add-place-format


MatMoore (approved):
How come this PR is blocked on the others? I think we can make stuff indexable at any time, as long we delete & repopulate it later.

PR #901, opened 2017-08-29
Delete documents by format rake task

brenetic:
In order to move each format over to the `govuk` index we need to purge all existing documents by format and then republish them.

https://trello.com/c/CRDK5Rmi/256-switch-over-mainstream-publisher-to-the-new-index


MatMoore (approved):
Looks good. Only thing I would add is a check that `args[:format]` is not nil, just in case elasticsearch does something weird with it.

PR #902, opened 2017-08-29
Use consitent sitemap names

dwhenry:
We believe this will allow search engines to better parse the sitemap

https://trello.com/c/bsvYLOaf/265-see-if-renaming-the-sitemap-affects-how-google-indexes-it

### DIFF
--- a/lib/sitemap/sitemap.rb
+++ b/lib/sitemap/sitemap.rb
@@ -1,16 +1,35 @@
 class Sitemap
+  SUB_DIRECTORY = "sitemaps".freeze
+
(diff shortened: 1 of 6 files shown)


sitemap_1.xml
```diff
@@ -0,0 +1 @@
+1a6c2065-bb98-4c0d-af0e-fef01759e9b1.xml
```
brenetic: Do we need this?

dwhenry: nope.. deleted

brenetic (approved):
Linting issues. Otherwise 🚀

PR #903, opened 2017-08-29
Make delete by format rake task Elasticsearch 2.4 compatible

brenetic:
`delete_by_query` is not supported in elasticsearch 2.4. We have swapped to use the scrollenumerator to delete documents in batches instead.

https://trello.com/c/3qwYPp4P/277-delete-and-republish-mainstream-formats-to-govuk-index

### DIFF
--- a/lib/tasks/delete.rake
+++ b/lib/tasks/delete.rake
@@ -67,21 +67,39 @@ namespace :delete do
   rake 'delete:by_format[format_name, elasticsearch_index]'
   "
   task :by_format, [:format, :index_name] do |_, args|
-    if args[:format].nil?
-      puts 'Specify format for deletion'
-    else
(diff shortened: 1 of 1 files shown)


MatMoore (approved):
Looks good 👍

lib/tasks/delete.rake
```diff
@@ -67,21 +67,37 @@ namespace :delete do
   rake 'delete:by_format[format_name, elasticsearch_index]'
   "
   task :by_format, [:format, :index_name] do |_, args|
-    if args[:format].nil?
+    format = args[:format]
+    index  = args[:index_name]
+
+    if format.nil?
```
dwhenry, line 73: do we want a check that an index has been provided?

PR #906, opened 2017-08-30
Use content_store_document_type in delete task

brenetic:
The `govuk` index contains documents that were added without some of the fields during moving the formats in mainstream publisher. This task is temporary until we clear up the mainstream documents. Once this is done it can be removed as we will usually only want to delete by format.

https://trello.com/c/3qwYPp4P/277-delete-and-republish-mainstream-formats-to-govuk-index

### DIFF
--- a/lib/tasks/delete.rake
+++ b/lib/tasks/delete.rake
@@ -102,4 +102,46 @@ namespace :delete do
       end
     end
   end
+
+  desc "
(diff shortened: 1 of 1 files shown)


dwhenry (approved):
I would suggest just deleting everything that has a "MISSING" format field, but this works too.. :)

PR #907, opened 2017-08-31
Make answer format fully migrated

brenetic:
We want to make sure that we can fully migrate all the mainstream formats over to use the new `govuk` index. Start with the answer format before doing the rest in bulk.

https://trello.com/c/XTRquJcn/286-make-answer-format-fully-migrated


dwhenry (approved):
Might be nice to extract the test change out to a separate commit and explain why it is required (i.e. search will no longer return data for answers as it is being migrated so we have changed the format to transaction)

PR #911, opened 2017-09-04
Use the new index for all mainstream publisher formats

MatMoore:


config/govuk_index/migrated_formats.yaml
```diff
@@ -1,10 +1,11 @@
...
 - help_page
-indexable:
 - guide
 - licence
 - local_transaction
 - place
 - transaction
 - simple_smart_answer
+
+indexable:
```
dwhenry: I think this will need 
```
indexable: []
```
as it is expected to an array.

test/integration/indexer/migration_test.rb
```diff
@@ -1,9 +1,15 @@
 require 'integration_test_helper'
 
+# FIXME: this tests the BulkLoader. We don't think we need this anymore.
+# The "Migration" in the name means creating new indexes and copying data from
+# the existing ones.
```
dwhenry, line 5: We need to actually switch this over in the search_analytics codebase which I haven't got to yet.

dwhenry (approved):
Just check on that comment I made, but look good..

PR #912, opened 2017-09-04
Upgrade gds-api-adpaters to 47.9

thomasleese:
This will allow the use of the updated special route publisher
which sends update_type as part of put content rather than publish,
which will shortly be deprecated.


cbaines (approved):
👍

thomasleese:
I've had a look through the change log and I can't see anything obvious in the upgrade that might be a problem, and since the tests pass I'm going to merge this.

PR #914, opened 2017-09-07
Make popularity updater work for non-migrated indices

dwhenry:
This allow us to pass in an override which causes all data
to be processed by the popularity updater.

This requires a change in search_analytics to pass in the
`PROCESS_ALL_DATA` env variable to work for non govuk indices.

https://trello.com/c/IeU8rFj3/293-switch-indexes-to-use-update-popularity-instead-of-bulk-update

### DIFF
--- a/lib/govuk_index/popularity_updater.rb
+++ b/lib/govuk_index/popularity_updater.rb
@@ -1,9 +1,10 @@
 module GovukIndex
   class PopularityUpdater < Updater
-    def self.update(index_name)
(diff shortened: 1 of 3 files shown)


MatMoore (approved):
Can we just remove the old query for doing just the migrated formats, and make it always do all of them?

dwhenry:
We could, but that would be duplicated effort as the data would just be overwritten by the sync task that runs next. It would mean the process task longer than it should.

We will remove the branch once we have migrated everything. :)

PR #916, opened 2017-09-07
Prevent duplicate smart answers results

Rosa-Fox:
- The smart-answers application is sending items to Rummager with type smart-answer.

- The smart-answers application is also sending items to content store and Rummager via the publishing-api with type transaction. This has caused duplication in results for smart-answer pages.

- We still allowed for the deletion of smart answers as when a smart answer is deleted, the entry for it in the govuk index which had been synced across from mainstream is also deleted. This mirrors the functionality for other unmigrated formats. This is needed as the sync process does not delete data from the govuk index.

- This is a temporary fix to filter smart-answers from the govuk index within Rummager. We will investigate a permanent solution later in the sprint, during our spike work for migrating other formats to govuk index.

Paired with @brenetic @dwhenry

### DIFF
--- a/lib/govuk_index/publishing_event_worker.rb
+++ b/lib/govuk_index/publishing_event_worker.rb
@@ -52,7 +52,7 @@ module GovukIndex
       if presenter.unpublishing_type?
         logger.info("#{routing_key} -> DELETE #{presenter.base_path} #{presenter.type}")
         actions.delete(presenter)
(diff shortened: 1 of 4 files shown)


brenetic (approved):
🚀  LGTM FFTMWR

dwhenry:
LFTM @MatMoore do you want to do a review?

PR #917, opened 2017-09-08
Bump govuk_document_types to 0.1.8

tuzz:


greysteil:
I think I've asked before but no-one got back to me. Would you be up for automating these dependency update PRs, the same way MoJ do? See https://github.com/ministryofjustice/prison-visits-2/pulls?q=is%3Apr+is%3Aclosed for examples. Hopefully a time-saver, and would mean Dependabot has more data for its compatibility scores, too.

tuzz:
This is probably a question for @boffbowsh, @timblair.

boffbowsh:
Thanks @greysteil. We're currently using Snyk for a similar use case, but are about to re-evaluate that. I'll add Dependabot to the list of things to look at when we do this.

boffbowsh (approved):

dwhenry:
@tuzz can you rebase this and then we can get it merged.

greysteil:
*Can't resist the urge to mention that Dependabot automatically rebases its PRs for you.* :wink:

PR #918, opened 2017-09-11
Move Specialist Publisher documents to govuk index

brenetic:
# This migration of Specialist formats is NOT complete.

This adds all the required fields, however we need to address the issues raised when running the comparer rake task before proceeding. We believe merging this code is the best thing to do, as it won't affect the production system and gives a clear direction of next steps.

https://trello.com/c/Byp54SuK/308-implement-specialist-publisher


lib/govuk_index/presenters/method_builder.rb
```diff
@@ -0,0 +1,26 @@
...
+      @data_object || :payload
+    end
+
+    def data_object=(data_object)
+      define_method :payload do
+        public_send(data_object)
+      end
+      # @data_object = data_object
+    end
+  end
```
dwhenry: this is a combination of approaches as we are trying to decide on an approach

lib/govuk_index/presenters/details_presenter.rb
```diff
@@ -1,5 +1,10 @@
 module GovukIndex
   class DetailsPresenter
+    extend MethodBuilder
+    self.data_object = :details
```
MatMoore: I think it's not obvious enough how `data_object` and `add_method` interact here. 

It feels a bit magic because you have to look at the code of the mixin to understand what this line does, and if I'm looking at what this presenter does I don't really want to care about `MethodBuilder`.

What I'd suggest instead is to use an argument of `add_method` to explicitly say where you are getting the data from. This method feels very similar to rails's `delegate`, which uses the named argument `to` for this.

If you're going with the `add_method` approach, what about something like this?

```ruby
delegate_to_hash :payload "licence_identifier"
delegate_to_hash :payload, public_timestamp: "public_updated_at"
```

or

```ruby
map_field :public_timestamp hash: :payload, key: "public_updated_at"
```

lib/govuk_index/presenters/method_builder.rb
MatMoore: As someone who doesn't understand a lot of ruby metaprogramming, I'd go with what looks simplest: `define_method`.

Another approach I've seen is to call attr_reader with an array of names. If you have a list of all things you want to delegate to the hash, then you can do something like this

```ruby
attr_reader *FIELDS
```

or 

```ruby
attr_accessor *FIELDS
```

Then you can do all your metaprogramming in the initializer:

```ruby
FIELDS.each do |field|
  instance_variable_set field payload[field]
end
```

I don't have a strong opinion on this, but we should keep it as simple as possible, and make sure that the syntactic sugar doesn't get in the way of understanding how the presenters are mapping from one to thing to the other.

:man_shrugging:

lib/govuk_index/presenters/details_presenter.rb
```diff
@@ -1,48 +1,30 @@
 module GovukIndex
   class DetailsPresenter
```
MatMoore, line 2: I think we should consider renaming this class, because `details` isn't something we're presenting, it's something we're consuming.

I think the job of this class is to make the incoming payload easier to pull apart.

In general, there's no reason why we couldn't have fields in our elasticsearch documents that are derived partly from the details hash and partly from other bits of the content item.

`indexable_content` in particular is one of the most complicated fields, and it's also really important. Maybe that should live in the main presenter, and delegate to this one for `indexable_content_parts`?

lib/govuk_index/presenters/common_fields_presenter.rb
```diff
@@ -1,70 +1,34 @@
 module GovukIndex
   class CommonFieldsPresenter
+    extend MethodBuilder
+
+    add_method :content_id
```
suzannehamilton: I wonder if it would be clearer to name this something like `map_field`? `add_method` makes it sound more generic than it is.

lib/govuk_index/presenters/details_presenter.rb
suzannehamilton: +1

@brenetic mentioned earlier that it seemed easier to structure the presenters around the payload structure when we originally started the mapping, but it might not make sense anymore when we have to map several very different formats.

lib/govuk_index/presenters/common_fields_presenter.rb
suzannehamilton: Oops, just spotted Mat's comment below which has a more detailed suggestion. Though we did come up with the same name. 🙃

dwhenry: I have refactored the naming here to try and make it clearer, plaese let me know if you think the new names works better/still need improvement?

lib/govuk_index/presenters/details_presenter.rb
dwhenry: I think we are still some way off on naming, however I think the current split is working pretty well. I will leave off renaming this class right now, with the plan to look at the name of this class next time we add any functionality to it (I had originally planned to put the specialist formats fields in this class, but changed my mind in the end.

lib/govuk_index/presenters/elasticsearch_presenter.rb
```diff
@@ -49,6 +49,64 @@ def document
...
+        medical_specialism:                 specialist.medical_specialism,
+        railway_type:                       specialist.railway_type,
+        fault_type:                         specialist.fault_type,
+        faulty_item_type:                   specialist.faulty_item_type,
+        alert_issue_date:                   specialist.alert_issue_date,
+        manufacturer:                       specialist.manufacturer,
+        faulty_item_model:                  specialist.faulty_item_model,
+        serial_number:                      specialist.serial_number,
+        build_start_date:                   specialist.build_start_date,
+        build_end_date:                     specialist.build_end_date,
```
brenetic: Are we going to send all of these fields even when the document is not a specialist document? I know we send nil for a few fields at the moment but that's a lot of extra fields which a high percentage of documents won't be using. It feels like we should be splitting common fields, expanded links, specialist etc out.

lib/govuk_index/presenters/specialist_presenter.rb
```diff
@@ -0,0 +1,74 @@
...
+    delegate_to_payload :medical_specialism
+    delegate_to_payload :railway_type
+    delegate_to_payload :fault_type
+    delegate_to_payload :faulty_item_type
+    delegate_to_payload :alert_issue_date
+    delegate_to_payload :manufacturer
+    delegate_to_payload :faulty_item_model
+    delegate_to_payload :serial_number
+    delegate_to_payload :build_start_date
+    delegate_to_payload :build_end_date
```
brenetic: Will we not therefore be sending all of these regardless of whether the specific document type requires them? Do we want to do that?

lib/govuk_index/presenters/details_presenter.rb
```diff
@@ -1,48 +1,32 @@
...
-
-    def licence_short_description
-      details['licence_short_description']
+      @sanitiser.clean(indexable_content_parts)
     end
 
   private
 
-    attr_reader :details, :format, :sanitiser
+    attr_reader :details
```
brenetic, line 23: Might as well remove this attr_reader too and use the escargot.

lib/govuk_index/presenters/elasticsearch_presenter.rb
dwhenry: Why.. 

Sending everything might make the code slightly slower, but the alternative is to have more complex code. 

Do we perceive speed or complexity to be the most important? It will be running on threaded background workers.

lib/govuk_index/presenters/specialist_presenter.rb
dwhenry: yep

lib/govuk_index/presenters/details_presenter.rb
dwhenry: can't as it is accessed from the MethodBuidler code, I remove the other two as the accessor is only used in a single location, so it seem overkill.

lib/govuk_index/presenters/elasticsearch_presenter.rb
dwhenry: NOTE: I am happy to change this so we limit the fields if we believe that gives us an advantage, it just for me the biggest advantage would be from simplified code which this approach gives us. 

The other advantage of only using fields specified in the logic is that we can ensure the validity of our data, but again I don't think that is a strong enough argument given that the publishing app, schemas, and publishing API should already be doing this.

lib/govuk_index/presenters/details_presenter.rb
```diff
@@ -1,54 +1,42 @@
...
-    }.freeze
+    extend MethodBuilder
 
-    def initialize(details:, format:, sanitiser:)
+    set_payload_method :details
+
+    delegate_to_payload :licence_identifier
+    delegate_to_payload :licence_short_description
+
+    def initialize(details:, indexable_content_keys:, sanitiser:)
```
dwhenry, line 10: we can remove format as public methiod from DocTypeInferer as it is no longer used

test/integration/govuk_index/specialist_formats_test.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def test_finders_email_signup_are_never_indexed
+    random_example = GovukSchemas::RandomExample
+                       .for_schema(notification_schema: "finder_email_signup")
+                       .merge_and_validate(document_type: "finder_email_signup")
+    @queue.publish(random_example.to_json, content_type: "application/json")
+
+    assert_raises(Elasticsearch::Transport::Transport::Errors::NotFound) do
+      fetch_document_from_rummager(id: random_example["base_path"], index: "govuk_test", type: 'edition')
```
dwhenry: type: '_all'

lib/govuk_index/presenters/details_presenter.rb
dwhenry: This was incorrect the public method is required

test/integration/govuk_index/specialist_formats_test.rb
dwhenry: done

test/integration/govuk_index/specialist_formats_test.rb
```diff
@@ -0,0 +1,90 @@
...
+      maib_report
+      medical_safety_alert
+      raib_report
+      service_standard_report
+      tax_tribunal_decision
+      utaac_decision
+      vehicle_recalls_and_faults_alert
+    )
+
+    # ideally we would run a test for document types, but this takes 3 seconds so I have limited
```
MatMoore: assume this is meant to say "for all document types"?

lib/indexer/migration_field_comparer.rb
```diff
@@ -0,0 +1,54 @@
+module Indexer
+  class MigrationFieldComparer
```
MatMoore: Is there a clearer name for this? Hesitant to keep naming stuff "migration" because there's at least 3 meanings of that word and I think it will confuse people later on.

MatMoore (approved):
LGTM. I think it's worth merging now as we intend to build on it.

Just a couple of minor comments.

test/integration/govuk_index/specialist_formats_test.rb
dwhenry: corrected

lib/indexer/migration_field_comparer.rb
dwhenry: renamed to GovukIndexFieldComparer and added a comment about reasons for existence and what it does.

PR #919, opened 2017-09-11
Configure Sentry error reporting

Rosa-Fox:
[Trello](https://trello.com/c/92eP2Rr9/65-update-rummager)

Paired with @suzannehamilton


lib/rummager.rb
```diff
@@ -3,13 +3,14 @@
...
 require 'active_support/core_ext/hash'
 require 'active_support/core_ext/string'
-require 'airbrake'
 require 'cgi'
 require 'csv'
 require 'elasticsearch'
 require 'erb'
 require 'fileutils'
 require 'gds_api/publishing_api_v2'
+require 'sentry-raven'
```
suzannehamilton, line 12: @tijmenb: Do you think [configure.rb in govuk_app_config](https://github.com/alphagov/govuk_app_config/blob/master/lib/govuk_app_config/configure.rb) should `require sentry-raven`? This works, but we were a bit surprised that we had to add it (and make sure that it gets required before govuk_app_config). Though I realise we have to do a bit of extra work because rummager isn't on Rails. 🙂

tijmenb: I think it already does, but in a roundabout way:

https://github.com/alphagov/govuk_app_config/blob/master/lib/govuk_app_config.rb#L3
->
https://github.com/alphagov/govuk_app_config/blob/master/lib/govuk_app_config/govuk_error.rb#L1

We should probably require it directly in https://github.com/alphagov/govuk_app_config/blob/master/lib/govuk_app_config.rb, but this still should've worked?

Gemfile
```diff
@@ -22,7 +22,7 @@ gem 'sidekiq-limit_fetch'
 gem "sidekiq-statsd", "0.1.5"
 gem "sinatra", "~> 2.0.0"
 gem "slop", "~> 3.4.5"
-gem "statsd-ruby", "~> 1.3.0"
+gem "statsd-ruby", "~> 1.4.0"
```
dwhenry, line 25: do we need to manually add this given it is a dependency of `govuk_app_config`

dwhenry (approved):
Looks good just the one small query

Gemfile
suzannehamilton: Rummager makes some calls to statsd directly (e.g. in the PublishingEventProcessor), so I think it's worth including this an explicit dependency.

lib/rummager.rb
suzannehamilton: Cheers. I'm going to merge this PR now so that we can do some more testing before the ES 2.4 upgrade, but I'll take a look in case we misunderstood something yesterday.

PR #923, opened 2017-09-14
Attempt to describe fields in the search API

MatMoore:
Add descriptions to some of the fields. All of them should have descriptions,
but I don't have the patience to do all them now. These are used to populate https://docs.publishing.service.gov.uk/apis/search/fields.html

There's various things that could probably be removed or simplified, but before
even thinking about this I'd like to improve the documentation quality and
start versioning the API.

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -5,22 +5,27 @@
   },
 
   "format": {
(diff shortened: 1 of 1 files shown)


suzannehamilton (approved):
👏

brenetic (approved):
🚀

PR #924, opened 2017-09-14
Add another procfile worker for the bulk-reindex queue

MatMoore:
At the moment, this uses the same sidekiq queue, but I'm going to make sure
this works end to end before changing anything else.

Trello: https://trello.com/c/nPiykYmS/270-get-rummager-reindex-working-in-dev


suzannehamilton (approved):
Linter error, but otherwise this looks good. 👍

PR #926, opened 2017-09-15
Fix intermittent test failures

Rosa-Fox:
[Trello](https://trello.com/c/r88Bmytd/310-fix-intermittent-integration-test-failures-caused-by-smart-answers-exclusion)

- Tests generate a publishing API message based on a random schema.
- The failures happened when the random schema generator generated a message with the publishing_app set to `smartanswers`, because we [deliberately exclude smart answers](https://github.com/alphagov/rummager/pull/916) in the new search index.

Paired with @dwhenry

### DIFF
--- a/test/support/test_helpers.rb
+++ b/test/support/test_helpers.rb
@@ -1,7 +1,10 @@
 require "gds_api/test_helpers/publishing_api_v2"
 
 module TestHelpers
+  class RandomExampleError < StandardError; end
+
   include GdsApi::TestHelpers::PublishingApiV2
+  EXAMPLE_GENERATOR_RETRIES = 5
 
   def teardown
     Timecop.return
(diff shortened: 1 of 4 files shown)


test/support/test_helpers.rb
```diff
@@ -34,11 +35,17 @@ def setup_page_traffic_data(document_count:)
...
-      .for_schema(notification_schema: schema)
-      .merge_and_validate(payload, excluded_fields)
+    retry_attempts.times do
+      random_example = GovukSchemas::RandomExample
+        .for_schema(notification_schema: schema)
+        .merge_and_validate(payload, excluded_fields)
+
+      return random_example unless regenerate_if.call(random_example)
+    end
+    raise "Could not find example"
```
brenetic: Might need to rescue this. Is it worth creating a specific error class for the message?

test/support/test_helpers.rb
```diff
@@ -34,11 +37,17 @@ def setup_page_traffic_data(document_count:)
     commit_index("page-traffic_test")
   end
 
-  def generate_random_example(schema: "help_page", payload: {}, excluded_fields: [])
+  def generate_random_example(schema: "help_page", payload: {}, excluded_fields: [], regenerate_if: ->(_example) { true }, retry_attempts: EXAMPLE_GENERATOR_RETRIES)
```
MatMoore: shouldn't the default lambda always return `false`?

MatMoore (approved):

PR #927, opened 2017-09-15
[DISCUSS] ADR for performing popularity updates in a separate index

MatMoore:
This ADR discusses some issues with the popularity update.

We haven't actually decided on the approach yet but we've rejected a couple of solutions.

[Rendered markdown](https://github.com/alphagov/rummager/blob/2b9149695b04c20b17ae99d1f71b2f261d586a0f/doc/arch/adr-005-rebuild-into-empty-indexes.md)

Trello: https://trello.com/c/saDruTlG/307-investigate-1-day-intermittent-rummager-timeouts


dwhenry:
As a note on other options we could export:

* We already have code that inserts data into a new index while locking the current index - This is the migration code, so we could pivot to using this.
* We could use a read alias and a write alias for each index, if we then create a new index, assign the write alias to it the do the popularity insert into that index (as well as any new writes). The read index would miss new updates while this process was running, but that is the same as for a locked implementation. The downside here is that if the new process fails we have no way to recover the new (non reindex) write data.

Similar to the above approach but writes (that aren't reindexing) are performed for both the read and write alias, this keeps everything up today without locks. This would require more complex coding.

MatMoore:
@dwhenry Option 1 is probably a decent fallback plan if we can't address it by adding more resources, so I'm going to make this the focus of the ADR while @suzannehamilton investigates. I've added the second suggestion as a rejected alternative.

If it turns out that this doesn't work, or we can fix it without a significant code change, I think I'll just commit this as a rejected proposal. 

I don't think we've established yet whether using a separate index actually makes a difference. We're assuming so but it could be just that the write workload has gotten bigger somehow when we changed the implementation.

PR #928, opened 2017-09-18
Refactor load from code

dwhenry:
**We need to resolve issues with the finder frontend spikes before we merge this code.**

This removes the BulkLoad class and replaces to use sidekiq workers
for threading

https://trello.com/c/6beFCfUH/311-remove-recode-the-pagetraffic-import-code-so-that-we-can-fully-delete-the-bulkloader-class


bin/bulk_load
```diff
@@ -7,15 +7,12 @@ LIBRARY_PATH = PROJECT_ROOT + "lib/"
...
 $LOAD_PATH << LIBRARY_PATH unless $LOAD_PATH.include?(LIBRARY_PATH)
 
-INDICES = %w{mainstream detailed government}.freeze
-
 require "rummager"
 
 Slop.parse(help: true) do
-  search_config = SearchConfig.new
-  banner %{Usage: #{File.basename(__FILE__)} {#{search_config.index_names.join("|")}}
+  banner %{Usage: #{File.basename(__FILE__)}
```
suzannehamilton: Missing a closing brace?

bin/bulk_load
```diff
@@ -7,15 +7,12 @@ LIBRARY_PATH = PROJECT_ROOT + "lib/"
...
 
-Bulk loads data from stdin to the specified index. The data should be in the
+Bulk loads data from stdin to the page traffic index. The data should be in the
```
suzannehamilton: I think it'd be worth renaming this script now that it just handles page traffic data. "bulk_load" makes it sound more generic than it is.

lib/govuk_index/page_traffic_loader.rb
```diff
@@ -0,0 +1,48 @@
...
+        # new index exists, but the old index is available for writes
+        index_group.switch_to(new_index)
+        old_index.close
+      end
+    end
+
+  private
+
+    # Breaks the given input stream into batches of documents
+    def in_even_sized_batches(iostream, batch_size = @iostream_batch_size, &_block)
```
suzannehamilton: Do you think we should add a link to the [bulk API docs](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/docs-bulk.html) to explain why things need to be split into even-sized batches? It's pretty counterintuitive if you haven't used the bulk API before.

suzannehamilton (approved):
Just a few minor comments, but otherwise this looks like a big improvement. 🎉

bin/bulk_load
dwhenry: Moved to using <<~DOC..DOC as the multiline string with `%{..}` is rubbiush

bin/bulk_load
dwhenry: Done, I have also added a sym link for the old file name until we update the script in puppet

MatMoore (approved):
LGTM, but I think the worker names are confusing.

GooglePopularityUpdater vs PopularityWorker. 

Maybe it should be PageTrafficWorker?

PR #930, opened 2017-09-19
Fix the rummager:clean rake task after 2.4 upgrade

dwhenry:
This fixes two issues with the clean command
* ES 2.4 does not return closed indices for the `get_aliases` method
* govuk was being excluded from the list of indices when run with RUMMAGER_INDEX=all

https://trello.com/c/k1wSV5AM/325-fix-clean-command-es-24-cleanup

### DIFF
--- a/lib/index_group.rb
+++ b/lib/index_group.rb
@@ -99,7 +99,7 @@ module SearchIndices
     end
 
     def clean
-      alias_map.each do |name, details|
+      alias_map(include_closed: true).each do |name, details|
         delete(name) if details.fetch("aliases", {}).empty?
       end
     end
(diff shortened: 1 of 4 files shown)


suzannehamilton (approved):
Linter error, but otherwise looks good. 👍

lib/index_group.rb
```diff
@@ -124,10 +124,12 @@ def generate_name
...
     end
 
-    def alias_map
+    def alias_map(include_closed: false)
+      expand_wildcards = ['open']
+      expand_wildcards << 'closed' if include_closed
       # Return a map of all aliases in this group, of the form:
       # { concrete_name => { "aliases" => { alias_name => {}, ... } }, ... }
-      indices = @client.indices.get_aliases
+      indices = @client.indices.get(index: "#{@name}*", expand_wildcards: expand_wildcards)
```
suzannehamilton, line 132: Will this not return full index names instead of alias names? Or have I misunderstood how this works?

dwhenry: As discussed, `get_aliases` returned the index name with aliases listed in the associated hash, `get` has the same return object.

PR #931, opened 2017-09-19
Force index commit before attempting to retrieve data

dwhenry:
This should allow the data to become available in the elasticsearch
index before we try and retrieve it. This will hopefully help fix
the random test failures we have been encountering.

https://trello.com/c/ph0nXo7H/326-intermittent-test-failures


brenetic (approved):
🚀

PR #932, opened 2017-09-20
Reduce the number of Sidekiq workers used by bulk processing

suzannehamilton:
Queries have been timing out when the popularity update runs it bulk indexing. This change to the Sidekiq worker count reduces the indexing load in the hope that it improves parallel search queries.

https://trello.com/c/saDruTlG/307-investigate-1-day-intermittent-rummager-timeouts

Surprisingly, this didn't change the time taken for the queue to be processed (45 minutes in each case).

There are still slow requests and frontend 500 errors with this change, but still it's a clear improvement.

Before:
screen shot 2017-09-20 at 13 44 50

After:
screen shot 2017-09-20 at 13 45 17

Marked as "do not merge" while we look into disk I/O issues, because we still think the underlying problem might be in the VM rather than the code.

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -10,4 +10,5 @@ production:
   - default
   - bulk
 :limits:
-  bulk: 12
+  bulk: 4
+  default: 4


suzannehamilton:
With a limit of 4, the errors were lower again still. This took 50 minutes rather than 45, but the Sidekiq graph makes it look like another process added a few documents during the popularity update so the difference might not be significant.

screen shot 2017-09-22 at 16 50 49

suzannehamilton:
I've also added the a limit to the default queue because of the errors we saw yesterday when processing a lot of data through the `AmendWorker` (which uses the default rather than the bulk queue).

MatMoore (approved):

PR #933, opened 2017-09-21
Add migrated govuk index documents to site map

brenetic:
This uses filtering to add the migrated govuk index documents to the
site map.

We also refactorered the sitemap generator a little bit.

https://trello.com/c/gVTzUj6v

@Rosa-Fox and @dwhenry were also on The Scroll Train

### DIFF
--- a/lib/sitemap/sitemap_generator.rb
+++ b/lib/sitemap/sitemap_generator.rb
@@ -1,8 +1,8 @@
 class SitemapGenerator
   EXCLUDED_FORMATS = ["recommended-link", "inside-government-link"].freeze
 
-  def initialize(sitemap_indices)
-    @sitemap_indices = sitemap_indices
+  def initialize(search_config)
+    @search_config = search_config
     @all_documents = get_all_documents
   end
 
@@ -11,17 +11,31 @@ class SitemapGenerator
   end
 
   def get_all_documents
+    search_body = {
+      query: {
+        bool: {
+          must_not: { terms: { format: EXCLUDED_FORMATS } }
+        }
+      },
+      filter: Search::FormatMigrator.new.call
+    }
     property_boost_calculator = PropertyBoostCalculator.new
 
+    # We need the extra enumerator here so that we can inject the homepage
+    # as the first item to be processed.
     Enumerator.new do |yielder|
-      # Hard-code the site root, as it isn't listed in any search index
       yielder << homepage
 
-      @sitemap_indices.each do |index|
-        index.all_documents(exclude_formats: EXCLUDED_FORMATS).each do |document|
-          yielder << SitemapPresenter.new(document, property_boost_calculator)
-        end
+      enum = ScrollEnumerator.new(
+        client: Services.elasticsearch(hosts: @search_config.base_uri, timeout: SearchIndices::Index::TIMEOUT_SECONDS),
+        search_body: search_body,
+        index_names: index_names,
+        batch_size: SearchIndices::Index.scroll_batch_size
+      ) do |hit|
+        SitemapPresenter.new(hit["_source"], property_boost_calculator)
       end
+
+      enum.each { |doc| yielder << doc }
     end
   end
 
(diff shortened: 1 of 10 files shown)


test/unit/sitemap/sitemap_presenter_test.rb
```diff
@@ -120,14 +120,14 @@ def test_page_with_boosted_format_has_adjusted_priority
   end
 
   def build_document(url:, timestamp: nil, format: nil, is_withdrawn: nil)
-    attributes = {
+    document = {
       "link" => url,
-      "_type" => "some_type",
+      # "_type" => "some_type",
```
dwhenry: we should remove this line.

dwhenry (dismissed):
Just need to remove the comment

lib/sitemap/sitemap_generator.rb
```diff
@@ -1,8 +1,9 @@
 class SitemapGenerator
   EXCLUDED_FORMATS = ["recommended-link", "inside-government-link"].freeze
 
-  def initialize(sitemap_indices)
-    @sitemap_indices = sitemap_indices
+  def initialize(index_names:)
+    # @sitemap_indices = sitemap_indices
```
suzannehamilton: Commented-out line to delete.

test/integration/sitemap/sitemap_generator_test.rb
```diff
@@ -111,22 +144,37 @@ def test_links_should_include_timestamps
...
 
-    priorities = Nokogiri::XML(sitemap_xml[0])
-      .css("url > priority")
+    pages = Nokogiri::XML(sitemap_xml[0])
+      .css("url")
+      .select { |item| item.css("loc").text == "http://www.dev.gov.uk/an-example-answer" }
 
-    assert_equal 6, priorities.count
+    assert_equal 1, pages.count
+    assert_equal "0.5", pages[0].css("priority").text
```
suzannehamilton: This assertion is a bit confusing because it looks like it's the priority of the test in the setup, whereas it's actually the priority of the homepage.

I deliberately didn't add an assertion of a real priority in this test because I didn't want to make the integration tests depend on a slightly arbitrary number. The unit tests do some more in-depth testing of the actual value.

brenetic: @suzannehamilton I have changed the assertion to check the priority is within a range between 0-10. This will make it a little bit less brittle

lib/tasks/sitemap.rake
```diff
@@ -8,7 +8,10 @@ namespace :sitemap do
 
     output_directory = File.join(PROJECT_ROOT, "public")
     sitemap = Sitemap.new(output_directory)
-    sitemap_index_path = sitemap.generate(search_server.content_indices)
+    search_config = SearchConfig.instance
+    sitemap.generate(
+      index_names: search_config.content_index_names + [search_config.govuk_index_name]
```
MatMoore: I think it makes sense here to pass in `search_config` rather than the index names, because the generator actually has a dependency on the whole thing (since it also accesses the singleton object).

spec/integration/sitemap/sitemap_generator_spec.rb
```diff
@@ -1,73 +1,71 @@
...
+          "link" => "/an-example-answer",
+          "indexable_content" => "I like my badger: he is tasty and delicious",
+          "public_timestamp" => "2017-07-01T12:41:34+00:00"
+        },
+        {
+          "title" => "Cheese on Ruby's face",
+          "description" => "Ruby weevils",
+          "format" => "answer",
+          "link" => "/an-example-answer-rubylol",
+          "indexable_content" => "I like my rubby badger: he is tasty and delicious"
```
MatMoore: ummm....

lib/sitemap/sitemap_generator.rb
```diff
@@ -11,17 +11,36 @@ def self.sitemap_limit
   end
 
   def get_all_documents
+    search_config = SearchConfig.instance
+    search_body = {
+      query: {
+        bool: {
```
MatMoore: I think this bit could also go in the filter context, since you don't care about scoring.

https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-filter-context.html

MatMoore (approved):
Looks good, couple of comments inline

dwhenry (approved):

PR #935, opened 2017-09-25
Remove mocha

dwhenry:
Becuase now that we are on Rspec we don;t really need to use mocha as well.

I have tried to do as little as possible to the actual tests and just adjust the syntax as required, however, this was not always possible and I did rewrite a couple of small tests.

I have also adjusted the way SearchConfig mocking works as Rspec wouldn't allow stubs in the before suite.


lib/search_config.rb
```diff
@@ -8,12 +8,15 @@ class SearchConfig
...
     base_uri
     govuk_index_name
+    default_index_name
   ].each do |config_method|
     define_method config_method do
       elasticsearch.fetch(config_method)
     end
   end
 
+  attr_writer :elasticsearch # this is used by integration testing to override the default
```
MatMoore, line 18: I would move this to an optional param in the initialiser and always set the instance variable (i.e. if nil, call `config_for("elasticsearch")`)

Otherwise there's a possibility of the config effectively changing between two reads which should never happen.

MatMoore (approved):
Looks good, just a minor comment about the config.

lib/search_config.rb
dwhenry: As discussed this is a temporary (hopefully) thing that will be removed by making the config file environment aware. :)

PR #939, opened 2017-09-28
[DO NOT MERGE] Remove global stub

dwhenry:
I have achieved this by adding a global method `allow_elasticsearch_connection` which needs to be called at the top of any test files that require access to the ES instance (i.e. most integration test files).

I think this is an improvement, but am happy to abandon this change depending on how others feel.

I am open to suggestions if anyone has a better idea on how to achieve the removal of the global stub.

### DIFF
--- a/spec/integration/analytics_data_spec.rb
+++ b/spec/integration/analytics_data_spec.rb
@@ -1,6 +1,8 @@
 require 'spec_helper'
 
 RSpec.describe 'AnalyticsDataTest', tags: ['integration'] do
+  allow_elasticsearch_connection(scroll: true)
+
   before do
     @analytics_data_fetcher = AnalyticsData.new(SearchConfig.instance.base_uri, ["mainstream_test"])
   end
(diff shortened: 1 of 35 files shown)


MatMoore (approved):
I think this makes sense as it makes it more explicit that the tests are doing something weird.

Do you know if we still need the `disable_net_connect` at all? I might be missing something, but for integration tests, I'd generally expect all requests to be made as normal, especially those to elasticsearch, which is the integration we actually want to test.

https://github.com/alphagov/rummager/pull/938 ensures we're isolated to indexes managed by the tests; is there anything else to worry about?

dwhenry:
I am less sure of the approach than I was. I feel it adds a lot of noise to the integration tests without adding enough value. I would consider merging just the first commit and ignoring the second one as that makes what is happening more explicit, but again not 100% sure this is a true improvement.

MatMoore:
Shall we close this?

dwhenry:
@MatMoore my plan is to drop the second commit and tidy up what remains and then resubmit. HAppy to close the PR for now and I will open a new one when it's done - not sure when that will be as this is pretty low priority

MatMoore:
Cool, I'll close for now to avoid clutter but feel free to reopen.

PR #941, opened 2017-09-29
Surpress server errors when benchmarking search

MatMoore:
There is a jenkins job that runs this nightly, but it can fail if
search is not responding properly. This is best monitored by other checks,
and the benchmark results are not something we depend on, so we'd
prefer for this not to trigger a 2nd line alert.


dwhenry (approved):
Looks good. If we care about it running fully we could potentially add a retry facility in the future

PR #942, opened 2017-09-29
Remove minitest expectations

dwhenry:
* extract `.include?` to `include`
* extract `.<method>?` to `be_<method>`
* `.is_a?` to `be_a`
* remained to `be_truthy`
* replace assert_equal
* replace assert_nil
* replace assert_document_is_in_rummager -> expect_document_is_in_rummager
* replace custom test methods rename assert_* to expect_*
* replace assert_match
* replace assert_raises
* replace refute_*

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,7 +37,6 @@ group :test do
   gem 'bunny-mock', '~> 1.7'
   gem "ci_reporter", "~> 1.7.1"
   gem 'govuk_schemas', '~> 2.2.0'
-  gem "minitest", "~> 5.6.0"
   gem "rack-test", "~> 0.6.3"
   gem 'rspec'
   gem "simplecov", "~> 0.10.0"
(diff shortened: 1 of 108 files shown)


h-lame:
This is pretty amazing.  One minor nitpick is that in minitest/test-unit world we'd write `assert_equal <expected>, <actual>` but in rspec we'd write `expect(<actual>).to eq(<expected>)`.  The conversion here hasn't flipped those around so we have `expect(<expected>).to eq(<actual>)`) which might confuse people and make the failure messages look weird.  Is it easy for you to also flip those around?

dwhenry:
@h-lame I know it does annoys me a little, however, due to the number of conversions a lot of it was done with manual regular expressions and I lost the will to make everything perfect.

So the short answer is there is no easy way to fix this.

I will be encouraging the team review the tests in any file the edit and sort out the remaining problems like this during our normal work. Given we were initially going to have to manually convert each file as we worked on it I don't think they will mind.

dwhenry:
@h-lame If I do manage to complete and merge everything else today I may try to do some fixes for this, but don;t feel confident just due to the number of variants.

MatMoore (approved):

h-lame:
@dwhenry - fair enough.  It's a shame there's no robust tool that does it, like `transpec` for upgrading old rspec syntax.

PR #943, opened 2017-09-29
Revise README and search API docs

MatMoore:
Split the search API guide into sections so it's easier to find stuff.

This'll be rendered at https://docs.publishing.service.gov.uk/apis/search/search-api.html

I'll probably split the stuff about filtering/aggregating into a separate page later,
but for now I've just separated it into its own section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,25 @@
 # Rummager
 
-Rummager is the internal GOV.UK API for search.
+Rummager is the GOV.UK API for search.
 
 ## Live examples
 
-- [alphagov/frontend](https://github.com/alphagov/frontend) uses Rummager to
-	serve the GOV.UK search at [gov.uk/search](https://www.gov.uk/search).
-- [alphagov/finder-frontend](https://github.com/alphagov/finder-frontend) uses
-	Rummager to serve document finders like
-	[gov.uk/aaib-reports](https://www.gov.uk/aaib-reports).
-
-This API is publicly accessible:
-
+### The public search API
 https://www.gov.uk/api/search.json?q=taxes
 ![Screenshot of API Response](doc/api-screenshot.png)
 
-You can read how to use the API in the blog post: ["Use the search API to get useful information about GOV.UK content"](https://gdsdata.blog.gov.uk/2016/05/26/use-the-search-api-to-get-useful-information-about-gov-uk-content/).
+For the most up to date query syntax and API output see the [Search API documentation](https://docs.publishing.service.gov.uk/apis/search/search-api.html).
+
+You can also find some examples in the blog post: ["Use the search API to get useful information about GOV.UK content"](https://gdsdata.blog.gov.uk/2016/05/26/use-the-search-api-to-get-useful-information-about-gov-uk-content/).
+
+### GOV.UK site search
+
+[alphagov/frontend](https://github.com/alphagov/frontend) uses the search API
+to render search results.
+
+### Finders
+[alphagov/finder-frontend](https://github.com/alphagov/finder-frontend) uses
+the search API to serve document finders like [gov.uk/aaib-reports](https://www.gov.uk/aaib-reports).
 
 ## Technical documentation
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,21 +1,25 @@
...
+
+You can also find some examples in the blog post: ["Use the search API to get useful information about GOV.UK content"](https://gdsdata.blog.gov.uk/2016/05/26/use-the-search-api-to-get-useful-information-about-gov-uk-content/).
+
+### GOV.UK site search
+
+[alphagov/frontend](https://github.com/alphagov/frontend) uses the search API to render search results.
+
+### Finders
+[alphagov/finder-frontend](https://github.com/alphagov/finder-frontend) uses
+	the search API to serve document finders like
```
dwhenry: whitespace

doc/search-api.md
```diff
@@ -46,176 +59,190 @@ The parameters supported are:
...
 
-   - `example_scope`: `global` or `query`.  If the `examples` option is supplied, the
-     `example_scope` option must be supplied too.
+## Error reporting
 
-     The value of `global` causes the returned examples to be taken from all
-     documents in which the facet field has the given slug.
+The search API supports many query string parameters.  It validates
+parameters strictly - any unknown parameters, or parameters with invalid
+options, will cause an HTTP 422 error.  This makes it likely that typos do not
```
dwhenry: This means that typos do not..

doc/search-api.md
```diff
@@ -46,176 +59,190 @@ The parameters supported are:
...
+
+Only a few fields are returned by default. You can override the fields returned using the `fields` parameter. For example:
+
+```
+https://www.gov.uk/api/search.json?q=passport&fields=mainstream_browse_pages&fields=title
+```
+
+Note that query parameters which are repeated may be specified in standard HTTP
+style (ie, `name=value&name=value`, where the same name may be used multiple
+times), or in Ruby/PHP array style (ie, `name[]=value&name[]=value`).  The `[]`
```
dwhenry: I don't think the last sentence is needed/adds value

doc/search-api.md
```diff
@@ -46,176 +59,190 @@ The parameters supported are:
...
-   backwards compatibility, a fairly long set of fields is currently returned,
-   but it is good practice to set this to only the fields you actually want
-   information on (doing this will normally increase performance).
+### Filter parameters
+You can filter search results using any of the filter_<field name> or reject_<field name> query parameters.
 
- - ab_tests: a/b test with selected variant type. This allows test to be configured
-   from upstream apps.
+- filter is used to only returns documents that match a value
+- reject excludes documents that don't match a value
```
dwhenry: I think this sentence is wrong. 

reject excludes documents that match a value

doc/search-api.md
```diff
@@ -46,176 +59,190 @@ The parameters supported are:
...
+
+Only some fields can be filtered on - an HTTP 422 error will be returned if the requested field is not a value sort field.
+
+If a filter and a reject are specified for the same field, an HTTP 422 error will be returned. However, it is valid to specify a reject for some fields and a filter for others - documents will be required to match the criteria on both fields.
+
+### Aggregation parameters
+
+Aggregations look at all the values of a a field and count up the number of times each one appears in documents matching the search.
+
+You can include facets in any search by specifying one of the facet_<field name> query parameters.
```
dwhenry: do we want to mention that this name is the old style and that we will be gradually moving to use `aggregate_*` which is the new terminology?

especially considering that is what you used in your example.

It might also be worth mentioning the returned data has different keys facet/aggregate

doc/search-api.md
```diff
@@ -46,176 +59,190 @@ The parameters supported are:
...
+
+## Other parameters
+- ab_tests: a/b test with selected variant type. This allows test to be configured
+  from upstream apps.
+
+  Each a/b test name should be followed by a ':' and then the variant type to
+  be used. If multiple a/b test setting are being passed in they should be
+  comma separated.
+
+  No validation is done to ensure the a/b test name provided is current implemented.
```
dwhenry, line 247: Do we want to mention cache busting?

dwhenry (changes requested):
Looks pretty good, just a couple of small comments.

doc/search-api.md
MatMoore: Have added a short description at the end. There's probably a couple of other params missing but I'm just focusing on the main ones for now.

dwhenry (approved):

PR #944, opened 2017-09-29
Fix expect(A).to eq(B) ordering

dwhenry:
These where migrated from assert_equal with A and B in the 
wrong positions.

This has been correct for the following criteria:

The assertion is on a single line.

* A starts with “ or `
* A is a number
* A starts with [ or {
* A = expected

And B does not start with the same value.

This will not fix the ordering on all tests, however it will 
cover a majority of case and result in reduced manual effort 
later.


h-lame (approved):
Thanks!

PR #945, opened 2017-09-29
ADR for incremental popularity updates (rejected)

MatMoore:
This replaces https://github.com/alphagov/rummager/pull/927 and documents an approach we rejected for https://trello.com/c/saDruTlG/307-investigate-1-day-intermittent-rummager-timeouts

[Rendered](https://github.com/alphagov/rummager/blob/9c290f6b0819a1aa2e567c088a0e28a5d0abffb8/doc/arch/adr-005-incremental-popularity-updates.md)


doc/arch/adr-005-incremental-popularity-updates.md
```diff
@@ -0,0 +1,68 @@
...
+to update the entire search index every day - we could get away with 10% of the current update workload.
+
+![Changes in VC-14 field day to day](images/changes-in-vc-14-metric.png)
+
+## Decision
+
+We decided against making any technical changes to the popularity update process now. Instead we're going to reduce the number of sidekiq processes that index into
+elasticsearch at the same time, which reduces the impact on search performance.
+
+We don't think that the incremental popularity updates would work
```
dwhenry: This feels more like "learnings" than "decision". And hide the fact the first paragraph here gives the information on why this ADR was rejected and what it was rejected in favour of.

dwhenry (approved):
I think it would be better to break up decision into learning and decision, but I don;t feel super strongly about this.

PR #946, opened 2017-10-02
Split docs for filter/aggregation params

MatMoore:
These are a bit complicated if you're unfamiliar with the API and deserve their own page in the docs.

Will raise a corresponding govuk-developer-docs PR for this.

[Rendered markdown](https://github.com/alphagov/rummager/blob/49dc3848431cfd55f6e4d94b62470b1482ead033/doc/public-api/faceted-search.md)

Currently the first subheading doesn't render but I'm not sure why 🤔


suzannehamilton (approved):
LGTM. The subheader rendered correctly when I pasted it into a gist, so I have no idea what's going on there. 🤔

PR #948, opened 2017-10-03
Create all indices

dwhenry:
This fixes a bug introduced into the "publishing e2e tests" by the removal of the Indexer::BulkLoader
class. As they only required the creation of empty indices this seems like the best approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,16 +67,17 @@ rebuild the index nightly to incorporate the latest analytics.
 
 ### Setup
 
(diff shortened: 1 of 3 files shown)


cbaines:
Would be good to also keep the documentation in the README up to date [1].

1: https://github.com/alphagov/rummager#setup

dwhenry:
@cbaines I have update the readme as requested. I will also look to update the `docs` if required

thomasleese (approved):

PR #949, opened 2017-10-03
Retry on failure when multiple ES hosts

dwhenry:
This will allow us to move away from the ES proxy

### DIFF
--- a/lib/govuk_index/client.rb
+++ b/lib/govuk_index/client.rb
@@ -32,7 +32,8 @@ module GovukIndex
     def client
       @_client ||= Services.elasticsearch(
         hosts: search_config.base_uri,
-        timeout: @_options[:timeout] || TIMEOUT_SECONDS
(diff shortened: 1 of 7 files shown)


brenetic:
How come we no longer need to parse the uri's?

dwhenry:
@brenetic we never needed to parse the URI, the gem is happy to receive string values. We want to stop it so that the puppet change I made (reference above) can pass in multiple hosts instead of the localproxy. :)

MatMoore (commented):
Are we happy with the default of 3 tries?

This seems reasonable for indexing but not search requests - if a search request takes too long I think it should time out rather than retry, as this could deny service to other users if the service is overloaded.

dwhenry:
@MatMoore updated this so the retry only happens when modifying data, gets should fail immediately.

MatMoore (approved):

PR #953, opened 2017-10-10
Support date-times in date filters

suzannehamilton:
Allow documents to be filtered by dates and times, as well as just by date. The time in a filter parameter used to just be ignored.

Examples:
 - `from:2014-04-01`
 - `to:2015-12-31 14:05`
 - `from:2017-09-01 00:00:00,to:2017-09-15 23:59:59`

As well as making search filtering more flexible, this fixes a timezone bug in finders like Announcements (https://www.gov.uk/government/announcements) which allow users to specify publishing dates. Before, if you entered a 'to' date which falls during BST, the results would include anything publishing between midnight and 1am on the following day.

https://trello.com/c/lSbGhlav/349-make-rummager-handle-times-as-well-as-dates

Probably easiest to review each commit because there are few which just add some test cases or validation. The main change is in the final commit.

### DIFF
--- a/lib/parameter_parser/search_parameter_parser.rb
+++ b/lib/parameter_parser/search_parameter_parser.rb
@@ -270,7 +270,7 @@ private
     def parse_date(string)
       Date.parse(string)
     rescue
-      @errors << %{Invalid value "#{string}" for parameter "#{field_name}" (expected ISO8601 date}
+      @errors << %{Invalid value "#{string}" for parameter "#{field_name}" (expected ISO8601 date)}
       null_date
     end
 
--- a/lib/parameter_parser/search_parameter_parser.rb
+++ b/lib/parameter_parser/search_parameter_parser.rb
@@ -257,6 +257,7 @@ private
       values.map { |combined_from_and_to|
         dates_hash = combined_from_and_to.split(",").reduce({}) { |dates, param|
           key, date = param.split(":")
+          validate_date_key(key)
           dates.merge(key => parse_date(date))
         }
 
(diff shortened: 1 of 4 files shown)


lib/parameter_parser/search_parameter_parser.rb
```diff
@@ -267,13 +270,29 @@ def parse_dates(values)
       }
     end
 
-    def parse_date(string)
-      Date.parse(string)
+    def validate_date_key(key)
+      if !%w(from to).include?(key)
+        @errors << %{Invalid date filter parameter "#{key}:" (expected "from:" or "to:")}
```
brenetic, line 275: Do we want to allow the a result to be returned even if there are invalid keys at this point?

brenetic (approved):
One comment which might warrant a little discussion.

lib/parameter_parser/search_parameter_parser.rb
suzannehamilton: We talked about this in person, but for everyone else: if `@errors` is populated, then the response is a 422 error.

So adding this line does add extra validation that wasn't there before. I think it's more useful to return an error than silently ignore the invalid filter, and it's consistent with how we treat invalid date formats. **But does anyone have any concerns with us releasing this?**

Rosa-Fox (approved):
I have scoured through with pry and have tested out in browser and it seems to work fine. No validations displayed to user when the format is incorrect, as you mentioned earlier but something to think about for another card.

suzannehamilton:
We left this on staging over the weekend to check for any errors in the traffic replay. I can't see any `HTTPUnprocessableEntity` errors in the frontend apps which look relevant, so I think this is OK to merge.

PR #955, opened 2017-10-16
Don't double-count documents in aggregates

Rosa-Fox:
Some documents exist in both the mainstream and govuk indexes because we are currently migrating all
content to the govuk index. Search queries already look in one index or the other, based on the
document's format, which prevents duplicate search results. But aggregates (which are used to power
filters like the organisation filter on the search homepage) were not doing this, so the counts
displayed on filters were too high because documents were being counted twice.

This fix applies the same format filtering to aggregates that is already applied to search queries.

This commit also deletes some unit tests. They tested the exact Elasticsearch query, which made
them quite fragile: adding the new aggregate filter added new (and fairly complex) terms to the
query, which made it difficult to update the tests. The same functionality was being tested by
integration tests, so it seems safe to just delete the unit tests and rely on integration tests
for checking how rummager uses Elasticsearch.

https://trello.com/c/lg2ZZunc/359-incorrect-counts-in-site-search-organisation-filter

Mobbed with @suzannehamilton and @brenetic.


spec/integration/search/search_spec.rb
```diff
@@ -216,6 +216,34 @@
     expect(aggregates_without_filter["mainstream_browse_pages"]["options"].size).to eq(2)
   end
 
+  it "aggregate_counting_with_filter_on_a_different_field" do
```
dwhenry: Rspec test name don't require the underscore. 

I have purposefully left the underscore in the migrated tests as a marker they were migrated, with the plan to remove them as they are reviewed/updated. 🍰 

should maybe have mentioned that somewhere.....

spec/unit/query_components/aggregates_spec.rb
```diff
@@ -1,166 +0,0 @@
-require 'spec_helper'
```
dwhenry, line 1: Good deletion

spec/unit/query_components/aggregates_spec.rb
```diff
@@ -1,166 +0,0 @@
...
-            "filtered_aggregations" => {
-              missing: { field: "organisations" }
-            }
-          }
-        },
-      ).to eq(result)
-    end
-  end
-
-  context "search with aggregate and filter on same field" do
```
dwhenry, line 45: Can I confirm we have an integration test for this feature?

dwhenry (dismissed):
As a side note it might be worth either grouping the aggregate tests into a context or extracting them to a separate integration test file. 

We have more options in terms of keeping the test suite clean now that we are on rspec so believe we should be at least considering what tidy up we could do as we edit each test.

spec/integration/search/search_spec.rb
Rosa-Fox: We discussed this but there are so many tests with underscores in the name, it would look inconsistent.

spec/unit/query_components/aggregates_spec.rb
Rosa-Fox: https://github.com/alphagov/rummager/blob/de6d015ea376f3b410c842ce284d18360efddfea/spec/integration/search/search_spec.rb#L202

spec/integration/search/search_spec.rb
Rosa-Fox: We have now decided to move the aggregates tests into their own file, removing those underscores in the process.

Rosa-Fox:
Thanks @dwhenry. Will move the aggregates tests into a separate file.

spec/integration/search/aggregates_spec.rb
```diff
@@ -0,0 +1,252 @@
+require 'spec_helper'
+
+RSpec.describe 'search queries' do
+  context 'with aggregates' do
+    it "returns document count" do
+      populate_content_indexes(section_count: 2)
```
MatMoore: I'm not sure I'm following what the test is doing, without knowing what `populate_content_indexes` does. Why do we expect 2 documents in each browse page? Would the number change if we removed the number of indexes?

I think we could probably make the setup clearer by not using this method (for example, by making a method call for each document instead).

spec/unit/query_components/aggregates_spec.rb
dwhenry: 😄

spec/integration/search/aggregates_spec.rb
dwhenry: Have renamed the method to be clearer. let me know if you still prefer inlining.

MatMoore (approved):

PR #956, opened 2017-10-17
Publish Finder Frontend site search special route

brenetic:
We are adding site search to Finder Frontend. As Rummager already talks to the publishing API it seemed like a reasonable place to be registering the routes as opposed to making another frontend app talk to the publishing API.

https://trello.com/c/lxKbO5Xw

dependent on https://github.com/alphagov/frontend/pull/1294

### DIFF
--- a/lib/special_route_publisher.rb
+++ b/lib/special_route_publisher.rb
@@ -1,0 +1,64 @@
+require 'gds_api/publishing_api/special_route_publisher'
+
+class SpecialRoutePublisher
+  def initialize(publisher_options)
+    @publisher = GdsApi::PublishingApi::SpecialRoutePublisher.new(publisher_options)
+  end
+
+  def publish(route)
+    @publisher.publish(
+      route.merge(
+        publishing_app: "rummager",
+        format: "special_route",
+        public_updated_at: Time.now.iso8601,
+        update_type: "major",
+      )
+    )
+  end
+
+  def routes
+    [
+      {
+        rendering_app: "finder-frontend",
+        content_id: "84e0909c-f3e6-43ee-ba68-9e33213a3cdd",
+        base_path: "/search",
+        title: "GOV.UK search results",
+        description: "Sitewide search results are displayed here.",
+        document_type: "search",
+        type: "exact",
+      },
+      {
+        rendering_app: "finder-frontend",
+        content_id: "9f306cd5-1842-43e9-8408-2c13116f4717",
+        base_path: "/search.json",
+        title: "GOV.UK search results API",
+        description: "Sitewide search results are displayed in JSON format here.",
+        type: "exact",
+      },
+      {
+        rendering_app: "finder-frontend",
+        content_id: "ba750368-8001-4d01-bd57-cec589153fdd",
+        base_path: "/search/opensearch.xml",
+        title: "GOV.UK opensearch descriptor",
+        description: "Provides the location and format of our search URL to browsers etc.",
+        type: "exact",
+      },
+      {
+        rendering_app: "rummager",
+        base_path: "/sitemap.xml",
+        content_id: "fee32a90-397a-4761-9f98-b06e47d2b798",
+        title: "GOV.UK sitemap index",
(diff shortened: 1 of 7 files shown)


lib/special_route_publisher.rb
```diff
@@ -0,0 +1,58 @@
...
+        public_updated_at: Time.now.iso8601,
+        update_type: "major",
+      )
+    )
+  end
+
+  def routes
+    [
+      {
+        publishing_app: "finder-frontend",
```
suzannehamilton: Is the publishing app technically rummager? I'm not quite sure what "publishing app" means for a special route.

lib/special_route_publisher.rb
```diff
@@ -0,0 +1,58 @@
...
+  def initialize(publisher_options)
+    @publisher = GdsApi::PublishingApi::SpecialRoutePublisher.new(publisher_options)
+  end
+
+  def publish(route)
+    @publisher.publish(
+      route.merge(
+        format: "special_route",
+        public_updated_at: Time.now.iso8601,
+        update_type: "major",
```
suzannehamilton: Out of curiosity, do other special routes get published as a "major" update?

lib/tasks/publishing_api.rake
```diff
@@ -1,46 +1,14 @@
 namespace :publishing_api do
   desc "Publish special routes such as sitemaps"
   task :publish_special_routes do
```
suzannehamilton, line 3: Now that rummager's publishing more pages, it would be good to add some unit tests to check that the routes being published are valid against the content schemas. Here's how frontend does it: https://github.com/alphagov/frontend/blob/master/test/unit/special_route_publisher_test.rb

lib/special_route_publisher.rb
brenetic: They always seem to be major updates

MatMoore (dismissed):

dwhenry (dismissed):
As discussed, we don't want to merge this to master as we need to be certain the testing endpoints do get indexed

lib/special_route_publisher.rb
```diff
@@ -1,10 +1,25 @@
+require 'gds_api/publishing_api'
 require 'gds_api/publishing_api/special_route_publisher'
 
 class SpecialRoutePublisher
   def initialize(publisher_options)
     @publisher = GdsApi::PublishingApi::SpecialRoutePublisher.new(publisher_options)
   end
 
+  def take_ownership_of_search_routes
+    publishing_api_v1 = GdsApi::PublishingApi.new(
```
suzannehamilton: What does `v1` mean here? I'm a bit hazy on the publishing API's history, but I thought everything used v2 by default?

spec/unit/special_route_publisher_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def assert_valid_content_item(payload)
+    validator = GovukContentSchemaTestHelpers::Validator.new(
+      "special_route",
+      "schema",
+      payload
+    )
+
+    expect(validator.valid?).to be true
```
suzannehamilton, line 39: A [custom message](https://relishapp.com/rspec/rspec-expectations/docs/customized-message) would be useful here, otherwise it won't be clear from the test failure which special route has failed validation.

You could use `validator.errors` like frontend does: https://github.com/alphagov/frontend/blob/master/test/unit/special_route_publisher_test.rb#L40

MatMoore (approved):
LGTM. I assume that the rake task (including the part that takes ownership of the search routes) can be run multiple times without changing anything?

lib/special_route_publisher.rb
MatMoore: I'm also unfamiliar with the v1 API. Is this functionality unavailable in v2?

dwhenry:
@MatMoore yep it can all be run multiple times, and I will remove the take ownershipo code once we are live. I also have https://github.com/alphagov/rummager/pull/992 as a fallback in case we need to revert. :)

PR #957, opened 2017-10-19
Add contact groups field

brenetic:
- Adds `contact_groups` field to `govuk` index as part of migrating `contacts-admin`.
- Add `indexable_content` to `govuk` index

The contact format was not a simple mapped field. We needed to extract the titles
for each of the contacts in the `contact_groups` field. As these title
fields were at a deeper level inside `details` we could not use the string value
of the required field as we currently do with `licence` and `transaction`.

We therefore refactored indexable content into it's own presenter. This
should help us in the future when migrating other more complicated
formats.

https://trello.com/c/enaSEdX4 &&
https://trello.com/c/MWgAx6MX


MatMoore (approved):
This approach looks good 👍 

Can you do a quick pass over the tests that were moved to make them more idiomatic? There's a few still like `it "details_with_govspeak_and_text_html"` from when we switched to rspec.

lib/govuk_index/presenters/indexable_content_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+    def indexable
+      case format
+      when 'contact'
+        contact
+      else
+        default_indexable_content
+      end
+    end
+
+    def default_indexable_content
```
dwhenry: does this method add any value

spec/unit/govuk_index/presenters/indexable_content_presenter_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+          "title" => "Title 1"
+        },
+        {
+          "slug" => "slug-2",
+          "title" => "Title 2"
+        }
+      ]
+    }
+
+    expect(indexable_content_presenter(details, 'contact').indexable_content).to eq("Title\n\n\nDescription\n\n\nTitle 1\n\n\nTitle 2")
```
MatMoore: How come there's so many newlines between the parts? I don't think it matters for searching as it will get split into tokens regardless, but seems a bit strange.

lib/govuk_index/presenters/indexable_content_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+  private
+
+    attr_reader :details, :format, :sanitiser
+
+    def indexable
+      case format
+      when 'contact'
+        contact
+      else
+        default_indexable_content
```
dwhenry: why do we need the conditional? if we do need the conditional shouldn't it be extended to cover formats that can have hidden content?

could we not just make this 

```indexable_content_parts + hidden_content + content_group_titles```

that would keep the existing pattern.

**Review by dwhenry — APPROVED** [review]
Looks pretty good, just some queries around the need for the conditional.

File: lib/govuk_index/presenters/indexable_content_presenter.rb [hunk 4]
```diff
@@ -0,0 +1,54 @@
...
+      'transaction' => %w(introductory_paragraph more_information),
+    }.freeze
+
+    def initialize(format:, details:, sanitiser:)
+      @format    = format
+      @details   = details
+      @sanitiser = sanitiser
+    end
+
+    def indexable_content
```
dwhenry, line 16: naming wise this is not idea `IndexableContent#indexable_content`. We should revisit how we are naming both the classes and the methods, but lets do that in the next PR. :)

spec/unit/govuk_index/presenters/details_presenter_spec.rb
```diff
@@ -50,31 +14,13 @@
...
-  end
-
-  def details_presenter(details, indexable_content_keys = %w(body parts))
-    described_class.new(
-      details: details,
-      indexable_content_keys: indexable_content_keys,
-      sanitiser: GovukIndex::IndexableContentSanitiser.new
-    )
+  def details_presenter(details)
+    described_class.new(details: details)
```
dwhenry: this could easily be moved to a `subject` definition, or given that it is only used once, inlined.

spec/unit/govuk_index/presenters/indexable_content_presenter_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+        { "content_type" => "text/html", "content" => "<strong>introductory paragraph</strong>" }
+      ],
+      "more_information" => "more information",
+      "start_button_text" => "Start now",
+    }
+
+    expect(indexable_content_presenter(details, "transaction").indexable_content).to eq("introductory paragraph\n\nmore information")
+  end
+
+  def indexable_content_presenter(details, format = 'help_page')
```
dwhenry: I would move this to a `subject` definition as that is more inline with standard rspec styling

dwhenry:
couple of small rspec style comments (these will keep appearing as we touch migrated files for the first time) but otehrwise looks good to 🚢

dwhenry:
🚢  :shipit:   🥅

PR #959, opened 2017-10-23
[DO NOT MERGE] - Delete contact documents with old id format

brenetic:
The _id field for most content is the same as the link (including the leading `/`), but contacts have an _id without a leading slash.

The publishing to the new govuk index will now make each contact document have an _id with the leading `/` therefore we need to delete all the old documents.

This task can be removed once it has been run in production after the contact format documents have been republished to the new govuk index.

https://trello.com/c/dLnAxXzs

### DIFF
--- a/lib/tasks/delete.rake
+++ b/lib/tasks/delete.rake
@@ -102,4 +102,41 @@ namespace :delete do
       end
     end
   end
+
+  desc "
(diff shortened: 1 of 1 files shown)


brenetic:
Going about this a different way.

PR #960, opened 2017-10-24
Remove id check for contact format in duplicate deleter

brenetic:
The contact format has been moved to the govuk index and republished with the correct `_id` which includes the preceeding `/`. We no longer need to check for this in the duplicate deleter.

https://trello.com/c/kt6kFrjO


dwhenry (commented):
:ship: once we have fully migrated

suzannehamilton (approved):
🎉 Deleted code. 🎉

PR #963, opened 2017-10-25
Add asciifolding filter to analyzers

Rosa-Fox:
- [Trello](https://trello.com/c/7XlHiC7T/361-make-search-ignore-diacritics)
- Search currently returns different results depending on whether terms include diacritics. It should be the same regardless.
- This PR strips diacritics by applying asciifolding filters to analyzers.
- In order for these changes to be applied to the exisiting search data, we will need to reindex the `govuk`, `mainstream`, `government` and `detailed` indexes.

Paired with @suzannehamilton

### DIFF
--- a/config/schema/elasticsearch_schema.yml
+++ b/config/schema/elasticsearch_schema.yml
@@ -5,13 +5,13 @@ index:
         default:
           type: custom
           tokenizer: standard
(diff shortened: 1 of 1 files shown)


brenetic (approved):
🚀  LGTM. FFTM, when the schemas test stuff is fixed

PR #964, opened 2017-10-25
Add A/B test of specialist sector (aka topic) boosting

brenetic:
We have previously run A/B tests which showed that boosting mainstream
content above specialist content helped more users to find the
information they were looking for. This new A/B test iterates on that by
boosting topics in search results, which we belive are useful navigation
pages for users to find in search.

The boost value of 1.5 is just an initial value, chosen to be similar to other mainstream boosts. We'll refine this based on how it performs in the healthcheck.

https://trello.com/c/cmxfgx1H/391-configure-format-weighting-a-b-test-in-rummager

Mobbed with @JonathanHallam and @suzannehamilton

### DIFF
--- a/lib/search/query_components/booster.rb
+++ b/lib/search/query_components/booster.rb
@@ -21,7 +21,16 @@ module QueryComponents
   private
 
     def boost_filters
-      property_boosts + [time_boost]
+      boosts = property_boosts + [time_boost]
+      boosts << specialist_sector_boost if @search_params.format_weighting_b_variant?
+      boosts
+    end
+
+    def specialist_sector_boost
+      {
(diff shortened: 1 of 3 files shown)


MatMoore (approved):
👍

suzannehamilton:
I've just changed the boost from 1.5 to 2.5. We compared a few values in the search healthcheck, and 2.5 seems like a good compromise between promoting topics and not pushing other results too far down.

Full comparison here: https://docs.google.com/a/digital.cabinet-office.gov.uk/spreadsheets/d/1H3ihfMngJHb3Ens_5b-lgMAMv_nHDu4y1K2xsW2zDBE/edit?usp=sharing

PR #967, opened 2017-10-27
Remove incorrect deprecation warning from rake task

suzannehamilton:
Remove deprecation warning from the `indices:migrate_schema` rake task. It was accidentally copied from the old `migrate_index` task which was deprecated and then deleted.

cc @dwhenry


brenetic (approved):
🚀

PR #978, opened 2017-11-02
Don't notify of missing basepath for contacts and world locations

dwhenry:
This formats don't expect this field to be populated so these notifications
are just noise.

### DIFF
--- a/lib/govuk_index/publishing_event_worker.rb
+++ b/lib/govuk_index/publishing_event_worker.rb
@@ -6,6 +6,8 @@ module GovukIndex
   class UnknownDocumentTypeError < StandardError; end
   class ValidationError < StandardError; end
 
+  DOCUMENT_TYPES_WITHOUT_BASE_PATH = %w(contact world_location).freeze
+
   class PublishingEventWorker < Indexer::BaseWorker
     notify_of_failures
 
@@ -16,6 +18,9 @@ module GovukIndex
       process_response(response) if response
     # Rescuing as we don't want to retry this class of error
     rescue ValidationError => e
+      return if e.message ==  "base_path missing from payload" &&
+          DOCUMENT_TYPES_WITHOUT_BASE_PATH.include?(payload["document_type"])
+
       GovukError.notify(
         e,
         extra: {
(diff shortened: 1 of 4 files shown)


lib/govuk_index/publishing_event_worker.rb
```diff
@@ -16,6 +18,9 @@ def perform(routing_key, payload)
       process_response(response) if response
     # Rescuing as we don't want to retry this class of error
     rescue ValidationError => e
+      return if e.message ==  "base_path missing from payload" &&
```
MatMoore: Can we subclass ValidationError instead of checking strings here?

MatMoore (changes requested):

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -6,6 +6,8 @@ class NotFoundError < StandardError; end
   class UnknownDocumentTypeError < StandardError; end
   class ValidationError < StandardError; end
 
+  DOCUMENT_TYPES_WITHOUT_BASE_PATH = %w(contact world_location).freeze
```
tijmenb: This duplicates knowledge from the content schemas. Maybe the validation here should be swapped for a schema validation?

MatMoore: I agree with @dwhenry here.

These are valid content items but anything without a base path (or some other way of generating a link) should be excluded from search. We can't rely only on content schemas because our data model is different.

tijmenb: Yeah I agree, I probably meant it should be done by checking the presence of a `base_path`, so you don't duplicate information in the schema.

lib/govuk_index/publishing_event_worker.rb
dwhenry: Done, I actually replaced the error class as we don't currently have any other types of validation errors and I doubt we will at any point soon.

lib/govuk_index/publishing_event_worker.rb
tijmenb: (but don't let this hold you off from merging, I'm just a tourist on this repo)

MatMoore (approved):

PR #981, opened 2017-11-03
Add core boost to A/B test in rummager

brenetic:
https://trello.com/c/1t0kDavf


suzannehamilton:
I've fixed the failing test by adding the supertype to the test setup, because this is what rummager uses at query time to boost the query.

The boost of 2.5 is just a starting point: we'll tune it using the benchmarking tool before we start the A/B test.

dwhenry (approved):
All looks to make sense

PR #986, opened 2017-11-07
Fix encoding bug

dwhenry:
Paired with @JonathanHallam 

https://trello.com/c/hVCAwMKq/399-specialist-publisher-employmenttribunaldecision

### DIFF
--- a/spec/unit/govuk_index/indexable_content_sanitiser_spec.rb
+++ b/spec/unit/govuk_index/indexable_content_sanitiser_spec.rb
@@ -1,70 +1,72 @@
 require 'spec_helper'
 
 RSpec.describe GovukIndex::IndexableContentSanitiser do
-  it "strips_html_tags_from_indexable_content" do
-    payload = [
-      [
-        { "content_type" => "text/govspeak", "content" => "**hello**" },
-        { "content_type" => "text/html", "content" => "<strong>hello</strong> <a href='www.gov.uk'>marmaduke</a>" }
-      ]
-    ]
-
-    expect(subject.clean(payload)).to eq("hello marmaduke")
-  end
-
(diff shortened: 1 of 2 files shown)


spec/unit/govuk_index/indexable_content_sanitiser_spec.rb
```diff
@@ -1,70 +1,72 @@
...
+        [
+          { "content_type" => "text/govspeak", "content" => "**hello**" },
+          { "content_type" => "text/html", "content" => "<strong>hello</strong> <a href='www.gov.uk'>marmaduke</a>" }
+        ]
+      ]
+
+      expect(subject.clean(payload)).to eq("hello marmaduke")
+    end
+
+    it "correctly processes multiple payload items" do
```
MatMoore: "correctly" makes it hard to tell if the test is looking for the right thing. Can we rename it to describe the behaviour better? Like "it joins multiple payload items with newlines"?

spec/unit/govuk_index/indexable_content_sanitiser_spec.rb
```diff
@@ -80,6 +82,32 @@
 
     expected_content = "title 1\n\nhello\n\ntitle 2\n\ngoodbye"
 
-    expect(expected_content).to eq(subject.clean(payload))
+    expect(subject.clean(payload)).to eq(expected_content)
+  end
+
+  context "Special character encoding" do
```
MatMoore: "Special character encoding" makes it sound like this it could be a character encoding thing, but it's specifically HTML encoding we're opting out of.

MatMoore (approved):
Looks good to me, just a couple of comments about the tests

PR #990, opened 2017-11-08
ESI Fund

dwhenry:
This required a small change to ensure that we get the expected/matching format for this document type in elasticsearch

https://trello.com/c/nGnqfSwv/404-specialist-publisher-europeanstructuralinvestmentfund

### DIFF
--- a/lib/govuk_index/presenters/common_fields_presenter.rb
+++ b/lib/govuk_index/presenters/common_fields_presenter.rb
@@ -1,12 +1,14 @@
 module GovukIndex
   class CommonFieldsPresenter
+    CUSTOM_FORMAT_MAP = {
+      "esi_fund" => "european_structural_investment_fund",
+    }.freeze
     extend MethodBuilder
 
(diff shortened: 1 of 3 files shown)


lib/govuk_index/presenters/common_fields_presenter.rb
```diff
@@ -1,12 +1,14 @@
 module GovukIndex
   class CommonFieldsPresenter
+    CUSTOM_FORMAT_MAP = {
```
MatMoore, line 3: So is this needed because specialist pub is currently sending
european_structural_investment_fund as the content_store_document_type to search, and pub api is using esi_fund? Or have I got that mixed up?

dwhenry: That is correct. I would suggest we should investigate if we can remove this in the future, but have chosen to leave it as it is for the time being

suzannehamilton (approved):

PR #993, opened 2017-11-09
Add a task for debugging synonyms

MatMoore:
This prints out how text is being analyzed, and shows example content that
matches the query with the synonym rules applied (ignoring all other factors
that cause results to be boosted).

I've been using this to figure out what our synonyms are doing.

If this proves useful I'll probably move it into search admin.

Example output:

## Old synonyms
screen shot 2017-11-09 at 17 38 50

## New synonyms
screen shot 2017-11-09 at 17 39 18

### DIFF
--- a/lib/debug/synonyms.rb
+++ b/lib/debug/synonyms.rb
@@ -9,25 +9,26 @@ module Debug
         @index = index
       end
 
-      def search(query)
+      def search(query, pre_tags:, post_tags:)
         search_query = {
           query: {
             multi_match: {
               "query" => query,
-              "fields" => %w(title^1000 description)
+              "fields" => %w(title^1000 description),
+              "analyzer" => 'query_with_old_synonyms',
             }
           },
           highlight: {
             "fields" => { "title" => {}, "description" => {} },
-            "pre_tags" => ["\e[32m"],
-            "post_tags" => ["\e[0m"]
+            "pre_tags" => pre_tags,
+            "post_tags" => post_tags
           }
         }
 
-        client.search(index: index, analyzer: 'query_with_old_synonyms', body: search_query)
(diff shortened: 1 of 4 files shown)


lib/debug/synonyms.rb
```diff
@@ -0,0 +1,69 @@
...
+        search_query = {
+          query: {
+            multi_match: {
+              "query" => query,
+              "fields" => %w(title^1000 description)
+            }
+          },
+          highlight: {
+            "fields" => { "title" => {}, "description" => {} },
+            "pre_tags" => ["\e[32m"],
```
dwhenry: would prefer to pass this value in as a param or have it as a constant - same for the post_tags

lib/debug/synonyms.rb
```diff
@@ -0,0 +1,69 @@
...
+          }
+        }
+
+        client.search(index: index, analyzer: 'with_search_synonyms', body: search_query)
+      end
+
+      def analyze(query)
+        index_tokens = client.indices.analyze text: query, analyzer: 'with_index_synonyms', index: index
+        search_tokens = client.indices.analyze text: query, analyzer: 'with_search_synonyms', index: index
+        [index_tokens, search_tokens]
```
dwhenry: would prefer this method to have the same output format for both models, with either a separate method or param to retreive the index_tokens

dwhenry (approved):
Just a couple of suggestions.

It does feel a little like we could remove some of the duplication with only a small amount of effort, but that depends on what our long term plan is. :)

MatMoore:
Am going to leave this with some duplication for now since I hope to be able to remove one of them once we have a single implementation in Rummager itself.

PR #996, opened 2017-11-10
Fix bug when index is locked that resulted in corrupt data

dwhenry:
The data missing retried when an index was locked was missing the id and type
fields and as a result would never be add/updated in the index.

Paired with @brenetic 

https://trello.com/c/xkPc6dZg/423-report-errors-to-the-user-in-search-admin-when-best-bet-updates-have-failed-timebox-1-day

### DIFF
--- a/spec/integration/indexer/indexing_spec.rb
+++ b/spec/integration/indexer/indexing_spec.rb
@@ -125,4 +125,41 @@ RSpec.describe 'ElasticsearchIndexingTest' do
     expect(last_response.status).to eq(202)
     expect_document_is_in_rummager(SAMPLE_DOCUMENT)
   end
+
+  it "rescheduling during locking does not corrupt the process" do
+    stubbed_client = client
+
+    locked_response = { "items" => [
+      { "index" => { "error" => { "reason" => "[FORBIDDEN/metasearch/index write" } } }
+    ] }
+
+    expect(stubbed_client).to receive(:bulk).and_return(locked_response)
+    expect(stubbed_client).to receive(:bulk).and_call_original
+    allow_any_instance_of(SearchIndices::Index).to receive(:build_client).and_return(stubbed_client)
+
+    details = <<~DETAILS
+      {\"best_bets\":[
+        {\"link\":\"/learn-to-drive-a-car\",\"position\":1},
+        {\"link\":\"/learn-to-drive-a-car\",\"position\":3},
+        {\"link\":\"/learn-to-drive-a-car\",\"position\":10}
+      ],\"worst_bets\":[]}", "stemmed_query_as_term"=>" learn to drive "}]
+    DETAILS
+
+    post "/metasearch_test/documents", {
+      "_id" => "learn+to+drive-exact",
+      "_type" => "best_bet",
+      "stemmed_query" => "learn to drive",
+      "details" => details
+    }.to_json
+
+    expect_document_is_in_rummager(
+      {
+        "stemmed_query" => "learn to drive",
+        "details" => details,
+      },
+      index: "metasearch_test",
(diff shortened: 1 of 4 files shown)


spec/integration/indexer/indexing_spec.rb
```diff
@@ -125,4 +125,41 @@
...
+
+    details = <<~DETAILS
+      {\"best_bets\":[
+        {\"link\":\"/learn-to-drive-a-car\",\"position\":1},
+        {\"link\":\"/learn-to-drive-a-car\",\"position\":3},
+        {\"link\":\"/learn-to-drive-a-car\",\"position\":10}
+      ],\"worst_bets\":[]}", "stemmed_query_as_term"=>" learn to drive "}]
+    DETAILS
+
+    post "/metasearch_test/documents", {
```
suzannehamilton: I wonder if it would be worth adding a similar test for the content indexing endpoint? I know the process is (currently) the same under the hood, but it would be good to have more confidence that updating a document isn't affected by locking, because that's even more important than best bets.

dwhenry: its the same piece of code so that "should" be covered by this test..

As a side note we did corrupt some data on friday due to index locking.. :(

spec/integration/indexer/indexing_spec.rb
```diff
@@ -125,4 +125,41 @@
...
   end
+
+  it "rescheduling during locking does not corrupt the process" do
+    stubbed_client = client
+
+    locked_response = { "items" => [
+      { "index" => { "error" => { "reason" => "[FORBIDDEN/metasearch/index write" } } }
+    ] }
+
+    expect(stubbed_client).to receive(:bulk).and_return(locked_response)
```
suzannehamilton: Did you consider any alternatives to this stubbing? Is it just difficult to write the test without stubbing because the Sidekiq retries are unpredictable?

I'm just worried that it makes the integration test more fragile. e.g. what happen if we upgrade Elasticsearch and the lock response changes?

dwhenry: it might be possible but not sure how as in test mode sidekiq jobs run immediately. I guess we could lock the index run the test and have a separate thread that unlock the index after a small timeout?? but that is more complex

spec/integration/indexer/indexing_spec.rb
suzannehamilton: Yeah, I know it's the same piece of code, and you shouldn't (and can't) test lots of combinations of things in integration tests. But the implementation can change, and I'm just wondering if it's worth adding a second test here because it's really important that the _content_ indices aren't affected by locking, not just the metasearch index?

Is that data corrupted or just stale? Can we fix it with the Gor traffic replay?

dwhenry: A test for content has been added. This has shown that the issue only occurs for bets-bet updates as everything else already creates a copy of the hash before the `_id` and `_type` delete

spec/integration/indexer/indexing_spec.rb
```diff
@@ -125,4 +125,90 @@
...
+        },
+        index: "metasearch_test",
+        type: "best_bet",
+        id: "learn+to+drive-exact",
+      )
+    end
+  end
+
+  context "when processing document POST" do
+    it "rescheduling during locking does not corrupt the process" do
```
suzannehamilton: Tiny nitpick: now that we're using rspec's `it "does a thing"` syntax, does it make sense to fit our tests into rspec's style for new tests?

suzannehamilton (approved):

suzannehamilton (approved):

PR #997, opened 2017-11-13
Map HMRC manual fields

suzannehamilton:
Add HMRC manuals and manual sections to the fields saved to the govuk index.

They are not marked as indexable yet: we need to check the `indexable_content` field first.

https://trello.com/c/y5AmkYeF/428-add-new-hmrc-manual-fields


brenetic (approved):
LGTM 🚀

spec/unit/govuk_index/presenters/hmrc_manual_presenter_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'spec_helper'
+
+RSpec.describe GovukIndex::HmrcManualPresenter do
```
dwhenry: I feel like this should belong in the `DetailsPresenter` so it is populated in a standard way rather than have a format specific presenter, something we have tried to avoid.

suzannehamilton: Ah, I thought were trying to move towards presenters that were focused around content rather than which part of the message they came from? Happy to move it to the DetailsPresenter if that's not right, though.

@brenetic Do you have any thoughts on this? I know you were looking at refactoring it a few weeks ago.

suzannehamilton: After talking to Brendan, I've moved this code into the DetailsPresenter. 👍 

We're not entirely sure how these classes should be structured, so let's leave things simple (if verbose) for now. We'll revisit these decisions when we start importing whitehall content, since it might completely change our data model.

brenetic:
🚀 Coolio! Bring on Whitehall 😱

PR #998, opened 2017-11-13
Check indexes all exist and report sizes to statsd

MatMoore:
I'm not sure if there's a better way of implementing this, but I want to periodically query the search indexes so that I can send data to statsd.

Using this data we can set up monitoring to alert when indexes suddenly increase or decrease in size.

I was thinking of creating a jenkins job for this but open to other ideas.

Trello: https://trello.com/c/XvzOqrHB/432-trigger-an-alert-if-there-is-high-growth-reduction-in-index-size

### DIFF
--- a/lib/tasks/indices.rake
+++ b/lib/tasks/indices.rake
@@ -173,4 +173,31 @@ You should run this task if the index schema has changed.
       index_name: args.index_name
     )
   end
+
+  desc "Monitor the search indices and send data to statsd"
+  task :monitor_indices do
+    client = Services.elasticsearch
+    statsd = Services.statsd_client
+    missing = []
+
+    SearchConfig.instance.all_index_names.each do |index|
+      begin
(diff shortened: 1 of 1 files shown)


lib/tasks/indices.rake
```diff
@@ -173,4 +173,31 @@ You should run this task if the index schema has changed.
...
+    statsd = Services.statsd_client
+    missing = []
+
+    SearchConfig.instance.all_index_names.each do |index|
+      begin
+        stats = client.indices.stats index: index, docs: true
+        docs = stats["_all"]["primaries"]["docs"]
+
+        count = docs["count"]
+        statsd.gauge("#{index}_index.docs.count", count)
```
suzannehamilton, line 189: Is there any value in just dumping all the index stats into statsd rather than extracting a couple of individual values?

I don't have any particular use in mind, but it might be useful for debugging in future.

suzannehamilton (approved):

PR #1003, opened 2017-11-16
Mark HMRC Manuals as indexable in govuk index

brenetic:
https://trello.com/c/64Nv2XP8


suzannehamilton (approved):
Good to merge when we've compared the data on integration. 👍

PR #1004, opened 2017-11-16
Log the type of anything deleted

MatMoore:
Some best bets aren't being deleted properly and I expect the type/id
isn't being passed correctly.

Also remove some temporary code from an old deploy.


brenetic (approved):
🚀

PR #1005, opened 2017-11-16
Prepend HMRC Manual Section ID to title

brenetic:
Manual section titles need to be prepended by their section ids when being added to the elasticsearch index.

Doing this is in the common fields presenter is not ideal. As we migrate more formats we are finding more and more instances of format specific foibles in our payload part approach to presenters.

https://trello.com/c/64Nv2XP8

### DIFF
--- a/lib/govuk_index/presenters/common_fields_presenter.rb
+++ b/lib/govuk_index/presenters/common_fields_presenter.rb
@@ -16,11 +16,14 @@ module GovukIndex
     delegate_to_payload :publishing_app
     delegate_to_payload :rendering_app
     delegate_to_payload :search_user_need_document_supertype
-    delegate_to_payload :title
     delegate_to_payload :user_journey_document_supertype
 
     def initialize(payload)
       @payload = payload
+    end
+
+    def title
+      [section_id, payload["title"]].compact.join(" - ")
     end
 
     def is_withdrawn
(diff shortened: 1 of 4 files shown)


lib/govuk_index/presenters/common_fields_presenter.rb
```diff
@@ -16,13 +16,18 @@ class CommonFieldsPresenter
...
-    delegate_to_payload :title
     delegate_to_payload :user_journey_document_supertype
 
     def initialize(payload)
       @payload = payload
     end
 
+    def title
+      payload_title = payload["title"]
+      payload_title.prepend("#{section_id} - ") if format == "hmrc_manual_section"
```
suzannehamilton: Is `section_id` ever missing for this format?

dwhenry: Dose this need to be format specific?

Can't we just append the `section_id` if it is present - which currently only happens for hmrc_manual_sections?

```
[section_id, payload_title].compact.join(' - ')
```

brenetic: Needs to be prepended. HMRC Manual Sections need them but Manual Sections don't seem to. We don't yet know for other formats which have sections.

MatMoore (approved):
I think this is fine for now, but maybe we should do a spike into refactoring the presenters?

I'm thinking that subclassing this would be cleaner than entirely separating the formats. There is definitely a standard way to handle some of the fields, but a lot of format specific stuff as well.

Another thing to keep in mind is that elasticsearch document types are going away in version 6, which will leave us with a single elasticsearch mapping with a huge list of fields. At that time we might want to consider merging any that serve similar purposes, and standardising the data a bit more.

PR #1008, opened 2017-11-20
Make Manual and Manual Sections indexable in govuk index

brenetic:
Manuals do not seem to care about change notes or section ids in the way that HMRC manuals do.

Indexable content is also just the description.

https://trello.com/c/mYeuXqjv

### DIFF
--- a/lib/govuk_index/presenters/details_presenter.rb
+++ b/lib/govuk_index/presenters/details_presenter.rb
@@ -19,11 +19,13 @@ module GovukIndex
     def latest_change_note
       return nil if details["change_notes"].nil? || details["change_notes"].empty?
 
-      note_info = details["change_notes"]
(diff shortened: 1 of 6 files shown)


lib/govuk_index/presenters/details_presenter.rb
```diff
@@ -19,11 +19,13 @@ def contact_groups
     def latest_change_note
       return nil if details["change_notes"].nil? || details["change_notes"].empty?
 
-      note_info = details["change_notes"]
-        .sort_by { |note| DateTime.parse(note["published_at"]) }
-        .last
+      unless format == "manual"
```
suzannehamilton: Would it be safer to lock this down to the only format that uses change notes? I think it's just HMRC manuals.

suzannehamilton (approved):

PR #1009, opened 2017-11-20
Allow new synonyms to be defined separately at index and search time

suzannehamilton:
Replace logic for synonym codes like `!S123` with more explicitly-defined synonyms that can be applied at index time or search time or both.
    
This gives us more flexibility over how different types of synonyms work. For example, typos should just be fixed at query time, but true synonyms like "abroad" and "overseas" should be merged into one term at both search and query time.
    
"New" synonyms are not being used in production yet, so this change won't have an immediate effect on search, but it will be used in an upcoming A/B test.

https://trello.com/c/NbRsaAD3/440-split-definitions-of-index-and-query-synonyms

Paired with @Rosa-Fox. @JonathanHallam mobbed with us on the validation commit.

Marked as DO NOT MERGE until we've moved the full list of new synonyms to the index/search/both format.


suzannehamilton:
Perhaps we should convert the validation to a test?

We were thinking of it as runtime validation because we were writing a parser, but the synonyms are checked into this repo so we can just validate them in CI.

brenetic (approved):
🚀 

If the overhead of runtime validation has little effect on performance then I think it's fine to leave it as it is.

suzannehamilton:
Cheers. The validation is run several times by the tests, and we didn't see any noticeable slowdown.

In production, the validation will only run when an index is created.

suzannehamilton:
I've imported the synonyms list, so this is now ready for a final review.

The synonyms list is complete apart from tweaks based on content feedback and the results of the healthcheck.

Rosa-Fox (approved):

PR #1010, opened 2017-11-22
Configure synonym A/B test

Rosa-Fox:
[Trello](https://trello.com/c/PqwdCw1d/447-configure-synonym-a-b-query-in-rummager)
- Change the field and query time analyzer for B variant

- 'Do not merge' as we need to test with the health check and Search Performance Explorer. Currently awaiting new synonym data.

Paired with @JonathanHallam and @suzannehamilton


lib/search/query_components/core_query.rb
```diff
@@ -118,10 +115,10 @@ def match_bigrams(fields)
   private
 
     def query_analyzer
```
suzannehamilton: I think this will do the right thing in the main case that we're interested in (i.e. the call to `minimum_should_match` from line 67 of `QueryBuilder`). 👍 

It might be using the new analyzer too much, though. There are several other calls to `minimum_should_match` and `query_analyzer` that use different fields like `title` and I don't think we want to make them all use the new analyzer.

Maybe we could have a B alternative to the `mininimum_should_match` method, and call that from the `if` statement in `query_builder`? Give me a shout if that's not clear. 🙂

brenetic (approved):
🚀

PR #1015, opened 2017-11-27
[REVIEW ONLY] Run all search integration tests with A and B variants

suzannehamilton:
This is just an idea. Feedback welcome on whether it's a _good_ idea, and also on how to make it easier to configure in RSpec.

Most of our A/B tests only affect a small part of the search functionality. For example, changing format weighting or synonyms shouldn't affect best bets or quoted search terms.

This means that most of the existing integration tests are applicable to both the A and the B variant, but we normally don't run them against B. Being able to run them for both would give us more confidence that we haven't caused any regressions in B. We worried about this we configured the synonyms A/B test because changing the analyzers is quite fiddly.

This change runs the integration tests which hit the `/search` endpoint twice: once for each variant. It does it by wrapping the test definitions in a block, which makes it easy to define separate tests which really do apply to just A or B.

Things that don't work so well (but might be fixable with some RSpec magic):
- The diff is enormous because the change indents all the tests. This would happen each time you set up or remove an A/B test.
- You have to change `get` to `get_with_variant` in each test, otherwise all the tests will silently use the A variant.

https://trello.com/c/PqwdCw1d/447-configure-synonym-a-b-query-in-rummager

Paired with @Rosa-Fox on the initial idea.

### DIFF
--- a/spec/integration/search/best_bets_spec.rb
+++ b/spec/integration/search/best_bets_spec.rb
@@ -3,12 +3,14 @@ require 'spec_helper'
 RSpec.describe 'BestBetsTest' do
   with_ab_variants do
     it "exact_best_bet" do
-      commit_document("mainstream_test",
+      commit_document(
+        "mainstream_test",
         "link" => '/an-organic-result',
         "indexable_content" => 'I will turn up in searches for "a forced best bet"',
       )
 
-      commit_document("mainstream_test",
+      commit_document(
+        "mainstream_test",
         "link" => '/the-link-that-should-surface',
         "indexable_content" => 'Empty.',
       )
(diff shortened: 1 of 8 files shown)


dwhenry:
I have looked at these sort of approaches in the past and as a general rule they are difficult to maintain. 

What I have seen work better is a manual approach that allows running of the test suite with a passed in variable. SO rather than the test suite running each test for A and B, it is run twice, once as per normal (A variant) and once passing the override in (B variant).

This would hopefully avoid the need to update the test suite when we change/add/remove variants.

suzannehamilton:
Closing this because we're not planning to run any more A/B tests for a while.

PR #1016, opened 2017-11-27
Delete healthcheck output file

suzannehamilton:
This file should not have been committed to the repo.

Also add it to gitignore so it won't be added again.


brenetic (approved):
🚀

PR #1018, opened 2017-11-27
Map collections fields for govuk index

Rosa-Fox:
[Trello](https://trello.com/c/8oQDr7ca/458-configure-rummager-for-moving-collections-to-govuk-index)

- sets up the ability to migrate the collections formats (`mainstream_browse_page` and `specialist_sector`) to `govuk` index.
-  Similar to mainstream but `slug` is generated from the `base_path`.

### DIFF
--- a/config/govuk_index/migrated_formats.yaml
+++ b/config/govuk_index/migrated_formats.yaml
@@ -40,4 +40,8 @@ migrated:
 - task_list
 
 indexable:
+- mainstream_browse_page
+- service_manual_guide
+- service_manual_topic
+- topic
 - travel_advice
--- a/config/govuk_index/migrated_formats.yaml
+++ b/config/govuk_index/migrated_formats.yaml
@@ -43,5 +43,5 @@ indexable:
 - mainstream_browse_page
 - service_manual_guide
 - service_manual_topic
-- topic
+- specialist_sector
 - travel_advice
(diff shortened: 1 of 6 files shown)


spec/integration/govuk_index/collections_spec.rb
```diff
@@ -32,6 +33,7 @@
     expected_document = {
        "link" => random_example["base_path"],
        "indexable_content" => "Mainstream browse page description",
+       "slug" => random_example["base_path"].gsub(%r{^/browse/}, ''),
```
suzannehamilton: Could this be hard-coded to "benefits"? It's normally better to use explicit values in tests so that you don't accidentally add the same bug to the app code and the test code.

Ditto for the test below.

config/govuk_index/migrated_formats.yaml
```diff
@@ -38,3 +38,5 @@ migrated:
 indexable:
 - service_manual_guide
 - service_manual_topic
+- topic
+- mainstream_browse_page
```
suzannehamilton: This should be alphabetical or @brenetic will be sad.

brenetic (changes requested):
Can you add a comment in the `mapped_document_types` for both types please: `# Collections Publisher`

brenetic (approved):

PR #1020, opened 2017-11-27
Check first 20 results in the health check

suzannehamilton:
Increase the search result count from the default (10) to 20 when running the healthcheck.

This will make it easier to judge the healthcheck results because users will see 20 results when the search on GOV.UK, so if a result is not found by the healthcheck then it will not be in the first page of results for a user.

https://trello.com/c/9VF3Rw5V/451-test-synonyms-with-healthcheck-and-search-performance-explorer


dwhenry (approved):
🚀

PR #1021, opened 2017-11-27
Reinstate flag for disabling synonyms

suzannehamilton:
The ability to debug search by disabling synonyms was removed in c611aff (PR #1010) because we thought it was part of an old synonym variation. This commit re-adds it so we can keep using it for debugging.

https://trello.com/c/PqwdCw1d/447-configure-synonym-a-b-query-in-rummager


dwhenry (approved):
🚀

PR #1022, opened 2017-11-27
Remove non breaking spaces from markdown files

cbaines:
These break markdown tooling, e.g. the highlighting in Emacs and the
processing done by the govuk-developer-docs.


suzannehamilton (approved):
So that's what causes this problem! 🎉 🎉 🎉

(I'm still wondering what text editor added the wrong kind of space, though. 🤔)

PR #1025, opened 2017-11-28
Fix flaky unpublishing test

suzannehamilton:
Ensure unpublishing test does not generate examples with the publishing app `smartanswers`, since these are currently excluded from the govuk index.

This workaround is already used elsewhere in the tests, and will be removed when we migrate smart answers to the govuk index.


brenetic (approved):
🚀

PR #1027, opened 2017-11-28
Add service manual custom format type

brenetic:
Both the service manual homepage and the service standard use `service_manual_guide` as their custom format type in elasticsearch.

https://trello.com/c/gJlxVo4x


suzannehamilton (approved):
👍 

As discussed, let's add the format map to the tech debt list in case we can just get rid of it. 🤞

PR #1032, opened 2017-11-30
[WIP] Extend synonyms A/B test to other subqueries

MatMoore:
This is a change to the synonym A/B test after debugging some queries.

I've also checked in some debug code I used, which I think could be useful in the future. I plan to use it to compare how the A and B versions treat specific documents.

See individual commit messages for more details.

This is still a WIP:
- I need to rerun the healthcheck with these changes
- I'm not convinced the `core_query.minimum_should_match_with_synonyms` part is working as expected, because we've debugged a couple of queries where matches on `all_searchable_text` don't show up in the query output, whereas with the A version we **can** see matches on the `_all` field.

Trello: https://trello.com/c/9VF3Rw5V/451-test-synonyms-with-healthcheck-and-search-performance-explorer

### DIFF
--- a/lib/debug/explainotron.rb
+++ b/lib/debug/explainotron.rb
@@ -18,7 +18,7 @@ module Debug
         metasearch_index: SearchConfig.instance.metasearch_index
       )
       search_query = query_builder.payload
-      puts search_query.inspect
+      pp search_query[:query]
 
       results = client.search(
         index: "govuk,mainstream,detailed,government",
--- a/lib/debug/explainotron.rb
+++ b/lib/debug/explainotron.rb
@@ -1,0 +1,33 @@
+require 'pry'
+
+module Debug
+  class Explainotron
+    def self.explain!(query, ab_tests: {})
+      client = Services.elasticsearch
+
+      parsed_params = {
+        query: query,
+        ab_tests: ab_tests,
+        debug: { explain: true, disable_best_bets: true, disable_popularity: true, disable_boosting: true }
+ }
+      search_params = Search::QueryParameters.new(parsed_params)
+
+      query_builder = Search::QueryBuilder.new(
+        search_params: search_params,
+        content_index_names: SearchConfig.instance.content_index_names,
+        metasearch_index: SearchConfig.instance.metasearch_index
+      )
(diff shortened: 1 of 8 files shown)


lib/tasks/debug.rake
```diff
@@ -78,4 +79,9 @@ namespace :debug do
       end
     end
   end
+
+  desc "Run the core of the search query with explain plans enabled and special boosting disabled"
+  task :explain, [:query] do |_, args|
+    Debug::Explainotron.explain!(args.query)
```
suzannehamilton: Should this pass in the A/B test parameter? (I realise this is a WIP, so you might be planning to add it later.)

lib/debug/explainotron.rb
```diff
@@ -0,0 +1,33 @@
+require 'pry'
+
+module Debug
+  class Explainotron
+    def self.explain!(query, ab_tests: {})
```
suzannehamilton, line 5: Excellent method name. 😄

lib/tasks/debug.rake
MatMoore: Possibly... I'm not sure what state I want to leave this in yet.

lib/debug/explainotron.rb
```diff
@@ -0,0 +1,33 @@
...
+      puts search_query.inspect
+
+      results = client.search(
+        index: "govuk,mainstream,detailed,government",
+        analyzer: 'with_search_synonyms',
+        body: search_query
+      )["hits"]["hits"]
+
+      # rubocop:disable Lint/Debugger
+      binding.pry
```
suzannehamilton, line 30: Out of curiosity, do you know what would happen if this ran on Jenkins? Would you have to manually stop the job?

MatMoore: I think so yeah, otherwise it will hang. I might replace this with some kind of output before merging this.

MatMoore:
Continuing in 

- https://github.com/alphagov/rummager/pull/1038
- https://github.com/alphagov/rummager/pull/1039
- https://github.com/alphagov/rummager/pull/1040

PR #1033, opened 2017-11-30
Add task to check state of Rummager document taxons

brenetic:
This is not entirely necessary but since I hacked this together to save me time I thought I'd raise it just in case in anyone else finds it useful.

It is just to check whether the taxons associated with a particular format were in draft state or not. There is a bug whereby documents in the mainstream index have taxons associated to them but
they are in draft state. As a result when republishing to the govuk index they are not present in the notification payload. This task just saves time checking them by hand.

### DIFF
--- a/lib/tasks/indices.rake
+++ b/lib/tasks/indices.rake
@@ -200,4 +200,47 @@ You should run this task if the index schema has changed.
 
     raise Exception.new("Missing index #{missing}!") unless missing.empty?
   end
+
+  desc "
+  Check for any taxons that are not in a draft state for a particular format.
+  Usage
+  rake 'rummager:check_for_non_draft_taxons[format_name, elasticsearch_index]'
+  "
+  task :check_for_non_draft_taxons, [:format, :index_name] do |_, args|
+    format = args[:format]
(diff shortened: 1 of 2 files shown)


lib/tasks/indices.rake
```diff
@@ -200,4 +200,47 @@ You should run this task if the index schema has changed.
...
+      publishing_api = Services.publishing_api
+
+      taxons = {}
+      ScrollEnumerator.new(
+        client: client,
+        search_body: { query: { term: { format: format } } },
+        batch_size: 500,
+        index_names: index
+      ) { |hit| hit }.map do |hit|
+        taxons[hit['_id']] = hit['_source']['taxons'] + hit['_source']['part_of_taxonomy_tree']
```
MatMoore: I think `part_of_taxonomy_tree` isn't needed here as a child taxon wouldn't be live if it's ancestors aren't live.

MatMoore (approved):
LGTM. Seems like it will be useful as we migrate stuff to the new index, but can delete it after that.

brenetic:
Thanks @MatMoore  Have updated 👍

PR #1034, opened 2017-11-30
Fix page traffic runner and exit with an error if called incorrectly

dwhenry:
Previously this was passed an index name to process which meant it had a single
argument, we are no longer passing in the index name (it was redundant) and as
a result we should have changed the logic.

### DIFF
--- a/bin/page_traffic_load
+++ b/bin/page_traffic_load
@@ -23,10 +23,11 @@ Slop.parse(help: true) do
   DOC
 
   run do |_opts, args|
-    if args.size == 1
+    if args.empty?
       GovukIndex::PageTrafficLoader.new.load_from($stdin)
     else
       puts self
+      exit 1
     end
   end
 end


bin/page_traffic_load
```diff
@@ -23,10 +23,11 @@ Slop.parse(help: true) do
   DOC
 
   run do |_opts, args|
-    if args.size == 1
+    if args.empty?
```
suzannehamilton, line 26: It'd be good to add something to the commit message explaining this change.

bin/page_traffic_load
```diff
@@ -23,10 +23,11 @@ Slop.parse(help: true) do
...
       GovukIndex::PageTrafficLoader.new.load_from($stdin)
     else
       puts self
+      exit 1
```
suzannehamilton, line 30: What does `puts self` print in this context? Is it helpful, or would it be better to raise an error with a message explaining what went wrong?

dwhenry: it prints the `slop` banner described above. 

I'm happy for this functionality to change but would do that as a separate task to this which is just getting it to run.

bin/page_traffic_load
dwhenry: have updated commit message to give more details

bin/page_traffic_load
dwhenry: That was due to the script being commented out on that date (this was how this job was initially disabled) If you go back to before that change you can see it in the older jobs.

brenetic (approved):
🚀

PR #1037, opened 2017-12-01
Fix error reporting when publishing special routes

suzannehamilton:
Fix `NameError`s which occur when logging errors caused by failed special route publishing.

It's generally safe to just log a warning because the special routes
published by rummager are updated so rarely:
- https://github.com/alphagov/rummager/pull/658
- cfcf6f9a1c6b6b9413a07f1484b98a72d71a21ce


brenetic (approved):
🚀

suzannehamilton:
Just pushed another commit to fix a typo on line 14.

PR #1039, opened 2017-12-01
Add a debug option to disable boosting

MatMoore:
With this option, we can disable all format boosting, and date boosting.
You can already disable popularity boosting and best bets, and by
disabling all of those you can isolate the main text matching part of
the query, i.e. the text analysis pipeline, TF-IDF, and the boolean
queries that combine various kinds of matching.

This is very useful at the moment, because this stuff is really broken,
and it's the foundation our search query is based on - by working on
this, we can tackle some of the really weird search behaviour, where
rummager returns documents that don't look anything like the query.

### DIFF
--- a/lib/parameter_parser/search_parameter_parser.rb
+++ b/lib/parameter_parser/search_parameter_parser.rb
@@ -337,6 +337,8 @@ private
         options[:disable_popularity] = true
       when "disable_synonyms"
         options[:disable_synonyms] = true
+      when "disable_boosting"
+        options[:disable_boosting] = true
       when "new_weighting"
         options[:new_weighting] = true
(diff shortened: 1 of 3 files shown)


lib/parameter_parser/search_parameter_parser.rb
```diff
@@ -337,6 +337,8 @@ def debug_options
         options[:disable_popularity] = true
       when "disable_synonyms"
         options[:disable_synonyms] = true
+      when "disable_boosting"
```
dwhenry, line 340: might be worth adding a `disable_all` option?

dwhenry (approved):
Looks good

PR #1040, opened 2017-12-01
Add a new debug command to inspect explain plans

MatMoore:
When investigating why a document scored higher/lower than anticipated, or why
a result matched at all, the boosting is fairly easy to understand, but it's
hard to see what the core of the query is doing. The explain plan sheds some
light on this, but it's very long and confusing when the whole query is used.

This is a minimal rake task that disables all boosting and best bets, enables
the explain plan, runs the query and then opens a repl with pry. This could
be extended to highlight the interesting parts of the explain output or diff
across A/B tests.

https://github.com/alphagov/rummager/pull/1039 should be merged first.

### DIFF
--- a/lib/debug/explainotron.rb
+++ b/lib/debug/explainotron.rb
@@ -1,13 +1,12 @@
-require 'pry'
+require 'rainbow'
 
 module Debug
   class Explainotron
-    def self.explain!(query, ab_tests: {})
+    def self.explain!(query)
       client = Services.elasticsearch
 
       parsed_params = {
         query: query,
-        ab_tests: ab_tests,
         debug: { explain: true, disable_best_bets: true, disable_popularity: true, disable_boosting: true }
  }
       search_params = Search::QueryParameters.new(parsed_params)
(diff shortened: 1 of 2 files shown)


lib/debug/explainotron.rb
```diff
@@ -0,0 +1,33 @@
+require 'pry'
+
+module Debug
+  class Explainotron
```
brenetic: FTW! 🙌

dwhenry:
I like the idea, my worry is that as a rake task it isn't super easy to use/interact with. I would prefer to see this implemented in https://github.com/alphagov/search-performance-explorer with a web frontend to simplify usage. :)

MatMoore:
Yeah that would be much better, but I'm probably not going to have time to implement that. I think it's a step up from using the debug API though, as that output is so hard to read, and this version removes all the boosting stuff that only affects ranking.

I found it useful when I wrote it so wanted to check it in some form even if it could be improved later. 

You can run it like `bundle exec rake debug:explain['self assess']`

and the output now looks like
screen shot 2017-12-14 at 17 14 25

dwhenry:
Move it to search performance and then merge I think as that is where it belongs. maybe with a note in the tech docs to make it a web based thing as part of next steps..

PR #1042, opened 2017-12-01
[PROOF OF CONCEPT] Handle batched data

dwhenry:
we currently process data off rabbitMQ a document at a time, this means that when we do large republishing (i.e. topic update) this can result in large volumes of records coming through and requiring processing.

Ideally we would use put multiple documents off the rabbit queue and put them on the sidekiq queue, however this is not possible so instead I am using redis as a tempory store to batch up data.

This does mean we could lose data is redis goes down.

This code adds some complexity as it makes the downstream processes handle both a single record and a batch of records.

### DIFF
--- a/lib/batcher.rb
+++ b/lib/batcher.rb
@@ -1,10 +1,25 @@
+# Due to the initial setup for the rabbit processing library we
+# have to do a couple of small hack so that we can overwrite the
+# calls to make everything work.
+#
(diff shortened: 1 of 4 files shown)


dwhenry:
closed as I think https://github.com/alphagov/govuk_message_queue_consumer/pull/47 is a cleaner solution for our problem and is more generic

PR #1043, opened 2017-12-04
Migrate to govuk_sidekiq gem

sihugh:
If we produce json formatted sidekiq logs, they will be automatically pulled
into kibana.  By moving to the govuk_sidekiq gem, we'll inherit this config.

Trello: https://trello.com/c/ODnBjU5V/528-cleanupdependencies-move-to-govuksidekiq-gem


MatMoore:
As discussed on slack - I think we need to explicitly call the initialiser somehow, but don't think there's any reason not to use this gem.

MatMoore:
I've rebased this and called the initializer from `initializers/sidekiq.rb`. I'll do some testing on integration tomorrow.

boffbowsh (approved):

PR #1046, opened 2017-12-05
Reconfigure synonyms for A/B test

suzannehamilton:
The new list of synonyms has not performed very well in the search healthcheck, so we have updated it to be more similar the old list of synonyms as a starting point for the A/B test:

- Revert the synonyms list back to search-time synonyms instead of a mix of search-time and index-time synonyms
- Leave a small group of multiword synonyms as index-time synonyms, because these generally performed better in the healthcheck
- Remove the my/your synonym, which had a bigger impact than we thought. Some results are significantly worse because of this synonym, so it needs some more thought

https://trello.com/c/9VF3Rw5V/451-test-synonyms-with-healthcheck-and-search-performance-explorer

cc @MatMoore

### DIFF
--- a/config/schema/synonyms.yml
+++ b/config/schema/synonyms.yml
@@ -1,7 +1,4 @@
 ---
-# Placeholder until we have real index time synonyms
-- index: searchotron => search
-
 # Multiple spellings
 - search: one, 1
 - search: two, 2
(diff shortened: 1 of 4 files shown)


config/schema/synonyms.yml
```diff
@@ -1,79 +1,80 @@
 ---
+# Placeholder until we have real index time synonyms
+- index: searchotron => search
```
brenetic: 😆

suzannehamilton: 😳 Fixed!

brenetic (approved):
There are some big gaps between groupings. If this is intentional then all good 🚀

suzannehamilton:
@MatMoore I cherry-picked your `*.synonym` commit. I've had to resolve some merge conflicts so could you review it please?

lib/search/query_components/core_query.rb
```diff
@@ -126,12 +124,26 @@ def match_bigrams(fields)
...
 
     def query_analyzer
       if search_params.disable_synonyms?
+        # this is the default defined in the mapping for regular fields
         DEFAULT_QUERY_ANALYZER_WITHOUT_SYNONYMS
+      elsif search_params.synonym_b_variant?
+        # this is the default defined in the mapping for *.synonym fields
+        QUERY_TIME_SYNONYMS_ANALYZER
       else
+        # this excludes the synonym filter
```
MatMoore: I wrote this, but I think this should say "includes" actually

MatMoore (approved):
Yup this looks like the behaviour I intended. Basically each field should switch to the ".synonym" variant in the B version and use the corresponding analyzer.

lib/search/query_components/core_query.rb
suzannehamilton: Maybe "this includes the old synonym filter"?

In any case, we'll hopefully be getting rid of this at the end of the A/B test.

PR #1047, opened 2017-12-05
Improve reindexing checks

suzannehamilton:
Add extra checks to reindexing to make sure that no documents have been added or removed by the reindexing process. The existing checks just looked for changes within a document.

Also add some integration tests of reindexing.

### DIFF
--- a/spec/integration/schema_migrator_spec.rb
+++ b/spec/integration/schema_migrator_spec.rb
@@ -67,8 +67,7 @@ RSpec.describe SchemaMigrator do
         migrator.reindex
 
         search_server.index_group("govuk_test").current_real.unlock
-        client.delete(index: "govuk_test", id: "/a-page-to-be-reindexed", type: "edition" )
-        commit_index("govuk_test")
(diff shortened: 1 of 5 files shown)


spec/integration/schema_migrator_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  context "index comparison" do
+    it "identifies when content has not changed" do
+      commit_document("govuk_test", { "link" => "/a-page-to-be-reindexed" })
+
+      SchemaMigrator.new("govuk_test", search_config, wait_between_task_list_check: 0.2) do |migrator|
+        migrator.reindex
+
+        expect(migrator.changed?).to be false
```
dwhenry: rspec convention would write this like:

```
expect(migrator).not_to be_changed
```

dwhenry (approved):
Look good, just the one syntax comment on rspec

spec/integration/schema_migrator_spec.rb
suzannehamilton: ✨ Today I learned. ✨

MatMoore (approved):

PR #1051, opened 2017-12-06
Remove Vehicle Recalls and Faults Alerts schema

tijmenb:
This specialist document type was added in https://github.com/alphagov/manuals-publisher/pull/480 as an MVP. Unfortunately no documents were ever published.

It's unused, so it's probably best to remove it to avoid unnecessary maintenance and avoid confusion.

The code was removed from specialist publisher in https://github.com/alphagov/specialist-publisher/pull/1072. The schema will be removed in https://github.com/alphagov/govuk-content-schemas/pull/675 (which will pass after this change is deployed).

https://trello.com/c/Jm6K9Tpx


config/schema/indexes/mainstream.json
```diff
@@ -25,7 +25,6 @@
     "service_manual_topic",
     "service_standard_report",
     "tax_tribunal_decision",
-    "utaac_decision",
-    "vehicle_recalls_and_faults_alert"
```
suzannehamilton, line 29: I think this means you can delete the type definition in vehicle_recalls_and_faults_alert.json, as well as some fields in field_definitions.json.

tijmenb: Updated, thanks. That's a lot of code for one document type!

suzannehamilton (approved):
🎉

PR #1052, opened 2017-12-08
 Highlight synonym version of description fields

suzannehamilton:
The description fields should have highlighting applied so that a user can see which of their search terms appear in the content. This was not working in the B version of the synonyms A/B test because the `description.synonym` fields did not have highlighting applied.

https://trello.com/c/yHuPBjBM/505-fix-term-highlighting-in-synonyms-b-version

We should fix title highlighting too for consistency, but it's not used in the search frontend so it's less critical than description highlighting which I'd like to deploy by the end of the day.


brenetic (approved):
LGTM 🚀

PR #1053, opened 2017-12-08
Make policies indexable

Rosa-Fox:
[Trello](https://trello.com/c/IXXgzDpD/497-make-policies-indexable)
- Maps `policy_groups`, `people`, `description` and `slug` fields.
- This is to set up the policy format to be migrated to the govuk search index.

### DIFF
--- a/lib/govuk_index/presenters/common_fields_presenter.rb
+++ b/lib/govuk_index/presenters/common_fields_presenter.rb
@@ -11,7 +11,6 @@ module GovukIndex
     delegate_to_payload :content_id
     delegate_to_payload :content_purpose_document_supertype
     delegate_to_payload :content_store_document_type, hash_key: "document_type"
-    delegate_to_payload :description
     delegate_to_payload :email_document_supertype
     delegate_to_payload :government_document_supertype
     delegate_to_payload :link, hash_key: "base_path"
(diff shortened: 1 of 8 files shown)


suzannehamilton:
I've picked this PR up from Rosa, and I've moved the `people` and `policy_group` logic from `PolicyPresenter` to `ExpandedLinksPresenter` for consistency.

lib/govuk_index/presenters/expanded_links_presenter.rb
```diff
@@ -62,6 +66,12 @@ def content_ids(collection)
       values_from_collection(expanded_links_item(collection), "content_id")
     end
 
+    def slugs(type, path_prefix)
```
dwhenry, line 69: I think this is a nice cleanup.

dwhenry: Only thought would be to change to to match: https://github.com/alphagov/rummager/pull/1053/files#diff-506cd05cce4cbe451fda1d266331a7bcR156

```
.gusb(%r{^/....
```

That way it will only match the prefix. :)

lib/govuk_index/presenters/common_fields_presenter.rb
```diff
@@ -26,6 +25,16 @@ def initialize(payload)
       @payload = payload
     end
 
+    def description
+      if format == "policy"
```
dwhenry, line 29: NOTE FOR THIS PR, but if we have time at the end of the quarter, it would be nice to try and remove the description where it is just a duplicate of content in the `indexable_content` field and see if it effect the score in any way.

That would enable us to remove this mess, plus some indexable content hacks.

dwhenry (approved):

lib/govuk_index/presenters/expanded_links_presenter.rb
suzannehamilton: I've made it more consistent. 👍 What do you think?

PR #1061, opened 2017-12-11
 Highlight matches in title field in B version of synonyms test

suzannehamilton:
Ensure that matching search terms in title are highlighted as they are for descriptions.

I don't think the frontend uses title highlighting at the moment (let me know if I'm wrong!), but it's part of the public API so it's worth fixing.

The last commit is a small refactoring of the highlighting code to separate the synonym handling from the rest of the formatting.

https://trello.com/c/yHuPBjBM/505-fix-term-highlighting-in-synonyms-b-version


dwhenry (approved):
I would potentially try and do some more tide up of the code, but as we will be deleting half of this once the test is complete this is as good a spot to leave it as any.

PR #1063, opened 2017-12-11
Update schema gems

suzannehamilton:
Update govuk_schemas and govuk-content-schema-test-helpers. This will update the json-schema dependency, which will fix the `Fixnum` deprecation warnings that we see when running the specs.


brenetic (approved):
🚀

PR #1069, opened 2017-12-12
Ignore differences in apostrophes when comparing content

suzannehamilton:
I spotted this issue when migrating policies, which has a few curly quotes in new content which the comparer should ignore. We already ignored differences in apostrophes in `indexable_content` - this just extends the comparison to other fields.

This change also adds some basic tests of the `GovukIndexFieldComparer`. It doesn't attempt to cover all the cases because the code is temporary and some of the checks are very specific to formats that we've already migrated.

https://trello.com/c/IXXgzDpD/497-make-policies-indexable


dwhenry (approved):
We should remember to delete this code before we finish up this quarter, just so we don't hurt the mind of the next group to work on this code. :)

PR #1076, opened 2017-12-13
Remove ci_reporter gem

suzannehamilton:
We stopped using this gem when we switched from MiniTest to RSpec (#869).


dwhenry (approved):
:godmode:

PR #1077, opened 2017-12-13
Make smart answers indexable

dwhenry:
Smart answers pass in `hidden_search_terms` which is an array of strings.

I have updated the indexable_content_presenter to handle this new format
which required a little reworking of the code.

https://trello.com/c/tSPsLjz3/489-make-smartanswers-indexable

### DIFF
--- a/lib/govuk_index/presenters/common_fields_presenter.rb
+++ b/lib/govuk_index/presenters/common_fields_presenter.rb
@@ -53,6 +53,9 @@ module GovukIndex
     end
 
     def format
+      # TODO: remove the special case for smart answers once it is fully migrated to
+      #   govuk as it's fallback `transaction` has the same implementation.
+      return 'smart-answer' if payload['publishing_app'] == 'smartanswers'
       document_type = payload['document_type']
       CUSTOM_FORMAT_MAP[document_type] || document_type
     end
--- a/lib/govuk_index/presenters/indexable_content_presenter.rb
+++ b/lib/govuk_index/presenters/indexable_content_presenter.rb
@@ -4,7 +4,7 @@ module GovukIndex
     BY_FORMAT = {
       'contact'       => %w(title description),
       'licence'       => %w(licence_short_description licence_overview),
-      'smart-answer'  => %w(introductory_paragraph),
+      'smart-answer'  => %w(introductory_paragraph more_information),
       'transaction'   => %w(introductory_paragraph more_information),
       'travel_advice' => %w(summary),
     }.freeze
--- a/lib/govuk_index/presenters/indexable_content_presenter.rb
+++ b/lib/govuk_index/presenters/indexable_content_presenter.rb
@@ -1,9 +1,10 @@
 module GovukIndex
   class IndexableContentPresenter
-    DEFAULTS = %w(body parts).freeze
+    DEFAULTS = %w(body parts hidden_search_terms).freeze
     BY_FORMAT = {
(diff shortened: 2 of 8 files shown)


lib/govuk_index/presenters/indexable_content_presenter.rb
```diff
@@ -34,7 +35,7 @@ def hidden_content
     def indexable_content_parts
       indexable_content_keys.flat_map do |field|
         indexable_values = details[field] || []
-        field == 'parts' ? parts(indexable_values) : [indexable_values]
+        field == 'parts' ? parts(indexable_values) : Array.new(indexable_values)
```
suzannehamilton: Was this broken before? It looks like `indexable_values` has always been an array.

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -50,7 +50,7 @@ def process_action(actions, routing_key, payload)
       if presenter.unpublishing_type?
         logger.info("#{routing_key} -> DELETE #{presenter.base_path} #{presenter.type}")
         actions.delete(presenter)
-      elsif MigratedFormats.indexable?(presenter.format) && payload['publishing_app'] != 'smartanswers'
+      elsif MigratedFormats.indexable?(presenter.format)
```
suzannehamilton, line 53: Does this mean we can get rid of all the `regenerate_if: ->(example) { example["publishing_app"] == "smartanswers" }` code? 😍

spec/unit/govuk_index/indexable_content_sanitiser_spec.rb
```diff
@@ -19,6 +19,30 @@
     expect(subject.clean(payload)).to eq("hello marmaduke")
   end
 
+  context "when array of string is passed in" do
```
suzannehamilton: Typo: string -> strings

lib/govuk_index/presenters/common_fields_presenter.rb
```diff
@@ -53,6 +53,7 @@ def popularity
     end
 
     def format
+      return 'smart-answer' if payload['publishing_app'] == 'smartanswers'
```
suzannehamilton: Would it be possible to use something like the schema instead? I'm worried that if the publishing app ever changed, it would silently (and mysteriously) break search.

lib/govuk_index/publishing_event_worker.rb
dwhenry: potentailly, but I haven;t tested it and would prefer to do that in a separate PR.

lib/govuk_index/presenters/common_fields_presenter.rb
dwhenry: I think one is as likely to change as the other. I will check with @MatMoore what he thinks the best field to use for this check would be.

spec/unit/govuk_index/indexable_content_sanitiser_spec.rb
dwhenry: corrected

lib/govuk_index/presenters/common_fields_presenter.rb
dwhenry: also just as a point, this field is useful now for the migration, however once that is complete it wont be used/required so this line could then be removed.

suzannehamilton (approved):

PR #1078, opened 2017-12-13
Handle batch of records into the Processor

dwhenry:
This is the first step to enabling the processing of batches from rabbit mq.
The idea is that this should increase the throughput speed when large volumes
of data are republished.

The code works by checking which of the events sent across succeed, and
manually resending any that failed. Unfortunately this means the sidekiq retry
count is reset whenever more than one item in the batch is successful, however
this is considered an acceptable compromise at this stage.

This code is designed to work in coordination with: https://github.com/alphagov/govuk_message_queue_consumer/pull/47


dwhenry:
This is teh wrong branch

PR #1080, opened 2017-12-13
Process message batches

dwhenry:
This is the first step to enabling the processing of batches from rabbit mq.
The idea is that this should increase the throughput speed when large volumes
of data are republished.

The code works by checking which of the events sent across succeed, and
manually resending any that failed. Unfortunately this means the sidekiq retry
count is reset whenever more than one item in the batch is successful, however
this is considered an acceptable compromise at this stage.

**UPDATE:**
This code now resends all messages in a batch if any fail. This allows us to reuse 
the sidekiq retry functionality at the expense of ES performance when an error occurs.
We plan to monitor this process once it is live to determine the number of errors that
occur.

This code is designed to work in coordination with: alphagov/govuk_message_queue_consumer#47

https://trello.com/c/OBhgB1Lw/531-code-tidy-up

### DIFF
--- a/lib/govuk_index/publishing_event_worker.rb
+++ b/lib/govuk_index/publishing_event_worker.rb
@@ -1,5 +1,7 @@
 module GovukIndex
+  class ElasticsearchRetryError < StandardError; end
   class ElasticsearchError < StandardError; end
+  class ElasticsearchInvalidResponseItemCount < StandardError; end
   class MissingTextHtmlContentType < StandardError; end
   class MultipleMessagesInElasticsearchResponse < StandardError; end
   class NotFoundError < StandardError; end
@@ -11,24 +13,16 @@ module GovukIndex
   class PublishingEventWorker < Indexer::BaseWorker
     notify_of_failures
 
-    def perform(routing_key, payload)
-      actions = ElasticsearchProcessor.new
-      process_action(actions, routing_key, payload)
-      response = actions.commit
-      process_response(response) if response
-    # Rescuing as we don't want to retry this class of error
-    rescue MissingBasePath => e
-      return if DOCUMENT_TYPES_WITHOUT_BASE_PATH.include?(payload["document_type"])
-      GovukError.notify(e, extra: { message_body: payload })
-    # Unpublishing messages for something that does not exist may have been
-    # processed out of order so we don't want to notify errbit but just allow
-    # the process to continue
-    rescue NotFoundError
-      logger.info("#{payload['base_path']} could not be found.")
-      Services.statsd_client.increment('govuk_index.not-found-error')
-    rescue UnknownDocumentTypeError
-      logger.info("#{payload['document_type']} document type is not known.")
-      Services.statsd_client.increment('govuk_index.unknown-document-type')
+    def perform(messages)
+      processor = ElasticsearchProcessor.new
+
+      messages.each do |routing_key, payload|
+        process_action(processor, routing_key, payload)
+      end
+
+      response = processor.commit
+      process_response(response, messages) if response
+
     # Rescuing exception to guarantee we capture all Sidekiq retries
     rescue Exception # rubocop:disable Lint/RescueException
       Services.statsd_client.increment('govuk_index.sidekiq-retry')
@@ -37,7 +31,7 @@ module GovukIndex
 
   private
 
-    def process_action(actions, routing_key, payload)
+    def process_action(processor, routing_key, payload)
       logger.debug("Processing #{routing_key}: #{payload}")
       Services.statsd_client.increment('govuk_index.sidekiq-consumed')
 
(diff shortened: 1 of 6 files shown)


lib/govuk_index/publishing_event_worker.rb
```diff
@@ -11,24 +11,16 @@ class MissingBasePath < StandardError; end
   class PublishingEventWorker < Indexer::BaseWorker
     notify_of_failures
 
-    def perform(routing_key, payload)
+    def perform(messages)
       actions = ElasticsearchProcessor.new
```
MatMoore: `actions` seems like a weird name here. Can we call it processor instead?

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -56,17 +48,48 @@ def process_action(actions, routing_key, payload)
...
+      logger.info("#{payload['document_type']} document type is not known.")
+      Services.statsd_client.increment('govuk_index.unknown-document-type')
     end
 
-    def process_response(response)
-      # we are only expecting to process a single message at a time
-      if response['items'].count > 1
-        Services.statsd_client.increment('govuk_index.elasticsearch.multiple_responses')
-        raise MultipleMessagesInElasticsearchResponse
+    def process_responses(responses, messages)
```
MatMoore: I think the name of this method should still be singular, since we're only sending one request to ES.

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -56,17 +48,48 @@ def process_action(actions, routing_key, payload)
...
+      messages_with_error = []
+
+      responses['items'].zip(messages).each do |response, message|
```
MatMoore: is the items array guaranteed to be in the same order as the request?

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -56,17 +48,48 @@ def process_action(actions, routing_key, payload)
...
+      Services.statsd_client.increment('govuk_index.not-found-error')
+    rescue UnknownDocumentTypeError
+      logger.info("#{payload['document_type']} document type is not known.")
+      Services.statsd_client.increment('govuk_index.unknown-document-type')
     end
 
-    def process_response(response)
-      # we are only expecting to process a single message at a time
-      if response['items'].count > 1
-        Services.statsd_client.increment('govuk_index.elasticsearch.multiple_responses')
```
MatMoore: Maybe we should still increment this counter, as we're not expecting there to be more than one message right now.

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -56,17 +48,48 @@ def process_action(actions, routing_key, payload)
...
+
+      responses['items'].zip(messages).each do |response, message|
+        messages_with_error << message unless valid_response?(response)
       end
 
-      item = response['items'].first
-      action_type, details = *item.first # item is a hash with a single key, value pair
+      if messages_with_error.any?
+        if messages_with_error.count == messages.count
+          # automatically retry
```
MatMoore: I don't think we should have a special case for complete failure. It would be simpler if the partial retry handled everything. Especially since the partial retry code is otherwise not expected to execute in production so we won't notice if there are any problems with it.

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -56,17 +48,48 @@ def process_action(actions, routing_key, payload)
...
+          )
+        else
+          Services.statsd_client.increment('govuk_index.sidekiq-partial-retry')
+          PublishingEventWorker.perform_async(messages_with_error)
+        end
+      end
+    end
+
+    def valid_response?(response)
+      action_type, details = *response.first # item is a hash with a single key, value pair
```
MatMoore: am confused about what this doing. Why is there a `.first` - I thought this was being called with a single item due to `responses['items'].zip(messages).each do |response, message|`

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -83,8 +106,16 @@ def process_response(response)
       else
         logger.error("#{action_type} not processed: status #{status}")
         Services.statsd_client.increment("govuk_index.elasticsearch.#{action_type}_error")
-        raise ElasticsearchError, action_type: action_type, details: details
+        # log something here
```
MatMoore: Unnecessary comment

spec/integration/govuk_index/publishing_event_processor_spec.rb
```diff
@@ -1,99 +1,137 @@
...
+        queue_name: "bigwig.test",
+        processor: GovukIndex::PublishingEventProcessor.new,
+        rabbitmq_connection: bunny_mock
+      )
+
+      @queue = @channel.queue("bigwig.test")
+      consumer.run
+    end
+
+    it "should_save_new_document_to_elasticsearch" do
```
MatMoore: can you make these strings into sentences?

MatMoore (changes requested):

lib/govuk_index/publishing_event_worker.rb
dwhenry: That is a good question, I also need to double check the results are in the same order as they were sent.

lib/govuk_index/publishing_event_worker.rb
dwhenry: The reason for this differentiation was that the partial retry loses the sidekiq retry implementation. I might have a look at how retry works and see if I can do partial retry with backoff and limited count of attempts.

lib/govuk_index/publishing_event_worker.rb
dwhenry: response if a hash (if my memory is correct)

```
response = { 'delete' => { details: 'hash'} }
response.first = ['delete', { details: 'hash'}]
```
happy to have a nicer approach to this, but was the best I could think of.

spec/integration/govuk_index/publishing_event_processor_spec.rb
dwhenry: done

lib/govuk_index/publishing_event_worker.rb
dwhenry: removed

lib/govuk_index/publishing_event_worker.rb
dwhenry: done

lib/govuk_index/publishing_event_worker.rb
dwhenry: done

lib/govuk_index/publishing_event_worker.rb
dwhenry: changed, but the variable names may still need some thought.

dwhenry:
@MatMoore still need:
- thoughts on the sidekiq args rewrite commit - not sure this is a good idea
- check response size and ordering matches the request

lib/govuk_index/publishing_event_worker.rb
MatMoore: Ah, I wasn't aware of `Hash#first` so it looked like you were working with an array.

Does the `*` do anything here though? Seems like you can also write
`action_type, details = response.first`

MatMoore:
Yeah I'm not convinced that modifying sidekiq jobs is a good idea. It's surprising behaviour and it will be difficult to maintain as part of rummager.

Maybe we should go back to how you originally had it, with one case for batches and one case for single documents. I think however we do it, the multiple-doc case won't be fully tested until we switch over to the batch consumer.

lib/govuk_index/publishing_event_worker.rb
dwhenry: I'm not totally sure I thought it was required, but it may splat automatically. I'll check

dwhenry:
@MatMoore as discussed this has been reverted and we are instead retrying all messages if any have an error.

MatMoore (approved):

MatMoore:
Changes look good, but worth another pair of eyes on this.

spec/unit/govuk_index/publishing_event_worker_spec.rb
```diff
@@ -1,166 +1,275 @@
...
+      it 'will notify for each message that fails' do
+        expect(Services.statsd_client).to receive(:increment).with('govuk_index.elasticsearch.index_error')
+        expect(Services.statsd_client).to receive(:increment).with('govuk_index.elasticsearch.index')
+        expect {
+          subject.perform([['routing.key', payload1], ['routing.key', payload2]])
+        }.to raise_error(GovukIndex::ElasticsearchRetryError)
+      end
+    end
+
+    it "raises an error is the number of response item does not match the number actions performed" do
```
suzannehamilton: Not quite sure I follow the test name. Is it supposed to be "raises an error if the number of response items does not match the number of actions performed"?

suzannehamilton (approved):
👍 This all looks sensible. I agree with the approach of requeueing everything to keep the Sidekiq interactions simple.

spec/unit/govuk_index/publishing_event_worker_spec.rb
dwhenry: maybe "..actions requested" is a better name.

Basically when I send 5 actions in the bulk message to ES if I get 4 or 6 responses then something has gone wrong.

dwhenry: updated as discussed

PR #1082, opened 2017-12-13
Use absolute path for sitemap files

dwhenry:
This fixes a bug caused by the symlinks pointing to the release path which is
deleted if more than 5 deployments happen on a given day.

https://trello.com/c/DmzPTudf/520-sitemapxml-files-become-unavailable


brenetic (approved):
🚀

PR #1088, opened 2017-12-14
Blacklisting functionality

dwhenry:
Add the ability to blacklist as well as whitelist formats. This also adds the ability to blacklist just some paths from a format which is required for `specialist_routes`.

https://trello.com/c/TZ6zhjUj/527-index-special-routes-to-govuk
https://trello.com/c/cBRbdmq1/521-blacklist-formats

### DIFF
--- a/lib/govuk_index/publishing_event_worker.rb
+++ b/lib/govuk_index/publishing_event_worker.rb
@@ -47,14 +47,17 @@ module GovukIndex
       )
       presenter.valid!
 
+      identifier = "#{presenter.base_path} #{presenter.type || "'unmapped type'"}"
       if presenter.unpublishing_type?
-        logger.info("#{routing_key} -> DELETE #{presenter.base_path} #{presenter.type}")
+        logger.info("#{routing_key} -> DELETE #{identifier}")
         actions.delete(presenter)
-      elsif MigratedFormats.indexable?(presenter.format)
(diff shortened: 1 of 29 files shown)


lib/govuk_index/publishing_event_worker.rb
```diff
@@ -47,14 +47,17 @@ def process_action(actions, routing_key, payload)
       )
       presenter.valid!
 
+      identifier = "#{presenter.base_path} #{presenter.type || "'unmapped type'"}"
```
brenetic, line 50: Can it not be `|| 'unmapped type'}"`?

Or do you want the inverted commas for extra noticability?

dwhenry: the second one, but happy to remove..

brenetic:
An alternative might be to use a separate list for the formats that need to be checked for routes. Then we could avoid having the `:all` and/or checking for an array.

lib/govuk_index/publishing_event_worker.rb
brenetic: shruggy mcshrugface

dwhenry:
@brenetic I did think about that, but felt it wouldn't offer any benefit over this approach, and for many things would be more complicated as I have to check two places for things.

However, that is gut feel only so have to pair on trying to do it that way if your keen. :)

brenetic (approved):
🚀

PR #1089, opened 2017-12-14
Delete smartanswers workaround

suzannehamilton:
Since #1077, rummager has been able to index documents published by smartanswers, so we can remove the workaround that prevented tests from using smartanswers documents as random examples.

### DIFF
--- a/spec/integration/govuk_index/collections_spec.rb
+++ b/spec/integration/govuk_index/collections_spec.rb
@@ -23,7 +23,6 @@ RSpec.describe "Collections publishing" do
         description: "Mainstream browse page description",
         base_path: "/browse/benefits",
       },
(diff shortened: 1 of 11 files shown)


suzannehamilton:
The tests are still failing intermittently. 😢 

It's because of this stub, which is in a lot of the tests:
```
allow(GovukIndex::MigratedFormats).to receive(:indexable_formats).and_return([search_document_type])
```

`search_document_type` isn't the right thing to use when the publishing app is `smartanswers`, because we override the format with `smart-answer`.

I'm going to think about this some more. It would be good to get rid of the workaround.

dwhenry (approved):

suzannehamilton:
#1092 means that the tests are now passing. Thanks @dwhenry! 👍

PR #1095, opened 2017-12-15
Remove description duplication in ES content

dwhenry:
This duplication added to the complexity of the code 
as we needed to look in multiple places on a format 
specific basis. Removing this functionality simplifies
the code, if we do need to re-introduce boosting for 
this formats we should do this using the standard 
boosting functionality.

We need to check if these changes have any effect on the health check sore before we merge.

https://trello.com/c/OBhgB1Lw/531-code-tidy-up

### DIFF
--- a/lib/govuk_index/presenters/common_fields_presenter.rb
+++ b/lib/govuk_index/presenters/common_fields_presenter.rb
@@ -11,6 +11,7 @@ module GovukIndex
     delegate_to_payload :content_id
     delegate_to_payload :content_purpose_document_supertype
     delegate_to_payload :content_store_document_type, hash_key: "document_type"
+    delegate_to_payload :description
     delegate_to_payload :email_document_supertype
     delegate_to_payload :government_document_supertype
     delegate_to_payload :link, hash_key: "base_path"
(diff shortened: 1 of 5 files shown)


lib/govuk_index/presenters/common_fields_presenter.rb
```diff
@@ -25,24 +26,10 @@ def initialize(payload)
       @payload = payload
     end
 
-    def description
```
MatMoore, line 28: I think removing this is the right thing to do, but it seems that policy pages don't have a `description` set, which means they won't display anything for the description on results pages after this change. Example data: https://www.gov.uk/api/content/government/policies/academies-and-free-schools

This is also affecting external search right now - eg googling brexit gives you

> Brexit. From: Department for Exiting the European Union. The UK voted to leave the European Union in the 23 June 2016 EU referendum. The government is now preparing to leave the EU in the best possible way for the UK's national interest. Subscribe to email alerts. Search. Document type. Case study

Instead of just 

> The UK voted to leave the European Union in the 23 June 2016 EU referendum. The government is now preparing to leave the EU in the best possible way for the UK’s national interest.

Before we merge this PR, we should make a change to policy publisher to actually populate descriptions.

dwhenry:
@MatMoore the policy update is now being deployed so can I get a approval on this please.

MatMoore (approved):

PR #1096, opened 2017-12-15
Remove lock_updates script

MatMoore:
This script duplicates functionality we can run via rake tasks.
It doesn't appear to be referenced anywhere.


brenetic (approved):
🚀

PR #1097, opened 2017-12-18
Remove alternate query

MatMoore:
This removes a lot of code that was never deployed, because it didn't perform better than the existing query.

Various components of this are still worth exploring (particularly shingles, ID code extraction, and different boosting of match fields), but we are not going to be able to do this in the foreseeable future, because the  current search team is disbanding in January.

I'd recommend improving tooling for evaluating search results before revisiting any of this work, and at this stage it's easier to start from scratch than reuse any of this code.

I'd like to also clean up the elasticsearch config to remove unused analyzers, but I'll do this in a future PR.


dwhenry:
I think that looks fine but would like a second set of eye if possible.

dwhenry (approved):

suzannehamilton (approved):
🎉

PR #1115, opened 2017-12-20
Remove slop

MatMoore:
We don't need it anymore. See discussion on https://github.com/alphagov/rummager/pull/1085


suzannehamilton (approved):
🎉

PR #1116, opened 2017-12-20
Create new V2 endpoint for use with best bets

dwhenry:
This adds new synchronous endpoints for inserting and deleting data from the
metasearch index.

https://trello.com/c/K5zkvpfi/537-delete-best-bet-query-and-items-not-working-in-search-admin


lib/metasearch_index/deleter.rb
```diff
@@ -0,0 +1,55 @@
...
+      class NotFound < StandardError; end
+      class UnknownError < StandardError; end
+
+      def initialize(id:, data:)
+        @id = id.presence || raise(MissingArgument, "ID and Type must be supplied.")
+        @data = data.presence || raise(MissingArgument, "Not record provided to insert into Elasticsearch.")
+      end
+
+      def delete
+        processor = ElasticsearchProcessor.new
```
MatMoore: Looks like you need to pass in a `MetasearchIndex::client` here (it defaults to the govuk one).

lib/metasearch_index/deleter.rb
```diff
@@ -0,0 +1,55 @@
...
+        processor.delete(self)
+        response = processor.commit
+        validate_response!(response.first)
+      end
+
+      def identifier
+        {
+          _type: _type,
```
MatMoore: Is this a typo? Or a _type-o...

lib/metasearch_index/deleter.rb
```diff
@@ -0,0 +1,55 @@
...
+          _id: base_path,
+        }
+      end
+
+      def validate_response!(response)
```
MatMoore: As discussed earlier - there is some logic duplicated between this method and `PublishingEventWorker` that we could potentially extract out, but I haven't thought too much about what that would look like. It would be good to have the core deletion logic, logging, stats, and error reporting implemented once in an index-independent way if there's a nice way of doing that.

Similar situation for the inserter.

lib/metasearch_index/deleter.rb
```diff
@@ -0,0 +1,55 @@
...
+          Services.statsd_client.increment("metasearch.elasticsearch.#{action_type}")
+        elsif status == 404 # failed while attempting to delete missing record so just ignore it
+          logger.info("Tried to delete a document that wasn't there; ignoring.")
+          Services.statsd_client.increment('metasearch.elasticsearch.already_deleted')
+          raise NotFound, "Document not found in index"
+        else
+          logger.error("#{action_type} not processed: status #{status}")
+          Services.statsd_client.increment("metasearch.elasticsearch.#{action_type}_error")
+
+          # manually send the error as we rescue it later in the process
```
MatMoore: Maybe we could send this at the point we rescue the exception, as whether we still want to notify seems related to the context its being caught in

lib/rummager/app.rb
```diff
@@ -73,7 +73,17 @@ def json_only
     halt(400, env['sinatra.error'].message)
   end
 
-  # Return results for the GOV.UK site search
+  error MetasearchIndex::Deleter::V2::UnknownError, MetasearchIndex::Deleter::V2::MissingArgument,
+    MetasearchIndex::Inserter::V2::UnknownError, MetasearchIndex::Inserter::V2::MissingArgument do
+    halt(500, env['sinatra.error'].message)
```
MatMoore: This should be a 400 error as it's only thrown if the client has passed us an invalid message

MatMoore (dismissed):
A few comments inline

lib/metasearch_index/deleter.rb
dwhenry: I agree, have started on this by extracting the code for these new classes to a shared location. will see if this I can include the govukindex code as well.

lib/metasearch_index/deleter.rb
dwhenry: yeah, I might even be able to drop this as need to look at what sinatra actually does in error mode.

lib/rummager/app.rb
dwhenry: changed

lib/index/response_validator.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+    end
+
+    def self.valid?(response)
+      action_type, details = response.first # response is a hash with a single [key, value] pair
+      status = details['status']
+
+      if (200..399).cover?(status)
+        logger.debug("Processed #{action_type} with status #{status}")
+        Services.statsd_client.increment("govuk_index.elasticsearch.#{action_type}")
```
MatMoore: do we want to parameterise the statsd key name? It always includes "govuk_index" at the moment.

lib/metasearch_index/deleter.rb
```diff
@@ -2,18 +2,16 @@ module MetasearchIndex
   module Deleter
     class V2
       class MissingArgument < StandardError; end
```
MatMoore: Could we just use the standard ArgumentError here?

lib/rummager/app.rb
```diff
@@ -73,12 +73,16 @@ def json_only
     halt(400, env['sinatra.error'].message)
   end
 
-  error MetasearchIndex::Deleter::V2::UnknownError, MetasearchIndex::Deleter::V2::MissingArgument,
-    MetasearchIndex::Inserter::V2::UnknownError, MetasearchIndex::Inserter::V2::MissingArgument do
+  error Index::ResponseValidator::ElasticsearchError do
+    # TODO: we want to one to be reported to sentry - check this is the case.
```
MatMoore: Still a TODO here

lib/index/response_validator.rb
```diff
@@ -0,0 +1,61 @@
+module Index
+  class ResponseValidator
+    class NotFound < StandardError; end
+    class ElasticsearchError < StandardError; end
+
+    def self.valid!(response)
```
MatMoore: How come we have both `valid!` and `valid?`? Shouldn't one call the other?

If they are meant to be index specific the name doesn't make that clear enough.

MatMoore (dismissed):
Looks good other than a couple of things with the ResponseValidator (see inline comments) and a TODO

lib/index/response_validator.rb
dwhenry: this is a temporary measure as i want both in the same spot, but still thinking about the best way to combine functionality.

lib/index/response_validator.rb
dwhenry: good point.. i'll think of a way to fix this.

lib/metasearch_index/deleter.rb
dwhenry: yes, that makes more sense..

lib/rummager/app.rb
dwhenry: it because I'm not sure how to test this. and ensure the error is reported.

lib/index/response_validator.rb
dwhenry: I plan this leave this as is, maybe try to refactor in the new year if i have time.

lib/rummager/app.rb
dwhenry: was going to push it live with the todo while I think about it, try to work out a plan. :(

Maybe integration testing will give me inspiration

lib/index/response_validator.rb
MatMoore: can you give them different names for now then? I think it's easy to confuse the two at the moment.

brenetic (approved):
🚀

PR #1117, opened 2017-12-20
Fully migrate external content to the govuk index

suzannehamilton:
Mark the `recommended-link` format (which corresponds to the `external_content` document type) as migrated because it is now indexed correctly in the govuk index.

https://trello.com/c/qzVTvlPe/539-mark-external-links-as-migrated

Marked as "do not merge" until we have run the comparison tool on production.


MatMoore (approved):
🎉

suzannehamilton:
We've deleted and republished all the the external links on production, so we can merge this now.

PR #1120, opened 2017-12-21
[DO NOT MERGE]Check for documents in mainstream that are not in govuk

Rosa-Fox:
- Check for documents that are in `mainstream` but not in `govuk`.
- The result should be 0.
- We have created this check to be run before and after rebuilding `govuk` so we can determine that all of the documents have been successfully migrated.

We are not planning on merging this but would like another pair of eyes @suzannehamilton and @dwhenry.

### DIFF
--- a/lib/tasks/indices.rake
+++ b/lib/tasks/indices.rake
@@ -243,4 +243,72 @@ You should run this task if the index schema has changed.
       end
     end
   end
+
+  desc "
+    Check to see what's in mainstream that isn't in govuk
+  "
+  task :compare_govuk_and_mainstream_yo do
+    # Arguments to run the publishing_api queue:requeue_document_type task with (aka all the migrated formats): mainstream_browse_page,topic,contact,answer,guide,help_page,licence,local_transaction,place,transaction,simple_smart_answer,aaib_report,asylum_support_decision,business_finance_support_scheme,cma_case,countryside_stewardship_grant,dfid_research_output,drug_safety_update,employment_appeal_tribunal_decision,employment_tribunal_decision,esi_fund,international_development_fund,maib_report,medical_safety_alert,raib_report,service_standard_report,tax_tribunal_decision,utaac_decision,calendar,finder,hmrc_manual,hmrc_manual_section,manual,manual_section,policy,service_manual_guide,service_manual_homepage,service_manual_service_standard,service_manual_topic,task_list,travel_advice,travel_advice_index,special_route,recommended-link
+
+    client = Services.elasticsearch(hosts: SearchConfig.new.base_uri, timeout: 5.0)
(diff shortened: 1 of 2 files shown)


lib/tasks/indices.rake
```diff
@@ -243,4 +243,68 @@ You should run this task if the index schema has changed.
...
+        batch_size: 500,
+        index_names: index.to_s
+      ) { |hit| hit }.map do |hit|
+        id = hit['_id']
+        id.prepend('/') unless id[0] == '/'
+        ids << id
+      end
+    end
+
+    not_in_govuk = indices[:mainstream] - indices[:govuk]
```
suzannehamilton, line 276: It'd be interesting to reverse this as well just to see if it finds anything.

lib/tasks/indices.rake
```diff
@@ -243,4 +243,68 @@ You should run this task if the index schema has changed.
...
+    actual_missing = not_in_govuk
+
+    puts "*************************************************************"
+    puts "Checking #{actual_missing.count} pieces of non unpublishing type content"
+    # puts actual_missing
+    puts "*************************************************************"
+
+    actual_missing.reject! do |base_path|
+      begin
+        content_item = JSON.parse(`curl https://www.gov.uk/api/content#{base_path}`)
```
suzannehamilton: Could you handle external content as well? It has a full URL as a document ID.

lib/tasks/indices.rake
```diff
@@ -243,4 +243,68 @@ You should run this task if the index schema has changed.
...
+        e.message.include?('Page no longer here - 410')
+      end
+    end
+
+    puts "*************************************************************"
+    puts "#{redirects} redirects"
+    puts "#{gones} gones"
+    puts "#{actual_missing.count} are actually actually actually missing"
+    actual_missing.each do |path|
+      puts "{#{path}}"
```
suzannehamilton: Not a big deal, but what are the extra curly braces for?

suzannehamilton (commented):
👍 This all looks sensible. Just a couple of extra things to consider.

lib/tasks/indices.rake
Rosa-Fox: Thanks, we have added in a check/counter.

lib/tasks/indices.rake
Rosa-Fox: Have removed!

lib/tasks/indices.rake
Rosa-Fox: It finds `2234` to be missing, I suppose because of the formats that have only been published to `govuk` due to migration.

MatMoore:
I think we can close this now

PR #1121, opened 2017-12-21
Remove unused analyzers, update synonyms

MatMoore:
cc @tarastockford 

Sorry for the mega PR. This branch does two things:
- removes elasticsearch configuration that is not used any more or has never been used in production
- changes and removes some synonym rules that weren't working well

I'd like to deploy both at the same time since we need to reindex for both.

`shingled_query_analyzer` is supposed to incorporate bigrams (pairs of adjacent words) into the search query.

This is useful because part of the meaning of a sentence comes from the word order, for example: `book a driving test for someone else` vs `driving someone else for a test book`.

However, this code never worked as intended because it only analyzed queries. So a query was broken down into single words and bigrams, but it compared those tokens to analyzed text that didn't contain any bigrams at all.

This means the bigram part of the query is only functioning as a single word match. This is very confusing when trying to understand what Rummager is doing.

I've changed it to use the normal query analyzer. This will change results slightly, because the shingles analyzer didn't include synonyms, but the new analyzer does.

Bigram matching was implemented properly as part of the 'new weighting' code a couple of years ago, but it never went live. This is something that could be revisited in future.

I've also removed:

- The reworked shingles analyzer (see https://trello.com/c/Zvdw9TLW/67-make-synonyms-work-properly-other-ranking-fixes)
- The ID codes analyser (see https://trello.com/c/lrYj5PdF/146-enable-searching-for-id-codes-with-or-without-spaces-punctuation)
- `query_default` which has been unused for some time
  
Trello:

- https://trello.com/c/VJXHCfYg/546-investigate-broken-search-term-highlighting
- https://trello.com/c/co7EVcJq/551-cleanuptimebox-remove-unused-analyzers-from-rummager

### DIFF
--- a/config/schema/synonyms.yml
+++ b/config/schema/synonyms.yml
@@ -39,6 +39,7 @@
 # Multiword synonyms "play havoc" with term positions, so avoid creating index rules that change the phrase length.
 # See: https://www.elastic.co/guide/en/elasticsearch/guide/current/multi-word-synonyms.html
 # If a phrase expands to a longer phrase, extra words will be highlighted in search descriptions.
+# NOTE: terms like `practical_driving_test` are not marked as keywords so they are still stemmed.
 - both: practical test, practical tests => practical test, practical_driving_test
 - both: practical driving test, practical driving tests => practical driving test, practical_driving_test
 - both: provisional licence, provisional license => provisional licence, provisional_driving_licence
--- a/config/schema/synonyms.yml
+++ b/config/schema/synonyms.yml
@@ -31,10 +31,18 @@
 - index: hazardous substances, dangerous substances
 - index: holiday pay, holiday entitlement
 - index: pension forecast, pension statement
-- index: practical driving test, practical test
-- index: provisional driving licence, provisional licence
 - index: tariff classification, trade tariff
 - index: tariff codes, commodity codes
+
+# Treat synonyms of differing lengths the same
+
+# Multiword synonyms "play havoc" with term positions, so avoid creating index rules that change the phrase length.
+# See: https://www.elastic.co/guide/en/elasticsearch/guide/current/multi-word-synonyms.html
+# If a phrase expands to a longer phrase, extra words will be highlighted in search descriptions.
+- both: practical test, practical tests => practical test, practical_driving_test
+- both: practical driving test, practical driving tests => practical driving test, practical_driving_test
+- both: provisional licence, provisional license, provisonal licence, provisonal license => provisional licence, provisional_driving_licence
+- both: provisional driving licence, provisional driving license, provisonal driving licence, provisonal driving license => provisional driving licence, provisional_driving_licence
 
 # Expanding queries to capture similar meanings.
 
--- a/config/schema/synonyms.yml
+++ b/config/schema/synonyms.yml
@@ -41,8 +41,8 @@
 # If a phrase expands to a longer phrase, extra words will be highlighted in search descriptions.
 - both: practical test, practical tests => practical test, practical_driving_test
 - both: practical driving test, practical driving tests => practical driving test, practical_driving_test
-- both: provisional licence, provisional license, provisonal licence, provisonal license => provisional licence, provisional_driving_licence
-- both: provisional driving licence, provisional driving license, provisonal driving licence, provisonal driving license => provisional driving licence, provisional_driving_licence
+- both: provisional licence, provisional license => provisional licence, provisional_driving_licence
+- both: provisional driving licence, provisional driving license => provisional driving licence, provisional_driving_licence
 
 # Expanding queries to capture similar meanings.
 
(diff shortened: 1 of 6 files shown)


suzannehamilton (commented):
I won't approve this yet because (as you say) we'll need to test it, but this looks sensible to me. 👍

MatMoore:
Have now tested this on integration - it's an overall improvement to the healthcheck score and it fixes some bad highlighting so I'm confident enough to deploy it.

suzannehamilton (approved):
🎉

config/schema/synonyms.yml
```diff
@@ -31,11 +31,19 @@
...
-- index: provisional driving licence, provisional licence
 - index: tariff classification, trade tariff
 - index: tariff codes, commodity codes
 
+# Treat synonyms of differing lengths the same
+
+# Multiword synonyms "play havoc" with term positions, so avoid creating index rules that change the phrase length.
+# See: https://www.elastic.co/guide/en/elasticsearch/guide/current/multi-word-synonyms.html
+# If a phrase expands to a longer phrase, extra words will be highlighted in search descriptions.
+- both: practical test, practical tests => practical test, practical_driving_test
```
dwhenry: I take it we don't need to use the special character in front of the underscored words to avoid it being stemmed?

config/schema/synonyms.yml
```diff
@@ -53,46 +61,46 @@
...
 - search: kiev => kiev, kyiv, ukraine
 - search: mot checker => mot check
 # Student finance login
-- search: my account => your account, online account
+- search: my account => my account, your account, online account
 - search: not arrived => not arrived, contact dvla
 - search: not received => not received, contact dvla
 - search: refuse collection => refuse collection, rubbish collection
-- search: self certificate => self certification
+- search: self certificate => self certificate, self certification
```
dwhenry, line 91: would this one be covered by normal stemming rules?
Also, would this add self to the resulting analysed sentence twice, artificially increasing its relevance?

dwhenry (approved):
Looks pretty good to me, just a couple of small questions.

config/schema/synonyms.yml
MatMoore: That's a good point - I don't expect this to be stemmed atm, but I'll add it in to make it clear what's going on.

config/schema/synonyms.yml
MatMoore: For whatever reason, `certificate` stems to `certif` and `certification` stems to `certificat`

config/schema/synonyms.yml
MatMoore: On second thoughts I'm going to skip this and just put in a comment, because we actually took out the protwords filter that does this: https://github.com/alphagov/rummager/commit/4f0013f43fff0e85a5e904d07fc53697fe7d0a3a

These terms do go through the stemmer but it shouldn't cause any issues since it's not stemming to an actual word stem.

```
elyzer --es 'http://dev.gov.uk:9200' --index 'detailed-2018-01-05t13:50:50z-6d703df4-b122-4a6a-ad6f-a55ebf635853' --text 'practical test' --analyzer 'with_index_synonyms'
CHAR_FILTER: normalize_quotes
{0:practical}	{1:test}
CHAR_FILTER: strip_quotes
{0:practical}	{1:test}
TOKENIZER: standard
{0:practical}	{1:test}
TOKEN_FILTER: standard
{0:practical}	{1:test}
TOKEN_FILTER: asciifolding
{0:practical}	{1:test}
TOKEN_FILTER: lowercase
{0:practical}	{1:test}
TOKEN_FILTER: index_synonym
{0:practical,practical_driving_test}	{1:test}
TOKEN_FILTER: stop
{0:practical,practical_driving_test}	{1:test}
TOKEN_FILTER: stemmer_override
{0:practical,practical_driving_test}	{1:test}
TOKEN_FILTER: stemmer_english
{0:practical,practical_driving_test}	{1:test}
```

```
elyzer --es 'http://dev.gov.uk:9200' --index 'detailed-2018-01-05t13:50:50z-6d703df4-b122-4a6a-ad6f-a55ebf635853' --text 'provisional licence' --analyzer 'with_index_synonyms'
CHAR_FILTER: normalize_quotes
{0:provision}	{1:licenc}
CHAR_FILTER: strip_quotes
{0:provision}	{1:licenc}
TOKENIZER: standard
{0:provisional}	{1:licence}
TOKEN_FILTER: standard
{0:provisional}	{1:licence}
TOKEN_FILTER: asciifolding
{0:provisional}	{1:licence}
TOKEN_FILTER: lowercase
{0:provisional}	{1:licence}
TOKEN_FILTER: index_synonym
{0:provisional,provisional_driving_licence}	{1:licence}
TOKEN_FILTER: stop
{0:provisional,provisional_driving_licence}	{1:licence}
TOKEN_FILTER: stemmer_override
{0:provisional,provisional_driving_licence}	{1:licence}
TOKEN_FILTER: stemmer_english
{0:provision,provisional_driving_lic}	{1:licenc}
```

PR #1124, opened 2017-12-22
 Refactor unpublishing presenter

suzannehamilton:
Extract a separate `ElasticsearchDeletePresenter` from the `ElasticsearchPresenter` to handle document unpublishing. Only the DeletePresenter now contains logic for looking up the existing document.

This will make it easier to implement unpublishing for external content, which also needs to look up the existing document.

This also takes us one step closer to refactoring this code into format-specific presenters, which we're planning to do soon.

https://trello.com/c/mwqMqK8s/542-allow-external-content-to-be-unpublished

It'll be easier to review the commits separately because the first one renames a file and lots of variables.


lib/govuk_index/document_type_inferrer.rb
```diff
@@ -1,5 +1,5 @@
 module GovukIndex
-  class DocumentTypeInferer
+  class DocumentTypeInferrer
```
brenetic, line 2: :man_facepalming:

brenetic:
This looks cool. Is there a way we can DRY the identifier? It's not likely to ever change is it?

lib/govuk_index/document_type_inferrer.rb
MatMoore: [At least it's easy to change](https://www.quora.com/What-is-the-origin-of-HTTP-header-%E2%80%9Creferer%E2%80%9D-misspelling) 😁

MatMoore (approved):
LGTM

MatMoore:
I've just rebased this so it can be merged. @brenetic I'll have a look at the identifier part and maybe raise another PR for it.

PR #1125, opened 2017-12-29
Allow deleting external content

MatMoore:
Currently, the delete presenter can't delete external content, because it doesn't have a base path.

External content used to be identified by the full URL, but if we keep that then we'll have to do a lookup by `link` to figure out what the `id` is when deleting, because publishing API uses a different schema for unpublishing, so we won't always get the full `details` of the original content item.

We're already doing lookups by ID to get the `type` so this seems like a lot of hassle to implement cleanly. Using URLs as IDs is also really annoying when making ad hoc queries because you have to escape all the slashes. To get around this I've changed this format to use `content_id` for the ID in elasticsearch. We'll have to update the comparer for this but I don't think anything else should have to change.

I've also done a bit more presenter refactoring.

A consequence of this PR is that we can no longer tell if an unpublishing is valid or not based on the presence of a base path. If it doesn't have a base path, but does have a content id, we assume its external content and attempt to delete.

Trello: https://trello.com/c/mwqMqK8s/542-allow-external-content-to-be-unpublished


suzannehamilton (commented):
The code all looks good. 👍 

I like the idea of using content IDs rather than full URLs  - they made it quite fiddly to work with external content.

Do you have the list of applications which depend on Rummager? It would be good to check that none of them will be affected by this change, either because they persist external document IDs or because they rely on the `link` always being the same as `_id`.

MatMoore:
I think these GOV.UK apps depend on rummager:

- whitehall
- manuals frontend
- finder frontend
- collections
- content tagger
- search admin

None of them persist data coming back from rummager, and this content gets filtered out of every query except the main site search.

It's impossible to tell for sure if anything depends on the link being the same as the ID, because the API is public, but I think the risk is low.

suzannehamilton (approved):
Yeah, that will make it it tricky to switch to using content IDs as rummager IDs for everything, but I agree that it's low risk for external content. 👍

PR #1129, opened 2018-01-02
Remove underscores from spec names

MatMoore:
These are still not always readable, but I think it's slightly better than before. Thoughts?


suzannehamilton (approved):
A few suggestions but I think this is definitely an improvement, if only to discourage people from writing more test names with underscores.

MatMoore:
Have updated those files and also made a small change to hide some STDOUT logging during rspec runs.

PR #1133, opened 2018-01-03
Migrate external content

MatMoore:
This reintroduces the change to start reading external content from the `govuk` index.

Best bets are now looked up by the `link` field, because that's what is actually written to the metasearch index.

This was previously relying on `link` and `id` being the same. We changed the `id` for external content to the `content_id` to make it easier to keep the data in sync with publishing api.

Trello: https://trello.com/c/qzVTvlPe/539-mark-external-links-as-migrated


dwhenry (approved):
I don't really understand this change, but assume it has been tested so happy to approve.

PR #1134, opened 2018-01-03
Throw an exception if a given format is not valid

Rosa-Fox:
[Trello](https://trello.com/c/UWRtjc3O/547-throw-an-exception-if-a-given-format-is-not-in-the-list-of-formats)

paired with @JonathanHallam

### DIFF
--- a/lib/govuk_index/publishing_event_worker.rb
+++ b/lib/govuk_index/publishing_event_worker.rb
@@ -7,6 +7,7 @@ module GovukIndex
   class UnknownDocumentTypeError < StandardError; end
   class NotIdentifiable < StandardError; end
   class MissingExternalUrl < StandardError; end
+  class InvalidFormatError < StandardError; end
 
   DOCUMENT_TYPES_WITHOUT_BASE_PATH = %w(contact world_location).freeze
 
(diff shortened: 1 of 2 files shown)


lib/govuk_index/publishing_event_worker.rb
```diff
@@ -60,6 +61,7 @@ def process_action(processor, routing_key, payload)
         processor.save(presenter)
       else
         logger.info("#{routing_key} -> UNKNOWN #{identifier}")
+        GovukError.notify(InvalidFormatError.new, extra: { error_message: "#{presenter.format} is an invalid format" })
```
dwhenry: would be good to get the identifier this is for in the error data. Also remember we plan to move to batch process at which point this record will be skipped so we ensure we think about how it would be sent for reprocessing if required. 

That can be as easy as having it resent from publishing API...

dwhenry (changes requested):
Looks good, but lets pad out what data we send with the error. Remember this is not a HTTP request so we don't get params for free.. :)

Rosa-Fox:
Thanks @dwhenry, good point! Have amended.

dwhenry (approved):
It might be worth doing an audit of how we send error data to sentry to ensure it is done in a consistent manner, but feel that is a different card @MatMoore

MatMoore:
@dwhenry we could definitely be using sentry better but I don't think there's anything worth doing now. The focus for now should be on making sure everything that's sent to sentry is actually an error ([RFC 87](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-087-dealing-with-errors.md))

PR #1135, opened 2018-01-03
Remove mainstream references

dwhenry:
Remove all reference to mainstream index from rummager. Ensuring the existing migration code is kept to assist with future migrations of government and detailed indices.

Reference to mainstream have been replaced with either government or govuk depending on the context of the tests, this has resulted in some data which does not match production, but does allow the tests to run.

https://trello.com/c/LiapHutO/553-remove-reference-to-mainstream-index


config/query/boosting.yml
```diff
@@ -1,10 +1,10 @@
 base:
   format:
-    # Mainstream formats
+    # Govuk formats
```
MatMoore: I think both of these comments are misleading, we should just get rid of them

doc/content_overview.md
```diff
@@ -8,9 +8,9 @@ and the apps currently responsible for publishing them as of June 2017.
 For a broader view of the content that is available, see [Document types on GOV.UK](https://docs.publishing.service.gov.uk/document-types.html).
 
 ## Publisher
-Publishes "mainstream" content on GOV.UK - about 3000 pages which target common user needs. This is the kind of stuff you can reach from the homepage.
+Publishes "govuk" content on GOV.UK - about 3000 pages which target common user needs. This is the kind of stuff you can reach from the homepage.
```
MatMoore: this mainstream is actually mainstream

doc/content_overview.md
```diff
@@ -22,17 +22,16 @@ Implemented in [search_payload_presenter.rb](https://github.com/alphagov/publish
  - 20 places
  - 8 help pages
 
-
 ## Travel-advice-publisher
 Publishes [foreign travel advice](https://www.gov.uk/foreign-travel-advice) on GOV.UK. There are 226 pages, but they have
 a format of "custom application" in Rummager.
 
-Implemented in [search_payload_presenter.rb](https://github.com/alphagov/travel-advice-publisher/blob/master/app/presenters/search_payload_presenter.rb).
+Published via the publishing API
```
MatMoore: We could probably just delete this page from the docs. I just wrote it so I could quickly compare what the apps were doing

lib/search_config.rb
```diff
@@ -102,7 +101,7 @@ def searcher
 
   def combined_index_schema
     @combined_index_schema ||= CombinedIndexSchema.new(
-      content_index_names,
+      all_index_names,
```
MatMoore: Why this change?

dwhenry: this all worked fine previously as govuk used the same field definitions as mainstream, however now that the mainstream schema has been removed this breaks in fun ways - like filter terms are removed as they lived on the mainstream schema. :)

doc/content_overview.md
dwhenry: done

doc/content_overview.md
dwhenry: reverted - nope.. removed

config/query/boosting.yml
dwhenry: removed

lib/search_config.rb
MatMoore: I was confused by the definition of `all_index_names`:

```
  def all_index_names
    # this is used to process data in the rake file when `all` is passed in as previous we skipped `govuk`
    # we can't update index_names at this stage as it is used in multiple spots including the index filtering
    content_index_names + auxiliary_index_names + [govuk_index_name]
  end
```

Is it ok to include `auxiliary_index_names` here and in the test helpers?

dwhenry: umm.. no that is a good point.

I thought that was content + govuk only.. let me correct that

spec/support/integration_spec_helper.rb
```diff
@@ -205,12 +205,12 @@ def stub_message_payload(example_document, unpublishing: false)
 private
 
   def build_sample_documents_on_content_indices(documents_per_index:)
-    SearchConfig.instance.content_index_names.each do |index_name|
+    SearchConfig.instance.all_index_names.each do |index_name|
```
MatMoore: this still says `all_index_names` - not sure if it matters

spec/unit/schema/combined_index_schema_spec.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.describe CombinedIndexSchema do
   before do
     search_config = SearchConfig.new
-    @index_names = search_config.content_index_names
+    @index_names = search_config.all_index_names
```
MatMoore: and here

MatMoore (approved):
LGTM other than a couple of test things.

PR #1136, opened 2018-01-04
Correct insert stemmed_query_as_term for metasearch index

dwhenry:
This value is generated on insertion which was missed from the initial
implementation. This creates a bug which effects multiple word stemmed
queries.

https://trello.com/c/9DQzn47B/554-bets-bets-bug-with-stemmed-car-hire-query-matching-any-queries-containing-car

### DIFF
--- a/lib/metasearch_index/inserter.rb
+++ b/lib/metasearch_index/inserter.rb
@@ -24,9 +24,17 @@ module MetasearchIndex
         {
           exact_query: @document["exact_query"],
           stemmed_query: @document["stemmed_query"],
-          stemmed_query_as_term: @document["stemmed_query_as_term"],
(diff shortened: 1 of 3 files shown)


lib/metasearch_index/inserter.rb
```diff
@@ -24,10 +24,18 @@ def document
         {
           exact_query: @document["exact_query"],
           stemmed_query: @document["stemmed_query"],
-          stemmed_query_as_term: @document["stemmed_query_as_term"],
+          stemmed_query_as_term: @document["stemmed_query"].presence && " #{analyzed_stemmed_query} ",
```
MatMoore, line 27: is this the same as `" #{analyzed_stemmed_query} if @document["stemmed_query"]`?

Also, how come there's spaces around it?

MatMoore (approved):
Looks good as far as I can tell, but not familiar with how this is used by rummager.

PR #1138, opened 2018-01-04
Do not allow actions to be made to the govuk index via the Rummager API

Rosa-Fox:
DO NOT MERGE - until [the code to remove mainstream from Rummager](https://github.com/alphagov/rummager/pull/1135) has been merged. Then the tests here should pass. 

[Trello](https://trello.com/c/RteOXf9T/555-do-not-allow-actions-to-be-made-to-the-govuk-index-through-the-rummager-apiI)

- We have moved over to a new indexing process from publishing applications via the publishing-api.
- As a result, the only formats that use the old process (aka indexing via the Rummager API) are formats sent from Whitehall to the `government` index.
- This commit prevents actions from being made to the `govuk` index via the Rummager API.
- Actions to the retired `mainstream` index are already prevented [here](https://github.com/alphagov/rummager/blob/master/lib/rummager/app.rb#L26).

### DIFF
--- a/lib/rummager/app.rb
+++ b/lib/rummager/app.rb
@@ -28,6 +28,12 @@ class Rummager < Sinatra::Application
     halt(404)
   end
 
+  def prevent_access_to_govuk
+    if index_name == 'govuk'
+      halt(403, "Actions to govuk index are not allowed via this endpoint, please use the message queue to update this index")
+    end
+  end
+
   def index_name
     @index_name ||= params["index"]
   end
@@ -166,6 +172,7 @@ class Rummager < Sinatra::Application
 
   # Insert (or overwrite) a document
   post "/:index/documents" do
+    prevent_access_to_govuk
     request.body.rewind
     documents = [JSON.parse(request.body.read)].flatten.map { |hash|
       hash["_type"] ||= "edition"
@@ -189,10 +196,12 @@ class Rummager < Sinatra::Application
   end
 
   post "/:index/commit" do
+    prevent_access_to_govuk
     simple_json_result(current_index.commit)
   end
 
   delete "/:index/documents/*" do
+    prevent_access_to_govuk
     document_link = params["splat"].first
 
     if (type = get_type_from_request_body(request.body))
@@ -223,6 +232,7 @@ class Rummager < Sinatra::Application
 
   # Update an existing document
   post "/:index/documents/*" do
+    prevent_access_to_govuk
     document_id = params["splat"].first
     updates = request.POST
     Indexer::AmendWorker.perform_async(index_name, document_id, updates)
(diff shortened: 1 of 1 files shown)


lib/rummager/app.rb
```diff
@@ -26,6 +26,12 @@ def current_index
     halt(404)
   end
 
+  def prevent_access_to_govuk
+    if index_name == 'govuk'
+      halt(403, "Actions to govuk index must be received by Rummager via the message queue, rather than the Rummager API")
+    end
+  end
+
```
dwhenry: Did you look at a before filter as an option:

```
before '/*' do
  if index_name
    if index_name =~ /^mainstream/
      halt(500, "The mainstream index has been removed and further updates are blocked")
    elsif index_name =~ /^govuk/
      halt(403, "Actions to govuk index are not allowed via this endpoint, please use the message queue to update this index")
    end
  end
end
```

dwhenry (approved):
It will work, but as per my comment I would prefer to use a before block that way I don't have to litter this check for each endpoint.

lib/rummager/app.rb
Rosa-Fox: @dwhenry I like the idea and think it is nice to also have a message that describes why actions can't be made to `mainstream` compared to just having the `SearchIndices::NoSuchIndex` error. I've tried to get this to work but it won't due to the params not being available to get the `index_name` from. I had a look at `before_filter` but it doesn't look like it will work for the Sinatra style routes. I can get it to work with `'/*'`  if I make it an `after` action as opposed to `before`. It does work but isn't ideal to run code and then prevent it, but I do think it is better than what I had before. What do you think?

Rosa-Fox: We investigated this further and it is working on [this branch](https://github.com/alphagov/rummager/compare/namespace-to-prevent-govuk-access-through-rummager-api). 

The workaround was to make use of a Sinatra namespace. Unfortunately this clashed with the rack namespace. It can be made to work (see `sinatra-namespace.rb` [here](https://github.com/alphagov/rummager/commit/6cca57b9113f4c4c80f06cb0c54ece19eeb11c3c#diff-31da4034a6a3ec4e21b0ffc240a45cdb)) but we are not sure if having to understand this extra complexity for people developing on this in the future is really worth it. Therefore we think it could be better to stick with the original approach.

@MatMoore @suzannehamilton would be great to hear either of your opinions on this. Thanks!

MatMoore: Your reasoning makes sense to me and I'm fine with a bit of duplication if it's simpler. These controllers don't change very often.

I can't really comment on the idiomatic-ness though as I haven't used sinatra much.

Rosa-Fox: 👍

PR #1151, opened 2018-01-12
Add ADR for bulk insertion of documents

dwhenry:


doc/arch/adr-007-batch-processing-of-rabbit-mq-messages.md
```diff
@@ -0,0 +1,74 @@
...
+
+We know that this causes a bottleneck with ES, as it performs balancing work 
+after each insert, it is for this reason that the [ES documentation](link here) recommends doing bulk
+insertion of data rather than individual documents. 
+
+We have have already implemented the following items to support batch processing:
+
+* Handling of batches of messages in the Rummager processor. This requeues all messages if 
+any of them fail during the ES update.
+* Batching of messages as they are read off rabbit MQ by the [govuk_message_queue_consumer]()
```
suzannehamilton: Missing link target.

doc/arch/adr-007-batch-processing-of-rabbit-mq-messages.md
```diff
@@ -0,0 +1,74 @@
...
+### New process 
+
+Processing 100 messages at a time with a 5 second timeout for approx 10000 documents (employment_tribunal_decision format)
+
+![Queue Size](images/100-queue.png)
+![Free CPU](images/100-free-cpu.png)
+
+### New process in bulk
+
+Processing 100 messages at a time with a 5 second timeout for approx 10000 documents (employment_tribunal_decision format)
```
suzannehamilton: What's the difference between this and "New process"? The descriptions are the same but the graphs are quite different.

doc/arch/adr-007-batch-processing-of-rabbit-mq-messages.md
```diff
@@ -0,0 +1,74 @@
...
+![Sidekiq Errors](images/100-bulk-errors.png)
+
+In addition I observed a number of SigTerm exceptions from the worker processes in Sentry
+which was unexpected.
+ 
+## Action plan
+
+The following items need to be completed before this can be rolled out:
+
+* Investigate the cause of the sentry SigTerm errors
```
suzannehamilton, line 71: Do we need to investigate the dependent app 500s as well? Do we see 500s when we _don't_ batch the messages?

boffbowsh:
Do we need to pull messages off the RabbitMQ queue then immediately requeue them in Sidekiq? I get it's for retry logic, but if we don't ack them in Rabbit, they'll get retried anyway if the consumer goes away.

MatMoore:
@boffbowsh I'm not following... are you asking why we're using sidekiq at all? There's some more context on that here: https://github.com/alphagov/rummager/blob/master/doc/arch/adr-001-use-of-both-rabbitmq-and-sidekiq-queues.md

Anything not acked in Rabbit will be retried but it will remain at the front of the queue and potentially block new messages from being processed. When we built this in q2, we saw sidekiq's exponential backoff behaviour as a good thing, since there are a lot of reasons indexing can fail, and it can be content specific.

boffbowsh:
Yeah ok, that makes sense 👍

doc/arch/adr-007-batch-processing-of-rabbit-mq-messages.md
dwhenry: Sorry this runs with about 1250k of records (I think) - basically the top 5/6 formats from mainstream by volume. - I'll update this to make it clear

doc/arch/adr-007-batch-processing-of-rabbit-mq-messages.md
dwhenry: fixed

doc/arch/adr-007-batch-processing-of-rabbit-mq-messages.md
dwhenry: We do see some either way and I don't feel the volume are worth investigating which is the reason for showing that graph - i.e. the extra load of processing @ twice the speed is not affecting downstream systems.

I will add a comment like this to the doc.

suzannehamilton (approved):

PR #1157, opened 2018-01-18
Add rake task to setup queues used by workers

RyanMacG:
We rely on the RabbitMQ queues that the workers use to be setup and
working for E2E testing. This ensure that the exchange and worker queues
are setup and ready for use.


Dockerfile
```diff
@@ -5,6 +5,7 @@ ENV GOVUK_APP_NAME rummager
 ENV REDIS_HOST redis
 ENV ELASTICSEARCH_URI http://elasticsearch:9200
 ENV PORT 3009
+ENV RABBITMQ_URL amqp://guest:guest@rabbitmq:5672
```
kevindew: Should this be a separate commit?

lib/tasks/message_queue.rake
```diff
@@ -6,6 +6,16 @@
 require 'rummager'
 
 namespace :message_queue do
+  desc "Create queues"
```
kevindew: I'd suggest adding a bit more explanation here e.g. "Create the queues that Rummager uses with Rabbit MQ" as it doesn't really add a lot being same as the task name.

lib/tasks/message_queue.rake
```diff
@@ -6,6 +6,16 @@
...
+  task :create_queues do
+    bunny = Bunny.new(ENV["RABBITMQ_URL"])
+    channel = bunny.start.create_channel
+    exch = Bunny::Exchange.new(channel, :topic, "published_documents")
+    channel.queue("rummager_to_be_indexed").bind(exch, routing_key: "*.links")
+    channel.queue("rummager_bulk_reindex").bind(exch, routing_key: "*.bulk.reindex")
+    channel.queue("rummager_govuk_index").bind(exch, routing_key: "*.*")
```
kevindew, line 18: Do we need to do any binding here?

I wonder if we should be using [#queue_declare](http://www.rubydoc.info/github/ruby-amqp/bunny/Bunny%2FChannel%3Aqueue_declare) ?

kevindew:
@MatMoore I'm tagging you here as I'm expecting you might still be the closest dev to this app - do correct me if I'm wrong.

At the moment there is a reliance on puppet to create the queues that rummager uses. This causes us a problem when we're running in an environment without puppet (which for us is [publishing-e2e-tests](https://github.com/alphagov/publishing-e2e-tests)) so want to put them in an app that's close to the queues (which is a bit tricky since they're for inter app communication). 

Does this seem good to you?

lib/tasks/message_queue.rake
RyanMacG: The queue will need to be bound to the exchange otherwise messages from the exchange won't be routed into the queue.  Are you proposing the binding gets done elsewhere?

We could switch over to ` queue_declare `, but ` queue ` command will have the benefit of being idempotent.

kevindew: Ah sure, I thought the govuk_message_queue_consumer did the bindings but I see that I'm wrong. Don't mind me 😄

Dockerfile
RyanMacG: It's only really relevant as part of the queue work as until now rummager didn't need to know about this ENV var until now

kevindew: Oh right I see as we use: https://github.com/alphagov/publishing-e2e-tests/blob/master/docker-compose.yml#L101-L104 as the way to connect.

It doesn't strike me as ideal to have different env vars that achieve the same task. 
Could we swap the Bunny.new to be either: `Bunny.new(GovukMessageQueueConsumer::RabbitMQConfig.from_environment(ENV))` or have the env vars inputted as a hash?

Ideally it'd be better still if GovukMessageQueueConsumer supported the RABBITMQ_URL env var, though we've had problems trying to make use of them since they support only a single host and we currently rely on a round robin collection of hosts (if I remember that stuff correctly)

MatMoore:
@kevindew we did used to set up bindings in the consumer, but changed it to puppet here: https://github.com/alphagov/govuk-puppet/commit/fc2ee2f807ac673c6ddb74948d21e405fe6f9358#diff-ab645e3a9fa5d9ee2104bac43666c3b0

> We are going to stop handling queue creation and binding in the consumer
processes, since it is too easy to mess up and leave a unintended binding
around.

I think that in the case where queues are durable, it's easier to monitor things if it's all managed in puppet, but if this makes things too complicated feel free to change it.

I'm also wondering if E2E testing is a special case where we should set up transient queues instead of durable ones. (I may be using the wrong terminology here - I mean the ones that go away when the consumer disconnects).

kevindew:
Thanks @MatMoore - Yeah I did some reading through that change and it sounded like relying on apps to build them automatically seemed to risk issues.

It's no bother running the rake task to create them separately. I think we're ok using the durable ones as we basically nuke RabbitMQ each time we build anyway, and best to keep it as close to production usage as possible.

kevindew (approved):

PR #1158, opened 2018-01-18
Add residential property tribunal decisions schema

elenatanasoiu:
The Residential Property Tribunal Judiciary publishes decisions that describe the outcome of a residential property tribunal case. This commit adds the schema for such a decision. 

For: https://trello.com/c/qzSMAFHZ/6-3-moj-residential-tribunals-finder

### DIFF
--- a/config/schema/elasticsearch_types/residential_property_tribunal_decision.json
+++ b/config/schema/elasticsearch_types/residential_property_tribunal_decision.json
@@ -1,0 +1,7 @@
+{
+  "fields": [
+    "tribunal_decision_category",
+    "tribunal_decision_sub_category",
+    "tribunal_decision_decision_date"
+  ]
+}
--- a/config/schema/indexes/govuk.json
+++ b/config/schema/indexes/govuk.json
@@ -21,6 +21,7 @@
     "medical_safety_alert",
     "policy",
     "raib_report",
+    "residential_property_tribunal_decision",
     "service_manual_guide",
     "service_manual_topic",
     "service_standard_report",


config/schema/elasticsearch_types/residential_property_tribunal_decision.json
```diff
@@ -0,0 +1,7 @@
+{
+  "fields": [
+    "tribunal_decision_category",
+    "tribunal_decision_category_name",
+    "tribunal_decision_decision_date"
```
h-lame: We don't have a `tribunal_decision_category_name` field, but we do have a `tribunal_decision_sub_category`.  We should also check with @dwhenry and @suzannehamilton if we need to add the `expanded_search_result_fields` attribute here to define how to map the parameterised values in `tribunal_decision_category` and `tribunal_decision_sub_category` into human readable values.  For example like we have in https://github.com/alphagov/rummager/blob/master/config/schema/elasticsearch_types/aaib_report.json (not sure why we don't have it in https://github.com/alphagov/rummager/blob/master/config/schema/elasticsearch_types/employment_appeal_tribunal_decision.json though).  You can rip these values from the finder definition in https://github.com/alphagov/specialist-publisher/pull/1136)

config/schema/indexes/govuk.json
```diff
@@ -24,6 +24,7 @@
     "service_manual_guide",
     "service_manual_topic",
     "service_standard_report",
+    "residential_property_tribunal_decision",
```
suzannehamilton: This should be in alphabetical order or @brenetic will be sad.

config/schema/elasticsearch_types/residential_property_tribunal_decision.json
suzannehamilton: I'm not actually sure where this is used. According to the [docs](https://github.com/alphagov/rummager/blob/master/doc/schemas.md#elasticsearch-document-types), `expanded_search_result_fields` is deprecated, and the labels are also defined in the [finder config](https://github.com/alphagov/specialist-publisher/blob/118db285035f46048063af5085da1f76b46261f1/lib/documents/schemas/aaib_reports.json). But @boffbowsh [had to update one recently](https://github.com/alphagov/rummager/pull/1150) so this information must be used somewhere.

Paul, do you know any more about this?

config/schema/indexes/govuk.json
brenetic: 👍 😢

config/schema/elasticsearch_types/residential_property_tribunal_decision.json
boffbowsh: I'm afraid I don't, I was (perhaps blindly) copying prior art.

config/schema/indexes/govuk.json
elenatanasoiu: Had alphabetical blindness there. Fixed now.

config/schema/elasticsearch_types/residential_property_tribunal_decision.json
elenatanasoiu: I'll leave it out then

elenatanasoiu: Thanks guys!

h-lame (approved):

PR #1169, opened 2018-01-31
Add expanded_search_result_fields to residential_property_tribunal_decision schema

h-lame:
For: https://trello.com/c/qzSMAFHZ/6-3-moj-residential-tribunals-finder

This allows rummager to translate the computer-readable values from the
document into human-readable values when it presents the search results as
json.  This may be a bit of tech debt that we should clean up as we
thought this config was deprecated.  The approach under discussion is that
we either:

1. Push both human and computer readable values into the documents
   provided to rummager
2. Make finder-frontend do the same as government-frontend and use the
   facet definitions of the finder to do the translations for us.

See alphagov/specialist-publisher#1147 and alphagov/govuk-content-schemas#713 for comparisons.

### DIFF
--- a/config/schema/elasticsearch_types/residential_property_tribunal_decision.json
+++ b/config/schema/elasticsearch_types/residential_property_tribunal_decision.json
@@ -3,5 +3,216 @@
     "tribunal_decision_category",
     "tribunal_decision_sub_category",
     "tribunal_decision_decision_date"
-  ]
+  ],
+
+  "expanded_search_result_fields": {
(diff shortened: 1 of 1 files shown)


elenatanasoiu (approved):

suzannehamilton (approved):
Once we've decided on an approach, it'd be good to write a short [ADR](https://github.com/alphagov/rummager/tree/master/doc/arch) so we remember our reasoning.

PR #1170, opened 2018-01-31
Add step_by_step_nav

sihugh:
step_by_step_nav is the new name for a task_list.  task_lists will still exist for a little while, but we'll clean them up when they are no more.


deborahchua (approved):
👍

PR #1175, opened 2018-02-14
Zip page traffic data on sidekiq

dwhenry:
As we need to process the entire index onto the sidekiq queue we have added
zipping which see a reduction of about 60% in the amount of data put into
memory. This solution has been coded into teh actual worker so it is easy to
see the deflate/inflate methods being called.

https://trello.com/c/JHcwDJc0/81-compress-rummager-data-in-sidekiq-when-updating-pageviews

### DIFF
--- a/lib/govuk_index/page_traffic_worker.rb
+++ b/lib/govuk_index/page_traffic_worker.rb
@@ -4,7 +4,14 @@ module GovukIndex
     QUEUE_NAME = 'bulk'.freeze
     sidekiq_options queue: QUEUE_NAME
 
-    def perform(records, destination_index)
+    def self.perform_async(records, destination_index)
+      data = Base64.encode64(Zlib::Deflate.deflate(Sidekiq.dump_json(records)))
+      super(data, destination_index)
+    end
+
+    def perform(data, destination_index)
+      records = Sidekiq.load_json(Zlib::Inflate.inflate(Base64.decode64(data)))
+
       actions = Index::ElasticsearchProcessor.new(client: GovukIndex::Client.new(timeout: BULK_INDEX_TIMEOUT, index_name: destination_index))
 
       records.each_slice(2) do |identifier, document|
--- a/lib/rummager.rb
+++ b/lib/rummager.rb
@@ -30,6 +30,7 @@ require 'time'
 require 'unf'
 require 'uri'
 require 'yaml'
+require 'zlib'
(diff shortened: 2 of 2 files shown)


lib/govuk_index/page_traffic_worker.rb
```diff
@@ -4,7 +4,14 @@ class PageTrafficWorker < Indexer::BaseWorker
     QUEUE_NAME = 'bulk'.freeze
     sidekiq_options queue: QUEUE_NAME
 
-    def perform(records, destination_index)
+    def self.perform_async(records, destination_index)
+      data = Base64.encode64(Zlib::Deflate.deflate(Sidekiq.dump_json(records)))
+      super(data, destination_index)
+    end
+
+    def perform(data, destination_index)
```
h-lame, line 12: I don't really know enough about what future developers might expect from a sidekiq job, but is it too weird to have the different signatures for `perform` and `perform_async`?

h-lame (commented):
I think the intention, and 60% reduction in size of the jobs, is great.  I'd like a 2nd opinion on the `perform` vs. `perform_async` thing though - I reckon it's pragmatic and good to have it encapsulated in the worker.  It is unusual though, but I don't have any better ideas.

lib/govuk_index/page_traffic_worker.rb
dwhenry: I did look at using a Sidekiq middleware instead, but it was very easy to understand that compress was happening, and as every middleware is used by every worker that made it all worse. Should also check if `perform_now` would just break.

lib/rummager.rb
```diff
@@ -30,6 +30,7 @@
 require 'unf'
 require 'uri'
 require 'yaml'
+require 'zlib'
```
boffbowsh, line 33: It'd be better to move this into the other file, I think. Keep the require where it's used.

boffbowsh (changes requested):
Small change but I think I agree with approach otherwise

lib/rummager.rb
dwhenry: we previously moved all requires that are used in the app to this file for clarity - basically, it helped identify strange includes in the project. and doesn't affect load time as they are required on load (in production). 

I personally prefer this approach.

boffbowsh (approved):

PR #1182, opened 2018-02-21
Revert using foreman in dev environment

kevindew:
Wider context: https://github.com/alphagov/publishing-e2e-tests/pull/202

This removes foreman from the Gemfile and instead installs foreman
separately via `gem install` in the Dockerfile so that it can be used in the
docker environment.

The reason for this is that running everything via unicorn in dev can
cause some wtfs (such as breaking better_errors) and installing foreman
via Gemfile does contradict some contenious advice given by the foreman
gem maintainer.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 FROM ruby:2.4.2
 RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential && apt-get clean
+RUN gem install foreman
 
 ENV GOVUK_APP_NAME rummager
 ENV REDIS_HOST redis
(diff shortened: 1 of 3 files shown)


adrianclay:
> This removes foreman from the Gemfile and from the startup.sh usage

`startup.sh` is using [mr-sparkle](https://github.com/MicahChalmer/mr-sparkle) which depends on Unicorn.

Is it using the bundle version of Unicorn or a gobal version?

kevindew:
> startup.sh is using mr-sparkle which depends on Unicorn.
> 
> Is it using the bundle version of Unicorn or a gobal version?

Whoops - copy/paste error to mention startup.sh - just not touched that.

thomasleese (approved):

PR #1187, opened 2018-02-27
Add new content purpose subgroup and supergroup fields

fmlharrison:
This is as a result of a change made to govuk_document_types
to add the new content_purpose_supergroup and content_purpose_subgroup
supertypes that will be used by the topic pages as part of the
navigation team's work in Q4


Gemfile.lock
```diff
@@ -272,4 +272,4 @@ DEPENDENCIES
   whenever (~> 0.10.0)
 
 BUNDLED WITH
-   1.16.0
```
vanitabarrett: We try and avoid committing changes to the Bundled With version.

vanitabarrett (changes requested):
One small comment on the Gemfile.lock. Otherwise looks good. 

I also noticed these files when digging around for uses of the old `content_purpose_document_supertype` in Github - do we need to update these files too?

https://github.com/alphagov/rummager/blob/6b668cc80132176521202127f3740f1facc8d045/lib/govuk_index/presenters/common_fields_presenter.rb

https://github.com/alphagov/rummager/blob/13502f89f3845b72c8080355f51bed6fa019cd9a/lib/govuk_index/presenters/elasticsearch_presenter.rb

Gemfile.lock
fmlharrison: ok thanks

fmlharrison:
@vanitabarrett I added in the missing content_purpose_sub and supergroup into the files you mentioned above

vanitabarrett (approved):

config/schema/base_elasticsearch_type.json
```diff
@@ -33,6 +33,8 @@
     "title",
     "topic_content_ids",
     "updated_at",
-    "user_journey_document_supertype"
+    "user_journey_document_supertype",
+    "content_purpose_supergroup",
```
leenagupte: It's a minor thing, but I'd put these in alphabetical order.

fmlharrison: Ok cool. Cheers

leenagupte: It all looks really good!

leenagupte (approved):
👍 💯

PR #1196, opened 2018-03-13
Publish advanced search finder

steventux:
Trello: https://trello.com/c/ACZBT7Dk/120-build-a-way-to-publish-the-new-finder

Adds a Rake task and underlying class to publish a finder content item from YAML config (similar to the mechanics of publishing finders in Specialist Publisher).

### DIFF
--- a/config/advanced-search.yml
+++ b/config/advanced-search.yml
@@ -1,0 +1,59 @@
+---
+base_path: "/search/advanced"
+content_id: 3df77dea-00c5-43f0-8f31-d08b8bd2a4d6
+content_purpose_document_supertype: navigation
+document_type: finder
+locale: en
+name: Advanced search
+phase: live
+pre_production: true
+publishing_app: rummager
+rendering_app: finder-frontend
+schema_name: finder
+search_user_need_document_supertype: government
+title: Latest on GOV.UK
+update_type: minor
+details:
+  document_noun: result
+  reject:
+    content_store_document_type:
+    - browse
+  show_summaries: true
+  facets:
+  - filter_key: content_purpose_subgroup
+    key: subgroup
+    name: Publication type
+    type: text
+    preposition: in
+    display_as_result_metadata: false
+    filterable: true
+    allowed_values: []
+  - key: public_timestamp
+    name: Publication date
+    short_name: '' 
+    type: date
+    preposition: published
+    display_as_result_metadata: true
+    filterable: true
+  - filter_key: taxons
+    key: topic
+    name: Taxonomy
+    type: hidden
+    preposition: ''
+    display_as_result_metadata: true
+    filterable: true
+    allowed_values: []
+  - filter_key: content_purpose_supergroup
+    key: group
+    name: Content purpose supergroup
+    type: hidden
+    preposition: ''
+    display_as_result_metadata: false
+    filterable: true
+    allowed_values: []
+  default_documents_per_page: 20
+routes:
+- path: "/search/advanced"
+  type: exact
+- path: "/search/advanced.json"
+  type: exact
--- a/config/advanced-search.yml
+++ b/config/advanced-search.yml
@@ -2,7 +2,7 @@
 base_path: "/search/advanced"
 content_id: 3df77dea-00c5-43f0-8f31-d08b8bd2a4d6
 content_purpose_document_supertype: navigation
-document_type: finder
+document_type: search
 locale: en
 name: Advanced search
 phase: live
--- a/lib/publishing_api_finder_publisher.rb
+++ b/lib/publishing_api_finder_publisher.rb
@@ -1,0 +1,80 @@
+require "gds_api/publishing_api_v2"
+
+class PublishingApiFinderPublisher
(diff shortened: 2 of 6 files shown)


lib/publishing_api_finder_publisher.rb
```diff
@@ -0,0 +1,70 @@
...
+    @finders = finders
+    @logger = logger
+  end
+
+  def call
+    finders.map do |finder|
+      if pre_production?(finder)
+        publish(finder)
+      else
+        logger.info("Not publishing #{finder[:file]['name']} because it is pre_production")
```
leenagupte: Should this read, "because it is not pre_production"?

lib/tasks/advanced_search.rake
```diff
@@ -0,0 +1,11 @@
+require "publishing_api_finder_publisher"
+
+namespace :publishing_api do
+  desc "Publish advanced-search finder."
+  task :publish_advanced_search_finder do
+    finder = YAML.load_file("config/advanced-search.yml")
+    timestamp = Time.now.iso8601
+
+    PublishingApiFinderPublisher.new([{ file: finder, timestamp: timestamp }]).call
+  end
```
leenagupte: `PublishingApiFinderPublisher` expects a logger as a potential second parameter, not a timestamp. The timestamp is already being calculated in the previous commit.

lib/publishing_api_finder_publisher.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+    end
+  end
+
+  def publishing_api
+    @publishing_api ||= GdsApi::PublishingApiV2.new(
```
leenagupte: This method is only being called by private methods, so should probably be private itself.

lib/publishing_api_finder_publisher.rb
leenagupte: Also, what is `finder[:file]`?

leenagupte (dismissed):
A few comments.  Also, can you add something to README about what these pages are for and how to publish them.

spec/unit/publishing_api_finder_publisher_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+        allow(GdsApi::PublishingApiV2).to receive(:new).and_return(publishing_api)
+        allow(publishing_api).to receive(:put_content)
+        allow(publishing_api).to receive(:patch_links)
+        allow(publishing_api).to receive(:publish)
+
+        instance.call
+      end
+
+      it "drafts the finder" do
+        puts payload
```
fmlharrison: Is there a reason you have puts the payload here?

fmlharrison (dismissed):

config/advanced-search.yml
```diff
@@ -0,0 +1,70 @@
+---
+analytics_identifier: 
+base_path: "/search/advanced"
+content_id: 3df77dea-00c5-43f0-8f31-d08b8bd2a4d6
+content_purpose_document_supertype: navigation
+document_type: finder
```
tijmenb: If we make this `search` it won't be indexed.

lib/tasks/advanced_search.rake
```diff
@@ -0,0 +1,11 @@
+require "publishing_api_finder_publisher"
+
+namespace :publishing_api do
```
tijmenb, line 3: Would be good to add a schema test for the payload here too.

lib/tasks/advanced_search.rake
steventux: The second (`:logger`) param is actually omitted here. The only param passed is an array of hashes (with one item in it). Mostly because this class was based on something similar from SpecialistPublisher which publishes multiple finders at once. Arguably this doesn't need to happen here.

lib/tasks/advanced_search.rake
steventux: Added here: https://github.com/alphagov/rummager/pull/1196/commits/c8a2b2d587ee5af9059968cdc52de207fe244f1e#diff-794b684fdca56d301fb5ed3d322928e2R89 ~this fails if I change the advanced search data to be `document_type: search` though.
Errors: `The property '#/details' contains additional properties [\"document_noun\", \"show_summaries\", \"f...` Perhaps this is how I'm configuring the schema test helper?~
Made the necessary schema changes now - all good.

lib/tasks/advanced_search.rake
leenagupte: I'm going to dismiss my review, so that I'm not a blocker, but there are couple of things that I would change to make it a bit easier to follow:

You can't pass any parameters to the rake task, so I'd move the loading of the YAML file and setting of the default timestamp to `PublishingApiFinderPublisher`.

However, if you do still decide to pass a finder object from the rake task; are you expecting to publish multiple finders using `PublishingApiFinderPublisher`? If not, I'd change finder parameter from an array to a single finder. 
 
I'd also extract `finder[:file]` to make `PublishingApiFinderPublisher` easier to read.

leenagupte (approved):
👍 LGTM!

steventux:
Thanks @leenagupte I've simplified things to pass one finder and a few other optional params to the publisher class. Just tested it on integration.

PR #1209, opened 2018-03-27
Update advanced search content item

steventux:
* Removes an empty label from the publication date facet on the advanced search finder
* Removes redundant `pre_production` check in the finder publisher


leenagupte (approved):
👍 LGTM

PR #1214, opened 2018-04-04
Set encoding to UTF-8

thomasleese:
This is an attempt to fix a problem we're seeing in Sentry where Sidekiq cannot JSON encoding a hash because it has non ASCII characters in it.

https://sentry.io/govuk/app-rummager/issues/450374046/

### DIFF
--- a/lib/rummager/config.rb
+++ b/lib/rummager/config.rb
@@ -31,3 +31,6 @@ Raven.configure do |config|
 
   use Raven::Rack
 end
+
+Encoding.default_external = "UTF-8"
+Encoding.default_internal = "UTF-8"


MatMoore:
Are you sure we need to set both of these? I'm not familiar with how these work but the docs say this for the external one:

> You should not set ::default_external in ruby code as strings created before changing the value may have a different encoding from strings created after the value was changed., instead you should use ruby -E to invoke ruby with the correct default_external.

https://ruby-doc.org/core-2.2.0/Encoding.html#method-c-default_external-3D

On production these are set like this:
```
irb(main):001:0> Encoding.default_external
=> #<Encoding:UTF-8>
irb(main):002:0> Encoding.default_internal
=> nil
```

and the system encoding also seems to be UTF-8, which I would have thought would be enough for ruby to do the right thing 😕 
```
locale
LANG=en_US.UTF-8
LANGUAGE=en_US:
LC_CTYPE="en_GB.UTF-8"
LC_NUMERIC="en_GB.UTF-8"
LC_TIME="en_GB.UTF-8"
LC_COLLATE="en_GB.UTF-8"
LC_MONETARY="en_GB.UTF-8"
LC_MESSAGES="en_GB.UTF-8"
LC_PAPER="en_GB.UTF-8"
LC_NAME="en_GB.UTF-8"
LC_ADDRESS="en_GB.UTF-8"
LC_TELEPHONE="en_GB.UTF-8"
LC_MEASUREMENT="en_GB.UTF-8"
LC_IDENTIFICATION="en_GB.UTF-8"
LC_ALL=en_GB.UTF-8
```

thomasleese:
To be honest I wasn't completely confident in this as a solution, but I basically found https://github.com/mperham/sidekiq/issues/1597#issuecomment-44760009 and a lot of other examples of people setting these to solve unicode issues when Sidekiq turns a hash into JSON. It does indeed look like production is configured to do the right thing though, so maybe the exception is happening for a different reason.

MatMoore:
I guess it's worth testing on integration but I'd be interested to know what sidekiq is doing here. If this doesn't fix it maybe we could put in some diagnostic logging to check the encoding of the job args?

MatMoore (approved):

thomasleese:
Not working on this for the moment.

barrucadu:
So this branch does seem to solve the problem on integration.  With `master` there are encoding errors coming in every few seconds to `sidekiq.log`, but with this branch there aren't any.

Before deploying this, I did a quick sanity check that sidekiq is actually using UTF-8 normally:

```
$ cat /tmp/sidekiq.yml 
---
:require: /tmp/sidekiq.rb

$ cat /tmp/sidekiq.rb 
puts "external: #{Encoding.default_external}"
puts "internal: #{Encoding.default_internal}"

$ bundle exec sidekiq -C /tmp/sidekiq.yml
external: UTF-8
internal: 
```

Which looks reasonable to me.

thomasleese:
@barrucadu thanks for doing this, it still seems very weird that we need to do this though 😢 feel free to reopen a new PR and I'll approve.

PR #1215, opened 2018-04-04
Don't send HTTPNotFound errors on the Publishing API to sentry

thomasleese:
This fills up Sentry with a lot of noise about a situation that we cannot fix or recover from.

### DIFF
--- a/lib/indexer/links_lookup.rb
+++ b/lib/indexer/links_lookup.rb
@@ -83,6 +83,11 @@ module Indexer
           }
         )
         raise Indexer::PublishingApiError
+      rescue GdsApi::HTTPNotFound => e
+        # If the Content ID no longer exists in the Publishing API, there isn't really much
+        # we can do at this point. There doesn't seem to be any compelling reason to record
+        # this in Sentry as there is no bug to fix.
+        @logger.error("HTTP not found error fetching expanded links for #{content_id}: #{e.message}")
       rescue GdsApi::HTTPErrorResponse => e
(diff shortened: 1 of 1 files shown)


MatMoore (approved):
👍

lib/indexer/links_lookup.rb
```diff
@@ -83,6 +83,12 @@ def find_links(content_id)
           }
         )
         raise Indexer::PublishingApiError
+      rescue GdsApi::HTTPNotFound => e
+        # If the Content ID no longer exists in the Publishing API, there isn't really much
+        # we can do at this point. There doesn't seem to be any compelling reason to record
+        # this in Sentry as there is no bug to fix.
+        @logger.error("HTTP not found error fetching expanded links for #{content_id}: #{e.message}")
+        raise Indexer::PublishingApiError
```
cbaines: What is the implication of raising `Indexer::PublishingApiError`, does this mean that this task will be retried?

MatMoore: I would expect so. I didn't notice this but I think it should just swallow the exception here?

thomasleese: Hmm, oh yeah, I guess it might be retried... I put that in because it's what the other `rescue` blocks do and [it doesn't get sent to Sentry](https://github.com/alphagov/rummager/blob/master/lib/rummager/config.rb#L24). Let me see if I can replace it with something else.

PR #1217, opened 2018-04-09
Allow more taxon faceting

tijmenb:
One of the x-factors with retirement of legacy topic is that they power the "services and information" pages like https://www.gov.uk/government/organisations/hm-revenue-customs/services-information.

I want to experiment with a similar page that uses the new taxonomy. For this we need to allow faceting and facet examples for both taxonomy fields. This is already allowed for topics ("specialist sectors") so that Collections can do this:

https://github.com/alphagov/collections/blob/6ebae1b5bf095c15258a26ce89882808093e0a04/app/lib/services_and_information_links_grouper.rb#L15-L19

### DIFF
--- a/lib/parameter_parser/base_parameter_parser.rb
+++ b/lib/parameter_parser/base_parameter_parser.rb
@@ -89,9 +89,11 @@ class BaseParameterParser
     manual
     navigation_document_supertype
     organisations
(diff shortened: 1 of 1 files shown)


cbaines (approved):
I don't have a lot of context on this, but it sounds good :+1:

PR #1222, opened 2018-04-10
Remove "navigation_document_supertype"

tijmenb:
This supertype is no longer used and has been superseded by https://docs.publishing.service.gov.uk/document-types/content_purpose_subgroup.html. Because the groupings aren't the same, we can't use that
in the boosting. To keep the boosting the same, we'll use the document type instead.

https://trello.com/c/6wRHXCyT


MatMoore (approved):
LGTM. Can you check in Kibana that nothing is actually requesting that field through the API? (eg `fields[]=navigation_document_supertype`, `filter_navigation_document_supertype`).

tijmenb:
I've checked Kibana, all good!

PR #1223, opened 2018-04-10
Show all content tagged to taxon and below

vanitabarrett:
Related to https://github.com/alphagov/finder-frontend/pull/467

We want to replace filter_taxons in the rummager query with filter_part_of_taxonomy_tree so that we get taxon content for the given taxon and below.

**Before:**
screen shot 2018-04-10 at 12 12 49

**After:**
screen shot 2018-04-10 at 12 12 57

### DIFF
--- a/config/advanced-search.yml
+++ b/config/advanced-search.yml
@@ -33,7 +33,7 @@ details:
     preposition: published
     display_as_result_metadata: true
     filterable: true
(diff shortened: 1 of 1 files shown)


leenagupte (commented):
👍 💯

leenagupte (approved):

PR #1224, opened 2018-04-11
Replace search_user_need_document_usertype boosting

tijmenb:
The removed `search_user_need_document_usertype` (https://github.com/alphagov/govuk_document_types/pull/44) contains these document types:

- answer
- guide
- local_transaction
- place
- simple_smart_answer
- smart_answer
- transaction
- travel_advice
- travel_advice_index

For each of those, I've multiplied the `search_user_need_document_usertype` boost with the existing boost. If I understand elasticsearch correctly, this way the boosts will stay the same.

`travel_advice_index` wasn't previously in the list, so it's not multiplied and gets 1.5.

https://trello.com/c/6wRHXCyT

### DIFF
--- a/config/query/boosting.yml
+++ b/config/query/boosting.yml
@@ -22,28 +22,27 @@ base:
     # Topics
     specialist_sector: 2.5
   content_store_document_type:
-    answer: 2.5
+    answer: 3.75
     contact: 2.5
     detailed_guide: 2.5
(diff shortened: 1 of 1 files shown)


MatMoore (approved):
👍 

(In theory) you should be able to verify the boosts are the same from the explain plan, which you can view by passing `debug=explain` to the API.

PR #1229, opened 2018-04-19
Enable the root_taxon link_type to be used in parts_of_taxonomy

1pretz1:
To be able to use the homepages content_id in a query using the
parts_of_taxonomy filter.

### DIFF
--- a/lib/govuk_index/presenters/expanded_links_presenter.rb
+++ b/lib/govuk_index/presenters/expanded_links_presenter.rb
@@ -81,7 +81,8 @@ module GovukIndex
     def parts_of_taxonomy(taxon_hash)
       parents = [taxon_hash["content_id"]]
 
-      direct_parents = taxon_hash.dig("links", "parent_taxons")
+      direct_parents = direct_parent_taxons(taxon_hash)
+
       while direct_parents && !direct_parents.empty?
         # There should not be more than one parent for a taxon. If there is,
         # make an arbitrary choice.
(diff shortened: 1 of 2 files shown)


lib/govuk_index/presenters/expanded_links_presenter.rb
```diff
@@ -81,15 +81,20 @@ def organisation_slugs(type)
     def parts_of_taxonomy(taxon_hash)
       parents = [taxon_hash["content_id"]]
 
-      direct_parents = taxon_hash.dig("links", "parent_taxons")
-      while direct_parents && !direct_parents.empty?
+      parent = taxon_hash
+
+      loop do
```
leenagupte: Just a minor comment. It would be easier to read this section if it was obvious what's being looped over. 

It could just be as simple as converting the `loop` back to the original `while` and move
getting the next parent to a function.

e.g.
```
direct_parents = parent_taxon(taxon_hash)
```
and in the `while` call:

```
direct_parents = parent_taxon(direct_parents)
```

and have `parent_taxon` return the result of the or:

```
def parent_taxon(taxon)
  taxon.dig("links", "parent_taxons") ||
    taxon.dig("links", "root_taxon") ||
    []
end
```

leenagupte (changes requested):

MatMoore:
Have you considered moving this logic into a gem? IMO you shouldn't have to know about how the taxonomy fits together to maintain the search code.

This code just needs to know two things:
- what taxons is the content item tagged to?
- what are the ancestor taxons of all those taxons?

We created [this gem](https://github.com/alphagov/govuk_taxonomy_helpers) a while back. It's only used by content tagger at the moment, but the intention was to make it a general purpose library for working with taxons. Is it still actively maintained?

1pretz1:
Yes govuk-taxonomy-helpers is still active and used, and i agree with you that it would be better for the search api not to contain specific information about the taxonomy. 

However i wouldn't know how to implement that without some thinking, but could look at doing this at a later point ?

leenagupte (approved):
👍 LGTM!

PR #1245, opened 2018-06-18
Add delete content result task

binaryberry:
This allows to delete a single result from the search results. This could also be accomplished using a [curl command](https://github.com/alphagov/rummager/blob/master/doc/content-api.md#delete-contentlinka-link), however running a rake task removes the need to know about the boxes to run the curl command on.

Paired with @steventux 

https://trello.com/c/j69WUORA/230-remove-duplication-from-search-results

### DIFF
--- a/lib/tasks/delete.rake
+++ b/lib/tasks/delete.rake
@@ -2,6 +2,23 @@ require 'rummager'
 
 namespace :delete do
   desc "
+  Delete a single content search result
+  Usage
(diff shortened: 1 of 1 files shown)


tijmenb:
Have you tried using search admin for this? It has a UI to delete search results.

binaryberry:
No, and that worked like a charm. Thanks Tijmen.

PR #1254, opened 2018-06-26
Add first_published_at to all content

sihugh:
This field is [sent by publishing-api](https://github.com/alphagov/govuk-content-schemas/blob/52b26b24cd105b1b23acd90915adfd59b4228a57/formats/shared/default_properties/notification.jsonnet#L5-L6) for everything (and [generated if it isn't present](https://github.com/alphagov/govuk-content-schemas/blob/d41b2b745393d229406f4dbb4060845a4445517d/formats/shared/definitions/publishing_api_base.jsonnet#L137) when an [application publishes something](https://github.com/alphagov/publishing-api/blob/1455c9e0a1fa5217f4857a9591fc86bc1e355a63/app/models/edition/timestamps.rb#L17).)

We're using rummager for more things like feeds, and it's useful to be able to tell if a piece of content is new, or just updated.

https://trello.com/c/PosIQiOn/45-we-no-longer-distinguish-between-latest-documents-updated-or-published-dates

### DIFF
--- a/config/schema/base_elasticsearch_type.json
+++ b/config/schema/base_elasticsearch_type.json
@@ -8,6 +8,7 @@
     "content_store_document_type",
     "description",
     "email_document_supertype",
+    "first_published_at",
     "format",
     "government_document_supertype",
     "indexable_content",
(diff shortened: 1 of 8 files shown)


rubenarakelyan (approved):

tijmenb (approved):
Looks sensible to me!

PR #1257, opened 2018-06-28
Increase max result count

sihugh:
We have over 1000 organisations.  These are commonly used as facets, so we need to be able to return all of them at once.  1500 is a bit arbitrary, but given that the org count only ever goes up, it's useful to have some headroom.


thomasleese (approved):
This seems fine, I'm sure we won't exceed millions of organisations... but I wonder if we could instead get the thing which uses them as facets to paginate correctly?

sihugh:
This would seem like a good approach for later.  I'll add it to the tech debt board

PR #1268, opened 2018-08-06
Add image hash to rummager

sihugh:
We want to be able to retrieve image information from search results for things like news_articles s that we can show them in navigation. (Example of a page that has an image section that we want: https://www-origin.integration.publishing.service.gov.uk/government/news/alternative-provision-progress-made-but-more-still-to-be-done )

Images are within the details hash and look like this:

```
"details" {
  "image": {
    "url": "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/image_data/file/41367/s300_holidygov.png",
    "caption": null,
    "alt_text": "Holiday image"
  }
}
```

An `opaque_object` is intended to be something that is not searchable and has an undefined format.  Given that we don't want to have to keep rummager schemas up to date if the format of an image section changes this seems OK.

The existing [`metadata` field also uses this type](https://docs.publishing.service.gov.uk/apis/search/fields.html)

https://trello.com/c/MoyJIc4n/36-index-image-information-for-news-pages

### DIFF
--- a/config/schema/elasticsearch_types/edition.json
+++ b/config/schema/elasticsearch_types/edition.json
@@ -15,6 +15,7 @@
     "has_command_paper",
     "has_official_document",
     "id",
+    "image",
     "important_to_policy",
     "is_historic",
     "is_political",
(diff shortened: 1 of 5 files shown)


sihugh:
We don't need all of this, so closing in favour of a simpler method

PR #1276, opened 2018-08-22
Make news_story indexable when published by Content Publisher

emmabeynon:
For https://trello.com/c/SlaLEmdQ/86-change-rummager-to-index-the-documents-were-publishing-l

The `news_story` document type will soon be publishable by Content Publisher as
well as Whitehall. Whitehall indexes its content to Rummager directly, whereas
Content Publisher will use the conventional way of relying on Publishing API to
take care of communicating with Rummager. In order to make it possible for
Content Publisher content to be indexable, we whitelist `news_story` as an
indexable format and specify Content Publisher as the publishing app for it.
`news_story` remains in the non_indexable list as well and we specify Whitehall
as the publishing app here. This approach will allow us to check the publishing
app for a piece of content and use this to determine whether it should be
indexed or not.

### DIFF
--- a/config/govuk_index/mapped_document_types.yaml
+++ b/config/govuk_index/mapped_document_types.yaml
@@ -25,6 +25,7 @@ mainstream_browse_page: edition # Collections Publisher
 manual: manual
 manual_section: manual_section
 medical_safety_alert: medical_safety_alert # Specialist Publisher
+news_story: edition
 place: edition
 policy: policy # Policy Publisher
 raib_report: raib_report # Specialist Publisher
--- a/config/govuk_index/migrated_formats.yaml
+++ b/config/govuk_index/migrated_formats.yaml
@@ -48,6 +48,8 @@ migrated:
 - statutory_instrument
 - step_by_step_nav
 - task_list
+- news_story: # not indexable when published by Whitehall
+    publishing_app: content-publisher
 - taxon:
   - /world/afghanistan
   - /world/albania
@@ -369,7 +371,8 @@ non_indexable:
 - ministerial_role
 - national_statistics
 - national_statistics_announcement
-- news_story
+- news_story: # indexable when published by Content Publisher
+    publishing_app: whitehall
 - notice
 - official_statistics
 - official_statistics_announcement
--- a/lib/govuk_index/migrated_formats.rb
+++ b/lib/govuk_index/migrated_formats.rb
@@ -5,7 +5,7 @@ module GovukIndex
     def non_indexable?(format, path, app)
       non_indexable_formats[format] &&
         (non_indexable_formats[format] == :all || non_indexable_formats[format].include?(path) ||
-          published_by_whitehall?(format, app))
+          published_by_non_indexable_app?(format, app))
     end
 
     def non_indexable_formats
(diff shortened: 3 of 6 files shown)


cbaines:
I'm not very clear on how Rummager works, but will the search function definitely work for news stories with the content in two different indexes?

emmabeynon:
@cbaines We know that search queries are performed across all indexes so that should be ok, but we have some unknowns around things like advanced search which we need to explore more (there's going to be a separate card written up for that).  At the moment we're not publishing Content Publisher content to production so we don't need to worry about surfacing it in things like advanced search yet, but it's something we'll need to solve before we get Content Publisher into private beta.

config/govuk_index/mapped_document_types.yaml
```diff
@@ -25,6 +25,7 @@ mainstream_browse_page: edition # Collections Publisher
 manual: manual
 manual_section: manual_section
 medical_safety_alert: medical_safety_alert # Specialist Publisher
+news_story: edition
```
kevindew, line 28: As we're not too sure on the why or the implications could we put this in a separate commit to document the uncertainty?

config/govuk_index/migrated_formats.yaml
```diff
@@ -48,6 +48,8 @@ migrated:
 - statutory_instrument
 - step_by_step_nav
 - task_list
+- news_story: # not indexable when published by Whitehall
+  - content-publisher
```
kevindew: maybe we could have this as an object of `publishing_app: content-publisher` as this might make the intent somewhat clearer?

lib/govuk_index/migrated_formats.rb
```diff
@@ -39,5 +41,15 @@ def convert_to_allowed_hash(formats)
...
+
+    def published_by_content_publisher?(format, app)
+      return false unless app == "content-publisher"
+      indexable_formats[format].include?("content-publisher")
+    end
+
+    def published_by_whitehall?(format, app)
+      return false unless app == "whitehall"
+      non_indexable_formats[format].include?("whitehall")
+    end
```
kevindew: Maybe we can make these methods a bit more generic to not need to have the publishing app hardcoded into them and instead take input from the yaml file?

config/govuk_index/mapped_document_types.yaml
emmabeynon: Hopefully this commit explains this decision: https://github.com/alphagov/rummager/pull/1276/commits/3983732cc5e71b8771fdd09eab045867528c148a

config/govuk_index/migrated_formats.yaml
emmabeynon: I've updated https://github.com/alphagov/rummager/pull/1276/commits/a58e8e6f4325494ec88e4244f785f6b83228b5cd with this new data format.

lib/govuk_index/migrated_formats.rb
emmabeynon: I've refactors this here https://github.com/alphagov/rummager/pull/1276/commits/784a28d31217fa8e607b7f2072b94a97d44290f9 - let me know what you think.

kevindew (approved):
LGTM 👍

PR #1277, opened 2018-08-23
Allow remaining news formats to be indexed

emmabeynon:
For https://trello.com/c/SlaLEmdQ/86-change-rummager-to-index-the-documents-were-publishing-l

This enables the remaining news formats that we will publish from Content Publisher to be indexed to the govuk index.

### DIFF
--- a/config/govuk_index/mapped_document_types.yaml
+++ b/config/govuk_index/mapped_document_types.yaml
@@ -13,6 +13,7 @@ employment_tribunal_decision: employment_tribunal_decision # Specialist Publishe
 esi_fund: european_structural_investment_fund # Specialist Publisher
 external_content: edition # Search Admin
 finder: edition # Specialist Publisher
+government_response: edition
 guide: edition
 help_page: edition
 hmrc_manual: hmrc_manual
--- a/config/govuk_index/mapped_document_types.yaml
+++ b/config/govuk_index/mapped_document_types.yaml
@@ -48,3 +48,4 @@ transaction: edition
 travel_advice: edition # Travel advice publisher
 travel_advice_index: edition # Travel advice publisher
 utaac_decision: utaac_decision # Specialist Publisher
+world_news_story: edition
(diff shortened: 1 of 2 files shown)


kevindew:
This looks good - we shouldn't need news_articles though as that is no longer used as a document_type and is just a schema name: https://github.com/alphagov/govuk-content-schemas/blob/master/formats/news_article.jsonnet#L2-L7

emmabeynon:
@kevindew ok cool - I used [this](https://github.com/alphagov/govuk_document_types/blob/master/data/supertypes.yml) to get the list, so will update this too.

kevindew:
Sure, it might make sense for it to still be in that list since they exist but no new ones are created.

emmabeynon:
@kevindew Ah I see, I will leave it be then :) I've removed the commit adding news_article from this PR.

kevindew (commented):
Oh sorry if I missed this last time. But rather than world_location_news_article we want government_response as per: https://github.com/alphagov/govuk-content-schemas/blob/master/formats/news_article.jsonnet#L6

I don't really know what we do about world_location_news_article. I'm going to research that format today to try learn how it differs to world_news_story :-/

kevindew:
Digging into things it looks like world_location_news_article isn't used anymore (which probably means it should be removed from content_schemas - I'll add a note in our trello) as there is only one published document on gov.uk with that type and if you filter Whitehall by it you get [0](https://whitehall-admin.publishing.service.gov.uk/government/admin/editions?type=world_location_news_article) whereas there are [lots](https://whitehall-admin.publishing.service.gov.uk/government/admin/editions?type=news_article_4) of world stories.

So I think we should just do nothing about world_location_news_article

emmabeynon:
@kevindew `world_location_news_article` removed, `government_response` added 👍

kevindew (approved):
Looks good 👍

PR #1284, opened 2018-09-12
Revert "Add first_published_at to all content"

sihugh:
Reverts alphagov/rummager#1254

This hasn't worked and appears to have caused a problem with the DFID finder.  The `first_published_at` used by the DFID finder lives within `["details"]["metadata"]["first_published_at"]` whereas the one I was hoping to populate lives at the root of the content item.

I don't think this is used anywhere else yet.  We did this at the time to be able to determine whether content had been updated or newly created, but didn't implement anything that used it.

Will fix https://govuk.zendesk.com/agent/tickets/3153878 when deployed, but will require republishing of the DFID specialist content.

https://trello.com/c/mgqxWPuz/167-fix-firstpublishedat-in-rummager

### DIFF
--- a/config/schema/base_elasticsearch_type.json
+++ b/config/schema/base_elasticsearch_type.json
@@ -8,7 +8,6 @@
     "content_store_document_type",
     "description",
     "email_document_supertype",
-    "first_published_at",
     "format",
     "government_document_supertype",
     "indexable_content",
--- a/config/schema/elasticsearch_types/dfid_research_output.json
+++ b/config/schema/elasticsearch_types/dfid_research_output.json
@@ -1,6 +1,7 @@
 {
   "fields": [
     "country",
+    "first_published_at",
     "dfid_authors",
     "dfid_review_status",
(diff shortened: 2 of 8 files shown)


rubenarakelyan (approved):

tijmenb:
Interesting, as far as I know `first_published_at` should indeed be populated for everything. We're using it in the frontend since https://github.com/alphagov/government-frontend/pull/996.

sihugh:
@tijmenb yeah, it's weird.  I haven't really worked out why it wasn't working - but I realise that it broke the existing behaviour for the dfid content.

I'll probably revisit this, but I might have to use a different key in rummager for the general `first_published_at` because (taking https://www.gov.uk/api/content/dfid-research-outputs/evidence-on-formative-classroom-assessment-for-learning as an example), the base `first_published_at` value looks like `"2018-09-12T11:15:02.000+00:00"` whereas the `[details][metadata][first_published_at]` value looks like `"2016-11-25"`.

PR #1298, opened 2018-10-17
Allow filtering with topical_events

bilbof:
This makes it possible to search for documents related to topical_events.
For example: `rummager.search(filter_topical_events: 'budget-2018')`.

`topical_events` are currently slugs like `budget-2018`, not full topical event resources.

This change enables some Whitehall finders to use rummager.

Related Trello card: https://trello.com/c/FSMKn70m/26-make-it-possible-to-filter-rummager-by-topical-events.

Example Trello card this unblocks: https://trello.com/c/Dl8JvGH4/16-allow-news-to-appear-on-topical-event-pages.

This currently is an incomplete task, as we will need to populate the topical_events field
we will need to also make a change in Whitehall and possibly also the publishing-api in
order to make filtering by topical events useful.

According to the [adding new field](https://github.com/alphagov/rummager/blob/master/doc/adding-new-fields.md) doc, the field will become usable after 24 hours:

> On production, this is done automatically every night by the search_fetch_analytics jenkins job.

### DIFF
--- a/config/schema/base_elasticsearch_type.json
+++ b/config/schema/base_elasticsearch_type.json
@@ -34,6 +34,7 @@
     "taxons",
     "title",
     "topic_content_ids",
+    "topical_events",
     "updated_at",
     "user_journey_document_supertype"
   ]
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -104,6 +104,11 @@
     "type": "identifiers"
   },
 
+  "topical_events": {
+    "description": "The topical events related to this page. These are \"slug\" identifiers.",
+    "type": "identifiers"
+  },
+
   "policies": {
     "description": "Policy content related to the page (https://www.gov.uk/government/policies).",
     "type": "identifiers"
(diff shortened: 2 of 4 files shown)


sihugh (approved):

sihugh:
🚀

bilbof:
Related PR: https://github.com/alphagov/whitehall/pull/4430

tijmenb:
Perhaps you're aware, but topical events are currently also indexed using the `policy_areas` field:

https://www.gov.uk/government/topical-events/budget-2018
https://www.gov.uk/api/search.json?filter_policy_areas=budget-2018

tijmenb:
> this is done automatically every night

This _might_ be a lie - the indexing process changed so this may have become outdated.

bilbof:
Thanks @tijmenb! I think that policy areas are going to be removed from rummager at some point, so this field should be useful.

Thanks for the heads up - I'll manually run the job if it doesn't get run overnight.

andrewgarner (approved):

PR #1299, opened 2018-10-18
Add batch search endpoint

oscarwyatt:
Trello: https://trello.com/c/3KbtCe3D/123-add-batchsearch-endpoint-to-rummager

Adds a batch search endpoint so that multiple searches can be processed with a single Rummager query. Uses Elasticsearch [multi query](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-multi-search.html) to process the searches efficiently as a single Elasticsearch call. 

It can process multiple searches with an example url such as:
```
/batch_search?search[][0][count]=10&search[][0][q]=ministry+of+magic&search[][0][order]=-popularity&search[][1][start]=0&search[][1][q]=floo+network+bandwith+enquries
```

It allows a maximum of 10 searches per request at the moment. There is no technical reason for this it's completely arbitrary so this can be changed in the future if the need arises


lib/search/batch_query.rb
```diff
@@ -0,0 +1,36 @@
+module Search
+  class BatchQuery < Query
+    class TooManyQueries < Error; end
+    def run(searches_params)
+      raise(TooManyQueries, 'Maximum of 10 searches per batch') unless searches_params.count <= 10
```
sihugh, line 5: Can we add something to the commit message to indicate that this is a fairly arbitrary number and it's small because we're cautious?

lib/search/batch_query.rb
```diff
@@ -0,0 +1,36 @@
...
+        )
+      end
+    end
+
+    def aggregate_payloads(builders)
+      builders.map { |builder| process_elasticsearch_errors { builder.payload } }
+    end
+
+    def log_search_count(searches)
+      GovukStatsd.increment "batch_search_#{searches.count}_searches"
```
sihugh, line 33: Nice 👍

lib/search/query.rb
```diff
@@ -61,39 +90,13 @@ def run_spell_checks(search_params)
       return unless suggestion_blocklist.should_correct?(search_params.query)
 
       query = {
-        size: 0,
-        suggest: QueryComponents::Suggest.new(search_params).payload
+          size: 0,
```
sihugh: Not sure why the linter didn't pick up extra whitespace 🤔 It'd be good to remove this, as it will preserve the `blame` trail better.

lib/search/query.rb
```diff
@@ -25,7 +25,36 @@ def run(search_params)
 
       payload     = process_elasticsearch_errors { builder.payload }
       es_response = process_elasticsearch_errors { index.raw_search(payload) }
+      process_es_response(search_params, builder, payload, es_response)
+    end
+
+  private
```
sihugh, line 31: I don't think very much has changed here.  I wonder if it's possible to split the method extraction into a separate commit.  It might make it easier to reason about when we come back to it in future?

spec/support/integration_spec_helper.rb
```diff
@@ -2,11 +2,11 @@
 
 module IntegrationSpecHelper
   SAMPLE_DOCUMENT_ATTRIBUTES = {
-    "title" => "TITLE1",
-    "description" => "DESCRIPTION",
-    "format" => "local_transaction",
-    "link" => "/URL",
-    "_type" => "edition",
+      "title" => "TITLE1",
```
sihugh: Same comment about ditching the whitespace as before - if nothing's changed we don't want to obscure the git history.

sihugh (dismissed):
This looks pretty good.  I've a few comments about style/lint type things, just so we can keep the git history clean.

vanitabarrett:
@oscarwyatt Would be good to get the linting errors fixed so we unblock merging of https://github.com/alphagov/gds-api-adapters/pull/855

oscarwyatt:
> @oscarwyatt Would be good to get the linting errors fixed so we unblock merging of [alphagov/gds-api-adapters#855](https://github.com/alphagov/gds-api-adapters/pull/855)

@vanitabarrett Apologies, fixed now

vanitabarrett (approved):

PR #1302, opened 2018-10-29
Remove spec relating to policy pages

cbaines:
These pages have not all been removed from GOV.UK, as the Topic pages
have superseded them.


bilbof (approved):
lgtm

PR #1307, opened 2018-11-01
Cap the limit of the "start" param

sihugh:
Values that are greater than around 900,000 result in a 500 error.  We currently have around 375,000 results in Rummager, so capping the maximum start value at 500,000 doesn't seem too bad.

We've experiencing repeated scanning requests to search pages which result in numbers (eg 29999800 from the below example) that are vexatious.  Platform Health have a card to relieve this in the longer term, but this will hopefully reduce the number of 500s we serve in the meantime and reduce the noise on 2nd line.

I've removed the test for expecting a 400 when submitting large integers to elasticsearch because in practice this results in a 503.  We're also putting defences in front of that now.

Example query:
`GET /search.json?count=100&start=29999800&fields=title%2Clink%2Cdescription%2Cpublic_timestamp%2Cpart_of_taxonomy_tree%2Corganisations&filter_organisations[]=department-for-work-pensions&order=-public_timestamp&facet_organisations=1500%2Corder%3Avalue.title HTTP/1.1`

### DIFF
--- a/lib/parameter_parser/search_parameter_parser.rb
+++ b/lib/parameter_parser/search_parameter_parser.rb
@@ -7,6 +7,10 @@ class SearchParameterParser < BaseParameterParser
   ].freeze
   MAX_RESULTS = 1500
 
+  # We currently have 375,000 items in Rummager.  Anything over around 900_000
+  # results in 5XX errors
+  MAX_START = 900_000
+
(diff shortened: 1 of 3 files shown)


koetsier (approved):
lgtm 👍

thomasleese (approved):

PR #1312, opened 2018-11-05
Index documents tagged to topical events in govuk index

1pretz1:
Previously, all tagging to topical events were handled in whitehall. When a document is published whitehall would update the government index in rummager which is the older index. The search endpoint would return results from both the new (govuk index) and the old (government index).

Content-publisher can now tag topical events to a document however the index it populates (govuk index) when publishing does not index this field currently, this change enables documents published by content publisher which are tagged to topical events to be indexed so the search endpoint can return these results.

## Local search result for a document published by content-publisher with topical events from the govuk index.
screen shot 2018-11-05 at 11 48 27

The topical events tagged to a document are saved as a hash with topical_events being the key and the slugs being the values in the govuk index, this follows the same pattern as how topical events are indexed in the the government index. E.g

## Search result returning topical events tagged to a document from the government index.
screen shot 2018-11-05 at 12 30 31

Trello:
https://trello.com/c/Nx5jjnIj/373-populate-topical-events-in-rummager-for-content-publisher-content

### DIFF
--- a/lib/govuk_index/presenters/elasticsearch_presenter.rb
+++ b/lib/govuk_index/presenters/elasticsearch_presenter.rb
@@ -97,6 +97,7 @@ module GovukIndex
         tiers_or_standalone_items:           specialist.tiers_or_standalone_items,
         title:                               common_fields.title,
         topic_content_ids:                   expanded_links.topic_content_ids,
+        topical_events:                      expanded_links.topical_events,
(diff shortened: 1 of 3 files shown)


leenagupte:
It looks good to me 🎉🎉🎉, but I've tagged @kevindew for a review.

kevindew (approved):
LGTM 👍 🎉

PR #1315, opened 2018-11-06
Index documents tagged to world locations in govuk index

1pretz1:
Previously, all tagging to world_locations were handled in whitehall.
When a document is published whitehall would update the government index
in rummager which is the older index. The search endpoint would return
results from both the new (govuk index) and the old (government index).

Content-publisher can now tag world_locations to a document however the
index it populates (govuk index) does not index this field currently,
this change enables documents published by content publisher which are
tagged to world_locations to be indexed so the search endpoint can
return these results.

## Local search result for a document published by content-publisher with world locations from the govuk index.
screen shot 2018-11-06 at 15 31 35

Trello:
https://trello.com/c/GByG04gX/387-populate-world-locations-in-rummager-for-content-publisher-content


bilbof (approved):
lgtm

emmabeynon (approved):
🎉

PR #1316, opened 2018-11-08
Add a path parameter to the analytics:create_data_import_csv

koetsier:
This rake task is run from the 'run_rake_task' job in Jenkins and
setting a parameter is easier this way than through an environment
variable

### DIFF
--- a/lib/tasks/analytics.rake
+++ b/lib/tasks/analytics.rake
@@ -10,13 +10,15 @@ namespace :analytics do
   The generated file is saved to disk, so you should run this task from the server and
   then use SCP to retrieve the file, which will be around 100 MB.
   "
-  task :create_data_import_csv do
-    elasticsearch_config = SearchConfig.new.elasticsearch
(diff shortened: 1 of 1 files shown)


lib/tasks/analytics.rake
```diff
@@ -10,13 +10,15 @@ namespace :analytics do
   The generated file is saved to disk, so you should run this task from the server and
   then use SCP to retrieve the file, which will be around 100 MB.
   "
-  task :create_data_import_csv do
-    elasticsearch_config = SearchConfig.new.elasticsearch
+  task :create_data_import_csv, [:path] do |_, args|
+    args.with_defaults(path: (ENV['EXPORT_PATH'] || 'data'))
+    path = args[:path]
```
benthorner, line 15: I think this would be a bit easier to read if we could avoid the mutator `with_defaults` method. Would `args[:path] || ENV['EXPORT_PATH'] || 'data'` work the same?

benthorner (approved):
👍

PR #1318, opened 2018-11-12
[WIP] Expand World Locations in Finder responses

bilbof:
This fetches world location data from https://www.gov.uk/api/world-locations
and uses it to provide titles and slugs to Finders for use in faceting with
world locations.

We need to use an external source as most world locations were removed from
rummager (commit: https://github.com/alphagov/whitehall/pull/3966/commits/45ce7295a731991f483f5ba238c3424549213746).

This is a short term fix. The longer term solution would be to bring world locations back
into the content-store and rummager. That may be tricky and is being discussed at the moment.

Only the title and slug of a world locations is required for Finder-frontend.

This is part of the migration of finders from whitehall to finder-frontend.
After this is merged we can add world_locations as an available filter to
finders in finder-frontend.

## Screenshots

screen shot 2018-11-12 at 14 29 32


lib/search/registries.rb
```diff
@@ -51,6 +51,10 @@ def organisations
       )
     end
 
+    def world_locations
+      @world_locations ||= WorldLocationsRegistry.new
```
bilbof, line 55: This is the 'entry point', in that we are switching from using a collection of rummager documents of the type 'world_location' to using an external source for world locations.

This is necessary because almost all world locations were removed from rummager and moved into the taxonomy.

andrewgarner (dismissed):

lib/search/world_locations_registry.rb
```diff
@@ -0,0 +1,42 @@
...
+      raise NotImplementedError
+    end
+
+    def [](slug)
+      all.find { |o| o['slug'] == slug }
+    end
+
+  private
+
+    def fetch
```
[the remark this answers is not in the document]
bilbof: Thanks for this! I didn't know about that. I needed to make it possible to paginate via the gds api adapter - could you approve this?

https://github.com/alphagov/gds-api-adapters/pull/865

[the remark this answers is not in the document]
andrewgarner: As discussed on Slack, you can use the following. (posting here for visibility)

```ruby
GdsApi.worldwide.world_locations.with_subsequent_pages
```

spec/spec_helper.rb
```diff
@@ -78,6 +78,7 @@
   # config.warnings = true
 
   config.before do
+    stub_any_world_locations_request
```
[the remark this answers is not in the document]
bilbof: Yeah good idea!

lib/services.rb
```diff
@@ -37,6 +37,10 @@ def self.elasticsearch(hosts: ENV['ELASTICSEARCH_HOSTS'] || 'http://localhost:92
   def self.statsd_client
     @statsd_client ||= Statsd.new.tap { |sd| sd.namespace = "govuk.app.rummager" }
   end
+
+  def self.worldwide_api
+    @worldwide_api ||= GdsApi::Worldwide.new(Plek.new.find('whitehall-admin'))
```
andrewgarner: How about `GdsApi.worldwide` – that way if the common way to connect to this API changes we inherit those changes too.
```suggestion
    @worldwide_api ||= GdsApi.worldwide
```

spec/spec_helper.rb
```diff
@@ -91,4 +93,11 @@
 
   config.order = :random
   Kernel.srand config.seed
+
+  def stub_any_world_locations_request
+    worldwide_api_has_locations %w(hogwarts privet-drive diagon-alley)
+    # worldwide_api_has_selection_of_locations()
```
andrewgarner: Shall we remove this commented out code?

lib/services.rb
bilbof: I get "undefined method `worldwide' for GdsApi:Module`" when making that change. Any idea why?

andrewgarner: Hmm, it should work 🤔 

https://github.com/alphagov/gds-api-adapters/blob/c9cae38a7b5d96837bc64c853420566f2668617b/lib/gds_api.rb#L181-L183

What version of the gem are you using?

andrewgarner (approved):

lib/services.rb
andrewgarner: Have you required the right files?

bilbof: I think so. Could it be because GdsApi is a module defined by Rummager, too?

andrewgarner: That would be it yeh! 🤦‍♂️

bilbof:
Closing this PR as I'm moving this logic downstream into finder-frontend (https://github.com/alphagov/finder-frontend/pull/659) so rummager doesn't depend on whitehall.

PR #1319, opened 2018-11-19
Add a synonym of "Brexit" for "EU Exit"

bilbof:
This ensures that searches for "brexit" return results that have the word "EU Exit" in the searchable fields.

This is useful because brexit is a common search term, but on production there are far fewer results for brexit than there are for EU Exit:

17234 results for [eu exit](https://www.gov.uk/search.json?q=eu%20exit)
1337 results for [brexit](https://www.gov.uk/search.json?q=brexit)

This should surface much more content when searching across gov.uk
for brexit, even if brexit has been omitted from the documents
searchable fields.

A schema migration will be necessary after this change across all
environments:
See https://github.com/alphagov/rummager#changing-the-schemareindexing.

https://trello.com/c/LgqvNg72/153-add-a-synonym-brexit-eu-exit-to-rummager

### DIFF
--- a/config/schema/synonyms.yml
+++ b/config/schema/synonyms.yml
@@ -18,6 +18,7 @@
 
 # Same meanings
 - search: abroad, overseas
+- search: brexit, eu exit => brexit, eu exit
 - search: motorbike, motorcycle
 - search: opening times, opening hours
 - search: sickness, illness


sihugh (approved):

bilbof:
Thanks Simon, I'm going to test this on integration before merging.

config/schema/synonyms.yml
```diff
@@ -102,6 +102,7 @@
 # Unofficial or deprecated terms should expand to the preferred term
 - search: ancestral visa => ancestry visa
 - search: bedroom tax => bedroom tax, spare bedroom, spare room
+- search: brexit => eu exit
```
andrewgarner: Should this be as follows so it includes both terms?

```suggestion
- search: brexit => brexit, eu exit
```

bilbof: Good catch!

config/schema/synonyms.yml
```diff
@@ -102,6 +102,7 @@
 # Unofficial or deprecated terms should expand to the preferred term
 - search: ancestral visa => ancestry visa
 - search: bedroom tax => bedroom tax, spare bedroom, spare room
+- search: brexit => brexit, eu exit
```
andrewgarner: This is also configured in such a way that "EU Exit" is the preferred term for "Brexit". Is this the case given https://www.gov.uk/government/brexit ?
```suggestion
- search: eu exit => eu exit, brexit
```

andrewgarner: This block of configuration has the comment "Unofficial or deprecated terms should expand to the preferred term". My understanding looking at how other synonyms have been entered is that the unofficial or deprecated term is entered first.

So `search: brexit => brexit, eu exit` would suggest that `brexit` is the deprecated term, but any search for `brexit` should also include matches for `eu exit`.

You're right though, `brexit, eu exit => brexit, eu exit` might be what is needed since neither of these terms are used consistently. Or maybe it should be `eu exit, brexit => eu exit, brexit` ?!

bilbof: Yeah sorry this is a bit confusing. I put this synonym under this comment as Brexit is an unofficial term (like `xmas => xmas, christmas`). 

I changed it and moved it under the 'same meanings' comment.

andrewgarner: 👍

config/schema/synonyms.yml
```diff
@@ -18,6 +18,7 @@
 
 # Same meanings
 - search: abroad, overseas
+- search: brexit, eu exit => brexit, eu exit
```
andrewgarner, line 21: Should this now just be `search: brexit, eu exit`

PR #1330, opened 2018-11-20
Amend base filter with new metadata filter

steventux:
https://trello.com/c/2BGHlD7M/13-make-it-possible-to-publish-finder

We'd like to include any document tagged to a business readiness
filtering facet, so use an internal metadata value as the base
filter.

DNM until release_827 of govuk-content-schemas

Depends on https://github.com/alphagov/govuk-content-schemas/pull/838

### DIFF
--- a/lib/publishing_api_finder_publisher.rb
+++ b/lib/publishing_api_finder_publisher.rb
@@ -77,7 +77,7 @@ class FinderContentItemPresenter
   end
 
   def details
-    schema["details"].except("reject", "subscription_list_title_prefix", "email_filter_facets")
+    schema["details"].except("email_filter_facets", "subscription_list_title_prefix")
   end
 end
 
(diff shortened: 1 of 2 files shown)


barrucadu:
I think this will pass CI when #1327 is merged.

lib/publishing_api_finder_publisher.rb
```diff
@@ -96,7 +96,7 @@ def details
     details.merge(
       "subscription_list_title_prefix" => details.fetch("subscription_list_title_prefix", {}),
       "email_filter_facets" => details.fetch("email_filter_facets", []),
-    ).except("document_noun", "facets", "reject")
+    ).except("document_noun", "facets", "filter")
```
barrucadu: Should this be `.except("document_noun", "facets", "reject", "filter")`?

thomasleese (approved):

PR #1338, opened 2018-11-21
Inject tags for content related to business readiness

barrucadu:
The rummager counterpart of https://github.com/alphagov/email-alert-api/pull/731


config/metadata.csv
```diff
@@ -0,0 +1 @@
+
```
thomasleese: Would you be able to add `/government/publications/test-business-readiness,yes` here so we can test the whole journey on integration with this piece of content?

barrucadu:
I think it would be good to test the full flow on integration before merging, publishing something from Whitehall and something from a different publishing app, and seeing that the metadata arrives in the search index.

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -36,6 +36,12 @@ def perform(messages)
       processor = Index::ElasticsearchProcessor.govuk
 
       messages.each do |routing_key, payload|
+        if payload.is_a?(Hash) && payload.has_key?("base_path")
+          base_path = payload["base_path"]
+          extra_metadata = Indexer::MetadataTagger.metadata_for_base_path(base_path)
+          payload = payload.merge(extra_metadata)
```
thomasleese, line 42: In `email-alert-api` we used slightly more complex logic to try and merge arrays nicely - https://github.com/alphagov/email-alert-api/pull/731/files#diff-9959dd9022669e7af3eea3a452bb2643R13. I'm not sure if you want to add this to Rummager.

I guess this is why we might want a Gem...

barrucadu: I think this is fine, because it'll only cause a problem if the content is both tagged by the publishing app and in the csv (and with different metadata in each).

thomasleese (approved):

PR #1340, opened 2018-11-21
Content updates to business readiness finder facets

steventux:
Update facet labels and values for business readiness finder and the associated email signup page.
[There's also a change to the values for one of the facets](https://github.com/alphagov/rummager/pull/1340/commits/2c8d19cd996b1072b9346d14a6fc65a1080df327) which will necessitate retagging with these values.

### DIFF
--- a/config/find-eu-exit-guidance-business.yml
+++ b/config/find-eu-exit-guidance-business.yml
@@ -208,27 +208,12 @@ details:
       radio_button_name: Wholesale (excl. Motor Vehicles)
       topic_name: Wholesale (excl. Motor Vehicles)
       prechecked: false
-  - facet_id: employ_eu_citizens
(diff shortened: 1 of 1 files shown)


config/find-eu-exit-guidance-business.yml
```diff
@@ -451,19 +451,6 @@ details:
     name: Sector / Business Area
     preposition: for businesses in
     type: text
-  - allowed_values:
```
thomasleese, line 454: I think you'll need to make these changes to the email facets as well.

steventux:
@thomasleese email facets should be in sync with finder values now https://github.com/alphagov/rummager/pull/1340/commits/791b2620f909031304238e128782c5de2c619251

thomasleese (approved):

PR #1343, opened 2018-11-22
Lookup facet config from finder definition

steventux:
https://trello.com/c/UhV3PPfO/87-metadata-tagger-should-use-finder-facet-definitions

Facet config is stored in YAML in this project and should be
treated as the single definition of the facets and values we
tag with. This should reduce the impact of changes to metadata
values and names.

### DIFF
--- a/lib/indexer/metadata_tagger.rb
+++ b/lib/indexer/metadata_tagger.rb
@@ -2,9 +2,9 @@ require 'csv'
 
 module Indexer
   class MetadataTagger
-    def self.initialise(metadata_file_path)
+    def self.initialise(metadata_file_path, facet_config_file_path)
       @metadata = {}
-      @config = YAML.load_file(File.join(Dir.pwd, 'config/find-eu-exit-guidance-business.yml'))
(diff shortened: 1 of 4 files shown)


lib/indexer/metadata_tagger.rb
```diff
@@ -2,10 +2,11 @@
 
 module Indexer
   class MetadataTagger
-    def self.initialise(file_name)
+    def self.initialise(metadata_file_path)
       @metadata = {}
+      @config = YAML.load_file(File.join(Dir.pwd, 'config/find-eu-exit-guidance-business.yml'))
```
barrucadu: I think this path should be passed in, if nothing else it makes testing easier.   It could be an argument with a default value.

steventux:
:+1: @barrucadu updated in https://github.com/alphagov/rummager/pull/1343/commits/6f40399010e1ff126829df12b53e9eb838a588cb

barrucadu (approved):

thomasleese (approved):

thomasleese (approved):

PR #1347, opened 2018-11-23
Present email filter facets derived from the finder facets

steventux:
Rather than maintain two sets of definitions in one config file it's possible to build the email filter facets from the finder definitions.

See also:
https://github.com/alphagov/govuk-app-deployment-secrets/pull/5


thomasleese (approved):
Woo!

barrucadu (approved):

PR #1354, opened 2018-11-30
Lookup organisation's default news image when setting image URL

emmabeynon:
For https://trello.com/c/w1QxJ7gk/500-update-rummager-to-lookup-organisation-default-news-image-when-setting-the-image-for-an-indexed-document

In order to be able to display a document's image on topic pages (e.g. https://www.gov.uk/society-and-culture), we need to set the image URL in the index to the organisation's default image when a lead image isn't available.


leenagupte (approved):
👍

PR #1360, opened 2018-12-04
Lookup default news image if newslike content

emmabeynon:
We should only provide a default news image if the content belongs to the
`news` subgroup.

### DIFF
--- a/lib/govuk_index/presenters/elasticsearch_presenter.rb
+++ b/lib/govuk_index/presenters/elasticsearch_presenter.rb
@@ -185,6 +185,7 @@ module GovukIndex
     end
 
     def newslike?
+      return false if common_fields.content_store_document_type == "fatality_notice"
       common_fields.content_purpose_subgroup == "news" ||
         common_fields.content_purpose_subgroup == "speeches_and_statements"
     end
--- a/lib/govuk_index/presenters/elasticsearch_presenter.rb
+++ b/lib/govuk_index/presenters/elasticsearch_presenter.rb
@@ -143,7 +143,7 @@ module GovukIndex
     end
 
     def image_url
-      details.image_url || expanded_links.default_news_image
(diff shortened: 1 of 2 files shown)


kevindew:
This looks good 👍 

The only thing I'm concerned with is whether our newslike criteria is enough. Looking up what documents use this in Whitehall I can see it comes from [LeadImagePresenterHelper](https://github.com/alphagov/whitehall/blob/7b5c5a086b89cb62ffba62b152a0a8dcfc10c8e6/app/presenters/lead_image_presenter_helper.rb) and the usage search pulls up: https://github.com/alphagov/whitehall/search?q=leadimagepresenterhelper&unscoped_q=leadimagepresenterhelper which would mean including in https://github.com/alphagov/govuk_document_types/blob/master/data/supertypes.yml#L411 and probably the document type of case study

Case study was a surprise for me, although I did expect speech so could leave them out since they're a) probably quite far down the road compared to speeches and b) probably need checking first for default image anyway

emmabeynon:
@kevindew I guess I was approaching this by thinking about what would surface under the news and communications section of topics pages specifically, because that's what is using rummager.  This section uses the "news_and_communications" content purpose supergroup (https://github.com/alphagov/govuk_document_types/blob/2645f543bea870b9469a70d6c93fde4db6e493dd/data/supertypes.yml#L500) but I wondered if that was too broad a list with lots of specialist documents in it?

kevindew:
Yeah, I'd agree that we don't want specialist documents included

emmabeynon:
@kevindew Shall we go with the "news" and "speeches_and_statements" subgroups?

kevindew:
@emmabeynon yup sounds good

emmabeynon:
@kevindew Now amended to include speeches and statements 👍

kevindew (approved):
LGTM 👍

PR #1363, opened 2018-12-06
Adds routes for UK nationals living in the EU

DilwoarH:


lib/special_route_publisher.rb
```diff
@@ -81,6 +81,14 @@ def routes
         title: "Prepare your business for the UK leaving the EU",
         description: "Q&A frontend for the 'Find EU Exit guidance for your business' finder",
         type: "exact",
+      },
+      {
+        rendering_app: "finder-frontend",
+        base_path: "/uk-nationals-living-in-the-eu",
+        content_id: "????",
```
DilwoarH: ^^ how would I get a content ID for this page?

thomasleese: You can generate one yourself, `SecureRandom.uuid` should give you one in a Ruby console.

steventux (approved):

lib/special_route_publisher.rb
DilwoarH: Thanks ! updated

thomasleese (approved):

PR #1365, opened 2018-12-07
Set the operator between facets to or

thomasleese:
This changes the find-eu-exit-guidance-business finder to set the operator between the facets to or meaning that ticking boxes will show up more content rather than and where the content filters down.

Depends on https://github.com/alphagov/govuk-content-schemas/pull/847

[Trello Card](https://trello.com/c/HoWR8kQM/150-or-between-and-in-facets-in-the-finder)


thomasleese:
Superseeded by https://github.com/alphagov/govuk-app-deployment-secrets/pull/18

PR #1366, opened 2018-12-07
Add finders for individuals

sihugh:
Adds finders for each of the top level taxons. We're doing this to simplify the code we need to write in finder-frontend to build queries.

We haven't included the Government taxon because the query doesn't seem to work because the citizen taxon is a child of Government.

We don't want to include email signups yet as they're not ready.

These finders have base filters of the citizen topic under /government/brexit which is draft at present, `AND` the relevant first level taxon topic.

run `rake publishing_api:publish_citizen_finders` to do the deed


config/topic_finders/prepare-individual-uk-leaving-eu.yml.erb
```diff
@@ -0,0 +1,36 @@
...
+locale: en
+public_updated_at: "<%= config[:timestamp] %>"
+publishing_app: rummager
+rendering_app: finder-frontend
+details:
+  beta: false
+  document_noun: publication
+  facets: []
+  filter:
+    all_part_of_taxonomy_tree:
```
bilbof, line 22: Is additional work required in `finder-frontend` in order for this filter to work? I remember it was necessary for faceting, but perhaps not for filtering.

bilbof (dismissed):

config/topic_finders/prepare-individual-uk-leaving-eu.yml.erb
sihugh: It works in exactly the same way as having a single taxon set as the `filter` and then one set as a facet (so like the `brexit` checkbox on the news and comms finder).

oscarwyatt (approved):

PR #1372, opened 2018-12-11
Add tasks to unpublish business readiness and UK citizens in the EU

bevanloon:
Unpublish the special routes as well as the document finder and email signup route.

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -24,4 +24,12 @@ namespace :publishing_api do
       end
     end
   end
+
+  desc "Unpublish special routes"
+  task :unpublish_prepare_business_and_uk_nationals_special_routes do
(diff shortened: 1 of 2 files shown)


lib/tasks/publishing_api.rake
```diff
@@ -24,4 +24,12 @@ namespace :publishing_api do
       end
     end
   end
+
+  desc "Unpublish special routes"
+  task :unpublish_special_routes do
```
thomasleese: Could we rename this to `unpublish_business_readiness_special_routes` since it only does those?

bevanloon: 👍 see what you think of the name in 7885a65. The routes aren't both related to business readiness so I've gone for a slightly more verbose and descriptive name without using the entire route names because that would become rather cumbersome.

thomasleese (approved):

PR #1375, opened 2018-12-12
Include travel advice index page in citizen finder

sihugh:
We can do this now that https://github.com/alphagov/govuk-content-schemas/pull/849
has been merged.

The supergroups should get expanded to document types and merged with the supplied content store document types since https://github.com/alphagov/finder-frontend/pull/738 was merged.

https://trello.com/c/UDfCG0jA/184-add-foreign-travel-advice-index-to-going-abroad-finder


bilbof (approved):
🚢

PR #1377, opened 2018-12-13
Add route for citizen finder atom feed

sihugh:
The feed won't work without the route. (Lines 39-40)

The first commit is a small tidy up, where we've extracted some common elements to variables.


bilbof (approved):
⚛️

PR #1390, opened 2019-01-02
Prevent email signup content item attempting to add `canonical_link` field

edwardkerry:
The Finder schema now has an optional `canonical_link` field. This commit
ensures we don't error by attempting to add the field to the email signup
content item.

### DIFF
--- a/lib/publishing_api_finder_publisher.rb
+++ b/lib/publishing_api_finder_publisher.rb
@@ -100,7 +100,7 @@ class FinderEmailSignupContentItemPresenter
     details.merge(
       "subscription_list_title_prefix" => details.fetch("subscription_list_title_prefix", {}),
       "email_filter_facets" => email_filter_facets,
-    ).except("document_noun", "facets", "filter", "generic_description", "reject", "summary")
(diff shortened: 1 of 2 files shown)


lib/publishing_api_finder_publisher.rb
```diff
@@ -100,7 +100,7 @@ def details
     details.merge(
       "subscription_list_title_prefix" => details.fetch("subscription_list_title_prefix", {}),
       "email_filter_facets" => email_filter_facets,
-    ).except("document_noun", "facets", "filter", "generic_description", "reject", "summary")
+    ).except("canonical_link", "document_noun", "facets", "filter", "generic_description", "reject", "summary")
```
barrucadu, line 103: I feel like, with a blacklist here, this is just going to keep happening.  But maintaining a whitelist could be just as painful...

barrucadu (approved):

lib/publishing_api_finder_publisher.rb
edwardkerry: I agree. I will add a tech-debt card to look at this in more detail :+1:

PR #1394, opened 2019-01-07
Add authentication for write endpoints

sophiebrown01:
This PR adds authentication to the write endpoints of Rummager, requiring consumer apps to provide a bearer token to be processed by Signon.

We do not use GDS SSO directly as the gem is written against Rails; it is instead far simpler to include only the necessary configuration for this Sinatra app to add bearer token auth.

Not to be merged until consumer apps have bearer token auth in place.

Trello card: https://trello.com/c/M9Xek3W4

Co-Authored-By: Kevin Dew @kevindew 
Co-Authored-By: Oscar Wyatt @oscarwyatt


spec/spec_helper.rb
```diff
@@ -38,6 +39,7 @@
 Sidekiq::Logging.logger = nil
 
 require 'webmock/rspec'
+require 'climate_control'
```
kevindew: oops think we've managed to merge in requiring this twice.

sophiebrown01: Done

lib/auth/gds_sso.rb
```diff
@@ -0,0 +1,39 @@
...
+    def initialize(token_string)
+      @token_string = token_string
+    end
+
+    def locate
+      access_token = OAuth2::AccessToken.new(oauth_client, token_string)
+      body = access_token.get("/user.json?client_id=#{CGI.escape(oauth_id)}").body
+      JSON.parse(body)["user"]
+    rescue OAuth2::Error
+      nil
```
bilbof, line 18: Could this result in `nil` being cached as the `api-user-cache`?

> Nil values can be cached ([source](https://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html)).

bilbof (commented):
Looks good. I have a question regarding the cacheing of nil values.

lib/auth/gds_sso.rb
sophiebrown01: Yes, but only for five minutes. I think the point being that if the access token doesn't exist then we don't want to be constantly querying Signon over that period.

bilbof: In that case, would it be worth sending this authentication error to monitoring/elsewhere? It might end up being permanently cached as nil if the token becomes invalid, but would this be noticeable?

There may be cases where the writes fail due to a broken token, but once the token is fixed there could be a 5 minute delay.

kevindew: > There may be cases where the writes fail due to a broken token, but once the token is fixed there could be a 5 minute delay.

Yup that's correct, but I don't think that's too big a problem and is consistent with how GDS-SSO works: https://github.com/alphagov/gds-sso/blob/master/lib/gds-sso/bearer_token.rb#L8-L12

It'd be consistent that the monitoring of the usage is handled by the consumer application rather than here.

bilbof (approved):

PR #1396, opened 2019-01-08
Revert "Add authentication for write endpoints"

oscarwyatt:
Reverts alphagov/rummager#1394

There are some parts of Whitehall that would not work were this to be deployed, reverting until this is fixed


sophiebrown01 (approved):
👍

PR #1398, opened 2019-01-09
Run end-to-end tests as part of Rummager build

kevindew:
As rummager is part of end-to-end tests changes to it can break the
suite of tests.


sophiebrown01 (approved):
👍

PR #1399, opened 2019-01-09
Remove usage of PublishingApiV1 adapter

kevindew:
This was removed from GDS API Adapters at version 56 and now the V2
adapter must be used. This same put_path method exists in the V2 adapter
as well.


oscarwyatt (approved):
LGTM 🥇

PR #1402, opened 2019-01-15
Allow release timestamp as a sort field

bilbof:
This allows the release_timestamp field to be a sort option.


andrewgarner:
Do we know if this value is consistently present and reliable for all items in the index?

bilbof:
When I look at the aggregates [locally](http://rummager.dev.gov.uk/search.json?aggregate_release_timestamp=700&count=0) (add `release_timestamp` to allowed aggregates) it seems like it is quite a well used field. How reliable it is I am not sure. How would one find this out?

andrewgarner (approved):

PR #1406, opened 2019-01-22
Allow finders to have `sort` under `details`

peteglondon:
Previously publishing the latest find-eu-exit-guidance-business.yml
causes a crash when publishing the finder.

This was because it was sending `sort` to the content store for
the finder_email_signup.

This commit filters out `sort` from the json sent to the publishing api for
find-eu-exit-guidance-business/email-signup

Trello: https://trello.com/c/JgohmaDW/3-add-sort-order-to-the-business-finder-results-page


peteglondon:
[Add sort order to the business finder results page](https://trello.com/c/JgohmaDW/3-add-sort-order-to-the-business-finder-results-page)

tijmenb (approved):
This is the second time this has bitten us, as far as I know. Probably worth refactoring this by splitting up the YAML for the email / finder signup content item.

PR #1410, opened 2019-01-24
Use CommonFieldsPresenter#link path extraction logic

steventux:
https://trello.com/c/spSQHcZ9/34-bug-some-items-of-content-are-disappearing-from-finder

Makes determining the path consistent with [the way Elasticsearch presenters work](https://github.com/alphagov/rummager/blob/e20595813043f8529f5136c902a467ca959fa80f/lib/govuk_index/presenters/common_fields_presenter.rb#L31).

We can see Kibana entries which coincide with the finder metadata disappearing which have the info:

```
@fields.worker : GovukIndex::PublishingEventWorker
@message : answer.links -> INDEX /taking-a-lorry-abroad edition
```
Not certain if this means the condition for tagging metadata was too `base_path`-centric but this change brings extracting the base path in line with other presenter functionality.

### DIFF
--- a/lib/govuk_index/publishing_event_worker.rb
+++ b/lib/govuk_index/publishing_event_worker.rb
@@ -36,8 +36,8 @@ module GovukIndex
       processor = Index::ElasticsearchProcessor.govuk
 
       messages.each do |routing_key, payload|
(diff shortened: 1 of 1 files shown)


steventux:
This was a red herring, we'd never have a tagged external link. Closing

PR #1419, opened 2019-02-01
Publish news and comms finder email signup

bilbof:
https://trello.com/c/HZHHKXc8/311-make-it-possible-to-publish-news-and-communications-email-alert-content-item-from-rummager

**Note:** this will need to be merged after https://github.com/alphagov/govuk-content-schemas/pull/860 is merged and the latest version of the gem is used by rummager. The tests wont pass until this is done.

This adds `ContentItemPublisher` to handle the publication of finders and their email signup pages from rummager. Additionally it adds a rake task to publish the content items.

### DIFF
--- a/lib/content_item_publisher/finder_presenter.rb
+++ b/lib/content_item_publisher/finder_presenter.rb
@@ -1,0 +1,21 @@
+module ContentItemPublisher
+  class FinderPresenter < ContentItemPresenter
+    def present_links
+      links = {}
+      links["email_alert_signup"] = email_alert_signup if email_alert_signup?
+      links["parent"] = Array(content_item["parent"]) if content_item.key?("parent")
+
+      { content_id: content_id, links: links }
+    end
+
+  private
+
+    def email_alert_signup?
+      content_item.key?("signup_content_id")
+    end
+
+    def email_alert_signup
+      [content_item["signup_content_id"]]
+    end
+  end
+end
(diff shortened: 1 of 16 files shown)


lib/content_item_publisher/finder_presenter.rb
```diff
@@ -0,0 +1,21 @@
+module ContentItemPublisher
+  class FinderPresenter < ContentItemPresenter
+    def present_links
+      links = {}
+      links["email_alert_signup"] = email_alert_signup if email_alert_signup?
```
koetsier, line 5: could make line 5 and 6 more consistent. - they do a similar thing but are implemented differently

sophiebrown01 (approved):
Discussed with @bilbof

lib/content_item_publisher/publisher.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    def call
+      publish(content_item)
+    end
+
+  private
+
+    attr_reader :content_item, :logger, :timestamp
+
+    def publishing_api
```
koetsier: Could use Services.publishing_api here

spec/unit/content_item_publisher/finder_email_signup_publisher_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  describe "#call" do
+    let(:publishing_api) { instance_double("GdsApi::PublishingApiV2") }
+    let(:payload) {
+      ContentItemPublisher::FinderEmailSignupPresenter.new(finder, timestamp).present
+    }
+
+    before do
+      allow(logger).to receive(:info)
+      allow(GdsApi::PublishingApiV2).to receive(:new).and_return(publishing_api)
```
koetsier: - Line 24 could be a bit different when using Services.publishing_api
- Could think about using test helpers defined in: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api_v2.rb

PR #1425, opened 2019-02-07
Add services finder and email signup

bilbof:
https://trello.com/c/e53rC9Ee/287-publish-services-finder-from-rummager

This can be published with the following rake task:

```
publishing_api:publish_finder FINDER_CONFIG=services.yml EMAIL_SIGNUP_CONFIG=services_email_signup.yml
```

### DIFF
--- a/config/finders/services.yml
+++ b/config/finders/services.yml
@@ -1,0 +1,52 @@
+---
+base_path: "/services"
+content_id: f6d779ac-5f78-413d-a1ff-da391944e6ec
+description: Find services from government
+signup_content_id: e72eda7d-8fff-498f-b0f4-c07d8cd8ceb5
+details:
+  default_documents_per_page: 20
+  document_noun: service
+  facets:
+  - key: "_unused"
+    filter_key: all_part_of_taxonomy_tree
+    keys:
+    - level_one_taxon
+    - level_two_taxon
+    name: topic
+    short_name: topic
+    type: taxon
+    display_as_result_metadata: false
+    filterable: true
+    preposition: about
+  - display_as_result_metadata: true
+    filterable: true
+    key: organisations
+    name: Organisation
+    preposition: from
+    short_name: From
+    type: text
+  filter:
+    content_purpose_supergroup:
+    - services
+  show_summaries: true
+  sort:
+  - key: title
+    name: A-Z
+  - default: true
+    key: "-popularity"
+    name: Most viewed
+  - key: "-relevance"
+    name: Relevance
+document_type: finder
+locale: en
+publishing_app: rummager
+rendering_app: finder-frontend
+schema_name: finder
+title: Services
+routes:
+- path: "/services"
+  type: exact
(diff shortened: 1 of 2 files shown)


config/finders/services.yml
```diff
@@ -0,0 +1,63 @@
+---
+analytics_identifier:
+base_path: "/services"
+content_id: f6d779ac-5f78-413d-a1ff-da391944e6ec
+content_purpose_document_supertype: navigation
```
tijmenb: I think this supertype and the others aren't used. Can we remove them if that's the case? Might be worth looking at the other entries as well.

bilbof: Thanks, @tijmenb! You're right - I removed them.

config/finders/services.yml
```diff
@@ -0,0 +1,57 @@
...
+  sort:
+  - key: title
+    name: A-Z
+  - default: true
+    key: "-popularity"
+    name: Most viewed
+  - key: "-relevance"
+    name: Relevance
+document_type: finder
+first_published_at: '2019-01-15T13:00:00.000+00:00'
```
sihugh: we can remove this

config/finders/services.yml
```diff
@@ -0,0 +1,57 @@
...
+links: {}
+locale: en
+public_updated_at: '2019-01-15T13:00:00.000+00:00'
```
sihugh: we can remove this

config/finders/services.yml
```diff
@@ -0,0 +1,57 @@
...
+publishing_app: finder-frontend
```
sihugh: this should be `rummager`

config/finders/services.yml
```diff
@@ -0,0 +1,57 @@
...
+rendering_app: finder-frontend
+schema_name: finder
+title: Services
+updated_at: '2019-01-15T13:00:00.000+00:00'
```
sihugh: we can remove this

config/finders/services.yml
```diff
@@ -0,0 +1,57 @@
...
+  - key: title
+    name: A-Z
+  - default: true
+    key: "-popularity"
+    name: Most viewed
+  - key: "-relevance"
+    name: Relevance
+document_type: finder
+first_published_at: '2019-01-15T13:00:00.000+00:00'
+links: {}
```
sihugh: we can remove this

config/finders/services.yml
```diff
@@ -0,0 +1,57 @@
+---
+analytics_identifier:
```
sihugh: we can remove this

sihugh (changes requested):
There's just a few bits that publishing api will generate for us.  They'll be present in content store, but we don't need to provide them here.

bilbof:
Thank you, @sihugh, I removed those unnecessary fields.

sihugh (approved):
🚀

PR #1426, opened 2019-02-07
Update Popularity worker to update supertypes

koetsier:
- Adds rake task ```update_supertypes``` to update supertypes in all documents. The Supertypes are retrieved from the https://github.com/alphagov/govuk_document_types Gem

Trello: https://trello.com/c/gVU3f7Xr/285-convert-supertypes-to-document-types-inside-rummager


tijmenb:
Is this a temporary fix? I thought that the supertypes would no longer be persisted in elasticsearch, @sihugh?

sihugh:
@tijmenb This is from the piece of work to make querying by supertype consistent across GOV.UK.

We looked at bringing the conversion from supertype to its consituent document types ([like we did here](https://github.com/alphagov/collections/pull/901/files)) into Rummager, but Jos's analysis of how supertype querying was used at the moment showed that it might be difficult to get right.  A simple example of this is a query that `rejects` the `other` group of `content_purpose_supertype`.  It's not stated anywhere what these document types are, so you'd have to replace it with a filter of the document types that are defined.

The main reason behind us wanting to query using a list of document types over their supertype was the fact that it was a pain to have to republish everything any time a supertype's membership changed.  Also, there was no guarantee that anything was up to date.  

We discovered the popularity worker, and thought that using a similar method to do the updating might be a neat way to achieve this update of supertype memberships regularly (without republishing) and thus negate the need for introducing additional complexity.

tijmenb:
Cool, thanks for the explanation! Good point about `other`, hadn't thought of that.

Would you mind updating the docs here: https://github.com/alphagov/govuk_document_types#how-to-add-or-update-a-supertype to say that it'll take a nightly reindex before it's available in Rummager?

lib/govuk_index/supertype_worker.rb
```diff
@@ -0,0 +1,36 @@
+module GovukIndex
+  class SupertypeWorker < Indexer::BaseWorker
+    BULK_INDEX_TIMEOUT = 60
+    QUEUE_NAME = 'bulk'.freeze
+    sidekiq_options queue: QUEUE_NAME
+
+    def perform(records, destination_index)
+      actions = Index::ElasticsearchProcessor.new(client: GovukIndex::Client.new(timeout: BULK_INDEX_TIMEOUT, index_name: destination_index))
+
+      filtered_records = records.reject { |record|
```
bilbof: Maybe call this something more specific, like 'changed_records'

bilbof (dismissed):
LGTM

bilbof (approved):

PR #1437, opened 2019-02-20
Remove unset metadata values from documents amended by MetadataTagger

bilbof:
Currently it is not possible to remove a metadata value by leaving a field
unset in the metadata CSV when running the rake task metadata_tagger:tag_metadata.

This will set an empty array for the fields which don't have a value set.

### DIFF
--- a/lib/indexer/metadata_tagger.rb
+++ b/lib/indexer/metadata_tagger.rb
@@ -49,9 +49,7 @@ module Indexer
         row_index = index + 1
         metadata[facet["key"]] = row.fetch(row_index, "").split(",").map(&:strip)
       end
-      metadata.reject do |_, value|
-        value == []
-      end
+      metadata
     end
 
     def self.all_nil_metadata_hash
--- a/spec/unit/indexer/metadata_tagger_spec.rb
+++ b/spec/unit/indexer/metadata_tagger_spec.rb
@@ -10,7 +10,13 @@ RSpec.describe Indexer::MetadataTagger do
     {
       "sector_business_area" => %w(aerospace agriculture),
       "business_activity" => %w(yes),
-      "appear_in_find_eu_exit_guidance_business_finder" => "yes"
+      "appear_in_find_eu_exit_guidance_business_finder" => "yes",
+      "employ_eu_citizens" => [],
+      "eu_uk_government_funding" => [],
+      "intellectual_property" => [],
+      "personal_data" => [],
+      "public_sector_procurement" => [],
+      "regulations_and_standards" => [],
     }
   }
   let(:mock_index) { double(:index) }
@@ -87,6 +93,12 @@ RSpec.describe Indexer::MetadataTagger do
           "appear_in_find_eu_exit_guidance_business_finder" => "yes",
           "business_activity" => %W(yes),
           "sector_business_area" => %W(aerospace agriculture),
+          "employ_eu_citizens" => [],
+          "eu_uk_government_funding" => [],
+          "intellectual_property" => [],
+          "personal_data" => [],
+          "public_sector_procurement" => [],
+          "regulations_and_standards" => [],
         }
       )
 


bevanloon:
Have you tested in rummager what actually happens if you pass the empty array in? I can't quite remember how rummager->elasticsearch works in this case.

If you need to remove some values then you could just do this https://docs.publishing.service.gov.uk/manual/business-readiness-update-content.html#removing-content-from-the-business-readiness-finder as part of the update process. You can destroy the metadata for a base path or set of base paths or for the whole csv. If you then do step 6 https://docs.publishing.service.gov.uk/manual/business-readiness-update-content.html#updating-content, it should pull itself together :)

steventux:
I think the reason we reject empty facet arrays like this is because it emulates the behaviour of the publishing apps when 'tagging' content with facet values. ie. we do not send empty form params. So the metadata generated by the tagger here will be consistent with metadata created and updated in the publishing apps.
Also it would be worth understanding the impact of allowing empty facet data like this on finders. There may be some assumptions in finder-frontend about the presence of facet keys...

kevindew:
We, and the team before us, on 2nd line got caught out for quite a while that changes to the spreadsheet that remove an entry to a field don't have any affect with the metadata tag script.

The other option considered than an empty array would be setting null but that seems more dangerous and, I expect, would still leave a trail in the index.

If this is undesirable should we be considering running a strip metadata query as part of this?

Alternatively it's a question of cranking up the docs more but this feels like something that would always be somewhat surprising behaviour.

bevanloon:
@kevindew  - certainly the first few incarnations we were dealing mostly with adding rather than removing data. It might be useful to run `Indexer::MetadataTagger.destroy_all_eu_exit_guidance!` before https://github.com/alphagov/rummager/blob/b67abd60b0eb44db0612e02d6e2495a67c94daea/lib/tasks/metadata_tagger.rake#L5

I think there will be a small blip where nothing will be returned by the finder though. Maybe it's good enough for now though? 

Otherwise we could incorporate a destroy per row before amending. That would need something like https://github.com/alphagov/rummager/blob/8709e6b7e431784493db128aa26bb253d5ac130c/lib/indexer/metadata_tagger.rb#L73-L76 to be put in place somewhere around https://github.com/alphagov/rummager/blob/8709e6b7e431784493db128aa26bb253d5ac130c/lib/indexer/metadata_tagger.rb#L29

I'm not sure if there's a more elegant way since that would call `Indexer::AmendWorker.new.perform` twice in pretty quick succession for a base path which feels annoying.

what do you think @steventux?

steventux:
I think regularly blowing away all the metadata could also have consequences in that email notifications are triggered for items considered _new_ to the finder. This is determined by something being in the Rummager index but not returned by the search for all guidance for the business readiness finder https://github.com/alphagov/rummager/blob/8709e6b7e431784493db128aa26bb253d5ac130c/lib/indexer/metadata_tagger.rb#L87

The emails are sent via a worker here
https://github.com/alphagov/rummager/blob/8709e6b7e431784493db128aa26bb253d5ac130c/lib/indexer/metadata_tagger.rb#L34

I think removing all metadata would mean that all content would be considered new. So preserving the `appear_in_find_eu_exit_guidance_business_finder = 'yes'` data would be necessary.

I also think we should be pouring effort into getting away from this tagging system now. We have a story in our backlog to improve the tagging process, @thomasleese 's suggestion of making facets / facet values a format we can link to being the best suggestion we have so far.

In the short term if we preserve `appear_in_find_eu_exit_guidance_business_finder` when cleaning the metadata the emails won't be triggered.

As far as content disappearing momentarily as the metadata is unset and reset, I'd imagine for a high visibility tool like this we'd need to be careful and could draw some negative attention. So blowing away metadata on a per case basis, though time consuming, seems to be better than removing it all for everything.

kevindew:
> I think regularly blowing away all the metadata could also have consequences in that email notifications are triggered for items considered _new_ to the finder. This is determined by something being in the Rummager index but not returned by the search for all guidance for the business readiness finder

Hmm does that mean that when we run the remove metadata command to clear fields that might also be triggering new emails when we re-add them? We had to do that on two docs yesterday.

> I also think we should be pouring effort into getting away from this tagging system now. We have a story in our backlog to improve the tagging process, @thomasleese 's suggestion of making facets / facet values a format we can link to being the best suggestion we have so far.

That sounds great 👍 
 
> In the short term if we preserve `appear_in_find_eu_exit_guidance_business_finder` when cleaning the metadata the emails won't be triggered.

Ok cool. 

Do you know how common this scenario we've hit here is? If I understand right, our remove then re-add approach isn't great as it will send an email. The approaches around that are relatively complex though.

So then I wonder if this empty array might be just be fine, so long as queries to rummager check there is a particular value in a facet rather than the facet itself exists?

steventux:
I think the scenario is relatively uncommon because the field stores a `yes` value or nothing, whereas tagging a document various values typically takes the form of updating with something else rather than removing a value.
I agree that setting with an empty array might be the best compromise for now if we can prove it doesn't have consequences for the search results.

steventux (approved):

PR #1439, opened 2019-02-25
Make Rummager Search API Again

barrucadu:
We don't want to lose the PR history that this repository has, so the search-api repository is only temporary: we use it while we have rummager and search-api running in parallel, then we:

- merge the changes back into rummager,
- remove the search-api repository,
- and rename this repository to "search-api".

This PR contains all the changes made to search-api.  Below is a summary of PR bodies:

## Do the bare minimum rummager/search-api renaming

> I don't want to go through and change all mentions of rummager, as that will increase chances of merge conflicts while we're still in a half-migrated state bringing commits over from alphagov/rummager to alphagov/search-api.

Trello card: [Set up new search-api app](https://trello.com/c/xNNHprUZ/20-set-up-new-search-api-app)

## Use elasticsearch 5

> Updates rummager to use elasticsearch 5, dropping elasticsearch 2 compatibility, whilst preserving the externally visible API.
>
> Elasticsearch 5 has changed the default similarity scoring metric to BM25, which behaves differently to the old one.  The old one is being phased out in [Lucene](https://issues.apache.org/jira/browse/LUCENE-7347), so we have no choice but to switch.

Trello card: [Make rummager able to query es5](https://trello.com/c/q5WUhqbt/12-make-rummager-able-to-query-es5)
Trello card: [Make rummager able to update es5](https://trello.com/c/Eblfvmke/16-make-rummager-able-to-update-es5)

## Use a single mapping type per index

> We previously created indices that contained more than one document type.  This feature has been deprecated in Elasticsearch 6.x for newly created indices.
>
> An additional field, `document_type` now contains the type that was previously handled by Elasticsearch.  This field is implicitly included into all schemas. All documents now have the `_type` of 'generic-document'.
>
> The externally facing API should not have changed:
>
> - For input, the `document_type` field is used if present otherwise it falls back to the old `_type` field.
>
> - For output, we abstracted over the elasticsearch `_type` field in results anyway
>
> This PR also removes the `DuplicateDeleter` and associated rake tasks, as that looks for documents which have the same ID but different types, which is now impossible.

Trello card: [Remove multiple document types in our ES5 instance](https://trello.com/c/gyGwwla1/54-remove-multiple-document-types-in-our-es5-instance)

## Use a different content ID for taxon removal test

> This same content ID is used in the external content tests.  We use the content ID as the document ID in these integration tests (normally we use the base path).  As we're now using a single mapping type, this means the document identifiers collide.
>
> We use versioning to ensure that a document in elasticsearch hasn't changed while we attempt to delete it.  I assume inserting a second document with the same ID is altering the version number, and so causing the delete to fail.
>
> I don't think this problem can arise during normal execution, as it requires two documents to have the same ID, and the publishing-api prevents that by ensuring that both base paths and content IDs are unique.

Trello card: [Fix flakey search-api integration test](https://trello.com/c/RIKZlw0k/66-fix-flakey-search-api-integration-test)

## End to end tests

> Make the e2e tests work, which involves switching away from the search-api port back to the rummager port.

Trello card: [Ensure end-to-end tests work with managed Elasticsearch](https://trello.com/c/oqlNhmVt/72-ensure-end-to-end-tests-work-with-managed-elasticsearch)

## Change Plek to use 'external: true' for publishing-api

> Publishing API is still in Carrenza, so we need to tell Plek to use `external: true` to generate the external URL (i.e. publishing-api.staging.publishing.service.gov.uk instead of publishing-api.staging.govuk-internal.digital).

## Do not use "must" with "should" or "filter"

> A matching "must" overrides a failing "should" and "filter":
>
> > If the `bool` query is in a query context and has a `must` or `filter` clause then a document will match the `bool` query even if none of the `should` queries match. In this case these clauses are only used to influence the score.
>
> So this PR makes two changes:
>
> 1. Fix the one case where we have a `must` / `should` pair (best bets)
> 2. Turn all the cases where we have a `must` / `filter` pair into a `query` with a `must` + a `post_filter`, which is closer to how a lot of these things were before I changed the queries in the first place.

Trello card: [Best bets are broken](https://trello.com/c/DrGWIXbo/85-best-bets-are-broken)

## Revert similarity metric and schema to ES2 versions

> We have a problem with search results being Weird.  These changes fix it.  We'll need to revisit this before moving to ES6 though.

Trello card: [Investigate why the ordering of search results is Weird](https://trello.com/c/C3STgB2P/86-ordering-of-search-results-is-weird)

## Change the index blocking method

> AWS have a process that runs regularly and updates `index.blocks.write` based on the cluster health.  This resulted in the blocking being reset periodically.
>
> After discussion with AWS, they proposed changing this setting to `index.blocks.read_only_allow_delete`, which is not reset by their processes.
>
> Additionally, the locking needed updating since `index.blocks.read_only_allow_delete` does not permit alias updates, so the index must be unlocked before changing aliases.

Trello card: [Something seems broken with AWS managed ES reindexing](https://trello.com/c/73ligyT2/88-something-seems-broken-with-aws-managed-es-reindexing)

## Review documentation

Trello card: [Update medium priority documentation](https://trello.com/c/MvemOpZV/93-update-medium-priority-documentation)

## Revert "Change Plek to use 'external: true' for publishing-api"

> The environment changes have been made in Puppet

## Expand all synonyms

> We had trouble with equivalent synonyms not being expanded. Explicitly setting `expand` to true in the schema didn't change anything, but writing the equivalent synonyms in the explicit `expand`-is-true form did.  Not really sure what's up with that.

Trello card: [Searching for department+of+health doesn't return all the matches](https://trello.com/c/Ivb7TKAZ/104-searching-for-departmentofhealth-doesnt-return-all-the-matches)

## Do not close old page-traffic index

> AWS managed Elasticsearch doesn't allow the close API.  Leaving the index open doesn't prevent it from being removed by the rummager:clean task, so this isn't a problem.

## Correctly set page traffic document type

No description.

## Keep rummager as the publishing_app for now

> The paths are reserved by rummager in the publishing-api, we'll change that (and this) when tearing down rummager.

## Bump limit of default sidekiq queue to 8

> We've occasionally seen a build up of sidekiq jobs (in both search-api
> and rummager).  I experimented by bumping the limit on the "default"
> queue to 8 via the app console during one such spike: the jobs cleared
> much faster than in rummager (which still had the limit of 4), and the
> Elasticsearch search latency increased by maybe a couple of
> milliseconds - it's hard to say because any increase is small enough
> to be obscured by the natural variability of the metric.
>
> These limits were added to solve a problem, but that problem occurred
> with an almost totally different search architecture, so I think it's
> worth experimenting with the limits a bit.  They could perhaps be
> increased further, but let's see how this change performs for now.
>
> Elasticsearch indexing rate (units are operations/minute), clear change showing when the number of workers went up:
> Screen Shot 2019-04-05 at 12 19 42
>
> Elasticsearch search latency (units are ms):
> Screen Shot 2019-04-05 at 12 20 19
>
> search-api queue with limit 8:
> Screen Shot 2019-04-05 at 12 17 17
>
> rummager queue with limit 4:
> Screen Shot 2019-04-05 at 12 17 52

## Include facet_group in the patch_links call

> When publishing a finder, we need the new `facet_group` links to be included in the links sent to `patch_links` if it exists.
> 
> Discovered during https://trello.com/c/ovMEU0aL/283-end-to-end-testing-on-integration-of-new-links-based-approach-for-tagging

This is a change which should probably have been made to rummager and then merged into search-api.

## Catch and ignore BadRequest exceptions in best bet text analysis

> Some strings when passed to `@client.indices.analyze`, like "*" and "", give a bad request error. I don't fully understand what the elasticsearch library is doing, because I can analyse such strings manually with curl just fine, but here they're a problem.
>
> Discard the exception and just say there are no best bets. If there is some more significant problem with the query, then it will probably be flagged up when the actual search is performed.

## Reject long queries

> We're seeing a few long, slow-running, queries which are clearly not something entered by a human. Stemming, synonyms, and other analyses aren't free - so spending 1.3s executing a 1677-character query which no human is looking at the result of seems like a waste of time and effort.
>
> We don't want to impact human users, so the new limit is quite high, and there will be follow-up work in finder-frontend to suggest the contact form when someone enters a long query.

Trello card: [Return a 422 through the public search API for any queries >512 characters](https://trello.com/c/taQuHzDe/124-return-a-422-through-the-public-search-api-for-any-queries-512-characters)

## Treat unmapped types as a missing value in sorting

> Sorting on an unmapped field throws an error. This is a problem because our indices have different fields in their mappings. So treat an unmapped field as an integer, which will be given a missing value, so the doc will then be caught by the missing: "_last".
>
> https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-sort.html#_ignoring_unmapped_fields

Trello card: [Look into "No mapping found for x errors"](https://trello.com/c/jz6vHhDv/120-look-into-no-mapping-found-for-x-errors)

## Change default port back to 3233

There's now a branch of the e2e tests making this work: https://github.com/alphagov/publishing-e2e-tests/pull/295

Note that this doesn't affect the port used when search-api / rummager is *deployed*, that is set in puppet.

## Only generate 25,000 links per sitemap file

> We've seen issues with the sitemap generation task running out of memory. Not sure why this has started to happen over the last few days, but a quick fix is to reduce the number of links per file - as the entire xml file is generated in memory before being written to disk.


README.md
```diff
@@ -1,4 +1,6 @@
-# Rummager
+# Search API
+
+**This is a fork of [rummager](https://github.com/alphagov/rummager), and is in a half-renamed state!  We will clean this up as the migration progresses.**
```
bilbof: You're probably aware of this but FYI the note above looks to be incorrect now ☝️

bilbof (approved):
Looks good!

README.md
barrucadu: I've changed the note

PR #1444, opened 2019-03-04
Stop filtering 'other' supergroup in all-content finder

koetsier:
Some small changes to the all-content finder

- include the 'other' content purpose supergroup.
- relevance as the default sort order
- remove email and atom feed link

Trello: https://trello.com/c/ftI0xd67/459-make-post-review-changes-to-all-content-finder

### DIFF
--- a/config/finders/all_content.yml
+++ b/config/finders/all_content.yml
@@ -13,9 +13,6 @@ details:
   default_documents_per_page: 20
   document_noun: result
   format_name: Documents
-  reject:
-    content_purpose_supergroup:
-    - other
   facets:
   - key: "_unused"
     display_as_result_metadata: false
--- a/config/finders/all_content.yml
+++ b/config/finders/all_content.yml
@@ -8,7 +8,6 @@ rendering_app: finder-frontend
 schema_name: finder
 title: All content
 description: Find content from government
-signup_content_id: 49838f97-5d2e-407a-8876-30c4abe2a6bc
 details:
   default_documents_per_page: 20
   document_noun: result
@@ -63,8 +62,4 @@ details:
     name: Updated (oldest)
 routes:
 - path: "/all-content"
-  type: exact
-- path: "/all-content.atom"
-  type: exact
-- path: "/all-content.json"
   type: exact
--- a/config/finders/all_content.yml
+++ b/config/finders/all_content.yml
@@ -51,12 +51,12 @@ details:
     type: date
   show_summaries: true
   sort:
-  - key: "-popularity"
+  - default: true
+    key: "-popularity"
     name: Most viewed
   - key: "-relevance"
     name: Relevance
-  - default: true
-    key: "-public_timestamp"
+  - key: "-public_timestamp"
     name: Updated (newest)
   - key: public_timestamp
     name: Updated (oldest)


config/finders/all_content.yml
```diff
@@ -6,16 +6,12 @@ locale: en
...
-title: All content
 description: Find content from government
 signup_content_id: 49838f97-5d2e-407a-8876-30c4abe2a6bc
 details:
   default_documents_per_page: 20
   document_noun: result
   format_name: Documents
-  reject:
-    content_purpose_supergroup:
-    - other
```
bilbof, line 18: 😢 can we remove `reject_content_purpose_supergroup` from the email alert api if we're not using this anymore?

config/finders/all_content.yml
```diff
@@ -6,16 +6,12 @@ locale: en
 publishing_app: rummager
 rendering_app: finder-frontend
 schema_name: finder
-title: All content
```
bilbof: Does it still work without a title? You might need to update the content schemas (either making title optional or adding a new field to specify whether the title should be displayed).

https://github.com/alphagov/govuk-content-schemas/blob/1c942c7e6f5101e69a563024630d2d27ce3de092/formats/finder.jsonnet

https://github.com/alphagov/govuk-content-schemas/blob/1c942c7e6f5101e69a563024630d2d27ce3de092/formats/shared/default_format.jsonnet#L7

koetsier: ah I did not know that. I'll add an empty field instead

koetsier: That does not work either. I put the title back for now. Removing the title is now a new card

config/finders/all_content.yml
```diff
@@ -55,19 +51,15 @@ details:
...
-    key: "-public_timestamp"
+  - key: "-public_timestamp"
     name: Updated (newest)
   - key: public_timestamp
     name: Updated (oldest)
 routes:
 - path: "/all-content"
   type: exact
-- path: "/all-content.atom"
-  type: exact
```
bilbof, line 71: Will this also remove the route, or is that a separate task?

config/finders/all_content.yml
```diff
@@ -8,14 +8,10 @@ rendering_app: finder-frontend
 schema_name: finder
 title: All content
 description: Find content from government
-signup_content_id: 49838f97-5d2e-407a-8876-30c4abe2a6bc
```
bilbof, line 11: Shall we also delete the email signup content item from rummager? We should unpublish it, too.

bilbof (dismissed):

config/finders/all_content.yml
koetsier: yes good idea

koetsier: Added a trello card for this

config/finders/all_content.yml
koetsier: Yes, I'll add a point to the trello card mentioning that we should do that as part of deployment

bilbof (approved):

config/finders/all_content.yml
koetsier: I suspect it will remove the route through the publishing api, but I'll make a note to check.

PR #1446, opened 2019-03-05
Fix small bugs in publish_finder rake task

koetsier:
Fixed typo and now email_signup_config and finder_config env
variables are properly optional

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -60,16 +60,21 @@ namespace :publishing_api do
     finder_config = ENV["FINDER_CONFIG"]
     email_signup_config = ENV["EMAIL_SIGNUP_CONFIG"]
 
-    unless finder_config || email_signup_finder_config
+    unless finder_config || email_signup_config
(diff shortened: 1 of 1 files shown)


lib/tasks/publishing_api.rake
```diff
@@ -60,16 +60,21 @@ namespace :publishing_api do
...
-    ContentItemPublisher::FinderPublisher.new(finder, timestamp).call if finder
+    unless finder_config.nil?
+      finder = YAML.load_file("config/finders/#{finder_config}")
+      ContentItemPublisher::FinderPublisher.new(finder, timestamp).call
+    end
+
+    unless email_signup_config.nil?
+      email_signup = YAML.load_file("config/finders/#{email_signup_config}")
+      ContentItemPublisher::FinderEmailSignupPublisher.new(email_signup, timestamp).call
+    end
```
bilbof, line 72: Should we continue to publish the email content item first? That way when the finder itself is published the email signup content item it refers to already exists.

sophiebrown01 (approved):

lib/tasks/publishing_api.rake
koetsier: good idea - I swapped them

bilbof (approved):

PR #1447, opened 2019-03-06
Fix bug in patching links in FinderPresenter

koetsier:
The publish_document_finder rake task publishes a given content item
and then performs a patch links api call to to the publishing api to update 
links in the content item.

Removing links from an existing content item requires a call to
patch links containing the relevant key and an empty array. Omitting
the key will leave the link in place.

This fix ensures that removing either email_alert_signup or parent links
works properly.


oscarwyatt (approved):
This sparks joy

PR #1449, opened 2019-03-06
Add sort option to prepare EU Exit finders

bilbof:
https://trello.com/c/C9rx0dwK/401-add-a-sort-facet-to-the-citizen-readiness-finders

This will enable users of the Prepare EU Exit finders to sort results. It will default to showing the most
popular results first.

I ran [the rake task](https://www-origin.integration.publishing.service.gov.uk/prepare-eu-exit/going-and-being-abroad?parent=&keywords=brexit&order=relevance) that publishes these finders on integration and it worked.

Example:
https://www-origin.integration.publishing.service.gov.uk/prepare-eu-exit/going-and-being-abroad

### DIFF
--- a/config/prepare-eu-exit.yml.erb
+++ b/config/prepare-eu-exit.yml.erb
@@ -30,6 +30,16 @@ details:
       - guidance_and_regulation
   subscription_list_title_prefix: <%= title %>
   show_summaries: true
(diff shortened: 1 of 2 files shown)


koetsier (approved):
lgtm

PR #1452, opened 2019-03-07
Allow specific facets to use the new checkbox filter feature

aliuk2012:
The new key added to the facets will control whether user see a filter textfield as shown in image below 

image

Ticket: https://trello.com/c/sEGGP7dF/470-turn-on-option-select-filters-for-finders


bilbof (approved):
LGTM

PR #1453, opened 2019-03-07
Query by facets as links

peteglondon:
# What
Allow rummager to index and query the following link types:

* facet_values
* facet_groups

As part of the work to address tech debt introduced by the [EU Exit business finder](https://www.gov.uk/find-eu-exit-guidance-business) and to [Improve the Business Finder tagging system](https://trello.com/c/zn5LBg3L/77-improve-the-business-finder-tagging-system).

We are introducing Facet Groups, Facets and FacetValues as content items.

Tagging to these things will then be done in `content-tagger` (similar to taxonomy) and implemented as links.

Trello: https://trello.com/c/RHl653OZ/245-rummager-query-on-facet-group-and-facet-value

### DIFF
--- a/lib/govuk_index/presenters/expanded_links_presenter.rb
+++ b/lib/govuk_index/presenters/expanded_links_presenter.rb
@@ -32,6 +32,10 @@ module GovukIndex
       content_ids("taxons")
     end
 
+    def facet_values
+      content_ids("facet_values")
+    end
+
     def topic_content_ids
       content_ids("topics")
     end
--- a/lib/govuk_index/presenters/expanded_links_presenter.rb
+++ b/lib/govuk_index/presenters/expanded_links_presenter.rb
@@ -32,6 +32,10 @@ module GovukIndex
       content_ids("taxons")
     end
 
+    def facet_groups
+      content_ids("facet_groups")
+    end
+
     def facet_values
       content_ids("facet_values")
     end
--- a/config/schema/base_elasticsearch_type.json
+++ b/config/schema/base_elasticsearch_type.json
@@ -8,6 +8,7 @@
     "content_store_document_type",
     "description",
     "email_document_supertype",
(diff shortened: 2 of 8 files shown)


peteglondon:
[Mapping Rummager request/response in Finder Frontend](https://trello.com/c/Sd7omEFH/263-mapping-rummager-request-response-in-finder-frontend)

peteglondon:
[Rummager query on facet group and facet value](https://trello.com/c/RHl653OZ/245-rummager-query-on-facet-group-and-facet-value)

lib/govuk_index/presenters/expanded_links_presenter.rb
```diff
@@ -32,6 +32,10 @@ def taxons
       content_ids("taxons")
     end
 
+    def facet_groups
+      content_ids("facet_groups")
```
steventux, line 36: At the moment, the link from facet to facet group is in the `parent` links array https://github.com/alphagov/govuk-content-schemas/blob/master/formats/facet.jsonnet#L90 
https://github.com/alphagov/govuk-content-schemas/blob/master/examples/facet/publisher_v2/facet_links.json#L7
I'm not sure `facet_groups` is a valid link type.

steventux (changes requested):
Looks good, one question around facet groups.

lib/govuk_index/presenters/expanded_links_presenter.rb
steventux: Having discussed this we've agreed that content either needs `facet_groups` links or the facet group to be present in the `facet_values` links, the former is more explicit see: https://github.com/alphagov/govuk-content-schemas/commit/90be6c695ae3f5827790c68d10cee7b61f16c7ae

steventux (approved):
LGTM

PR #1459, opened 2019-03-12
Change finder paths to have '/search' prefix

bilbof:
https://trello.com/c/SjudMnoT/499-update-urls-across-all-7-finders

We will need to republish all of these finders and email signup pages. I removed a commit that added a rake task to redirect finders paths, as it wasn't necessary. This will create redirects automatically (see Simon's comment below).

### DIFF
--- a/config/finders/all_content.yml
+++ b/config/finders/all_content.yml
@@ -1,5 +1,5 @@
 ---
-base_path: "/all-content"
+base_path: "/search/all"
 content_id: dd395436-9b40-41f3-8157-740a453ac972
 document_type: finder
 locale: en
@@ -64,5 +64,7 @@ details:
   - key: public_timestamp
     name: Updated (oldest)
 routes:
-- path: "/all-content"
+- path: "/search/all"
+  type: exact
+- path: "/search/all.json"
   type: exact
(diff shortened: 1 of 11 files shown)


sihugh (commented):
Publishing api should handle redirects for us, so we don't need step 2.  We just need to change the base path and routes and republish 🙏

bilbof:
Thanks, @sihugh I removed that second commit and updated the description. :)

config/finders/all_content.yml
```diff
@@ -64,5 +64,5 @@ details:
   - key: public_timestamp
     name: Updated (oldest)
 routes:
-- path: "/all-content"
+- path: "/search/all-content"
```
sihugh: The json route got removed at some point.  Live search uses this to update the results.  Any chance you could add this back in please?  (And for the guidance and regulation finder too please?).

I think those routes are still present in router which is why these finders still work, but they'll be broken when we republish these unless we add them back in.

bilbof: I added the JSON route for the all content finder.

Guidance and regulation appears to have a JSON route:
https://github.com/alphagov/rummager/pull/1459/files#diff-761566fda268c4381877f7ab723294cbR68

sihugh: Cool, just my eyes then 👍

sihugh (approved):

PR #1464, opened 2019-03-14
Use update instead of index for updating popularity data

cbaines:


cbaines:
I'm not looking at this at the moment, so closing.

PR #1466, opened 2019-03-14
Have Rummager wait while updating the popularity

cbaines:
Currently updating the popularity values involves putting the
documents in the index on to the Sidekiq queue, so that they can be
sent back to Elasticsearch with the updated popularity value.

This uses lots of memory, and recently, enough memory to cause Redis
to run out.

I've attempted to switch this to use the update, rather than index
action, but this doesn't work due to the use of `external_gte`
versioning, as the update increments the version, which could cause a
problem with processing messages from the Publishing API. ADR 002
describes one way to workaround this issue "External versioning with
multiplier", but changing the versioning scheme seems like a big
change to me.

Instead, this commit changes the popularity updating Rake task to
block while the jobs are being processed. As the nightly search
analytics job updated 3 indexes, I'm hoping that this will reduce the
peak Redis memory usage.

### DIFF
--- a/lib/govuk_index/popularity_updater.rb
+++ b/lib/govuk_index/popularity_updater.rb
@@ -6,6 +6,8 @@ module GovukIndex
         destination_index: index_name,
         process_all: process_all,
       ).run
+
+      worker.wait_until_processed
     end
 
     def self.worker
--- a/lib/tasks/indices.rake
+++ b/lib/tasks/indices.rake
@@ -88,7 +88,8 @@ correctly
   desc "Update popularity data in indices.
 
 Update all data in the index inplace (without locks) with the new popularity
-data using sidekiq workers.
+data using sidekiq workers. The Rake task will exit when the workers
+have finished.
 
 This does not update the schema.
 "


benthorner:
This looks good to me - I wonder if it would be more self-documenting to put the wait in the search-analytics repo, as that's where the work is being 'managed'. Also, is there a way to profile this on integration or locally, to see that it works?

cbaines:
> This looks good to me - I wonder if it would be more self-documenting to put the wait in the search-analytics repo, as that's where the work is being 'managed'. 

Yeah, updating the job description was my attempt. Maybe the waiting could be a separate Rake task which is run by the nightly job script? It could then run that between each task that enqueues jobs.

> Also, is there a way to profile this on integration or locally, to see that it works?

I presume just deploy it, run the nightly job, and compare the Sidekiq queue lengths and Redis memory usage with data from a week or two ago.

cbaines:
The search-analytics script has been modified to have a similar effect https://github.com/alphagov/search-analytics/commit/8f24581b95af9b37581f16988f466364d5de5808 . So I'll close this now.

PR #1470, opened 2019-03-18
Publish business readiness finder and email signup separately

leenagupte:
Trello: https://trello.com/c/0rDl6oiK

Related to: 
* https://github.com/alphagov/finder-frontend/pull/954
* https://github.com/alphagov/govuk-app-deployment-secrets/pull/90
* https://github.com/alphagov/govuk-developer-docs/pull/1623

## Motivation
The way email signups work for the business readiness finder is changing. This is because the finder works differently to other finders.
Rather than narrowing the results, the finder expands them, doing an "or" query rather than an "and". That is not how email-alert-api works. It creates an "and" query from the facets it's passed.

This means that most users subscribed to the business readiness finders were not receiving updates.

## What's changed

"appear_in_find_eu_exit_guidance_business_finder" is being added to the email-signup content item:

```
    "email_filter_by": "appear_in_find_eu_exit_guidance_business_finder",
    "email_filter_name": "appear_in_find_eu_exit_guidance_business_finder",
    "email_signup_choice": [
      {
        "key": "yes",
        "radio_button_name": "yes",
        "prechecked": true
      }
    ],
```
This replaces `email_filter_facets` in the content_item. If `email_filter_facets` are present, finder-frontend forces the user to subscribe to a facet, rather than letting them subscribe to the whole list.

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -78,6 +78,30 @@ namespace :publishing_api do
     puts "FINISHED"
   end
 
+  desc "
+    Publish business readiness finder and email signup content items
+
+    Usage:
+    rake publishing_api:publish_eu_exit_business_finder
+  "
+  task :publish_eu_exit_business_finder do
+    finder_config = File.join(Dir.pwd, "config", "find-eu-exit-guidance-business-email-signup.yml")
+    email_signup_config = File.join(Dir.pwd, "config", "find-eu-exit-guidance-business.yml")
+
+    timestamp = Time.now.iso8601
+
+    unless email_signup_config.nil?
+      email_signup = YAML.load_file(email_signup_config.to_s)
+      ContentItemPublisher::FinderEmailSignupPublisher.new(email_signup, timestamp).call
+    end
+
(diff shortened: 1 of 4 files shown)


lib/tasks/publishing_api.rake
```diff
@@ -78,6 +78,30 @@ namespace :publishing_api do
...
   end
 
+  desc "
+    Publish business readiness finder and email signup content items
+
+    Usage:
+    rake publishing_api:publish_eu_exit_business_finder
+  "
+  task :publish_eu_exit_business_finder do
+    finder_config = File.join(Rails.root, "config", "find-eu-exit-guidance-business-email-signup.yml")
```
steventux: Is `Rails.root` defined in a Sinatra app?

steventux (changes requested):

lib/tasks/publishing_api.rake
leenagupte: I keep forgetting it's a Sinatra app!

steventux (approved):

PR #1472, opened 2019-03-19
Change path of stats finder

bilbof:
https://trello.com/c/VMPLPT1X/538-change-stats-finder-url

We will need to run publishing_api:publish_finder once this is deployed.


sophiebrown01 (approved):
👍

PR #1485, opened 2019-03-25
Update sectors in fixtures and business config yml

vanitabarrett:
Trello: https://trello.com/c/WoCAgoHq/153-improve-the-sector-list

### DIFF
--- a/config/find-eu-exit-guidance-business.yml
+++ b/config/find-eu-exit-guidance-business.yml
@@ -25,6 +25,8 @@ details:
       default: true
     - name: Most viewed
       key: -popularity
+    - name: Relevance
+      key: -relevance
     - name: Most recent
       key: -public_timestamp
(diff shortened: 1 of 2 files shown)


config/find-eu-exit-guidance-business.yml
```diff
@@ -1,9 +1,4 @@
 ---
-#
-# NOTE: This is a development copy of the find-eu-exit-guidance-business finder config
```
leenagupte: I think we should keep the comment for a few weeks, until we no longer need govuk-app-deployment-secrets

leenagupte (approved):

config/find-eu-exit-guidance-business.yml
vanitabarrett: Oops, that wasn't intentional. Put back 🙂

PR #1489, opened 2019-03-26
Update statistics title

sihugh:
We need to indicate that there are research documents in this finder too.

https://trello.com/c/kyjyR5RB/573-update-title-of-statistics-finder

### DIFF
--- a/config/finders/statistics.yml
+++ b/config/finders/statistics.yml
@@ -8,7 +8,7 @@ publishing_app: rummager
 rendering_app: finder-frontend
 schema_name: finder
 signup_content_id: 119db584-0ae7-45e4-8f3a-fd79316c6921
-title: Statistics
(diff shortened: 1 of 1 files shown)


sophiebrown01 (approved):
Guessing we should probably update https://github.com/alphagov/finder-frontend/blob/master/features/fixtures/statistics.json too 

👍

bilbof (approved):

sihugh:
@karlbaker02 I'm not really fussed about keeping finder-frontend in step with the content items.  Those files are there for development purposes and for test fixtures.  As long as our test fixtures allow us to exercise the functionality we've added that's fine, isn't it?

I'd rather rely on these files being a single source of truth, and the ones in finder-frontend being indicative, but not real.

PR #1491, opened 2019-03-27
Adds hidden clearable manual facet to all content finder

oscarwyatt:
Adds hidden clearable manual facet to all content finder

Trello: https://trello.com/b/yCngBbCU/bravigation-doing-q4-q3

### DIFF
--- a/config/finders/all_content.yml
+++ b/config/finders/all_content.yml
@@ -31,7 +31,7 @@ details:
     name: Manual
     preposition: in manual
     short_name: in
-    type: hidden
+    type: hidden_clearable
     show_option_select_filter: false
     allowed_values: []
   - display_as_result_metadata: true


config/finders/all_content.yml
```diff
@@ -31,9 +25,7 @@ details:
     name: Manual
     preposition: in manual
     short_name: in
-    type: hidden
-    show_option_select_filter: false
-    allowed_values: []
+    type: hidden_clearable
```
sihugh, line 34: Isn't this the only change we need?  I think the other changes look more like a frontend schema - for example we don't need the supertypes, and we definitely do need the routes.

In other news, I noticed yesterday that the finderpresenter tests aren't running properly, so they're not picking up schema violations where they should here: https://github.com/alphagov/rummager/blob/master/spec/unit/content_item_publisher/finder_presenter_spec.rb#L30.  I was going to have a poke to fix them.

sihugh (changes requested):
I'm not sure this is quite right - we can't directly translate a frontend schema to a backend schema, sadly.

config/finders/all_content.yml
oscarwyatt: Apologies, act in haste, repent at leisure, fixed now

sihugh (approved):

config/finders/all_content.yml
sihugh: We've all done it 🤦‍♂️

PR #1494, opened 2019-03-28
Adds filter values to research and statistics finder as well as changing

oscarwyatt:
Adds filter values to research and statistics finder as well as changing description to make it clearer to users why they can't sign up to upcoming
statistics

Trello: https://trello.com/c/qBt4q04L/548-prevent-users-from-subscribing-to-email-alerts-of-upcoming-statistics


sophiebrown01 (approved):
As discussed 👍

PR #1500, opened 2019-04-02
Fix finder definition testing

sihugh:
_This is a good candidate for [viewing without whitespace changes](https://github.com/alphagov/rummager/pull/1500/files) because the tests are mostly the same apart from being wrapped in an additional `context` block_

These tests weren't running against all the definition files, so invalid schemas were sneaking through.

I've updated them to iterate through all files in the folder with the amended naming convention that finder definitions end with `_finder`

I've also updated the failing definitions to pass schemas.

This should pass tests when https://github.com/alphagov/govuk-content-schemas/pull/880 is deployed.


sophiebrown01 (approved):
Looks good 👍 (subject to fixing-up / squashing commits)

sihugh:
I'll squash them up, thanks @karlbaker02 🙇

PR #1501, opened 2019-04-02
Allow and filter across facet values

peteglondon:
The problem with moving to the new `facet_values`
type of query, instead of the existing metadata facets,
is that Rummager treats queries on the same column
as `OR` queries.

All of the existing facets are in different fields,
the `facet_values` approach makes this a single field.

This commit introduces a second copy of the `facet_values`
called `and_facet_values`

When populated with identical data this will allow queries like:

&filter_facet_values=uuid-from-first-facet[....]&filter_and_facet_values=uuid-from-different-facet[...]

### DIFF
--- a/config/schema/base_elasticsearch_type.json
+++ b/config/schema/base_elasticsearch_type.json
@@ -10,6 +10,7 @@
     "email_document_supertype",
     "facet_groups",
     "facet_values",
+    "and_facet_values",
     "format",
     "government_document_supertype",
     "indexable_content",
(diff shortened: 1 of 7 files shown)


steventux (approved):
As discussed we know this approach is temporary, ideally we'll be able to pass Rummager parameters in a way that supports arbitrary numbers of AND clauses. We don't currently have a use case for > 2 facets using AND clauses so this will cover immediate needs.

PR #1504, opened 2019-04-03
Adds virtual filter 'research_and_statistics'

oscarwyatt:
Adds virtual filter 'research_and_statistics'

Filtering by upcoming statistics, published statistics or by research requires multiple filters (eg format and release timestamp), which cannot be easily done in finder frontend. This adds a 'virtual filter' that of 'research_and_statistics' which can be easily used by finder frontend to filter by 'upcoming_statistics', 'published_statistics' or 'research' to filter to these results without having to provide more than one filter.

Some effort has been made to make it easy to extend this if necessary in the future

Trello: https://trello.com/c/QpM8Vhwv/590-bug-stats-finder-is-only-displaying-a-few-upcoming-stats-docs
Trello: https://trello.com/c/mkf4eUXF/593-bug-stats-finder-displaying-wrong-number-of-published-stats-docs


sophiebrown01:
I have a general question around how these changes _should_ be giving us more results, because as long as content items are tagged to the correct document types (the announcement document types for upcoming, non-announcement document types for published) (which seems to be the case on Whitehall prod), then we shouldn't need to do anything?

On the other hand, this solves the issue with only 36 upcoming statistics being shown at present in the new finder. Interestingly, the dev docs for the upcoming stats document types sum up to a total of 36 documents when there are in fact many more - it is using the GOV.UK search index for this (example: https://docs.publishing.service.gov.uk/document-types/official_statistics_announcement.html).

This is perhaps warrants an investigation into what's happening, but in the meantime to publish the research and stats finder now with this PR.

sophiebrown01 (approved):

sophiebrown01 (approved):
.

PR #1506, opened 2019-04-04
Uses new research_and_statistics filter

oscarwyatt:
Research and stats finder uses new research_and_statistics filter

Trello: https://trello.com/c/QpM8Vhwv/590-bug-stats-finder-is-only-displaying-a-few-upcoming-stats-docs
Trello: https://trello.com/c/mkf4eUXF/593-bug-stats-finder-displaying-wrong-number-of-published-stats-docs

### DIFF
--- a/config/finders/statistics_finder.yml
+++ b/config/finders/statistics_finder.yml
@@ -68,33 +68,19 @@ details:
     display_as_result_metadata: false
     filterable: true
     preposition: about
-  - key: content_store_document_type
+  - key: research_and_statistics
     name: Statistics
     type: radio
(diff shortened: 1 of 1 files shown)


sihugh (approved):

bilbof (commented):
I suppose this means that we can filter rummager using `content_store_document_type=published_statistics`? I get no results at the moment (https://www.gov.uk/api/search.json?filter_content_store_document_type=published_statistics), unless I've missed a change in finder-frontend!

Can we remove option_lookup from the finder schema in this case?

PR #1519, opened 2019-04-18
statistics finder uses the new reasearch_and_statistics filters

koetsier:
trello: https://trello.com/c/mkf4eUXF/593-bug-stats-finder-displaying-wrong-number-of-published-stats-docs
https://trello.com/c/QpM8Vhwv/590-bug-stats-finder-is-only-displaying-a-few-upcoming-stats-docs
github: alphagov/finder-frontend#1053

### DIFF
--- a/config/finders/statistics_finder.yml
+++ b/config/finders/statistics_finder.yml
@@ -68,21 +68,13 @@ details:
     display_as_result_metadata: false
     filterable: true
     preposition: about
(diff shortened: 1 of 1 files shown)


bilbof (approved):
Looks good.

Sounds like we need to merge in https://github.com/alphagov/govuk-content-schemas/pull/886 before the tests will pass.

PR #1526, opened 2019-04-25
Time round-trips to elasticsearch

barrucadu:
We saw a series of spikes in the 99th percentile of search-api request durations.  We want to know if it was elasticsearch.


cbaines (approved):
Looks good to me :+1:

sihugh (approved):

PR #1531, opened 2019-05-01
Filters business finder by new facet_groups

oscarwyatt:
Using the old filter meant content newly tagged to
the business finder wasn't showing up on the finder
as the new tagging system uses facet_groups. This
fixes that.


brucebolt (approved):

steventux (approved):
If it produces the right query params then :+1: this was evidently omitted from the finder-frontend work, no feature covering it either.

PR #1534, opened 2019-05-02
Update docs to refer to search-api

barrucadu:
I don't think it's worth changing all the other references to "rummager" (file names, function names, etc) as there's a *lot* and it will make `git blame` all but useless.  Things can be changed as they're worked on as part of other tasks.

---

[Trello card](https://trello.com/c/QFqYu8XR/656-%F0%9F%92%80-update-rummager-code-and-documentation-to-refer-only-to-search-api-%F0%9F%92%80)


bilbof (approved):
👍

PR #1544, opened 2019-05-13
Bump elasticsearch gem to version 6

barrucadu:
We're still using elasticsearch 5, but by using version 6 of the gem
we can start to make changes to make search-api compatible with both.

---

This has now been on staging for about a day.  I've done some manual testing:

- Republishing all detailed guides from whitehall
- Republishing all transactions from publishing-api
- Publishing new guidance from whitehall & checking it appears in search
- Withdrawing guidance from whitehall & checking it disappears from search.
- Searching for a thing with a best bet

---

[Trello card](https://trello.com/c/uTsa3fpt/163-make-search-api-talk-to-both-es5-and-es6-%F0%9F%8D%90)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source "https://rubygems.org"
 
 gem "activesupport", "~> 5.2.3"
-gem "elasticsearch", "~> 5"
+gem "elasticsearch", "~> 6"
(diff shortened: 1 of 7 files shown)


sihugh:
Good plan 👍

bilbof (approved):

PR #1549, opened 2019-05-21
Enable filtering on boolean fields

bilbof:
This enables filtering on fields such as has_official_document, which are boolean fields.

This is possible with advanced_search, but is now possible with batch_search and search.

This is on integration (it may not be by the time this is reviewed): see `/api/search.json?filter_has_official_document=true`.

The fields that are now available are already documented:
https://docs.publishing.service.gov.uk/apis/search/fields.html

https://trello.com/c/0SfRvUJX/752-enable-boolean-filters-in-search-api

### DIFF
--- a/config/schema/field_types.json
+++ b/config/schema/field_types.json
@@ -166,6 +166,7 @@
 
   "boolean": {
     "description": "A boolean value which can be returned and used for filtering and aggregating",
+    "filter_type": "boolean",
     "es_config": {
       "type": "boolean",
       "include_in_all": false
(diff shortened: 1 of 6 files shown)


barrucadu (approved):
LGTM, though it's initially a bit confusing that the number of documents where the field is `true` + the number of documents where the field is `false` doesn't correspond to the number of documents in ES, or to the number of documents in the `government` index.

The field isn't present in the `govuk` index, and I guess it's probably `null` for a lot of the docs in the `government` index.

bilbof:
E2E tests need looking into. They're breaking on all branches.

bilbof:
@barrucadu I have a feeling that `has_official_document` is unset in some places. Some documents don't have this field, so it's neither false nor true, and that's why those docs don't appear in the results, which I think is the desired behaviour.

PR #1553, opened 2019-05-24
Switch from string to text / keyword

barrucadu:
Following the suggested migration here:
https://www.elastic.co/blog/strings-are-dead-long-live-strings

Analyzed string fields:

    { "type" "string"
    , "index": "analyzed"
    }

Become text fields:

    { "type" "text"
    , "index": true
    }

Non-analyzed string fields:

    { "type" "string"
    , "index": "not_analyzed"
    }

Become keyword fields:

    { "type" "keyword"
    , "index": true
    }

---

[Trello card](https://trello.com/c/lSa4fDiJ/718-make-elasticsearch-index-schema-definitions-es6-compatible-%F0%9F%8D%90)

### DIFF
--- a/config/schema/field_types.json
+++ b/config/schema/field_types.json
@@ -3,8 +3,8 @@
     "description": "An identifier, can be used for exact (case and whitespace sensitive) lookup and aggregating",
     "filter_type": "text",
     "es_config": {
-      "type": "string",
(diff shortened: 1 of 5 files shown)


koetsier (approved):
👍 💯

PR #1557, opened 2019-05-29
Remove use of include_in_all

barrucadu:
~This branches off #1553, and should be rebased after that is merged.~

---

We're not using the `_all` field at all, in fact the `all_searchable_text` and `spelling_text` fields are described as alternatives to `_all` that we do use.  We don't generate any queries for `_all` (outside of in a few of the tests), so remove the configuration.

---

[Trello card](https://trello.com/c/lSa4fDiJ/718-make-elasticsearch-index-schema-definitions-es6-compatible-%F0%9F%8D%90)

### DIFF
--- a/config/schema/field_types.json
+++ b/config/schema/field_types.json
@@ -4,8 +4,7 @@
     "filter_type": "text",
     "es_config": {
       "type": "keyword",
-      "index": true,
-      "include_in_all": false
+      "index": true
     }
   },
 
@@ -15,8 +14,7 @@
     "multivalued": true,
     "es_config": {
       "type": "keyword",
-      "index": true,
-      "include_in_all": false
+      "index": true
     }
   },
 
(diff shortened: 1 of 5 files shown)


koetsier:
according to https://www.elastic.co/guide/en/elasticsearch/reference/5.4/include-in-all.html this field defaults to true - are you going to disable this field, i.e. https://www.elastic.co/guide/en/elasticsearch/reference/5.4/mapping-all-field.html#disabling-all-field ?

barrucadu:
I was going to leave it unspecified, because https://www.elastic.co/guide/en/elasticsearch/reference/6.8/mapping-all-field.html says "`_all` may no longer be enabled for indices created in 6.0+".

Though then it gives some examples of enabling the field, so I'm not sure what is true.

koetsier (approved):
lgtm 🥇

barrucadu:
Rebased against master now that #1553 is merged

PR #1562, opened 2019-05-30
Finders are published with the correct phase tag

hannako:
At the moment, the phase tag is being reset to `live` when the finder config files are published by the rake task. This PR corrects that. 

Trello https://trello.com/c/GU9oNJot/722-add-beta-banner-to-search-research-and-statistics-s


bilbof (commented):
It might be better to do this `ContentItemPresenter`, as we may see the same problem with other things we want to publish:

```
    def present
      {
        ...
        phase: content_item["phase"] || "live",
```

hannako:
@bilbof 

Thanks Bill, will do. 👍

barrucadu (approved):

PR #1568, opened 2019-06-04
Replace `indices` query with `should` query

barrucadu:
The format migrator prevents users from getting duplicate results when
formats are migrated into the "govuk" index from the "government" or
"detailed" ones: it ensures that only non-migrated formats are
returned from the "government"/"detailed" indices and only migrated
formats are returned from the "govuk" index.

Previously it did this with an `indices` query.  They look like
this:

    { indices:
      { indices: ["list", "of", "indices"]
      , query: { query to perform against docs from the indices }
      , no_match_query: { query to perform against other docs }
      }
    }

However, the `indices` query has been deprecated in ES5 and removed in
ES6, in favour of just searching over the `_index` field in docs
directly.

This PR changes the `indices` query into a `should` with two clauses
where one clause matches documents in the "govuk" index and the other
clause matches documents in the other indices.  It's a bit more
cumbersome because the `terms` query I'm using [doesn't handle index
aliases](https://github.com/elastic/elasticsearch/issues/23306), and involves another round-trip to elasticsearch when
constructing the query to find out the real name of the "govuk" alias.

Index real names only change very infrequently (whenever we reindex
the cluster), so in principle they could be cached.  But we don't
currently have a good way for the reindexing rake task to signal to
the web processes that they need to purge their cache.

---

[Trello card](https://trello.com/c/iHfzqX4K/715-rewrite-deprecated-elasticsearch-indices-queries-%F0%9F%8D%90)

### DIFF
--- a/lib/legacy_client/index_for_search.rb
+++ b/lib/legacy_client/index_for_search.rb
@@ -13,6 +13,15 @@ module LegacyClient
       @search_config = search_config
     end
 
+    def real_index_names
+      index_names.map do |index_name|
+        # this may throw an exception if the index name isn't found,
+        # but we want to propagate the error in that case as it
+        # shouldn't happen.
+        @client.indices.get_alias(index: index_name).keys.first
+      end
+    end
+
     def raw_search(payload)
       logger.debug "Request payload: #{payload.to_json}"
       @client.search(index: @index_names, type: 'generic-document', body: payload)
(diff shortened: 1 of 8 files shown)


sihugh (approved):
We'll need to be a bit careful about this because of the extra calls.  Hopefully it's quick enough to not make much difference, but it might cause a little bit more pressure on the downstream apps, so worth monitoring for a bit when we deploy.

sihugh:
Really good though, thanks @barrucadu ! 👏

barrucadu:
It's mostly us who change search-api, so we can probably leave it on staging for a day and see how it fares.

PR #1569, opened 2019-06-04
Allow Search API to talk to multiple elasticsearch clusters

bilbof:
This enables search-api to send requests to multiple
elasticsearch clusters. You can specify the clusters
you would like to query using the query param cluster.

E.g. /search.json?cluster=B will query elasticsearch6.
The default behaviour is to query cluster A, which is
currently elasticsearch5.

This is needed now because we want to upgrade from
elasticsearch version 5 to 6, by querying 5 or 6 ad hoc.

This will have the side effect that we will be able to
do a/b tests and compare indexes/clusters more easily.

All writes to one cluster should be performed across all clusters

It also returns `es_cluster` with search responses.

Trello: https://trello.com/c/uTsa3fpt/745.


elasticsearch.yml
```diff
@@ -1,5 +1,6 @@
 production: &default
   base_uri: <%= ENV["ELASTICSEARCH_URI"] || 'http://localhost:9200' %>
+  base_uri_search_b: <%= ENV["ELASTICSEARCH_B_URI"] %>
```
barrucadu: Could these two variables have more similar names?  Why not just `base_uri_b`?

bilbof: Yeah good point. Changed.

elasticsearch.yml
```diff
@@ -16,6 +17,7 @@ development:
 
 test:
   base_uri: <%= ENV.fetch('ELASTICSEARCH_URI', 'http://localhost:9200') %>
+  base_uri_b: <%= ENV["ELASTICSEARCH_B_URI"] || 'http://localhost:9200' %>
```
barrucadu: re the `|| '...'`, does it make sense for the A and B clusters to actually refer to the same thing?

I feel like it'll be confusing as soon as indexing is added, and we start to add tests like "after indexing a document it can be retrieved from both clusters".

bilbof: `base_uri` is also the fallback if `base_uri_b` is unset (as is the case outside of the test environment). Rather than returning an error when say, `cluster=B` is part of the request but this isn't defined in the environment, we fall back to the default cluster.

bilbof: What do you think we should do here? This is mainly there so that any tests that want to speak to a non-default cluster can do so. When we add indexing and so on for other clusters, perhaps we should handle this better there?

barrucadu: It's probably fine for now, but we'll need to rethink this somehow when indexing is added.  The easiest way to do that might be to run both ES5 and ES6 in CI (and the dev VM) on different ports.

lib/search_config.rb
```diff
@@ -131,6 +157,14 @@ def combined_index_schema
     )
   end
 
+  def meta_assign(method, cluster, &block)
```
barrucadu, line 158: This is pretty clever, but I'm not sure of a more obvious way to do it.

bilbof: Yeah I asked around to see if anyone has ideas on a nicer way to do this (see our internal ruby channel 😄). If you have an idea I'd be interested!

lib/tasks/indices.rake
```diff
@@ -156,46 +180,54 @@ You should run this task if the index schema has changed.
...
+      missing = []
+      cluster_name = "cluster_#{cluster.key}"
+
+      SearchConfig.instance.all_index_names.each do |index|
+        begin
+          stats = client.indices.stats index: index, docs: true
+          docs = stats["_all"]["primaries"]["docs"]
+
+          count = docs["count"]
+          statsd.gauge("#{cluster_name}.#{index}_index.docs.count", count)
```
barrucadu, line 217: Joining up metrics will be a bit awkward, especially when the B cluster becomes the A cluster, but I'm not sure there's a good way to solve that here.

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -39,9 +39,11 @@ def perform(messages)
         process_action(processor, routing_key, payload)
       end
 
-      response = processor.commit
-      process_response(response, messages) if response
+      responses = processor.commit
```
barrucadu, line 42: Why does this now return multiple responses?  It doesn't look like the `ElasticsearchProcessor` has changed

lib/govuk_index/publishing_event_worker.rb
```diff
@@ -39,9 +39,11 @@ def perform(messages)
...
 
+      (responses || []).each do |response|
+        process_response(response, messages)
```
barrucadu, line 45: Does this make the `if response['items'].count > 1` check in `process_response` redundant?

spec/integration/indexer/indexing_spec.rb
```diff
@@ -180,9 +188,13 @@
...
       ] }
 
-      expect(stubbed_client).to receive(:bulk).and_return(locked_response)
-      expect(stubbed_client).to receive(:bulk).and_call_original
-      allow_any_instance_of(SearchIndices::Index).to receive(:build_client).and_return(stubbed_client)
+      index = SearchConfig.instance.search_server.index_group(index_name).current
+      index.lock
+
+      Thread.new do
+        sleep 0.1
```
barrucadu: Why doesn't the original way of stubbing the response work anymore?

lib/tasks/indices.rake
bilbof: One option would be to have version-like numbers? Rather than clusters having 'A' or 'B' names that signify priority. I figure we could join cluster metrics up with some clever use of `*` in our graphite queries. What do you think?

barrucadu: That makes sense, all the infrastructure is already called `elasticsearch5` or `elasticsearch6`.  Though if we ever want to do an A/B test with the same version of elasticsearch (but some other changes to the cluster) then it wouldn't quite fit.

spec/integration/indexer/indexing_spec.rb
bilbof: `any_instance_of` works great when we expect only one instance of a class to receive the messages. However, when we want multiple classes to be called (all instances of this class should be called once, and in total twice (as an example, because I have two clusters)), there's a gotcha.

From [rspec docs](https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/any-instance):

> This feature is sometimes useful when working with legacy code, though in general we
discourage its use for a number of reasons:
> The rspec-mocks API is designed for individual object instances, but this feature operates on entire classes of objects. As a result there are some semantically confusing edge cases. For example, in expect_any_instance_of(Widget).to receive(:name).twice it isn't clear whether a specific instance is expected to receive name twice, or if two receives total are expected. (It's the former.)

This means that `allow_any_instance_of(SearchIndices::Index).to receive(:build_client)` wouldn't work with `.twice` (it'd expect each instance to get called twice).

As this is an integration test (similar to the worker specs), and `allow_any_instance_of` is discouraged, I rewrote the test to more closely mirror the index being locked.

lib/tasks/indices.rake
bilbof: Perhaps we could also version the versions?

It's versions all the way down: `elasticsearch5-v1`

spec/integration/indexer/indexing_spec.rb
barrucadu: I'm not sure this is reliably testing the desired behaviour anymore, because there's no guarantee that the update will be rejected by the lock (and so be rescheduled) - the lock could expire before the `post` happens, in which case all we're testing is that we can insert a document by `post`ing.

lib/tasks/indices.rake
barrucadu: Let's just keep it simple and use the cluster name for now.

lib/govuk_index/publishing_event_worker.rb
bilbof: The client (ElasticsearchProcessor.new.govuk) is a subclass of `lib/index/client.rb` which now performs write actions against all active clusters:

```
    def bulk(params)
      Clusters.active.map do |cluster|
        client(cluster: cluster).bulk(
          params.merge(index: index_name)
        )
      end
    end
```

When `commit` is called on the processor (using the client to make the requests), it calls the modified bulk method:

```
@client.bulk(
        body: @actions
      )
```

Sorry, it's a bit convoluted. Let me know if I have misunderstood how this works.

barrucadu: Ok, that makes sense.

lib/govuk_index/publishing_event_worker.rb
barrucadu: No

spec/integration/indexer/indexing_spec.rb
bilbof: You're right, I could set the delay to way longer, but that still doesn't guarantee that the lock causes a retry. We only check that a lock happened and that the request was eventually successful.

One option would be to test it with only only active cluster?

```
allow(Clusters).to receive(:active).and_return([Clusters.default_cluster])
```

Then I could put back all the `allow_any_instance_of` stuff (:cry:). What do you think?

lib/tasks/indices.rake
bilbof: Is cluster name (e.g.) `elasticsearch5`?

spec/integration/indexer/indexing_spec.rb
barrucadu: I think that's probably the best solution

lib/tasks/indices.rake
barrucadu: No, the `cluster_A`, `cluster_B`.  We can always come up with a better way later.

spec/integration/indexer/indexing_spec.rb
bilbof: Done

barrucadu (approved):
We should put this on integration and exercise it a bit

PR #1570, opened 2019-06-04
Update url for Official Documents Finder

hannako:
Following convention of other specialist finders, the new official docs finder url is `/official-documents`

Following this change I will run rake `publishing_api:publish_supergroup_finders`

Ref [Trello](https://trello.com/c/6xFJqfAw/724-spin-up-finder-for-content-that-has-official-doc-status-s)

### DIFF
--- a/config/finders/official_documents_finder.yml
+++ b/config/finders/official_documents_finder.yml
@@ -1,5 +1,5 @@
 ---
-base_path: "/search/official-documents"
+base_path: "/official-documents"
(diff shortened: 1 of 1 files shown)


sihugh (approved):
You can just do `rake publishing_api:publish_finder FINDER_CONFIG=official_documents_finder.yml` for this, as we don't need to republish everything else

hannako:
@sihugh thanks, that sounds like a better idea :) Need the official [doc finder PR](https://github.com/alphagov/finder-frontend/pull/1154) to be merged before this can be deployed.

PR #1581, opened 2019-06-12
Allow per-cluster search config and use new text similarity for es6

barrucadu:
```
$ ELASTICSEARCH_A_URI='http://localhost:9200' ELASTICSEARCH_B_URI='http://localhost:9201' bundle exec rake search:create_all_indices SEARCH_INDEX=all
...
$ curl localhost:9200/govuk/_settings 2>/dev/null | json_pp | grep -C5 similarity
               "total_fields" : {
                  "limit" : "2000"
               }
            },
            "number_of_replicas" : "1",
            "similarity" : {
               "default" : {
                  "type" : "classic"
               }
            },
            "number_of_shards" : "3"
$ curl localhost:9201/govuk/_settings 2>/dev/null | json_pp | grep -C5 similarity
$
```

---

[Trello card](https://trello.com/c/MX1MtsEf/717-switch-to-new-elasticsearch-text-similarity-metric-%F0%9F%8D%90-l)

### DIFF
--- a/config/schema/elasticsearch_schema_b.yml
+++ b/config/schema/elasticsearch_schema_b.yml
@@ -1,0 +1,115 @@
+index:
+  settings:
+    mapping:
+      total_fields:
+        limit: 2000
+    analysis:
+      analyzer:
+        # At index time, elasticsearch will use an analyzer named default in the index settings
+        # if no analyzer is specified in the field mapping:
+        # https://www.elastic.co/guide/en/elasticsearch/reference/current/analyzer.html
+        #
+        # We don't need this at query time unless synonyms are disabled
+        # using a debug flag.
+        default:
+          type: custom
+          tokenizer: standard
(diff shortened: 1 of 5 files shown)


bilbof (approved):

bilbof:
Is there a way to test the clusters have different schemas applied?

barrucadu:
I'm not sure, I can't see an obvious way.

bilbof:
I suppose in an integration test, you could instantiate all clusters, then use the ES client to inspect the settings and check they're what they should be?

barrucadu:
That's a bit tricky because we don't set the *entire* config, so the result of `curl localhost:9200/govuk/_settings` isn't the same as what we have in our yaml file.

We could check the yaml file is a subset of the ES-reported index config, but that wouldn't help in this case where we want to check that removing a key from the config file worked.

PR #1591, opened 2019-06-19
Remove combine_mode from business finder email signup

brucebolt:
The `combine_mode` value was added to enable the use of the OrJoinedFacetSubscriberList model when signing up to business finder email alerts.  This model has been removed, so the toggle parameter is no longer needed, since we can identify this type of finder using the `email_filter_by` value.

Trello card: https://trello.com/c/udPtHvz2/158-remove-orjoinedfacetsubscriberlist-from-email-alert-api


ChrisBAshton (approved):
Approved

leenagupte (approved):

PR #1593, opened 2019-06-19
Load page traffic into all active clusters

bilbof:
This loads page traffic data into all indexes, not just the primary one.

Just wraps the calls in a `Clusters.active.each do |cluster|`. It'll probably take twice as long, but the strain on the indexes should be the same.

There's possibly better ways to do this (e.g. copying from one index to another?). But this is consistent with the way data enters into all clusters elsewhere in search-api.

One thing that might be nice is to create a worker for each cluster.

https://trello.com/c/15tBcoJJ/809-get-es6-page-traffic-sync-working

### DIFF
--- a/lib/govuk_index/page_traffic_loader.rb
+++ b/lib/govuk_index/page_traffic_loader.rb
@@ -7,28 +7,30 @@ module GovukIndex
     end
 
     def load_from(iostream)
-      new_index = index_group.create_index
-      @logger.info "Created index #{new_index.real_name}"
+      Clusters.active.each do |cluster|
+        new_index = index_group(cluster).create_index
+        @logger.info "Created index #{new_index.real_name}"
 
-      old_index = index_group.current_real
-      @logger.info "Old index #{old_index.real_name}"
+        old_index = index_group(cluster).current_real
+        @logger.info "Old index #{old_index.real_name}"
 
-      old_index.with_lock do
-        @logger.info "Indexing to #{new_index.real_name}"
+        old_index.with_lock do
+          @logger.info "Indexing to #{new_index.real_name}"
 
-        in_even_sized_batches(iostream) do |lines|
-          GovukIndex::PageTrafficWorker.perform_async(lines, new_index.real_name)
+          in_even_sized_batches(iostream) do |lines|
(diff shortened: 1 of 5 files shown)


lib/govuk_index/page_traffic_loader.rb
```diff
@@ -7,28 +7,30 @@ def initialize(iostream_batch_size: 250)
...
 
-      old_index.with_lock do
-        @logger.info "Indexing to #{new_index.real_name}"
+        old_index.with_lock do
+          @logger.info "Indexing to #{new_index.real_name}"
 
-        in_even_sized_batches(iostream) do |lines|
-          GovukIndex::PageTrafficWorker.perform_async(lines, new_index.real_name)
+          in_even_sized_batches(iostream) do |lines|
+            GovukIndex::PageTrafficWorker.perform_async(lines, new_index.real_name)
```
barrucadu: I think the `PageTrafficWorker` needs to know the cluster as well.  See https://github.com/alphagov/search-api/blob/master/lib/govuk_index/page_traffic_worker.rb#L15

barrucadu (changes requested):

lib/govuk_index/page_traffic_loader.rb
bilbof: Thanks, you're right. I've updated the commit.

spec/unit/govuk_index/page_traffic_loader_spec.rb
```diff
@@ -20,10 +20,11 @@
     line2 = [{ "val" => "c" }, { "data" => 1 }, { "val" => "d" }, { "data" => 1 }]
     line3 = [{ "val" => "e" }, { "data" => 1 }]
 
-    expect(GovukIndex::PageTrafficWorker).to receive(:perform_async).with(line1, 'new_index_name')
-    expect(GovukIndex::PageTrafficWorker).to receive(:perform_async).with(line2, 'new_index_name')
-    expect(GovukIndex::PageTrafficWorker).to receive(:perform_async).with(line3, 'new_index_name')
-
+    Clusters.active.each { |cluster|
```
barrucadu: I think `do`/`end` is a bit clearer for a multi-line thing like this.  Also, you have angered the linter

barrucadu (approved):

PR #1594, opened 2019-06-20
Remove and_facet_values

brucebolt:
Now that the EU Exit Business Readiness finder only has OR logic between all facets, the `and_facet_values` isn't needed any more (this was used to create the AND logic between the first two facets).

Trello card: https://trello.com/c/Xnxtwdnx


ChrisBAshton (approved):
I love deleting code! 🎉

PR #1596, opened 2019-06-20
Add 'topic' as a hidden parameter.

koetsier:
Topic refers to a taxon's content_id and filters the the finder accordingly.

Trello: https://trello.com/c/rjrtLu4N/821-allow-topic-parameter-in-supergroup-finder
See also: https://github.com/alphagov/finder-frontend/pull/1209

### DIFF
--- a/config/finders/guidance_and_regulation_finder.yml
+++ b/config/finders/guidance_and_regulation_finder.yml
@@ -38,6 +38,14 @@ details:
     display_as_result_metadata: false
     filterable: true
     preposition: about
+  - key: topic
+    filter_key: all_part_of_taxonomy_tree
+    name: topic
+    short_name: topic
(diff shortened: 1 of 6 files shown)


config/finders/guidance_and_regulation_finder.yml
```diff
@@ -38,6 +38,14 @@ details:
     display_as_result_metadata: false
     filterable: true
     preposition: about
+  - key: topic
+    filter_key: all_part_of_taxonomy_tree
+    name: topic
+    short_name: topic
```
barrucadu, line 44: Is it ok that this `name` and `short_name` are the same as for the facet above?  What are the `name` and `short_name` actually used for?

bilbof (approved):

PR #1598, opened 2019-06-24
Make FormatMigrator fetch migrated_indices from specified cluster

bilbof:
Currently FormatMigrator fetches migrated_indices index names from the default cluster. Now that we want results from specified cluster, FormatMigrator needs to know that it should get the index names for the correct cluster.

Trello: https://trello.com/c/9bjDVKo6/824

### DIFF
--- a/lib/search/format_migrator.rb
+++ b/lib/search/format_migrator.rb
@@ -1,7 +1,8 @@
 module Search
   class FormatMigrator
-    def initialize(base = nil)
-      @base = base
+    def initialize(base_query: nil, cluster: Clusters.default_cluster)
+      @base_query = base_query
+      @cluster = cluster
     end
 
     def call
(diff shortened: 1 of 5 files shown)


rubocop-205f737.html
```diff
@@ -0,0 +1,400 @@
+<!DOCTYPE html>
```
barrucadu: 🔪

bilbof: oOOOps!

lib/search/format_migrator.rb
```diff
@@ -1,7 +1,8 @@
 module Search
   class FormatMigrator
-    def initialize(base = nil)
-      @base = base
+    def initialize(base_query: nil, cluster: Clusters.default_cluster)
```
barrucadu, line 3: Given that all the `cluster` is used for is to get the migrated indices, would be be better to have a `migrated_indices` parameter?  It means `FormatMigrator` doesn't need to know about clusters.  `QueryBuilder` still needs to (for now) but that's at least a little further towards the boundaries of the system.

bilbof: I like the idea. It would mean duplicating `SearchConfig.instance.new_content_index(cluster: cluster).real_index_names` everywhere we use `FormatMigrator` though, and it might not be obvious why.

This way at least we know that FormatMigrator depends on a specific cluster through the index names.

barrucadu: Hmm, the duplication would be unfortunate.  Maybe leave it as it is for now then and we can think more later about how to handle this cluster proliferation.  I'm still leaning towards factory methods as the right thing to do.

barrucadu (approved):

lib/search/format_migrator.rb
bilbof: OK. If you have ideas on how to do the factory methods in a nice way I'd be keen to add them!

PR #1607, opened 2019-06-26
Boost step by steps

sihugh:
In the absence of any boosting factor, step by steps will have it set to 1 which is lower than document collections and topical events. 😿 

2.5 is equivalent to things like answers and guides (they're in the navigation_document_supertype of guidance)

This is also used in sitemaps - so may affect relative ranking of these pages in external search engines.

### DIFF
--- a/config/query/boosting.yml
+++ b/config/query/boosting.yml
@@ -2,6 +2,7 @@ base:
   format:
     service_manual_guide: 0.3
     service_manual_topic: 0.3
+    step_by_step_nav: 2.5
     transaction: 1.5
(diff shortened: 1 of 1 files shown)


koetsier (approved):
look good

barrucadu (approved):
We should definitely review the boosting at some point

leenagupte (approved):

bilbof (approved):

sihugh:
We're going to work out how to measure this before we merge it.  Outputs of the measuring of this change will inform our future strategy for format/other weighting

PR #1612, opened 2019-07-01
Add "/help/cookie-details" to block list

chao-xian:
This page needs to be hidden from search, as it's additional
information if the user has come form "/help/cookies/".
Visiting this page directly is not useful.

Note that you will still have to remove the page from the search index manually through the Search Admin. This only prevents it from being re-indexed.

### DIFF
--- a/config/govuk_index/migrated_formats.yaml
+++ b/config/govuk_index/migrated_formats.yaml
@@ -296,6 +296,9 @@ migrated:
 
 indexable: []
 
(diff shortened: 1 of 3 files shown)


chao-xian:
[Remove cookie details from internal search](https://trello.com/c/NEWUawQr/197-remove-cookie-details-from-internal-search)

vanitabarrett (approved):

PR #1618, opened 2019-07-04
Fix class variable caching in tests

oscarwyatt:
Our work on [govuk-docker](https://github.com/alphagov/govuk-docker/pull/100) found some problems where running tests on search-api with two clusters of elasticsearch weren't passing, this fixes that

### DIFF
--- a/lib/search_config.rb
+++ b/lib/search_config.rb
@@ -80,10 +80,10 @@ class SearchConfig
       # index schema uses) is unaffected by the 'elasticsearch_settings'
       # field (which is what can be overridden per-cluster).
       Cache.get(Cache::COMBINED_INDEX_SCHEMA) do
-      CombinedIndexSchema.new(
-          content_index_names + [govuk_index_name],
-          default_instance.schema_config
-        )
+        CombinedIndexSchema.new(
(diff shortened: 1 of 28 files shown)


lib/search_config.rb
```diff
@@ -71,7 +67,9 @@ def run_batch_search(searches)
     end
 
     def elasticsearch
-      @elasticsearch ||= ElasticsearchConfig.new.config
+      Cache.get(Cache::SEARCH_CONFIG) do
+        @elasticsearch = ElasticsearchConfig.new.config
```
barrucadu: Does that `@elasticsearch =` want removing?

barrucadu (commented):
Approach looks good to me, my only concern is that we should be sure we've not lost caching of anything which was previously cached.

barrucadu:
~A thing for me to do: get the cluster B URI environment variable set in CI so we can actually test this.~

Done in this PR because on reflection breaking builds of every search-api branch seems like a bad idea.

koetsier:
Closing this because I opened a new PR https://github.com/alphagov/search-api/pull/1619 which is the same but with a cleaned history

PR #1619, opened 2019-07-05
Remove memoization

koetsier:
Currently most caching of large configuration files is done by using
  memoization in several classes.

  This PR adds a very simple caching solution so that all memoizations
  can be done in this one class.

  This has as an advantage that caching can be controlled centrally
  and that before each test we only need to clear this one cache
  to ensure that all caching throughout the app has been cleared

### DIFF
--- a/lib/clusters/cluster.rb
+++ b/lib/clusters/cluster.rb
@@ -10,7 +10,7 @@ module Clusters
     end
 
     def uri
-      @uri ||= elasticsearch_config[uri_key]
+      @uri ||= ElasticsearchConfig.new.config[uri_key]
     end
 
     def inactive?
@@ -22,9 +22,5 @@ module Clusters
   private
 
     attr_reader :uri_key
-
-    def elasticsearch_config
-      @elasticsearch_config ||= ElasticsearchConfig.new.config
-    end
   end
 end
--- a/lib/services.rb
+++ b/lib/services.rb
@@ -2,7 +2,7 @@ require 'active_support/cache'
 
 module Services
   def self.publishing_api
-    @publishing_api ||= GdsApi::PublishingApiV2.new(
+    GdsApi::PublishingApiV2.new(
       Plek.find('publishing-api'),
       bearer_token: ENV['PUBLISHING_API_BEARER_TOKEN'] || 'example',
 
@@ -37,17 +37,13 @@ module Services
   end
 
   def self.statsd_client
-    @statsd_client ||= Statsd.new.tap { |sd| sd.namespace = "govuk.app.search-api" }
+    Cache.get(Cache::STATSD_CLIENT) do
+      Statsd.new.tap { |sd| sd.namespace = "govuk.app.search-api" }
+    end
   end
 
   def self.cache
-    @cache ||= if ENV['RACK_ENV'] == 'production'
-                 # Using a memory store as this is expected to store signon
-                 # tokens, this will be cleared each time the app starts
-                 ActiveSupport::Cache.lookup_store(:memory_store)
-               else
-                 ActiveSupport::Cache.lookup_store(:null_store)
-               end
+    ActiveSupport::Cache.lookup_store(:memory_store)
   end
 end
 
--- a/spec/integration/workers/indexer/amend_worker_spec.rb
+++ b/spec/integration/workers/indexer/amend_worker_spec.rb
@@ -14,18 +14,16 @@ RSpec.describe Indexer::AmendWorker do
   end
 
   it "amends documents" do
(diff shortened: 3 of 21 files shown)


lib/clusters/cluster.rb
```diff
@@ -24,7 +24,7 @@ def inactive?
     attr_reader :uri_key
 
     def elasticsearch_config
-      @elasticsearch_config ||= ElasticsearchConfig.new.config
+      ElasticsearchConfig.new.config
```
barrucadu: This involves parsing a YAML file, can we keep this cached?

lib/services.rb
```diff
@@ -37,17 +37,11 @@ def self.elasticsearch(cluster: nil, hosts: ENV['ELASTICSEARCH_HOSTS'] || 'http:
...
 
   def self.cache
-    @cache ||= if ENV['RACK_ENV'] == 'production'
-                 # Using a memory store as this is expected to store signon
-                 # tokens, this will be cleared each time the app starts
-                 ActiveSupport::Cache.lookup_store(:memory_store)
-               else
-                 ActiveSupport::Cache.lookup_store(:null_store)
-               end
+    ActiveSupport::Cache.lookup_store(:memory_store)
```
barrucadu, line 46: Why are we not using the `null_store` for non-prod environments any more?

lib/services.rb
```diff
@@ -37,17 +37,11 @@ def self.elasticsearch(cluster: nil, hosts: ENV['ELASTICSEARCH_HOSTS'] || 'http:
   end
 
   def self.statsd_client
-    @statsd_client ||= Statsd.new.tap { |sd| sd.namespace = "govuk.app.search-api" }
+    Statsd.new.tap { |sd| sd.namespace = "govuk.app.search-api" }
```
barrucadu: Is instantiating this class every time `self.statsd_client` is called going to cause any problems (performance or semantic)?

spec/integration/workers/indexer/amend_worker_spec.rb
```diff
@@ -8,34 +9,42 @@
...
-    described_class.new.perform(index_name, link, updates)
+      described_class.new.perform(index_name, link, updates)
 
-    doc_with_updates = document.merge(updates)
+      doc_with_updates = document.merge(updates)
 
-    assert_requested request, times: cluster_count
-    expect_document_is_in_rummager(doc_with_updates, id: link, index: index_name)
+      assert_requested request, times: cluster_count
+      expect_document_is_in_rummager(doc_with_updates, id: link, index: index_name)
```
barrucadu: If the fake testing only enqueues jobs and doesn't execute them, why is the updated document found in search here?

spec/support/index_helpers.rb
```diff
@@ -3,7 +3,7 @@ def self.setup_test_indexes
     puts "Setting up test indexes on clusters #{Clusters.active.map(&:key).join(',')}..."
     start_time = Time.now
 
-    clean_all
+
```
barrucadu: Extra blank line

barrucadu (commented):
Looks generally good, just a few small comments and questions.

spec/integration/workers/indexer/amend_worker_spec.rb
koetsier: because the class is called directly here. I suppose in this case the fake! might not be necessary because the index isn't locked and the infinite loop that the fake! is meant to avoid won't happen here - I'll remove this one.

spec/support/index_helpers.rb
koetsier: fixed

lib/clusters/cluster.rb
koetsier: it's private and only used in the ```uri``` method which is cached

lib/services.rb
koetsier: We now clear the cache before tests. I thought this was sightly easier

lib/clusters/cluster.rb
barrucadu: I wonder if it would be clearer to remove the method then and have

```ruby
    def uri
      @uri ||= ElasticsearchConfig.new.config[uri_key]
    end
```

lib/services.rb
koetsier: semantic I wouldn't think so. performance-wise I'm not sure really. I'll add it to the cache to be sure

lib/clusters/cluster.rb
koetsier: Probably. I'll change it

barrucadu (approved):
LGTM, thanks for figuring this out

PR #1622, opened 2019-07-11
Add documentation for relevancy ranking

bilbof:
Rendered: https://github.com/alphagov/search-api/blob/relevancy-documentation/doc/relevancy.md

---

This document should explain what happens when you perform
a search query without a specified order.

This is an updated version of the Wiki doc at
/wiki/spaces/GS/pages/183107586/Relevancy+investigation

Trello: https://trello.com/c/ItZwMLum/810


doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
+# Relevancy ranking
+
+This document explains how documents are ranked when you perform a search
+with results sorted by relevancy.
+
+## Stopwords
```
barrucadu: I think this section is up-to-date

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+
+A synonym is a word or phrase that means exactly or nearly the same as another
+word or phrase in the same language, for example shut is a synonym of close.
+
+Synonyms are defined in the [synonyms.yml file](https://github.com/alphagov/rummager/blob/master/config/schema/synonyms.yml).
+
+Each synonym group is represented as comma separated lists of synonyms, and
+optionally a "=>" symbol.
+
+### Grouping of synonyms
```
barrucadu: This subsection needs rewriting as the non-`=>` synonyms didn't work when we switched to ES5 (for unknown reasons).  There's a commit message somewhere with more detail.

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+Comma separated "foo, bar, baz" means to return any of those terms.
+
+### Filtering
+
+Additional configuration is defined in the elasticsearch_schema.yml and
+stems.yml files.  This configuration is merged with the JSON configuration,
+and then passed to elasticsearch directly.
+
+Within elasticsearch_schema.yml, as well as an index_synonym filter there is
+also a synonym_protwords filter.
```
barrucadu: Inaccurate: https://github.com/alphagov/search-api/commit/4f0013f43fff0e85a5e904d07fc53697fe7d0a3a

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+Popularity is derived from recent page views. The gov.uk search uses page
+popularity information extracted from Google. The data is gathered by the
+search-analytics application. That data is imported into Search API.
+
+### Popularity score
+
+`popularity_score` seems to be set in `PopularityLookup` class.
+
+```ruby
+popularity_score = 1.0 / (ranks[link] + @search_config.popularity_rank_offset)
```
bilbof: This is still correct: `popularity_score = 1.0 / (ranks[link] + SearchConfig.popularity_rank_offset)`

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+This blog post suggests that as well as using keywords, it can be useful to use
+'keep words' to only filter the phrases we want.
+
+The post suggests to first generate shingles, followed by synonyms, followed
+by keepwords.
+
+In other words: generate candidate keyphrases by shingling, expand them with
+synonyms, then cull out any non-synonyms with keepwords.
+
+### Old synonyms
```
barrucadu: I don't think the old synonyms are used.  The configuration is loaded: https://github.com/alphagov/search-api/blob/b6e0400c7e2397a927d702e24302b55bb58722d3/lib/schema/schema_config.rb#L49-L62

But nothing actually uses it.  There's nothing in the queries or analysers about the old synonyms.

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+The analyzer which use the old_synonyms filter do not also use the
+synonym_protwords which could mean less effective results.
+
+Not too sure why we have not migrated the old_synonyms.
+
+### Summary
+
+Things to further investigate:
+
+- Are the synonymn groupings effective. Does simple contraction make much difference?
```
barrucadu: This is still unknown (probably)

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+- Currently with_index_synonyms filter looks like this: filter: [standard,
+asciifolding, lowercase, index_synonym, synonym_protwords, stemmer_override,
+stemmer_english]. Maybe we could try applying shingles, then synonyms and then
+applying keepwords (like suggested in the blog post)
```
barrucadu: This is inaccurate, but the shingle / keepwords suggestion could still be something to try.

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+- Figure out what is going on with old_synonyms.
```
barrucadu: I don't think it's used

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+- Is popularity boosted for synonyms of the term?
```
barrucadu: This is still unknown (probably)

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+| query_with_old_synonyms | ✔                  | Main analyzer if synonyms are enabled                                   | ✔                | ✔            | standard                 | ✖    | ✔         |              | ✔                   | ✔                 | ✖                   | ✖                            | ✔                | ✔               | ✖             | ✖                 |
+| default                 | ✔                  | Main analyzer if synonyms are disabled                                  | ✔                | ✔            | standard                 | ✖    | ✔         |              | ✖                   | ✔                 | None                | None                         | ✔                | ✔               | ✖             | ✖                 |
+| shingled_query_analyzer | ✔                  | Useless as not used at index time                                       | ✖                | ✖            | standard                 | ✖    | ✔         |              | ✖                   | ✔                 | ✖                   | ✖                            | ✔                | ✔               | old_shingles  | ✖                 |
```
bilbof: `shingled_query_analyzer` can be removed, it no longer exists.

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+
+| Analyzer                | Used by main query | Details                                                                 | Normalize quotes | Strip Quotes | Tokenize                 | Trim | Lowercase | Asciifolding | Old school synonyms | Remove stop words | Synonyms            | Mark protwords (not stemmed) | Stemmer override | Porter2 stemmer | Shingles      | ID codes analysis |
+|-------------------------|--------------------|-------------------------------------------------------------------------|------------------|--------------|--------------------------|------|-----------|--------------|---------------------|-------------------|---------------------|------------------------------|------------------|-----------------|---------------|-------------------|
+| query_with_old_synonyms | ✔                  | Main analyzer if synonyms are enabled                                   | ✔                | ✔            | standard                 | ✖    | ✔         |              | ✔                   | ✔                 | ✖                   | ✖                            | ✔                | ✔               | ✖             | ✖                 |
```
bilbof: `query_with_old_synonyms ` no longer exists

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+slop, the words don’t necessarily need to be in the same sequence in order
+to match. With a high enough slop value, words can be arranged in any order.
+
+https://www.elastic.co/guide/en/elasticsearch/guide/current/slop.html
+
+Shingles are more lenient since we're only looking for pairs of words instead
+of requiring the whole phrase to match. **But our implementation is broken.**
+This is one of the things that was fixed in the new search weighting.
+
+See Richard's commit message: https://github.com/alphagov/search-api/commit/6ece793190932fee31be5437c4c56773dc358ff6
```
barrucadu: This references stuff which has been removed, needs further investigation

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+>  The existing weighting only applies this analysis at search time, so the
+>  pairs it looks for never exist in the index. This commit instead
+>  creates a new pseudo-subfield for searchable text fields which is
+>  indexed and searched with shingles applied. The weights from searches
+>  across this field are used to boost scores of documents.
+
+### There is no boosting of individual field matches
+
+We treat a match in title, description, acronym and indexable content the same,
+but these are completely different things.
```
barrucadu: The new text similarity algorithm should basically do this for us, probably worth A/B testing removing this field boosting.

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+is a quoted phrase, but not for normal searches.
+
+```
+PHRASE_MATCH_TITLE_BOOST = 5
+PHRASE_MATCH_ACRONYM_BOOST = 5
+PHRASE_MATCH_DESCRIPTION_BOOST = 2
+PHRASE_MATCH_INDEXABLE_CONTENT_BOOST = 1
+```
+
+### We have lots of code stuck behind feature flags
```
barrucadu: I don't think this is true any more

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+In the past this has caused problems when we know content is better but
+something else still gets high traffic.
+https://github.com/alphagov/search-api/commit/03c4bfa0a1a816a57d38b71ac5cb22c3a107c275
+
+If we had other signals of content quality we might be able to reduce our
+dependence on the page views.
+
+### Recommendations
+
+- Replace shingles implementation with one that works
```
barrucadu: I think this is done, but we need to double check

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+- A/B test with/without `match_phrase` component
```
barrucadu: If this was done, the result was to keep the `match_phrase` component.

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+- Boost title/description matches in the match_all_terms, and match_bigrams subqueries
```
barrucadu: There is no `match_bigrams`.  We should A/B test the boosting of `match_all_terms`, as new text similarity should do that already.

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+- Remove unused code behind feature flags by end of quarter
```
barrucadu: I think this has been done

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+The Porter2 stemmer for english text
+
+### Recommendations
+
+1. Share custom stemming list with James
+2. Show how queries are analyzed in search admin or search performance explorer
+so that anyone can identify stemming problems and update the override list
+
+
+## ID codes
```
barrucadu: This is gone now

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+
+Which basically finds "codes" which consist of some letters and some digits,
+possibly split into two by whitespace. Requires there to be at least one
+digit and a minimum length of 2 characters.
+
+This does mean that if someone was to search for 'P45 and P46' it is possible
+to end up with a token of 'p45and'.
+
+
+## Long queries
```
barrucadu: I don't think we've tried any of these.

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+potentially improve result ordering without significantly decreasing
+search performance.
+
+In addition to this phrase matching can be limited to sentences/paragraphs,
+which again could potentially improve results.
+
+Further details can be found here: https://www.elastic.co/guide/en/elasticsearch/guide/current/_improving_performance.html
+
+
+## Weight top of page more
```
barrucadu: I don't think we've tried this

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+boosting the score of documents that match the search text in the first
+paragraph - in a similar way to how we boost based on title and description.
+
+The easiest way to do this is a create a special field specifically for this
+purpose which we would then add to the query - in a similar way to how we add
+a boost for title.
+
+This would require reindexing to extra the data into a separate field.
+
+## Clean Slate
```
barrucadu: It would be nice to try

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+- A/B test with/without `match_phrase` component
+- Boost title/description matches in the match_all_terms, and match_bigrams subqueries
+- Remove unused code behind feature flags by end of quarter
+
+
+## Analyzer usage
+
+This table describes what each of the analyzers do:
+
+| Analyzer                | Used by main query | Details                                                                 | Normalize quotes | Strip Quotes | Tokenize                 | Trim | Lowercase | Asciifolding | Old school synonyms | Remove stop words | Synonyms            | Mark protwords (not stemmed) | Stemmer override | Porter2 stemmer | Shingles      | ID codes analysis |
```
bilbof: Should link to the code at config/schema/elasticsearch_schema.yml, as this won't get out of date.

Should also briefly explain what an analyzer is and link to https://www.elastic.co/guide/en/elasticsearch/reference/current/analyzer.html.

An index has various analyzers. At index time, elasticsearch will use an analyzer named default in the index settings if no analyzer is specified in the field mapping: https://www.elastic.co/guide/en/elasticsearch/reference/current/analyzer.html (will this still be true in version 6?).

We don't need this at query time unless synonyms are disabled using a debug flag.

doc/relevancy.md
```diff
@@ -0,0 +1,442 @@
...
+### 6) Apply the english Stemmer to each token
+
+https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-stemmer-tokenfilter.html
+The Porter2 stemmer for english text
+
+### Recommendations
+
+1. Share custom stemming list with James
+2. Show how queries are analyzed in search admin or search performance explorer
+so that anyone can identify stemming problems and update the override list
```
bilbof: Can remove these bits.

doc/relevancy.md
bilbof: Removed

doc/relevancy.md
bilbof: I removed the ID codes section

doc/relevancy.md
bilbof: Should the Long queries section be removed?

doc/relevancy.md
bilbof: I've added a detail on what an analyzer is, and a link to the docs.

doc/relevancy.md
barrucadu: We should probably note them down somewhere as things we could try, but they don't really belong in this doc

barrucadu: Same with weighting the top of the page more and a clean slate approach

doc/relevancy.md
```diff
@@ -0,0 +1,272 @@
...
+boost, which would let us do partial popularity updates. This is a bigger
+change than we want to make right now, as it would need careful measurement
+to ensure we don't break queries.
+
+https://github.com/alphagov/search-api/blob/master/doc/arch/adr-005-incremental-popularity-updates.md
+
+
+## Query DSL usage
+
+Here is a map of what goes into our elasticsearch query: https://www.draw.io/#G0B3LRehFrM0QuaVRxV0hiZzhTbG8
```
barrucadu: This diagram is inaccurate, we don't do shingling any more (and possibly other problems).  We should remake it to be higher res and also accurate.

doc/relevancy.md
```diff
@@ -0,0 +1,281 @@
...
+| spelling_analyzer       | ✖                  | Used in spelling_text field                                             | ✔                | ✔            | standard                 | ✖    | ✔         |              | ✖                   | ✖                 | ✖                   | ✖                            | ✖                | ✖               | shingle       | ✖                 |
+| string_for_sorting      | ✖                  | Used for storing a sortable subfield                                    | ✔                | ✔            | keyword                  | ✔    | ✔         | ✖            | ✖                   | ✖                 | ✖                   | ✖                            | ✖                | ✖               | ✖             | ✖                 |
+
+
+The main analyzer used at the moment is called query_with_old_synonyms.
```
bilbof: This sentence can be removed as it's no longer correct.

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
+# Relevancy
+
+This document explains how documents are ranked when you perform a search
```
barrucadu: I'm not sure of a better way to word it but "this document explains how documents are ranked" feels a bit off, because "document" is used in two different senses in the same clause.

Maybe something like

> This document explains how relevancy ordering works when performing a search

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+
+## Impacts on relevancy
+
+Out of the box, elasticsearch comes with a decent scoring algorithm. They have an excellent [guide on scoring relevancy](https://www.elastic.co/guide/en/elasticsearch/guide/master/scoring-theory.html) which is worth reading.
+
+We've done some work in Search API to [tune relevancy](https://www.elastic.co/guide/en/elasticsearch/guide/master/relevance-conclusion.html), which we go into below.
+
+'Tuning' essentially means overriding elasticsearch defaults for our purposes.
+
+### Relevancy impact contents
```
barrucadu: I'd merge this with the list above, so all the contents are in one place

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
+# Relevancy
+
+This document explains how documents are ranked when you perform a search
+with results sorted by relevancy.
+
+## Contents
+
+1. What is relevancy?
```
barrucadu: I think it would be good for these to link to the sections

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+```ruby
+# Response for a search for 'Harry Potter'
+[
+  { title: "Harry Potter", es_score: 1 },
+  { title: "Harry Kane", es_score: 0.5 },
+  { title: "Ron Weasley", es_score: 0.05 }
+]
+```
+
+The `es_score` value represents the rank of the result or, more generally, how relevant we think a result is to your query.
```
barrucadu: It's the score, not the rank.  Ranks would be 1 for the first result, 2 for the second, etc.

> The `es_score` value represents the rank of the result or, more generally, how relevant we think a result is to your query.

is repetitive, so maybe

> The `es_score` value is used for ranking results and represents how relevant we think a result is to your query.

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+### Debugging es_score
+
+If you want to understand why a result has a given `es_score`, you can use the elasticsearch [Explain API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-explain.html). This is exposed by the Search API.
+
+You can see the reasons behind an `es_score` by including the `debug=explain` query parameter in your query.
+This will add an `_explanation` field to every result, similar to a SQL-like [EXPLAIN](https://wiki.postgresql.org/images/4/45/Explaining_EXPLAIN.pdf) but a little easier to understand.
+
+Example: https://www.gov.uk/api/search.json?debug=explain&q=harry%20potter
+
+The query params `debug=show_query&count=0` are also sometimes handy.
```
barrucadu: I'd separate those and say something like

> You can see the query Search API sends to Elasticsearch with the `debug=show_query` parameter.  Debug parameters can be combined, like `debug=show_query,explain`.  The debug output is verbose, so sometimes restricting to only a handful of results, with `count=0` or `count=1`, is useful.

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+
+## Impacts on relevancy
+
+Out of the box, elasticsearch comes with a decent scoring algorithm. They have an excellent [guide on scoring relevancy](https://www.elastic.co/guide/en/elasticsearch/guide/master/scoring-theory.html) which is worth reading.
+
+We've done some work in Search API to [tune relevancy](https://www.elastic.co/guide/en/elasticsearch/guide/master/relevance-conclusion.html), which we go into below.
+
+'Tuning' essentially means overriding elasticsearch defaults for our purposes.
```
barrucadu: This doesn't need to be a separate paragraph.  Something like

> We've done some work in Search API to [tune relevancy](https://www.elastic.co/guide/en/elasticsearch/guide/master/relevance-conclusion.html), overriding the default Elasticsearch behaviour, which we go into below.

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+You can see the reasons behind an `es_score` by including the `debug=explain` query parameter in your query.
+This will add an `_explanation` field to every result, similar to a SQL-like [EXPLAIN](https://wiki.postgresql.org/images/4/45/Explaining_EXPLAIN.pdf) but a little easier to understand.
+
+Example: https://www.gov.uk/api/search.json?debug=explain&q=harry%20potter
+
+The query params `debug=show_query&count=0` are also sometimes handy.
+
+## Impacts on relevancy
+
+Out of the box, elasticsearch comes with a decent scoring algorithm. They have an excellent [guide on scoring relevancy](https://www.elastic.co/guide/en/elasticsearch/guide/master/scoring-theory.html) which is worth reading.
```
barrucadu: Elasticsearch is a proper noun, so it should have a leading capital everywhere

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+  content, and can make judgements on what sorts of content are more
+  likely to be useful to the average user.
+
+#### Popularity
+
+This is updated nightly from recent Google Analytics data, gathered by
+the [search-analytics][] application.
+
+Here is the initial reasoning for the numbers (May 2014)
+([6cbd84f][]):
```
barrucadu: Avoid adjacent parentheses and re-order:

> ([6cbd84f][], May 2014)

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+
+> This field is populated from the page-traffic index, which in turn
+> is populated from a dump file produced by the python scripts in the
+> search-analytics repository.
+
```
barrucadu: Add missing `>`

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+If we want to boost the popularity of a query then we do so by 0.001
+(`POPULARITY_OFFSET = 0.001` in [popularity.rb][]). It is not clear
+where this number came from.
+
+We also boost using `boost_mode: :multiply` (multiply the `_score`
+with the function result (default)).
+
+The rank of a page relative to other pages on GOV.UK is stored in its
+`rank_14` field.  At one point, its number of page views was stored in
+the `vc_14` field, which could have been used to allow us to perform
```
barrucadu: We actually still do have `vc_14` (though it's called `vf_14`) - these fields are in the `page-traffic` index, not in the search indices.

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+partial popularity updates, but that field doesn't seem to exist any
+more.
+
+#### Recency
+
+This is an implementation of [this curve][], and is applied to
+documents of the "announcement" type.  It serves to increase the score
```
barrucadu: Link to the code: "and is applied in filename.rb"

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+of new documents and decrease the score of old documents.
+
+The curve came from the solr documentation, but the link to the
+original source is broken, so it is unclear why this particular curve
+was chosen.
+
+#### Properties
+
+This is defined by the [boosting.yml][] file.
```
barrucadu: Link to the code: "and is applied in filename.rb"

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+Algeria: /foreign-travel-advice/algeria
+Citizenship: /becoming-a-british-citizen
+```
+
+There's also a concept called **worst bets**. These are useful if our
+search results are so bad that we have to manually downrank a result
+for a given query. E.g. 'vehicle' shouldn't result in 'unicycle' being
+the top result.
+
+See the code in `QueryComponents::BestBets` for more details.
```
barrucadu: Link to the file

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Ignoring stopwords
```
barrucadu: This should be "Stopwords" for consistency with the other headings

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+They're words we filter out when processing a search query, since they're so common they're not useful.
+They also enable elasticsearch to maintain a smaller index.
+
+The stopwords themselves are provided by elasticsearch.
+There's nothing in the Search API repo that looks at a list of omitted terms. This is handled by the default [stop token filter](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-stop-tokenfilter.html).
+
+**History of stopwords**
+
+There has been some talk of not including stopwords in the past
+([0fe6e52][]) during this time a no_stop method was implemented that
```
barrucadu: Add a date, like the commit link above

barrucadu: "no_stop" should be `no_stop`

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+seems to stop stopwords from being used. Again this seems to be served
+by ES because it's not defined in search-api.
```
barrucadu: Use "Search API" for consistency

barrucadu: Use "Elasticsearch" for consistency

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+
+
+### Synonyms
+
+We use synonyms to show relevant results related to 'vehicle tax' regardless of whether you searched for 'car tax' or 'auto tax'.
+
+A synonym is a word or phrase that means exactly or nearly the same as another
+word or phrase in the same language, for example shut is a synonym of close.
+
+Synonyms are defined in the [synonyms.yml][] file.
```
barrucadu: Link to the code where the synonyms are applied to the index config

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+terms are mapped to the same single value that exists in the index.
+
+### Filtering
+
+Additional configuration is defined in the [elasticsearch_schema.yml][] and
+[stems.yml][] files.  This configuration is merged with the JSON
+configuration, and then passed to elasticsearch directly.
+
+[This blog post](https://opensourceconnections.com/blog/2016/12/02/solr-elasticsearch-synonyms-better-patterns-keyphrases/) suggests that as well as using keywords, it can be useful to use
+'keep words' to only filter the phrases we want.
```
barrucadu: Change `'keep words'` to `"keepwords"`

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+   Removes words in the english stopword list
+
+5. Apply our custom stemmer override to each token
+
+   Customises stemming using this list
+   https://github.com/alphagov/search-api/blob/master/config/schema/stems.yml
+
+6. Apply the english Stemmer to each token
+
+   The Porter2 stemmer for english text
```
barrucadu: Add a small bit of description to this, just a clause.

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+the query and document match?_
+
+Although all words need to be present in phrase matching, even when using
+slop, the words don’t necessarily need to be in the same sequence in order
+to match. With a high enough slop value, words can be arranged in any order.
+
+[Shingles][shingles] are another way to take phrases into account,
+which allow matching n-grams.  This was implemented in the past, but
+was broken ([6ece793][]) and the fixed implementation never went live
+([806a68c][]).
```
barrucadu: Add dates to commit links

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+
+#### There is no boosting of individual field matches
+
+We treat a match in title, description, acronym and indexable content
+the same, but these are different things.  A user might expect title
+and description matches to be much more important than body content.
+
+We're using field-specific boosts when the query is a quoted phrase,
```
barrucadu: Link to the code

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+but not for normal searches.
+
+```
```
barrucadu: Syntax highlighting

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+PHRASE_MATCH_ACRONYM_BOOST = 5
+PHRASE_MATCH_DESCRIPTION_BOOST = 2
+PHRASE_MATCH_INDEXABLE_CONTENT_BOOST = 1
+```
+
+#### Popularity is the main proxy for "good"
+
+Popularity is one of the biggest factors influencing the weighting. It's
+based on the overall page views, so there's a positive feedback loop.
+`Thing gets popular -> thing becomes more visible -> thing gets more popular`
```
barrucadu: This isn't code

doc/relevancy.md
```diff
@@ -0,0 +1,383 @@
...
+
+In the past this has caused problems when we know content is better
+but something else still gets high traffic ([03c4bfa][]).
```
barrucadu: Add date to commit link

suganya-s:
might be good to note somewhere `harry potter` becoming `harri potter` in the explainer is an example of stemming "replace suffix y or Y by i if preceded by a non-vowel which is not the first letter of the word (so cry -> cri, by -> by, say -> say)" from here http://snowball.tartarus.org/algorithms/english/stemmer.html

doc/relevancy.md
```diff
@@ -0,0 +1,426 @@
...
+
+| Analyzer                | Used by main query | Details                                                                 | Normalize quotes | Strip Quotes | Tokenize                 | Trim | Lowercase | Asciifolding | Old school synonyms | Remove stop words | Synonyms            | Mark protwords (not stemmed) | Stemmer override | Porter2 stemmer | Shingles      | ID codes analysis |
+|-------------------------|--------------------|-------------------------------------------------------------------------|------------------|--------------|--------------------------|------|-----------|--------------|---------------------|-------------------|---------------------|------------------------------|------------------|-----------------|---------------|-------------------|
+| default                 | ✔                  | Main analyzer if synonyms are disabled                                  | ✔                | ✔            | standard                 | ✖    | ✔         |              | ✖                   | ✔                 | None                | None                         | ✔                | ✔               | ✖             | ✖                 |
```
barrucadu: I think all these analyzer names should be in code font

doc/relevancy.md
```diff
@@ -0,0 +1,426 @@
...
+| spelling_analyzer       | ✖                  | Used in spelling_text field                                             | ✔                | ✔            | standard                 | ✖    | ✔         |              | ✖                   | ✖                 | ✖                   | ✖                            | ✖                | ✖               | shingle       | ✖                 |
+| string_for_sorting      | ✖                  | Used for storing a sortable subfield                                    | ✔                | ✔            | keyword                  | ✔    | ✔         | ✖            | ✖                   | ✖                 | ✖                   | ✖                            | ✖                | ✖               | ✖             | ✖                 |
+
+
+The main analyzer used at the moment is called with_search_synonyms.
```
barrucadu: with_search_synonyms should be `with_search_synonyms`

stevenjmesser:
This looks good! It's especially useful to know about the `explain` method. 

However, I think there's a few key elements we've perhaps missed. For example, are announcements the only document type given a recency score? What percentage of a document is analysed for indexing? Are there any document types we don't allow users to search against, e.g. HTML publications?

I don't have any more questions beyond those though, it looks like pretty complete documentation!

bilbof:
Thanks, @stevenjmesser, I've added some changes to address your comments.

bilbof:
> What percentage of a document is analysed for indexing?

@stevenjmesser I don't know the exact percentage of documents that are used unfortunately. We put various fields into the search indexes. All of the fields defined [here](https://github.com/alphagov/search-api/blob/master/config/schema/field_definitions.json) can be used when deciding what results to return, though not all are guaranteed to be used (e.g. date fields are ignored unless you're filtering / sorting by date).

By analysed for indexing do you mean what fields are looked at in a keyword search? If so, this would be fields like `title` with e.g. the [searchable_sortable_text](https://github.com/alphagov/search-api/blob/master/config/schema/field_definitions.json#L64). These all seem to get lumped together into one field type `all_searchable_text_type`, which is used for keyword searches.

doc/relevancy.md
```diff
@@ -0,0 +1,459 @@
...
+If you want to understand why a result has a given `es_score`, you can
+use the Elasticsearch [Explain API][explain].  This is exposed by the
+Search API.
+
+You can see the reasons behind an `es_score` by including the
+`debug=explain` query parameter in your query.  This will add an
+`_explanation` field to every result, similar to a SQL-like `EXPLAIN`.
+
+For example, see the explanation produced by [searching for "harry
+potter"][explain-example].  This shows an example of stemming, where
```
andysellick, line 49: Where is this explanation? I had a look through the URL linked here and couldn't see it, maybe worth clarifying?

doc/relevancy.md
```diff
@@ -0,0 +1,459 @@
...
+
+The stopwords themselves are provided by Elasticsearch.  There's
+nothing in the Search API repo that looks at a list of omitted terms.
+This is handled by the default [stop token filter][].
+
+**History of stopwords**
+
+There has been some talk of not including stopwords in the past
+([0fe6e52][], May 2015) during this time a `no_stop` method was
+implemented that seems to stop stopwords from being used.  Again this
```
andysellick: Maybe `seems to prevent stopwords from being used`?

doc/relevancy.md
```diff
@@ -0,0 +1,459 @@
...
+Additional configuration is defined in the [elasticsearch_schema.yml][] and
+[stems.yml][] files.  This configuration is merged with the JSON
+configuration, and then passed to Elasticsearch directly.
+
+[This blog post][synonyms-blog] suggests that as well as using
+keywords, it can be useful to use "keepwords" to only filter the
+phrases we want.
+
+The post suggests to first generate shingles, followed by synonyms,
+followed by keepwords.
```
andysellick: Would it help here to briefly explain some of these terms? Maybe like this, but with better explanations...

```
The post suggests to first generate shingles (a group every
possible 1 - N set of words as a single token), followed by
synonyms, followed by keepwords (a list of words to keep).
```

andysellick (commented):
That is some detailed documentation, good job. I'm not sure I fully understood all of it, so just a few comments on that.

doc/relevancy.md
bilbof: There is an `_explanation` field which displays the source of each result's `es_score`. It's generated by the [Explain API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-explain.html).

bilbof:
@andysellick thanks! I've added some commits to address your comments. Please let me know what you think.

doc/relevancy.md
andysellick: Ah, okay, I see it now.

doc/relevancy.md
```diff
@@ -0,0 +1,460 @@
...
+This is handled by the default [stop token filter][].
+
+**History of stopwords**
+
+There has been some talk of not including stopwords in the past
+([0fe6e52][], May 2015) during this time a `no_stop` method was
+implemented that seems to prevent stopwords from being used.  Again this
+seems to be served by Elasticsearch because it's not defined in Search
+API.
+
```
hannako, line 184: The paragraph below seems to belong elsewhere as it's written in the first person and is a discussion on what we might do in the future. Maybe changing the title from `History of Stopwords` to something else would help. Eg `Future improvements around stopwords` sort of thing. Might be better to remove the `I'm not sure ` and say `It's not immediately clear how we'd go about changing this, but it's likely that we'd need to look into what words could be useful to omit/not omit`

doc/relevancy.md
```diff
@@ -0,0 +1,460 @@
...
+- If you search for nothing then you get the some of the more
+  frequently clicked links: http://www.gov.uk/api/search.json?q=
+
+### Synonyms
+
+A synonym is a word or phrase that means exactly or nearly the same as
+another word or phrase in the same language, for example shut is a
+synonym of close.
+
```
hannako, line 205: You can lose the `For example` below:

doc/relevancy.md
```diff
@@ -0,0 +1,460 @@
...
+| `best_bet_stemmed_match` | ✖                  | Used for triggering best bets                                           | ✔                | ✔            | standard                 | ✖    | ✔         |              | ✖                   | ✖                 | ✖                   | ✖                            | ✔                | ✔               | ✖             | ✖                 |
+| `spelling_analyzer`      | ✖                  | Used in spelling_text field                                             | ✔                | ✔            | standard                 | ✖    | ✔         |              | ✖                   | ✖                 | ✖                   | ✖                            | ✖                | ✖               | shingle       | ✖                 |
+| `string_for_sorting`     | ✖                  | Used for storing a sortable subfield                                    | ✔                | ✔            | keyword                  | ✔    | ✔         | ✖            | ✖                   | ✖                 | ✖                   | ✖                            | ✖                | ✖               | ✖             | ✖                 |
+
+
```
hannako, line 307: Why are these other ones defined but not used? Maybe needs a line of explanation.

doc/relevancy.md
```diff
@@ -0,0 +1,460 @@
...
+```
+with_search_synonyms:
+  type: custom
+  tokenizer: standard
+  filter: [standard, asciifolding, lowercase, search_synonym, stop, stemmer_override, stemmer_english]
+  char_filter: [normalize_quotes, strip_quotes]
+```
+
+These are the steps, ignoring asciifolding, which have been added:
```
hannako, line 324: I found this a bit confusing. Maybe it would help if you reference the various elements of the custom analyser under each of the smaller headings. Eg Under`Split into tokens` mention that this is defined by the `filter: [stop]` .

doc/relevancy.md
```diff
@@ -0,0 +1,460 @@
...
+These no-indexed paths and formats are defined in [`config/govuk_index/migrated_formats.yaml`](https://github.com/alphagov/search-api/blob/master/config/govuk_index/migrated_formats.yaml).
+
+## Possible problems with queries and relevance
+
+### The way we handle longer search terms is broken
+
+We use [phrase queries][] to take phrases into account in search queries.
+
+We're not using the _[slop][]_ parameter, so the phrase component will
+only work where the query uses the exact same phrasing. I'm not sure
```
hannako: Don't think you should use first person.

doc/relevancy.md
```diff
@@ -0,0 +1,460 @@
...
+this is all that useful for us, except maybe for names of
+services/forms/departments.
+
+The `slop` parameter tells the `match_phrase` query how far apart terms are
```
hannako, line 401: Think this section could use an example.

hannako (commented):
Hi Bill, I made some comments, hope they're helpful!

doc/relevancy.md
```diff
@@ -0,0 +1,460 @@
...
+
+## What impacts relevancy?
+
+Out of the box, Elasticsearch comes with a decent scoring algorithm.
+They have a [guide on scoring relevancy][scoring] which is worth
+reading.
+
+We've done some work in Search API to [tune relevancy][relevancy],
+overriding the default Elasticsearch behaviour, which we go into
+below.
```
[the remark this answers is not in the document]
bilbof: This is not completely true, as we have various bits that do relevancy tuning elsewhere (e.g. index time optimisations, like synonyms).

doc/relevancy.md
bilbof: Thanks, I'll add a header to this section **Potential improvements to stopwords**.

doc/relevancy.md
bilbof: Sure thing I will add a note

bilbof:
Thanks, @hannako. I've made some changes.

hannako (approved):
Great, thanks for explaining that - much clearer to me 👍

PR #1625, opened 2019-07-11
Add a synonym of Myanmar for Burma

MahmudH:
This change ensures that searches for Myanmar also returns results for Burma and vice versa. This will surface more content when searching across gov.uk.

Searching Burma returns 692 results:
Screen Shot 2019-07-11 at 13 13 28

Searching Myanmar returns 8 results:
Screen Shot 2019-07-11 at 13 13 48

Trello: https://trello.com/c/DauneS2z

### DIFF
--- a/config/schema/synonyms.yml
+++ b/config/schema/synonyms.yml
@@ -1111,3 +1111,7 @@
 
 # Common misspellings of key people, in both publishing and searching
 - both: asad, assad => asad, assad
+
+# Country changes
+- search: burma => burma, myanmar
+- search: myanmar => burma, myanmar


cbaines:
I don't follow what you mean by "Burma changed to Myanmar". This doesn't seem to match up to the change you're making. What are you trying to change?

MahmudH:
We want consistent search results. When Burma is searched we also want Myanmar results to show up and vice versa.

cbaines:
> We want consistent search results. When Burma is searched we also want Myanmar results to show up and vice versa.

Well, that sounds like a good reason to make the change, so having that in the commit message would be useful.

Looking now at the first line of the commit message ("Burma changed to Myanmar"), it would be good if that could be changed to describe, or at least begin to explain what exact change is being made.

MahmudH:
@cbaines I have updated the commit and PR message, let me know if it explains it better?

cbaines:
> @cbaines I have updated the commit and PR message, let me know if it explains it better?

Thanks, while that's longer, I'm still not sure what's it's saying about what the commit's actually changing.

I guess when I look at the change, I think something like...

  If applied, this commit will **Add an entry for Burma to the synonyms config**

Rather than:

  If applied, this commit will **Name change of Burma to Myanmar not reflected in search results**

Does that make sense?

cbaines:
Also, have a look at the commits that have touched this file. Have a read of the commit messages, and see if they make sense to you: https://github.com/alphagov/search-api/commits/a572cc96a20e438cc009ecd738326ee98307b161/config/schema/synonyms.yml

cbaines (approved):

PR #1630, opened 2019-07-18
Set a useful user agent for the Search API OAuth client

cbaines:
This configuration propagates through to the Faraday gem, and should
replace the default user agent there.

I'm trying to track down requests to Signon, and passing a user agent
through which includes the app name will help with that.


AlanGabbianelli (approved):
Makes sense to me 👍

PR #1638, opened 2019-07-25
Add Burma news path to block list

MahmudH:
Currently, "/world/burma/news" shows up in search as a duplicate to "/world/myanmar/news". This change will stop this duplication in search. 

Trello: https://trello.com/c/DauneS2z


AlanGabbianelli (approved):
As already discussed offline, I'd prefer to understand why it's showing up again after being deleted and put a proper solution in place. But I understand that we also need to be pragmatic, so if this works, let's do it.

PR #1646, opened 2019-08-01
When given a sort order, break ties by sorting on content ID

barrucadu:
If you have two items with the exact same timestamp, and sort by timestamp, they can't appear at the same position in the output.  The tie needs to be broken somehow.

We don't currently have a tie breaker, so it's done by Elasticsearch.  We'd like to have an explicit tie-break in our code, so we can ensure that the order of things remains consistent across changes to Elasticsearch (reindexes, upgrades, etc).  So, if we've got an explicit ordering, break ties by sorting on content ID.

I picked content ID because a document's content ID can't change---that would be a new document.

We were initially going to do this in finder-frontend just for Atom feeds, but I feel it makes more sense to go here, as a general solution.

---

[Trello card](https://trello.com/c/nTGzIEpy/914-ordering-of-atom-feeds-generated-by-finder-frontend-is-arbitrary)

### DIFF
--- a/lib/search/query_components/sort.rb
+++ b/lib/search/query_components/sort.rb
@@ -10,7 +10,9 @@ module QueryComponents
         # Sort by popularity when there's no explicit ordering, and there's no
         # query (so there's no relevance scores).
         elsif search_term.nil? && !search_params.disable_popularity?
-          return [{ "popularity" => { order: "desc" } }]
+          return [{ "popularity" => { order: "desc" } }, tie_breaking_sort]
+        # If there's no explicit ordering, but there is a query (and
+        # so there are relevance scores), sort by relevance
         else
           return nil
         end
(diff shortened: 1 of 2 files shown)


sihugh (approved):
Looks good.  Are there things in search that don't have content IDs that this might affect weirdly?  I was thinking about "external search results" as an example, but don't think they'd be likely to be subject to the same issue that we're trying to solve here.

Essentially, I guess I'm wondering if we might need a second tie breaker that we could guarantee existed - so _id or link or something?

barrucadu:
`content_id` is in the base Elasticsearch type common to all documents: https://github.com/alphagov/search-api/blob/9335550d6005d2d6598b92a5f71cff40d7850c44/config/schema/base_elasticsearch_type.json#L4

And it looks like external links do get assigned one: https://www-origin.integration.publishing.service.gov.uk/api/search.json?q=The%20European%20Citizens%27%20Initiative&fields=title,format,content_id,link

```
{
"title": "The European Citizens' Initiative",
"format": "recommended-link",
"content_id": "012cd8ea-64db-40b5-8ceb-c4131a0d3d6b",
"link": "http://ec.europa.eu/citizens-initiative",
"index": "govuk",
"es_score": 0.0029646938,
"_id": "012cd8ea-64db-40b5-8ceb-c4131a0d3d6b",
"elasticsearch_type": "edition",
"document_type": "edition"
},
```

PR #1647, opened 2019-08-01
Automate Google Analytics import of indexed pages

bilbof:
https://trello.com/c/xJP8a5KE/905-automate-data-import-into-ga-for-enhanced-ecommerce-reporting

Currently we generate a CSV and then save it to disk, then an analyst manually uploads it to Google Analytics.

This change will mean that the CSV will be uploaded into GA directly, replacing the final manual step.

The importing stuff that I've added has been copied from what we have in local links manager for importing data into Google Analytics.

I have extra tasks once this is merged:
- add environment variables so this can be run later.
- create a scheduled Jenkins job to run the rake task daily
- remove the old code for generating the CSV and so on


lib/analytics/load_service.rb
```diff
@@ -0,0 +1,84 @@
...
+      @service.authorization = GoogleAnalyticsExportCredentials.authorization(SCOPES)
+    end
+
+    def upload_csv(csv_data)
+      assert_has_required_env_vars
+
+      response = ''
+      Tempfile.create(['search_api_indexed_pages', '.csv'], export_path) do |file|
+        file.write(csv_data)
+        response = service.upload_data(
```
bilbof, line 25: We call this method here https://github.com/googleapis/google-api-ruby-client/blob/0.30.8/generated/google/apis/analytics_v3/service.rb#L2991

lib/analytics/load_service.rb
```diff
@@ -0,0 +1,84 @@
...
+          content_type: 'application/octet-stream',
+        )
+      end
+      response
+    end
+
+    def delete_previous_uploads
+      assert_has_required_env_vars
+
+      upload_list = service.list_uploads(
```
bilbof, line 40: We call this method here https://github.com/googleapis/google-api-ruby-client/blob/0.30.8/generated/google/apis/analytics_v3/service.rb#L2946

lib/analytics/load_service.rb
```diff
@@ -0,0 +1,84 @@
...
+        ENV['GOOGLE_EXPORT_CUSTOM_DATA_SOURCE_ID'],
+      )
+
+      return unless upload_list.items.count.positive?
+
+      old_file_ids = upload_list.items.map(&:id)
+
+      delete_upload_data_request_object = Google::Apis::AnalyticsV3::DeleteUploadDataRequest.new(custom_data_import_uids: old_file_ids)
+
+      service.delete_upload_data(
```
bilbof, line 52: We call this method here https://github.com/googleapis/google-api-ruby-client/blob/0.30.8/generated/google/apis/analytics_v3/service.rb#L2860

lib/analytics/export_pages_to_google_analytics.rb
```diff
@@ -0,0 +1,11 @@
+require 'analytics/extract'
+require 'analytics/transform'
+require 'analytics/load'
+
+module Analytics
+  def self.export_pages_to_google_analytics
+    data = Extract.new(ALL_CONTENT_SEARCH_INDICES)
```
koetsier, line 7: Very nice! 👍

koetsier (approved):
looks great! 👍

PR #1648, opened 2019-08-02
Change ES6 search query, gated behind a new A/B test

barrucadu:
Also removes the old A/B test.

**Reasoning:**

- ES6 removed coordination factors, but if a doc matches two subqueries that doesn't *necessarily* mean it's twice as good a match as a doc which matches one subquery.  So use a `dis_max` query with a tie-break of 0.7.

- Copy over the field boosting we use for quoted search queries - these were chosen for  reason (hopefully).

- Decrease the boosting of `match_any_terms` and `minimum_should_match` - these only ensure a doc contains *some* of the search terms, so they should pull a doc into the results, but they shouldn't be very significant factors in scoring.

These changes seem to work well, we've eyeballed ~100 queries, and there was nothing that looked particularly bad.

---

[Trello card](https://trello.com/c/dvLAhldZ/929-figure-out-result-ordering-weirdness-with-es6)

### DIFF
--- a/lib/search/query_components/core_query.rb
+++ b/lib/search/query_components/core_query.rb
@@ -63,13 +63,15 @@ module QueryComponents
 
     def unquoted_phrase_query
       [
-        match_phrase("title"),
-        match_phrase("acronym"),
-        match_phrase("description"),
-        match_phrase("indexable_content"),
-        match_all_terms(%w(title acronym description indexable_content)),
-        match_any_terms(%w(title acronym description indexable_content)),
-        minimum_should_match("all_searchable_text")
+        dis_max_query([
+          match_phrase("title"),
+          match_phrase("acronym"),
+          match_phrase("description"),
+          match_phrase("indexable_content"),
+          match_all_terms(%w(title acronym description indexable_content)),
+          match_any_terms(%w(title acronym description indexable_content)),
+          minimum_should_match("all_searchable_text")
+        ])
       ]
     end
 
--- a/lib/search/query_components/core_query.rb
+++ b/lib/search/query_components/core_query.rb
@@ -69,7 +69,7 @@ module QueryComponents
           match_phrase("description", PHRASE_MATCH_DESCRIPTION_BOOST),
           match_phrase("indexable_content", PHRASE_MATCH_INDEXABLE_CONTENT_BOOST),
           match_all_terms(%w(title acronym description indexable_content)),
-          match_any_terms(%w(title acronym description indexable_content)),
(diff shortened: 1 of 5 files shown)


lib/search/query_components/core_query.rb
```diff
@@ -61,6 +61,18 @@ def quoted_phrase_query
       ]
     end
 
+    def unquoted_phrase_query
+      [
+        match_phrase("title"),
+        match_phrase("acronym"),
+        match_phrase("description"),
+        match_phrase("indexable_content"),
+        match_all_terms(%w(title acronym description indexable_content)),
```
sihugh: It's not part of this work, but I do wonder about including acronym in `match_all_terms`. I'm probably underselling it wildly, but my brain can only think of organisations that might populate this field.

lib/search/query_components/core_query.rb
```diff
@@ -71,7 +71,7 @@ def unquoted_phrase_query
           match_all_terms(%w(title acronym description indexable_content)),
           match_any_terms(%w(title acronym description indexable_content)),
           minimum_should_match("all_searchable_text")
-        ])
+        ], tie_breaker: 0.7)
```
sihugh, line 74: https://www.elastic.co/guide/en/elasticsearch/reference/6.7/query-dsl-dis-max-query.html 👍

lib/search/query_components/core_query.rb
```diff
@@ -99,11 +99,12 @@ def match_phrase(field_name, boost)
       }
     end
 
-    def match_all_terms(fields)
+    def match_all_terms(fields, boost = 1.0)
```
sihugh, line 103: I guess this is added to give the 3 methods a similar signature.  I approve

sihugh (approved):
Looks good to me, and reflects what we discussed this morning 👍

lib/search/query_components/core_query.rb
barrucadu: Actually it was added because I originally changed the boost here, but forgot to undo the change to the signature.  🤷‍♂

lib/search/query_components/core_query.rb
barrucadu: Magic numbers straight from the horse's mouth

lib/search/query_components/core_query.rb
barrucadu: A search for `hmrc sign on` will have `match_phrase("acronym")` fail, because it's only one word (not the full phrase) in the acronym; but will match the `match_all_terms` (assuming "sign on" is in the page somewhere)

PR #1650, opened 2019-08-05
Add hidden content_store_document_type facet to all-content finder

barrucadu:
We don't want users to think about document types, because they're
confusing and departments use them inconsistently, but sometimes a
department does want to be able to link users to (eg) a search scoped
to their decisions and nothing else.

Screen Shot 2019-08-05 at 16 32 26

No visible doctype facet, but nevertheless restricted to decisions.

---

[Trello card](https://trello.com/c/jpLSQGAS/923-discover-facets-and-urls-relevant-for-scrutineer-y-finders)

### DIFF
--- a/config/finders/all_content_email_signup.yml
+++ b/config/finders/all_content_email_signup.yml
@@ -17,6 +17,8 @@ details:
   email_filter_facets:
   - facet_id: content_purpose_supergroup
     facet_name: content_purpose_supergroup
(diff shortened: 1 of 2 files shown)


sophiebrown01 (approved):
Discussed with Michael. Perhaps we also want to add the `manual` filter to the `email_filter_facets` as a separate commit?

🚢

PR #1658, opened 2019-08-16
Add display_date to edition

sihugh:
We co-locate this with release date because they are related items. They seem to only be used in stats announcements at present though. 

Missed this from https://github.com/alphagov/search-api/pull/1657
https://trello.com/c/zucpgkPW/966-update-statistics-announcements-finder-to-use-provisional-dates


SamJamCul (approved):
Looking good, can't see why this would cause any issues!

PR #1654, opened 2019-08-13
Add find brexit guidance finder definition

theseanything:
This adds a definition file for a finder to support dynamic lists for Brexit guidance.

PR #1665, opened 2019-08-19
Removed obsolete files

Tetrino:
Removes two files no longer required by the system.


tijmenb (changes requested):
This is removing too much, only the files added in https://github.com/alphagov/search-api/pull/1654/files need to be removed.

theseanything (approved):

tijmenb (approved):

PR #1667, opened 2019-08-20
Adjust Brexit topic page title / description

JonathanHallam:
We need a title and description for the Brexit topic page that does not
match that of the taxon. This adds logic to hardcode what should show up
in search

Co-authored-by: Bruce Bolt <bruce.bolt@digital.cabinet-office.gov.uk>


emmabeynon (commented):
I have a couple of suggestions, let me know what you think.

emmabeynon (approved):
Nice one ⭐️

PR #1669, opened 2019-08-21
Rework ES6 query, again

barrucadu:
See commit messages for details.

---

[Trello card](https://trello.com/c/vpCrUBqO/988-test-michaels-new-search-query-l)

### DIFF
--- a/lib/search/query_components/core_query.rb
+++ b/lib/search/query_components/core_query.rb
@@ -9,6 +9,15 @@ module QueryComponents
     # to make the various possible queries more consistant with each other.
     QUERY_TIME_SYNONYMS_ANALYZER = "with_search_synonyms".freeze
 
+    # Clause boosts for a search query
+    MATCH_ALL_TITLE_BOOST = 5
+    MATCH_ALL_ACRONYM_BOOST = 5
+    MATCH_ALL_DESCRIPTION_BOOST = 2
+    MATCH_ALL_INDEXABLE_CONTENT_BOOST = 1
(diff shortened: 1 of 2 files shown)


lib/search/query_components/core_query.rb
```diff
@@ -62,17 +71,15 @@ def quoted_phrase_query
...
-          match_phrase("title", PHRASE_MATCH_TITLE_BOOST),
-          match_phrase("acronym", PHRASE_MATCH_ACRONYM_BOOST),
-          match_phrase("description", PHRASE_MATCH_DESCRIPTION_BOOST),
-          match_phrase("indexable_content", PHRASE_MATCH_INDEXABLE_CONTENT_BOOST),
-          match_all_terms(%w(title acronym description indexable_content)),
-          match_any_terms(%w(title acronym description indexable_content), 0.2),
-          minimum_should_match("all_searchable_text", 0.2)
-        ], tie_breaker: 0.7)
-      ]
+      should_coord_query([
```
sihugh, line 74: Unquoted search queries on the ES6 cluster get sent here, but I think we're using the same quoted search query for both ES5 and 6.  Is that right?  Is that OK?

barrucadu: Yes, I haven't changed the quoted search query.  That does exact matching, as it should, so I'm not sure there's anything to change there.  We might want to combine the two types of query in the future, but that's a separate card.

sihugh: Cool, I'm happy then 👍

sihugh (approved):

PR #1677, opened 2019-08-28
Remove `cluster` argument from docs

barrucadu:
We've been doing this with an A/B test for a while


SamJamCul (approved):
Looking good, in it goes!

PR #1686, opened 2019-09-10
Phrase-match and require quoted query fragments

barrucadu:
These queries all give different results in Google:

- `marzipan shoes` (most results)
- `"marzipan" shoes`
- `marzipan "shoes"`
- `"marzipan shoes"` (least results)

The behaviour seems to be that all quoted fragments are phrase-matched whereas unquoted fragments are not required to be present.

We currently treat `"marzipan" shoes` and `marzipan "shoes"` the same as `marzipan shoes`.  Quotes only affect the query if the *entire* query is quoted.

*However* the finder-frontend UI does display quoted fragments differently to unquoted fragments, which is misleading to users if the quotes don't actually matter.

---

This PR separates quoted and unquoted query fragments, performing a `must` `quoted_phrase_query` for each quoted fragment, and a `should` `unquoted_phrase_query` for any unquoted fragment.

An open question is how to handle the relative weighting of the required and non-required query fragments.  How much more significantly should a result which matches both the required and optional parts of the query rank than a result which only matches the required bits?  Can a result which only matches the required bits do so *really well* and outrank a result which matches both the required and the optional bits?

---

[Trello card](https://trello.com/c/3KVL5I44/989-add-support-for-mixing-quoted-and-unquoted-query-fragments)

### DIFF
--- a/lib/parameter_parser/search_parameter_parser.rb
+++ b/lib/parameter_parser/search_parameter_parser.rb
@@ -37,12 +37,14 @@ private
     # Dummy field that can be used to bypass caching when testing/debugging
     @used_params << "c"
 
+    query = normalize_query(single_param("q"), "query")
     @parsed_params = {
       start: capped_start,
       count: capped_count,
       cluster: cluster,
       search_config: SearchConfig.instance(cluster),
-      query: normalize_query(single_param("q"), "query"),
+      query: query,
+      parsed_query: parse_query(query),
       similar_to: normalize_query(single_param("similar_to"), "similar_to"),
       order: order,
       return_fields: return_fields,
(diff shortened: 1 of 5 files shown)


lib/parameter_parser/search_parameter_parser.rb
```diff
@@ -101,6 +103,37 @@ def capped_count
...
+          quoted << current_keyword unless current_keyword.empty?
+        else
+          unquoted << current_keyword unless current_keyword.empty?
+        end
+        is_inside_quote = !is_inside_quote
+        current_keyword = ''
+      else
+        current_keyword << c
+      end
+    end
```
bilbof, line 127: Could use a regex here, but I spent too long trying to get an array of unquoted strings without success. Your solution is better!

```
quoted = '"S" "A" "s" z y x "'.scan(/"([^"]*)"/).flatten
```

bilbof (approved):

lib/parameter_parser/search_parameter_parser.rb
barrucadu: Well, you know what they say about regex...

> Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.

PR #1689, opened 2019-09-13
Tweak B query variant

barrucadu:
There are three changes here:

**Phrase matching:**
It seems that phrase matching contributes strongly to DCG score.  Even
though we know phrase matching has problems (the "brexit business"
example), it makes sense that it is useful in many cases, as there are
many search queries like "income tax" and "vehicle tax", which are
phrases which appear on the page.

**Incomplete matching:**
Removing the match_any_terms seems to improve DCG score.  The
minimum_should_match kind of covers the same niche anyway.

**Boosting:**
Much simpler boosting, only for the phrase matching.

With these changes, comparing to the A variant with all of our
relevancy judgements, we get:

- A wins on 49
- B wins on 49
- A and B tie on 10

Of the 49 A wins, only 27 of them have a >5% difference in DCG score.
I want to investigate those further, but I think it's also useful to
get this query in front of real users, as our relevancy judgements are
limited in scope.

---

[Trello card](https://trello.com/c/UIHH2BUZ/1005-figure-out-which-fields-are-most-correlated-with-dcg-score)

### DIFF
--- a/lib/search/query_components/core_query.rb
+++ b/lib/search/query_components/core_query.rb
@@ -88,13 +88,16 @@ module QueryComponents
 
     def unquoted_phrase_query_abvariant(query = search_term)
       should_coord_query([
-        match_all_terms(%w(title), query, MATCH_ALL_TITLE_BOOST),
-        match_all_terms(%w(acronym), query, MATCH_ALL_ACRONYM_BOOST),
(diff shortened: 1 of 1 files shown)


sihugh (approved):
Looks good to me 👍

PR #1695, opened 2019-09-17
Add Myanmar to IDF Locations

callumknights:
Add Myanmar in elasticsearch_types, this will correct the
finder indexing for Myanmar in advance of the documents
being retagged.

### DIFF
--- a/config/schema/elasticsearch_types/international_development_fund.json
+++ b/config/schema/elasticsearch_types/international_development_fund.json
@@ -78,6 +78,10 @@
         "value": "mozambique"
       },
       {
+        "label": "Myanmar",
+        "value": "myanmar"
+      },
+      {
         "label": "Nepal",
         "value": "nepal"
       },


thomasleese (approved):

config/schema/elasticsearch_types/international_development_fund.json
```diff
@@ -37,10 +37,6 @@
         "label": "Bangladesh",
         "value": "bangladesh"
       },
-      {
-        "label": "Burma",
-        "value": "burma"
-      },
```
cbaines: As we're now thinking of manually re-tagging the content, it's probably good to avoid removing Burma here until the content is re-tagged.

callumknights: Didn't notice this, thanks!

cbaines (approved):

PR #1700, opened 2019-09-20
Add popularity_b field to documents

bilbof:
The new field will be the 'raw' rank of popularity, rather
than a computed version that we set as popularity.

The field will be useful, since we can test new popularity
functions at query time rather than computing the values
when they enter the index, requiring us to reindex all docs.

Trello: https://trello.com/c/Dbcmxucw/

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -84,6 +84,11 @@
     "type": "float"
   },
 
+  "popularity_b": {
+    "description": "Popularity rank used to bias search results. Used at query time to compute a popularity score. A higher number indicates more pageviews. Updated nightly.",
+    "type": "integer"
+  },
+
   "organisations": {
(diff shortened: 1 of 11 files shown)


config/schema/field_definitions.json
```diff
@@ -84,6 +84,11 @@
     "type": "float"
   },
 
+  "popularity_b": {
+    "description": "Popularity rank used to bias search results. Used at query time to compute a popularity score. A higher number indicates more pageviews. Updated nightly.",
+    "type": "float"
```
barrucadu: This is going to be an integer, is there an integer type we can use?

bilbof: Changed it to integer

barrucadu (approved):

bilbof:
I deployed it to integration and am running the reindex process there. I'll leave it there over the weekend.

PR #1702, opened 2019-09-23
Add rake task to generate sitemaps & upload to s3

barrucadu:
It's kind of silly that we need to generate files on disk for every search machine.  This made sense when we had 3 VMs in Carrenza which never got replaced, but with AWS we can bring up new VMs on the fly.  Those won't have sitemaps until the nightly cronjob runs.

Phase 1 of getting rid of that is to store the sitemap files in s3, which can be done by *one* search machine, by Jenkins triggering a rake task.  When that is happening, we can switch to serving sitemaps from s3.

Tested in integration: see the `govuk-integration-sitemaps` bucket.

---

[Trello card](https://trello.com/c/KuBZtH1v/1002-generate-sitemaps-when-a-new-search-machine-is-provisioned)


lib/sitemap/sitemap.rb
```diff
@@ -8,18 +8,23 @@ def initialize(directory, timestamp = Time.now.utc)
...
+
   def generate(search_config)
     FileUtils.mkdir_p(@output_path)
 
-    # generate and link the sitemap data files
     sitemap_writer = SitemapWriter.new(@output_path, @timestamp)
-    sitemap_filenames_with_linkname = sitemap_writer.write_sitemaps(search_config)
-    update_links(sitemap_filenames_with_linkname)
+    sitemaps = sitemap_writer.write_sitemaps(search_config)
+    index = write_index(sitemaps.map(&:last))
```
bilbof, line 20: Is `index` a filename, or the name of an index?

barrucadu: It's the filename of the sitemap index file (https://www.gov.uk/sitemap.xml)

lib/tasks/sitemap.rake
```diff
@@ -8,8 +10,35 @@ namespace :sitemap do
...
+          "#{output_directory}/#{Sitemap::SUB_DIRECTORY}/#{filename}",
+          "#{Sitemap::SUB_DIRECTORY}/#{link_filename}"
+        )
+      end
+
+      upload_to_s3(
+        "#{output_directory}/#{Sitemap::SUB_DIRECTORY}/#{filenames[:index]}",
+        "sitemap.xml"
+      )
+    end
```
bilbof, line 37: Out of interest, why are there multiple sitemaps?

lib/tasks/sitemap.rake
```diff
@@ -8,8 +10,35 @@ namespace :sitemap do
...
+  end
+
+  def upload_to_s3(source, target)
+    o = Aws::S3::Object.new(bucket_name: @bucket_name, key: target)
+    raise "Failed to upload sitemap file '#{source}' as '#{target}'" unless o.upload_file(source)
```
bilbof, line 42: Can `upload_file` raise an error to explain why the upload failed?

lib/sitemap/sitemap.rb
bilbof: Ah I see, in the context of search api index always makes me think of search indexes.

bilbof (approved):
Looks good, nice solution 👍

lib/tasks/sitemap.rake
barrucadu: I think so.  Earlier I had a problem where I got the source file path wrong, and got an error to that effect.  Unsure about AWS-side errors, but I'd assume so.

lib/tasks/sitemap.rake
barrucadu: Two reasons: firstly, there's a limit of how big a sitemap file can be (50,000 entries); and secondly, we've run out of memory previously trying to generate 50,000-entry sitemaps.

lib/tasks/sitemap.rake
bilbof: It would be great to capture that, especially as I guess we'll want to have an Icinga alert should this upload fail. I can see myself retrying it if that happens, since it'll be hard to debug.

Or, will the error get raised anyway? For some reason I thought it wouldn't, and we'd just get a boolean for success back.

lib/tasks/sitemap.rake
bilbof: Awesome, thanks for explaining 👍

lib/tasks/sitemap.rake
barrucadu: It throws at least one type of exception - possibly more.  But it also returns a boolean.  Maybe every error will raise an exception, so it'll only ever return `true`, but maybe not.

PR #1703, opened 2019-09-23
Enable automatic slicing on Elasticsearch reindexing

Tetrino:
This PR enabled slicing on the reindexing migration task, which in
localised testing provides a 20% performance improvement. This
improvement scales based on the number of shards available to
an index, so should be higher still on our Elasticsearch cluster.

Slicing enables Elasticsearch to split large scrolls into multiple
chunks, processing each one in parallel. It is safe and relatively
memory efficient, and should lead to no detriment to performance
even in cases where we see no gain.

This took some time as other approaches were experimented with
but they proved unstable or detrimental.

Trello: https://trello.com/c/XGUXvcRO/1016-spike-improve-performance-of-search-reindexing

### DIFF
--- a/lib/schema_migrator.rb
+++ b/lib/schema_migrator.rb
@@ -10,7 +10,7 @@ class SchemaMigrator
 
   def reindex
     index_group.current.with_lock do
-      response = Services.elasticsearch(cluster: cluster, timeout: 60).reindex(
+      response = Services.elasticsearch(hosts: "#{cluster.uri}?slices=auto", timeout: 60).reindex(
         wait_for_completion: false,
         body: {
           source: {


lib/schema_migrator.rb
```diff
@@ -10,7 +10,7 @@ def initialize(index_name, cluster: Clusters.default_cluster, wait_between_task_
 
   def reindex
     index_group.current.with_lock do
-      response = Services.elasticsearch(cluster: cluster, timeout: 60).reindex(
+      response = Services.elasticsearch(hosts: 'http://localhost:9200?slices=auto', cluster: cluster, timeout: 60).reindex(
```
barrucadu: The URL isn't `localhost:9200` in AWS, and the presence of the `cluster` argument means that the `hosts` argument [gets ignored anyway](https://github.com/alphagov/search-api/blob/2c92aecd8f557121f4e8d42d2a6e95902afa93ca/lib/services.rb#L31).  This should be:

```ruby
response = Services.elasticsearch(hosts: "#{cluster.uri}?slices=auto", timeout: 60).reindex(
```

barrucadu (changes requested):

lib/schema_migrator.rb
bilbof: `Services.elasticsearch` takes `hosts` or `cluster`, but only uses one of them. If hosts is present it will ignore cluster:

```
    Elasticsearch::Client.new(
      hosts: cluster ? cluster.uri : hosts,
```

In this case, you'll need to add `slices=auto` to the `cluster.uri` (or find some other way to make this config change).

bilbof (changes requested):
Need to make a small change, but good work, I had to [read this](https://stackoverflow.com/questions/43211387/what-does-elasticsearch-automatic-slicing-do) to understand :)

lib/schema_migrator.rb
Tetrino: Actual documentation is pending, when I can refine it ;)

I'm going to take Michael's option, as it does what you want nice and simply.

bilbof (approved):

barrucadu (approved):

PR #1706, opened 2019-09-24
Add people as a field to `all_searchable_text`

Tetrino:
A simple modification that enables documents to be searched by the people field identifier.

Trello: https://trello.com/c/nT2yAoYH/1020-add-person-as-a-field-to-allsearchabletext-%F0%9F%8D%90

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -324,7 +324,7 @@
 
   "people": {
     "description": "Links to people associated with this page (https://www.gov.uk/government/people).",
-    "type": "identifiers"
+    "type": "searchable_identifiers"
   },
 
   "policy_groups": {


barrucadu (commented):
We tried this out on staging for a bit today, and it didn't seem that helpful unfortunately.  We got more results for name searches (eg, 14,800 vs  14,272 for `david cameron`), but nothing new showed up on the first page of results.  Which makes sense because the `all_searchable_text` has such low boosting.

Searching over the `people` field directly would help there, as we could boost it pretty highly---but we can't do that with it being an identifier field, and I don't know what the ramifications of changing it to a text field are.

There's a further problem in that quoted searches don't look at the `all_searchable_text`, so `"david cameron"` wouldn't find any of the new results.

Tetrino:
> Searching over the `people` field directly would help there, as we could boost it pretty highly---but we can't do that with it being an identifier field, and I don't know what the ramifications of changing it to a text field are.

Is this worth further investigating? Perhaps allowing us to boost an ID field in some way?

> There's a further problem in that quoted searches don't look at the `all_searchable_text`, so `"david cameron"` wouldn't find any of the new results.

I'm assuming fixing the above will solve this issue?

PR #1707, opened 2019-09-24
Serve sitemap files from s3

barrucadu:
Until the sitemap files are removed from the filesystem, nginx will
continue to serve those (rather than fall back to the app).

You can test this by talking to search-api directly, eg deploy this branch to integration, ssh in, and do `curl localhost:3233/sitemap.xml`

---

[Trello card](https://trello.com/c/KuBZtH1v/1002-generate-sitemaps-when-a-new-search-machine-is-provisioned)


lib/rummager/app.rb
```diff
@@ -343,6 +344,27 @@ def get_type_from_request_body(request_body)
...
+  get "/sitemap.xml" do
+    serve_from_s3("sitemap.xml")
+  end
+
+  get "/sitemaps/:sitemap" do |sitemap|
+    serve_from_s3("sitemaps/#{sitemap}")
+  end
+
+  def serve_from_s3(key)
+    o = Aws::S3::Object.new(bucket_name: ENV["AWS_S3_BUCKET_NAME"], key: key)
```
bilbof, line 356: Do you need to guard against `AWS_S3_BUCKET_NAME` not being defined?

bilbof (approved):
Will the `whenever` gem clear up the existing cron jobs?

lib/rummager/app.rb
barrucadu: `Aws::S3::Object.new` will throw an error about a missing bucket name if it's `nil`.

barrucadu:
Good question, I don't know.  I'll deploy this to integration and re-run puppet.

barrucadu:
Turns out a change to govuk-app-deployment to remove then`whenever` call is needed first: https://github.com/alphagov/govuk-app-deployment/pull/329

PR #1710, opened 2019-09-25
Set cache headers on sitemaps

barrucadu:
These are the headers nginx used for the files on disk: https://github.com/alphagov/govuk-puppet/pull/9628/files

### DIFF
--- a/lib/rummager/app.rb
+++ b/lib/rummager/app.rb
@@ -1,4 +1,5 @@
 require "aws-sdk-s3"
+require "date"
 require "sinatra"
 set :root, File.dirname(__FILE__)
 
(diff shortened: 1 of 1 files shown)


sihugh (approved):

sihugh (changes requested):
The RFC says:
> MUST be in RFC 1123 date format

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
Example from linked page:
`Expires: Thu, 01 Dec 1994 16:00:00 GMT`

sihugh (approved):

sihugh:
📅 🚀 🎸

barrucadu:
It's occurred to me that there's a race condition here, as the sitemaps are generated after midnight - you could get yesterday's sitemaps cached for a day if you request them in that window.  But is having a day old sitemap a problem?  *Probably* not.

sihugh:
Some of our pages haven't been indexed by Google for months, and not every page is in the sitemap anyway (for example chapters in guides), so I think it's fine

PR #1713, opened 2019-10-01
Switch to B variant, remove A cluster

barrucadu:
We'll rename the "B" cluster to "A" later.

---

[Trello card](https://trello.com/c/Rud9WSa1/1035-disable-a-cluster)


lib/parameter_parser/search_parameter_parser.rb
```diff
@@ -458,15 +458,8 @@ def debug_options
...
-        variants = character_separated_param("ab_tests")
-        variants = variants.map { |variant| variant.split(":", 2) }
-
-        variants.each_with_object({}) do |(variant_name, variant_code), variants_hash|
-          if variant_code.blank?
-            @errors << %{Invalid ab_tests, missing type "#{variant_name}"}
-          end
-          variants_hash[variant_name.to_sym] = variant_code
-        end
+        character_separated_param("ab_tests")
```
sihugh: Does this do anything?

barrucadu:
I've come up with a more sensible way of doing this, force-push incoming

lib/parameter_parser/search_parameter_parser.rb
barrucadu: Yes, it stops the parser complaining about an unused parameter

sihugh (approved):

PR #1722, opened 2019-10-03
Generate sitemaps individually

bilbof:
Generating and writing the XML files to temp files individually, rather than generating them all at once and then writing them to temp files, is less memory intensive.

https://trello.com/c/tGJcHr2f/1041-investigate-sitemap-generation-failures

### DIFF
--- a/lib/sitemap/sitemap_generator.rb
+++ b/lib/sitemap/sitemap_generator.rb
@@ -1,10 +1,35 @@
 class SitemapGenerator
-  EXCLUDED_FORMATS = %w[recommended-link].freeze
-
   def initialize(search_config)
     @search_config = search_config
     @all_documents = get_all_documents
   end
+
+  def sitemaps
+    @all_documents.each_slice(self.class.sitemap_limit).map do |chunk|
+      generate_xml(chunk)
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


lib/sitemap/sitemap_generator.rb
```diff
@@ -1,15 +1,35 @@
...
+      end
+    end
+
+    builder.to_xml
+  end
+
+private
+
+  EXCLUDED_FORMATS = %w[recommended-link].freeze
+  SITEMAP_LIMIT = 25_000
```
barrucadu, line 31: We may be able to up this to 50,000 now - which is the max amount of entries one sitemap file can have.  It used to be 50,000, but when we switched to ES5 the rake task was regularly running out of memory so I halved it.

barrucadu (approved):

hannako (approved):
👍

lib/sitemap/sitemap_generator.rb
bilbof: That's interesting, is there a benefit to having fewer files?

barrucadu: Not sure.  I imagine having 50,000 1-entry files would be bad, but I don't know if 1 50,000-entry file is better than 2 25,000-entry files.

PR #1726, opened 2019-10-04
Remove Burma from finder

callumknights:
Last part of the work for this ticket https://trello.com/c/C818g8vk/1218-3-switch-burma-to-myanmar-in-specialist-publisher

I didn't remove Burma initially so I could safely retag the documents without causing any issues, this is now done so any references to Burma can be safely removed. 

### Related PR's

- https://github.com/alphagov/search-api/pull/1726
- https://github.com/alphagov/specialist-publisher/pull/1513
- https://github.com/alphagov/specialist-publisher/commit/5e8963af29ef5640b31d36ddf5720e4d2252f449
- https://github.com/alphagov/search-api/commit/e8b38c7a82e5e03f8184a811228b9f18960b06df
- https://github.com/alphagov/govuk-content-schemas/commit/24aafbc3ba5c16b334595c9f15015119a4b98980

### DIFF
--- a/config/schema/elasticsearch_types/international_development_fund.json
+++ b/config/schema/elasticsearch_types/international_development_fund.json
@@ -38,10 +38,6 @@
         "value": "bangladesh"
       },
       {
-        "label": "Burma",
-        "value": "burma"
-      },
-      {
         "label": "Democratic Republic of the Congo",
         "value": "democratic-republic-of-congo"
       },


bevanloon (changes requested):
Could we provide a slightly better description for the PR please. Something about "why" we are making the change (and maybe why we didn't make the change as part of https://github.com/alphagov/search-api/pull/1695) and then link to the previous PRs in this series. Also a link to the trello card would be useful.

bevanloon (approved):

PR #1728, opened 2019-10-04
Add popularity b variant

bilbof:
Uses the suggested method of boosting by popularity from https://www.elastic.co/guide/en/elasticsearch/guide/current/boosting-by-popularity.html

This is not the default but can be enabled with `/api/search.json?ab_tests=popularity:B`.

Why make this change?

Currently our popularity booster gives massive gains to the most popular content, but doesn't help slightly less popular content. It produces [a bit of a cliff](https://www.wolframalpha.com/input/?i=x+*+%28%281.0+%2F+%28y+%2B+10%29%29+%2B+0.001%29+from+y%3D450000+to+0+from+x%3D0.01+to+2+) (lower ranks are better below):
 
gains

This change aims to prevent popular content showing up in results when it is not relevant.

For example, a query for "education" currently has "Universal Credit" and "Apply to the EU Settlement Scheme" on the first page of results. With this change "Education Maintenance Allowance" and "Contact the Department for Education" appear on the first page of results.

A logarithmic scale decreases the cliff. This change reduces the impact of popularity on the most popular documents. I've tried to keep parity for the less popular results. I also added a cap of `5` to the boost, so there's a limit to the amount of boosting being popular will get your doc.

After (higher ranks are better - most popular doc has a rank of 450k):

after

Screen Shot 2019-10-04 at 15 24 39

Result relevancy is improved for some queries but made worse for others, so before enabling this by default I'll do some testing. I'll do some manual testing with the search performance explorer and the rank eval API. If it looks good we can run an AB test.

For this to take effect we must run the following rake task:

```
PROCESS_ALL_DATA=true SEARCH_INDEX=all search:update_popularity
```

https://trello.com/c/tjrbWhML/1045-try-different-ways-of-computing-rank-based-popularity-boosts

### DIFF
--- a/lib/indexer/popularity_lookup.rb
+++ b/lib/indexer/popularity_lookup.rb
@@ -40,11 +40,13 @@ module Indexer
           popularity_score = 1.0 / (ranks[link] + SearchConfig.popularity_rank_offset)
         end
 
+        popularity_rank = traffic_index_size - (ranks[link] || traffic_index_size)
+
         [
           link,
           {
             popularity_score: popularity_score,
-            popularity_rank: ranks[link],
+            popularity_rank: popularity_rank,
           },
         ]
       }]
@@ -63,11 +65,13 @@ module Indexer
     end
 
     def traffic_index_size
-      results = traffic_index.raw_search({
-        query: { match_all: {} },
-        size: 0,
-      })
-      results["hits"]["total"]
+      @traffic_index_size ||= begin
+        results = traffic_index.raw_search({
+          query: { match_all: {} },
+          size: 0,
+        })
+        results["hits"]["total"]
+      end
     end
 
     def open_traffic_index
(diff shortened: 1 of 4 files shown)


lib/indexer/popularity_lookup.rb
```diff
@@ -40,11 +40,13 @@ def lookup_popularities(links)
           popularity_score = 1.0 / (ranks[link] + SearchConfig.popularity_rank_offset)
         end
 
+        popularity_rank = traffic_index_size - (ranks[link] || traffic_index_size)
```
barrucadu, line 43: See https://github.com/alphagov/search-api/pull/1731#discussion_r332426000

bilbof: This is fixed in that PR. Are you OK for this to be merged?

barrucadu (approved):

PR #1731, opened 2019-10-07
Add popularity C variant

bilbof:
This adds view_count to content indexes. The view count is already in the popularity index, but not specified as an indexed field. This makes it available in all indexes for query time popularity boosting.

View count is used for query time boosting, to see if using the absolute view count rather than popularity rank will have an impact on relevancy.

I've enabled this as an AB test so we can test it manually with the search-performance-explorer and go live with an ab test if it works.

For this to take effect we must run the following rake task:

```
PROCESS_ALL_DATA=true SEARCH_INDEX=all search:update_popularity
```

Trello: https://trello.com/c/pNxIeTFc/1044

### DIFF
--- a/lib/indexer/popularity_lookup.rb
+++ b/lib/indexer/popularity_lookup.rb
@@ -16,37 +16,44 @@ module Indexer
             path_components: links,
           },
         },
-        _source: { includes: %w[rank_14] },
+        _source: { includes: %w[rank_14 vc_14] },
         sort: [
           { rank_14: { order: "asc" } },
         ],
         size: 10 * links.size,
       })
 
-      ranks = Hash.new(traffic_index_size)
+      default_rank = Hash.new(
+        rank: traffic_index_size,
+        view_count: 1,
+      )
 
-      results["hits"]["hits"].each do |hit|
+      ranks = results["hits"]["hits"].each_with_object(default_rank) do |hit, hsh|
         link = hit["_id"]
-        rank = Array(hit["_source"]["rank_14"]).first
+        rank = Array(hit.dig("_source", "rank_14")).first
+        view_count = hit.dig("_source", "vc_14") || 1
         next if rank.nil?
 
-        ranks[link] = [rank, ranks[link]].min
+        hsh[link] = { rank: rank, view_count: view_count }
       end
 
       Hash[links.map { |link|
-        if ranks[link].zero?
+        if ranks[link][:rank].zero?
           popularity_score = 0
+          popularity_rank = 1
         else
-          popularity_score = 1.0 / (ranks[link] + SearchConfig.popularity_rank_offset)
+          popularity_score = 1.0 / (ranks[link][:rank] + SearchConfig.popularity_rank_offset)
+          popularity_rank = traffic_index_size - (ranks[link][:rank] || traffic_index_size)
         end
 
-        popularity_rank = traffic_index_size - (ranks[link] || traffic_index_size)
+        view_count = ranks[link][:view_count]
 
         [
           link,
           {
             popularity_score: popularity_score,
             popularity_rank: popularity_rank,
+            view_count: view_count,
           },
         ]
       }]
--- a/spec/unit/elasticsearch_index_spec.rb
+++ b/spec/unit/elasticsearch_index_spec.rb
@@ -56,8 +56,8 @@ RSpec.describe SearchIndices::Index do
     stub_popularity_index_requests(["/foo/bar", "/foo/baz"], 1.0, 20)
 
     json_documents = [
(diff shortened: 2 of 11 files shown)


spec/unit/elasticsearch_index_spec.rb
```diff
@@ -56,8 +56,8 @@
     stub_popularity_index_requests(["/foo/bar", "/foo/baz"], 1.0, 20)
 
     json_documents = [
-      { "document_type" => "edition", "link" => "/foo/bar", "title" => "TITLE ONE", "popularity" => "0.09090909090909091" },
-      { "document_type" => "edition", "link" => "/foo/baz", "title" => "TITLE TWO", "popularity" => "0.09090909090909091" },
+      { "document_type" => "edition", "link" => "/foo/bar", "title" => "TITLE ONE", "popularity" => "0.09090909090909091", "view_count" => "0.09090909090909091" },
+      { "document_type" => "edition", "link" => "/foo/baz", "title" => "TITLE TWO", "popularity" => "0.09090909090909091", "view_count" => "0.09090909090909091" },
```
barrucadu: Given that `view_count` is an integer field in the schema, I'd be happier if the tests also used integers

lib/indexer/popularity_lookup.rb
```diff
@@ -16,37 +16,43 @@ def lookup_popularities(links)
...
+        rank: traffic_index_size,
+        view_count: 1,
+      )
 
-      results["hits"]["hits"].each do |hit|
+      ranks = results["hits"]["hits"].each_with_object(default_rank) do |hit, hsh|
         link = hit["_id"]
-        rank = Array(hit["_source"]["rank_14"]).first
+        rank = Array(hit.dig("_source", "rank_14")).first
+        view_count = hit.dig("_source", "vc_14") || 1
```
barrucadu, line 34: Should this also have a fallback value for `rank`?  And if `rank` doesn't need one, why does `view_count`?

lib/indexer/popularity_lookup.rb
```diff
@@ -16,37 +16,43 @@ def lookup_popularities(links)
...
         next if rank.nil?
 
-        ranks[link] = [rank, ranks[link]].min
+        hsh[link] = { rank: rank, view_count: view_count }
       end
```
barrucadu, line 38: Previously, `ranks[link]` would have always returned a value.  Now `ranks[link]` will only return a value if the `link` is the `_id` of some result.  Given that, later on, we don't check if `ranks[link]` is non-`nil` before using it, is this a problem?

lib/indexer/popularity_lookup.rb
```diff
@@ -16,37 +16,43 @@ def lookup_popularities(links)
...
-        if ranks[link].zero?
+        if ranks[link][:rank].zero?
           popularity_score = 0
         else
-          popularity_score = 1.0 / (ranks[link] + SearchConfig.popularity_rank_offset)
+          popularity_score = 1.0 / (ranks[link][:rank] + SearchConfig.popularity_rank_offset)
         end
 
-        popularity_rank = traffic_index_size - (ranks[link] || traffic_index_size)
+        popularity_rank = traffic_index_size - (ranks[link][:rank] || traffic_index_size)
```
barrucadu: I'm not sure this is right (and that was also the case previously) if `ranks[link][:rank].zero?` holds:

```
traffic_index_size - (ranks[link][:rank] || traffic_index_size)
= traffic_index_size - (0 || traffic_index_size)         ; ranks[link][:rank].zero? == true
= traffic_index_size - 0                                 ; 0 || x == 0
= traffix_index_size                                     ; x - 0 = 0
```

So if something has a rank of zero (which, to be fair, I don't think should happen because [ranking starts from 1](https://github.com/alphagov/search-analytics/blob/master/analytics_fetcher/fetch.py#L64)) then it gets given the greatest possible `popularity_rank`.

lib/indexer/popularity_lookup.rb
bilbof: Rank has one, sorry it's a bit convoluted. There's a `next if rank.nil?` in there, so we provide the default rank if there's no rank provided in the popularity index.

```ruby
      default_rank = Hash.new(
        rank: traffic_index_size,
        view_count: 1,
      )
```

We don't do the same for view_count because I want to preserve the behaviour we have live. So if rank is present but view_count isn't (I'm not sure why that'd happen), then `view_count` will still get a default value.

lib/indexer/popularity_lookup.rb
bilbof: ranks[link] should still always return a value, since the hash has a default value: `Hash.new(rank: traffic_index_size, view_count: 1)`

barrucadu: Introducing a new variable confused me far too much 🤦‍♂

barrucadu (approved):

PR #1732, opened 2019-10-07
Highlight quoted query fragments in titles and descriptions

barrucadu:
Highlighting is done on the title and description fields (or their
.synonym fields), but quoted queries are run against the .no_stop
fields.

One approach would be to add the .no_stop fields to the highlighting
query, and then using that field in Search::HighlightedField if the
other fields aren't present.  But that wouldn't help with queries that
mix quoted and unquoted fragments.

For queries which match quoted and unquoted fragments, we want to
combine highlighting from two different fields.  We can do this by
using a (much simpler!) query to highlight results.  I've used
"match_phrase" for quoted fragments (so they're highlighted in their
entirety) and "match" for any unquoted fragment (so individual
keywords are highlighted).

Highlighting quoted fragments isn't perfect:

- Stopwords aren't highlighted
- Synonyms are applied (so a quoted query could match something new)

I'm not sure how best to work around those issues.  We could do all
highlighting against the no_stop field and apply all the index-time
synonyms at query-time for the unquoted fragment.  But this way is
better than not working at all.

---

[Trello card](https://trello.com/c/Gshiupx4/1056-highlight-quoted-search-terms)

### DIFF
--- a/lib/search/query_components/highlight.rb
+++ b/lib/search/query_components/highlight.rb
@@ -10,17 +10,21 @@ module QueryComponents
         fields: {
           title: {
             number_of_fragments: 0,
+            highlight_query: highlight_query(:title),
           },
           "title.synonym".to_sym => {
             number_of_fragments: 0,
+            highlight_query: highlight_query("title.synonym".to_sym),
           },
           description: {
             number_of_fragments: 1,
             fragment_size: 285,
+            highlight_query: highlight_query(:description),
           },
           "description.synonym".to_sym => {
             number_of_fragments: 1,
             fragment_size: 285,
+            highlight_query: highlight_query("description.synonym".to_sym),
           },
         },
       }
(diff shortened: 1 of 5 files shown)


bilbof (dismissed):
This looks good. Though I wonder how many quoted queries we see, since this adds a bit of complexity to the highlighting logic?

barrucadu:
I think Steve said it was only 0.18% of queries.  Also, I think I spotted a bug: see new commit.

lib/search/query_components/highlight.rb
```diff
@@ -10,17 +10,21 @@ def payload
         fields: {
           title: {
             number_of_fragments: 0,
+            highlight_query: highlight_query(:title),
           },
           "title.synonym".to_sym => {
             number_of_fragments: 0,
+            highlight_query: highlight_query("title.synonym".to_sym),
```
sihugh: I _think_ the behaviour of highlighting synonyms is a benefit as it'll hopefully improve understanding of why a result has been returned.  It might confuse some people though.  🤷‍♂ 

It'd be interesting to sneak this into a benchmarking test

spec/unit/query_components/highlight_spec.rb
```diff
@@ -2,36 +2,36 @@
...
 RSpec.describe QueryComponents::Highlight do
   describe "#payload" do
-    it "enables highlighting on title" do
+    it "enables highlighting on title with synonyms" do
       parameters = Search::QueryParameters.new(parsed_query: {}, return_fields: %w[title_with_highlighting])
 
       payload = described_class.new(parameters).payload
 
-      expect(payload[:fields].keys).to include(:title)
+      expect(payload[:fields].keys).to include("title.synonym")
```
sihugh, line 10: It's slightly weird that we sometimes use `"title.synonym".to_sym` (as in the previous commit)

sihugh: Ah, and I see that those bits have been removed in this commit

sihugh (approved):
Looks good to me 👍

PR #1740, opened 2019-10-14
Generate sitemap lazily to limit memory consumption

laurentqro:
The nightly job for generating a sitemap falls over every now and then. The process runs out of memory because we attempt to load all the data into memory.

Follows the investigation https://trello.com/c/tGJcHr2f/1041-investigate-sitemap-generation-failures.

This PR makes sure we generate the sitemaps in a lazy way.

It also entails quite a bit of refactoring, the biggest part of which is removing all the logic related to file creation and symlink shuffling. This is not needed anymore because sitemap files and index are hosted on S3 rather than on disk. We now generate the XML in-memory and upload it directly.

---

https://trello.com/c/wihXuUoT/1057-generate-sitemap-in-chunks


lib/s3_client.rb
```diff
@@ -0,0 +1,10 @@
+class S3Client
```
barrucadu: This class is only used in `lib/sitemap`, so I think it should be moved there.

barrucadu:
This tightly couples sitemap generation with sitemap uploading (eg, in `sitemap_generator.rb` which uploads files as soon as they've been created), so I think it would be clearer if this PR also removed the `sitemap:generate_and_replace` task and its associated symlink-shuffling logic.

When that is removed, it might make sense to generate the XML in-memory and upload it directly, rather than write to a file on disk and then immediately have the S3 library read that file.

laurentqro:
@barrucadu Implemented your recommendations, feels good!

lib/tasks/sitemap.rake
```diff
@@ -1,44 +1,23 @@
 require "aws-sdk-s3"
-require "sitemap/sitemap"
 require "tmpdir"
```
barrucadu: This `require "tmpdir"` can go away now

lib/tasks/sitemap.rake
```diff
@@ -1,44 +1,23 @@
...
-    end
-  end
-
-  def upload_to_s3(source, target)
-    o = Aws::S3::Object.new(bucket_name: @bucket_name, key: target)
-    raise "Failed to upload sitemap file '#{source}' as '#{target}'" unless o.upload_file(source)
+    search_client = Services.elasticsearch(
+      cluster: Clusters.default_cluster,
+      timeout: 10,
+    )
```
barrucadu: I think it would be better to get the `search_client` from the `search_config` inside `Sitemap::Generator`.  This should work:

```ruby
@search_client = Services.elasticsearch(hosts: search_config.base_uri, timeout: 10)
```

And remove the `search_client` argument from `initialize`.

lib/sitemap/generator.rb
```diff
@@ -0,0 +1,135 @@
...
+        from: batch_index * SITEMAP_LIMIT,
+        size: SITEMAP_LIMIT,
+        query: {
+          bool: {
+            must_not: { terms: { format: EXCLUDED_FORMATS } },
+          },
+        },
+        post_filter: Search::FormatMigrator.new(@search_config).call,
+      }
+    end
```
barrucadu: I have some concerns about this. 

1. We might need to change the index config for it to work:

    > Note that `from` + `size` can not be more than the `index.max_result_window` index setting, which defaults to 10,000.

    https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#request-body-search-from-size

2. We have 400,000+ documents, which is a *lot* larger than 10,000, so I wonder if increasing the window size that much would cause performance issues.

3. How will this handle new documents being inserted during sitemap generation?  We tried to do something like this to import ES2 data into ES5, and ended up with many missing documents.  Though now that I think about it I don't think we changed the result window size...

Can't we use the [scroll API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#request-body-search-scroll)?  It's designed for processing large numbers of documents and handles insertions/deletions mid-scroll.  We could set a scroll timeout of 10m or something, which should be far more than enough to process a batch of results.

We [already have code for a scroll-based enumeration](https://github.com/alphagov/search-api/blob/master/lib/scroll_enumerator.rb).  I think that's the right thing to use, even if it needs to be made lazier to work.

lib/sitemap/generator.rb
```diff
@@ -0,0 +1,135 @@
...
+            }
+          end
+        end
+      end
+      builder.to_xml
+    end
+
+    def batches_of_documents
+      Enumerator::Lazy.new(0...batch_total_count) do |yielder, index|
+        yielder << [homepage] + get_documents_for_batch(index)
```
barrucadu: Does this yield the homepage on every iteration, or just the first one?

lib/sitemap/generator.rb
laurentqro: I can look into making scroll API lazier 👌

lib/sitemap/generator.rb
```diff
@@ -0,0 +1,150 @@
...
+        if more_documents.any?
+          documents.push(*more_documents)
+        else
+          sitemaps = documents.map do |document|
+            SitemapPresenter.new(document["_source"], property_boost_calculator)
+          end
+        end
+
+        yielder << sitemaps if sitemaps
+      end
```
barrucadu: I don't really understand this enumerator body, I'd have expected something like:

```ruby
# yield individual documents and have `generate_sitemap_xml` group them into chunks of SITEMAP_LIMIT
Enumerator::Lazy.new(0..total_page_count) do |yielder|
  break if documents.empty?

  documents.map do |document|
    yielder << SitemapPresenter.new(document["_source"], property_boost_calculator)
  end

  documents = scroll(scroll_id)
end
```

Or

```ruby
# yield chunks of size SITEMAP_LIMIT
chunk = []
Enumerator::Lazy.new(0..total_page_count) do |yielder|
  documents.map do |document|
    chunk << SitemapPresenter.new(document["_source"], property_boost_calculator)
    if chunk.size == SITEMAP_LIMIT
      yielder << chunk
      chunk = []
    end
  end

  documents = scroll(scroll_id)

  if documents.empty?
    yielder << chunk
    break
  end
end
```

Also, it looks like the `scroll_id` changes between requests: https://github.com/alphagov/search-api/blob/master/lib/scroll_enumerator.rb#L47-L49

Can't the existing `ScrollEnumerator` class be re-used somehow?

barrucadu: Given that `break` doesn't work and there's nothing built in to group into chunks, which would make option 1 nice, this is a candidate:

```ruby
chunk = []
Enumerator::Lazy.new(0..total_page_count) do |yielder|
  next if documents.empty? and chunk.empty?

  documents.map do |document|
    chunk << SitemapPresenter.new(document["_source"], property_boost_calculator)
    if chunk.size == SITEMAP_LIMIT
      yielder << chunk
      chunk = []
    end
  end

  documents = scroll(scroll_id)

  if documents.empty?
    yielder << chunk
    chunk = []
  end
end
```

lib/sitemap/generator.rb
```diff
@@ -0,0 +1,150 @@
...
+            yielder << chunk
+            chunk = []
+          end
+        end
+
+        documents = scroll(scroll_id)
+
+        if documents.empty?
+          yielder << chunk
+        end
```
barrucadu: If we get multiple empty `documents` lists at the end of the scroll (I don't know if that can happen), then this will yield the same chunk multiple times.  So we should do something like

```ruby
if documents.empty?
  yielder << chunk unless chunk.empty?
  chunk = []
end
```

lib/sitemap/generator.rb
```diff
@@ -0,0 +1,150 @@
...
+      end
+    end
+
+    def initial_scroll
+      @search_client.search(scroll_query)
+    end
+
+    def scroll(scroll_id)
+      @search_client.scroll(scroll_id: scroll_id, scroll: "1m")["hits"]["hits"]
+    end
```
barrucadu: The `scroll_id` changes between requests (see https://github.com/alphagov/search-api/blob/master/lib/scroll_enumerator.rb#L47-L49), so this function needs to return the documents and also the new `scroll_id`.

lib/sitemap/presenter.rb
```diff
@@ -36,6 +36,10 @@ def priority
     withdrawn_status_boost * property_boost_calculator.boost(document)
   end
 
+  def root?
```
barrucadu: This method isn't used anywhere

lib/sitemap/generator.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    def run
+      create_sitemap_index(create_sitemaps(get_all_documents))
+    end
+
+    def create_sitemaps(enumerator)
+      enumerator.with_index.map do |documents, index|
+        batch_number = index + 1
+        documents.unshift(homepage) if batch_number == 1
```
barrucadu: Instead of `unshift`ing the homepage, would it be less confusing for line 76 to be `chunk = [homepage]`?  The problem with doing it here is that the first sitemap file will have 1 more entry than all the others, which could be confusing (especially if we decide we want all the sitemap files to have 50,000 entries, as that's the maximum).

laurentqro: I really like that, let's do it.

barrucadu (approved):
:+1:

PR #1742, opened 2019-10-15
Use HTTParty to make rank eval request

bilbof:
We need to make a basic http request here rather than
using the elasticsearch ruby client.

This is due to two factors:
- AWS Elasticsearch service has a bug that prevents
calling the rank_eval API without an index. We've raised
a ticket with them but have not been given an ETA on when
this will be fixed.
- The Elasticsearch ruby client has a bug that prevents
specifying an index when using the API. We raised a PR
to fix this (https://github.com/elastic/elasticsearch-ruby/pull/724).

This makes the call specify all indexes '*' and doesn't use the
elasticsearch client library. Once either issue is fixed
we can remove this.

Trello: https://trello.com/c/Uq13lWgG

### DIFF
--- a/lib/debug/rank_eval.rb
+++ b/lib/debug/rank_eval.rb
@@ -1,4 +1,6 @@
 require "csv"
+require "httparty"
+require "json"
 
 module Debug
   class RankEval
@@ -47,10 +49,7 @@ module Debug
         }
       end
 
-      result = @search_config.rank_eval(
-        requests: requests,
-        metric: { dcg: { k: 10 } },
-      )
+      result = rank_eval(requests)
 
       {
         score: result["metric_score"],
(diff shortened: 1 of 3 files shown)


lib/debug/rank_eval.rb
```diff
@@ -74,6 +73,29 @@ def queries
...
+      # AWS doesn't permit us to speak to all indexes with the rank eval api,
+      # and the elasticsearch client has a bug that prevents us calling individual
+      # indexes (https://github.com/elastic/elasticsearch-ruby/pull/698).
+      # This is a fix for that situation.
+      # @search_config.rank_eval(
+      #   requests: requests,
+      #   metric: { dcg: { k: 10 } },
+      # )
+
+      uri = Clusters.default_cluster.uri
```
barrucadu: This is `@search_config.base_uri`

bilbof: Thanks, I made that change

barrucadu (approved):

PR #1749, opened 2019-10-22
Document how to find underperforming queries

laurentqro:
https://trello.com/c/KtVv06e8/1072-document-process-to-find-underperforming-queries


doc/relevancy.md
```diff
@@ -458,6 +459,26 @@ but something else still gets high traffic ([03c4bfa][], May 2015).
...
 dependence on the page views.
 
+### Finding underperforming queries
+
+1. Go to the [datastudio report][]
+
+2. Apply the following filters:
+
+* `click_count >= 100`
+* `median_position = ~4`
```
suganya-s: would be interested to see how you came up with the method, I think anything with a reasonably high click count and a high median is underperforming. I would recommend ignoring the mean most of the time as it can be skewed by one person clicking on result 140 for example.

doc/relevancy.md
```diff
@@ -458,6 +459,26 @@ but something else still gets high traffic ([03c4bfa][], May 2015).
 If we had other signals of content quality we might be able to reduce our
 dependence on the page views.
 
+### Finding underperforming queries
+
+1. Go to the [datastudio report][]
```
suganya-s, line 464: The report does not auto-update so will be for the dates displayed, but you can rerun the BigQuery queries to get new data

doc/relevancy.md
laurentqro: @suganya-s I copied the text from this card: https://trello.com/c/KtVv06e8/1072-document-process-to-find-underperforming-queries.

bilbof (commented):
Nice work!

I'm not sure whether the percentages are accurate in the last step.

For example, for the query UTR we users are mostly clicking on "Find a lost UTR number" in position 7, but that isn't reflected by the chrome extension:

Screenshot 2019-10-23 at 11 40 18

I think it's because the query string isn't considered, so the click through rate is for all searches rather than a particular query.

Perhaps we could replace the final step (checking the chrome extension) with a link to the GA dashboard, and the steps to check how a query is performing from within GA.

E.g. in for the query UTR, but one could change the query string in the URL to something else:

```
https://analytics.google.com/analytics/web/#/report/conversions-ecommerce-product-list-performance/a26179049w50705554p53872948/_u.date00=20191016&_u.date01=20191023&explorer-table.plotKeys=%5B%5D&explorer-table.rowCount=50&explorer-table.secSegmentId=analytics.customDimension71&explorer-table.advFilter=%5B%5B0,"analytics.productListPosition","RE","%5E%5B0-9%5D%7B1,1%7D$",0%5D,%5B0,"analytics.customDimension71","EQ","utr",0%5D%5D&_r.drilldown=analytics.productListName:Search&explorer-table-dataTable.sortColumnName=analytics.productListCTR&explorer-table-dataTable.sortDescending=true&explorer-segmentExplorer.segmentId=analytics.productListPosition/
```

Also, another way for us to find underperforming queries, is to look at the most popular queries users do. Rough estimate of unique queries relative to the percent of searches:

```
Top 10: 6.4%
Top 100: 22.6%
Top 1,000: 51%
Top 10,000: 84%
```

doc/relevancy.md
suganya-s: @bilbof @laurentqro would recommend just looking at high medians and ignoring means

laurentqro: Thanks @suganya-s I'll update 👌

laurentqro: @suganya-s what would you say is a highish median? >4?

suganya-s: yep sounds good

bilbof (approved):

doc/relevancy.md
suganya-s: if you change the `~` to a `>` this is great

suganya-s (approved):

PR #1757, opened 2019-10-25
Ranking evaluation documentation

bilbof:
This documents our efforts to find a 'data based method for judging the performance of changes to the search algorithm' without requiring AB tests.

Trello: https://trello.com/c/OwuWu73I

### DIFF
--- a/doc/rank_evaluation.md
+++ b/doc/rank_evaluation.md
@@ -1,0 +1,137 @@
+# Rank evaluation
+
+One way we assess the affect of search-api changes to relevancy, is by using
+Elasticsearch's [Ranking Evaluation API](ranking_evaluation_api).
+
+The API enables us to score how well search-api ranks results for a given query.
+
+## What is rank evaluation?
+
+Rank evaluation is a process by which we rate how well search-api ranks results for
+a given query.
+
+For example, given the query `Harry Potter` in an ideal situation, the results
+would be returned in this order:
+
+1. Harry Potter
+2. Harry Potter World
+3. Who is Harry Potter?
+
+However, the results might come back in this order:
+
+1. Who is Harry Potter?
+2. Sign in to your HMRC account
+3. Harry Potter
+
+The first situation is good and the second is bad. But what we really
+want is a metric to say *how bad* the results are.
+
+Rank evaluation provides us with a metric which tells us how good or bad
+the results are for a set of queries that we have already manually ranked.
+
+So we have a rake task `debug:rank_evaluation` that does this.
+
+This is useful, because we can tell how good a change is before running an
+AB test on real users.
+
+For example:
+
+```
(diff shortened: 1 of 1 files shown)


doc/rank_evaluation.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+See the Search Relevancy grafana dashboard.
+
+## How do we collect relevancy judgements?
+
+We're working on that at the moment. So far we've manually generated them.
+
+In future we could gather more internally in the organisation, or use end-user
+click data as a signal for relevancy judgements.
```
MatMoore, line 131: Hi @bilbof! I investigated this for my data science accelerator project a couple of years ago and have some code on my github for doing this. I don't know if the team still has the documentation from it, but feel free to hit me up on x-gov slack if you are looking at this again and have any questions. 😄 

It's very cool to see GOV.UK experimenting with search again!

doc/rank_evaluation.md
```diff
@@ -0,0 +1,131 @@
...
+
+By using DCG we make two assumptions:
+
+> - Highly relevant documents are more useful when appearing earlier in a search engine result list (have higher ranks)
+> - Highly relevant documents are more useful than marginally relevant documents, which are in turn more useful than non-relevant documents.
+
+DCG requires a query and a list of relevancy judgements (rated documents).
+
+We manually give a rating between 0 and 4 to documents in search results.
+4 = Totally relevant, 0 being not relevant.
```
barrucadu: It's actually 0 to 3 in our current judgements:

- 0 = irrelevant (which is equivalent to unrated in DCG)
- 1 = misplaced
- 2 = near
- 3 = relevant

bilbof: Thanks @barrucadu, I added this to the doc!

doc/rank_evaluation.md
bilbof: Hi @MatMoore! That would be fantastic, thanks! I'll send you a message. 😄

doc/rank_evaluation.md
```diff
@@ -0,0 +1,137 @@
...
+debug:rank_evaluation[my-local-relevancy-judgements.csv]
+```
+
+We also report the rank evaluation scores to graphite.
+This enables us to plot how relevancy changes over time (overall
+  and for a given query).
+
+This runs every 3 hours in all environments:
+```
+SEND_TO_GRAPHITE=true debug:rank_evaluation
```
Tetrino, line 121: Minor thing, but now you've swapped this elsewhere should it be swapped here?

bilbof: I think the order doesn't matter. I reordered them for clarity in puppet - it was the quotes that were causing the issue there.

Tetrino (approved):

PR #1762, opened 2019-10-29
Get search result CTR from Google Analytics

bilbof:
This adds rake tasks to get click-through-rate (CTR) data on search
queries from Google Analytics, and optionally send these to graphite
with statsd.

These will be displayed in the search relevancy dashboard.

`relevancy:show_overall_ctr` gets the overall CTR on site search (top_1, top_3, top_5).

`relevancy:show_underperfoming_queries` gets the top 1,000 queries and their CTR.

`relevancy:show_top_queries` gets the top 1,000 queries and their view count.

In graphite the stats will be represented as (e.g.):

`relevancy.query.top_3_ctr`
`relevancy.query.universal_credit.viewcount`
`relevancy.query.universal_credit.top_1_ctr`

Added to dashboard here: https://github.com/alphagov/govuk-puppet/pull/9756.

What the result is:

ctr

Trello: https://trello.com/c/VY85d6Ks/1069

### DIFF
--- a/lib/analytics/load_service.rb
+++ b/lib/analytics/load_service.rb
@@ -7,7 +7,7 @@ module Analytics
     include Google::Apis::AnalyticsV3
     include Google::Auth
 
-    SCOPES = ["https://www.googleapis.com/auth/analytics.edit"].freeze
+    SCOPES = [AUTH_ANALYTICS_EDIT].freeze
 
     attr_reader :service
 
--- a/lib/analytics/overall_ctr.rb
+++ b/lib/analytics/overall_ctr.rb
@@ -1,0 +1,83 @@
+require "analytics/reporting_requester"
+
+module Analytics
+  class OverallCTR
+    include ReportRequester
+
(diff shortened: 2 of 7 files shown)


lib/analytics/load_service.rb
```diff
@@ -7,7 +7,7 @@ class LoadService
     include Google::Apis::AnalyticsV3
     include Google::Auth
 
-    SCOPES = ["https://www.googleapis.com/auth/analytics.edit"].freeze
+    SCOPES = [AUTH_ANALYTICS_EDIT].freeze
```
barrucadu, line 10: Where is this constant defined?

bilbof: It's in the `Google::Apis::AnalyticsV3` module:

https://github.com/googleapis/google-api-ruby-client/blob/master/generated/google/apis/analytics_v3.rb#L34

https://www.rubydoc.info/github/google/google-api-ruby-client/master/Google/Apis/AnalyticsV3/AnalyticsService

barrucadu (approved):

PR #1764, opened 2019-10-31
Make "post" and "postal" stem the same

barrucadu:
We noticed that, if you search for "postal vote", the word "post" in the description doesn't get highlighted.


bilbof (approved):
Excellent

PR #1768, opened 2019-11-04
Learning To Rank

bilbof:
Paired with @barrucadu on this

https://trello.com/c/LcKC6rZR/1125-learning-to-rank-mvp

This PR implements [Learning To Rank](https://en.wikipedia.org/wiki/Learning_to_rank) using Tensorflow’s [ranking module](https://github.com/tensorflow/ranking) as described in https://github.com/alphagov/search-api/pull/1778. See [the ADR](https://github.com/alphagov/search-api/pull/1778) for more details.

End users will not see a difference as a result of this change.

This adds an AB test `relevance:B` which uses a pre-trained model to rerank search results. This also adds rake tasks one can use to: generate training data, train a model with the data, and serve and evaluate a trained model.

```
rake learn_to_rank:generate_relevancy_judgements                                # Export a CSV of relevancy judgements generated from CTR on popular queries
rake learn_to_rank:generate_training_dataset[judgements_filepath]               # Export a CSV of SVM-formatted relevancy judgements for training a model
rake learn_to_rank:reranker:evaluate[relevancy_judgements]                      # Evaluate search performance using nDCG with and without the model
rake learn_to_rank:reranker:serve[model_dir]                                    # Serves a trained model
rake learn_to_rank:reranker:train[svm_dir,model_dir]                            # Train a reranker model with relevancy judgements
```

The new rake tasks and AB test are disabled by default. They require the environment variable `ENABLE_LTR` to be set.

The rake tasks calls classes in the new `LearnToRank` module and scripts in the `ltr_scripts` directory.

The next steps will be to be improve the model and data pipeline, and make the reranker production ready. See the LTR Epic on the search Trello board for more details of future work.


lib/learn_to_rank/ctr_to_judgements.rb
```diff
@@ -0,0 +1,67 @@
...
+        }
+        .flatten
+        .compact
+    end
+
+  private
+
+    attr_reader :ctrs
+
+    def ctr_to_relevancy_score(query_ctrs)
```
Tetrino, line 34: Did the refinement I suggest end up not working out? I could well believe it!

barrucadu:
> Hand-tuning relevance doesn't work very well because field boosts are complex, and difficult to maintain.

I think a more important point is that field boosts might have dependencies: some document types might benefit from a stronger title boost than other document types, for instance.  Expressing that just with Elasticsearch would be prohibitively complex, and be almost impossible to maintain.

It doesn't matter how complex the *ideal* scoring function is, as it's managed by a black box we don't really need to understand.  What this approach gains us is only needing to be *not terrible* with the scoring function our Elasticsearch query gives rise to.

lib/learn_to_rank/reranker.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  private
+
+    DEFAULT_COUNT = 20
+    MAX_MODEL_BOOST = 5
+
+    def reorder_results(search_results, new_scores)
+      search_results
+        .map
+        .with_index { |result, index|
```
barrucadu, line 22: As a minor point, I think `begin` / `end` is clearer when the block is multiple lines.

lib/learn_to_rank/reranker.rb
```diff
@@ -0,0 +1,35 @@
...
+          m_score = [new_scores[index], MAX_MODEL_BOOST].min
+          es_score = result.fetch("_score", 0)
+          result.merge(
+            "model_score" => m_score,
+            "original_rank" => index + 1,
+            "combined_score" => m_score * es_score,
```
barrucadu, line 28: I'm not certain about multiplying the scores.  Does the tensorflow model know it's producing a multiplicative factor, or does it think it's producing entirely new scores?

lib/learn_to_rank/reranker.rb
```diff
@@ -0,0 +1,35 @@
...
+          )
+        }
+        .sort_by { |res| res["combined_score"] }
+        .reverse
```
barrucadu: Rather than doing two traversals (I don't think Ruby is smart enough to fuse them?), sort by `-res["combined_score"]`

lib/search/presenters/result_presenter.rb
```diff
@@ -87,9 +87,14 @@ def add_debug_values(result)
...
       result[:es_score] = raw_result["_score"]
+      # LearnToRank generated values
+      result[:model_score] = raw_result["model_score"]
+      result[:original_rank] = raw_result["original_rank"]
+      result[:combined_score] = raw_result["combined_score"]
+
       result[:_id] = raw_result["_id"]
 
-      if raw_result["_explanation"]
+      if raw_result["_explanation"] && search_params.debug[:explain]
```
barrucadu, line 96: Why is this `&& ...` bit needed?

lib/search/query.rb
```diff
@@ -1,3 +1,5 @@
+require "learn_to_rank/reranker"
```
barrucadu: I think all the other `require`s are in `lib/rummager.rb`

lib/learn_to_rank/reranker.rb
bilbof: The model assumes it is producing new scores (not something to be multiplied). This was chosen so we could treat the model as a booster while we're not certain about its results.

I agree that it shouldn't be a multiplicative factor. We should just the new scores.

It has one benefit though, that best bets continue to work. But I think we should replace the best bet process with generating new relevancy judgements for a query.

lib/search/query.rb
```diff
@@ -77,6 +80,15 @@ def process_elasticsearch_errors
       end
     end
 
+    def rerank(es_response, search_params)
```
barrucadu, line 81: There should be a timer metric on this

lib/search/query_builder.rb
```diff
@@ -23,7 +23,7 @@ def payload
         sort: sort,
         aggs: aggregates,
         highlight: highlight,
-        explain: search_params.debug[:explain],
+        explain: true,
```
barrucadu: Ah, I see, this is needed to get the features the model needs

lib/search/presenters/result_presenter.rb
bilbof: We don't want to return explanation to end users unless they've requested them in their search params. However, the model currently requires the scores within the explanation, so we always need to extract it from ES (potential performance problem there to address).

lib/search/query.rb
bilbof: Yes very good idea.

ltr_scripts/tf_ranking_libsvm.py
```diff
@@ -0,0 +1,466 @@
+# Copyright 2019 The TensorFlow Ranking Authors.
```
barrucadu, line 1: Rather than include this file verbatim, would it be better to pull it in as a git submodule or something?

lib/learn_to_rank/ctr_to_judgements.rb
bilbof: Have you tried it in your PR? https://github.com/alphagov/search-api/pull/1775

lib/learn_to_rank/ranker.rb
```diff
@@ -0,0 +1,33 @@
...
+    def ranks
+      fetch_new_scores(feature_sets).fetch('results', [])
+    end
+
+  private
+
+    attr_reader :feature_sets
+
+    def fetch_new_scores(examples)
+      url = "http://172.17.0.3:8501/v1/models/ltr:regress"
```
barrucadu: I don't think it's guaranteed that docker containers will always use the same subnet, so it would be better if the container mapped a port from localhost

lib/learn_to_rank/judgements_to_svm.rb
```diff
@@ -0,0 +1,34 @@
+module LearnToRank
+  class JudgementsToSvm
```
barrucadu, line 2: It would be great if this linked to some documentation on the SVM format and what it's used for - is this a tensorflow-specific thing, or is it wider than that?

lib/learn_to_rank/features.rb
```diff
@@ -0,0 +1,31 @@
...
+module LearnToRank
+  class Features
+    # Features takes some values and translates them to features
+    def initialize(explain:, popularity: 0, es_score: 0)
+      @popularity = popularity
+      @es_score = es_score
+      @explain_scores = LearnToRank::ExplainScores.new(explain)
+    end
+
+    def as_hash
```
barrucadu, line 12: This should probably be an array rather than a hash if we're using numeric keys

lib/learn_to_rank/explain_scores.rb
```diff
@@ -0,0 +1,54 @@
...
+      description = explain["description"]
+      if description.include? "PerFieldSimilarity"
+        # is a bm25 score so return that value
+        field = (description.split("(")[1].split(":")[0].split(".")[0] + "_score").downcase.to_sym
+        return {} unless field_permitted? field
+
+        value = explain["value"]
+        { field => value }
+      elsif ["sum of:", "max of:", "function score, product of:"].include? description
+        # is not a bm25 score so return the sum of scores contained within
```
barrucadu, line 36: Why the sum of the scores within and not the aggregate score Elasticsearch already gives us? (sum, max, etc)

lib/learn_to_rank/embed_features.rb
```diff
@@ -0,0 +1,92 @@
...
+      flush_cached_queries
+      res
+    end
+
+  private
+
+    attr_reader :judgements
+
+    def features(judgement)
+      return unless doc = fetch_document(judgement)
```
barrucadu: Nitpicky: I think using assignment as an expression is confusing

lib/learn_to_rank/ctr_to_judgements.rb
```diff
@@ -0,0 +1,67 @@
...
+          2
+        else
+          3
+        end
+      end
+    end
+
+    def fetch_results(query)
+      q = query.gsub("_", "%20")
+      url = "https://www.gov.uk/api/search.json?fields=link&q=#{q}"
```
barrucadu: This should be able to use any instance of search-api

lib/analytics/total_query_ctr.rb
```diff
@@ -0,0 +1,119 @@
...
+              ],
+            ),
+          ],
+          order_bys: [
+            OrderBy.new(
+              field_name: "ga:productListPosition",
+              sort_order: "ASCENDING",
+            ),
+          ],
+        )
```
barrucadu, line 112: 🙀

lib/learn_to_rank/reranker.rb
barrucadu: I'm not sure the new ranks we're getting are particularly valid: new scores and boosting factors to existing scores are very different things.

We can keep best bets working by using the original score if it's greater than whatever the smallest best bet score is.

lib/learn_to_rank/ctr_to_judgements.rb
```diff
@@ -0,0 +1,67 @@
+module LearnToRank
+  class CtrToJudgements
```
barrucadu, line 2: Is this based on the "Unbiased Learning-to-Rank from biased feedback data." mentioned in the https://github.com/tensorflow/ranking README?  If not, why not?

bilbof: No it was a hack to get scores from CTRs. The reason was for experimentation - to see how the model would react to some user-generated data (even bad data as this is).

My plan is to get this PR into a mergeable state (e.g. robust, tested, but with hacks like this). We may eventually remove this class entirely, and replace it with the usage of a click model or something else. We need to get some reliable user generated ratings we can use for training the model. But I think having something at this level of hackiness as a starting point will be useful.

lib/learn_to_rank/explain_scores.rb
bilbof: We want the summed scores for each field e.g. title, description. Without going into "function score, product of" or a "sum of: " we don't know whether the value is for a filter / different field. If we did use `sum of:` we might get the fields combined (e.g. title and description) but I think there is value in having the title score and description score as separate features. However, this is quite a hack which may even be broken. It'd be nice to prove that it works as expected and do some work to make it faster.

E.g. https://www.gov.uk/api/search.json?q=national%20insurance&fields=title&debug=explain&count=1

Tetrino (approved):

PR #1775, opened 2019-11-06
[Spike] Learn to Rank Refinements

Tetrino:
The current spike is a large step towards using LTR values to generate
relevancy score. However the scoring itself is not particularly
precise and ignores situations wherein we have very high click through
rates, or very low max click through rates on given searches.

This pull request attempts to solve these problems by:

- Setting any result with an excessively high CTR to the highest
relevancy score automatically, ignoring any further calculation.

- Creating a benchmark value from the highest non-excessive CTR to which
all further rates are compared.

- Using this benchmark to judge relevancy scores for further options,
based on its value as a percentage of the benchmark. The percentages can
be further tuned as needed to refine our resultant scores.


lib/learn_to_rank/ctr_to_judgements.rb
```diff
@@ -29,15 +29,34 @@ def relevancy_judgements
...
+        # benchmark. Allows fine tuning of boundaries and takes into account
+        # weak searches where highest judgment is k < 2. Risks flattening the
+        # results if CTR calculation results are close.
+        h[position] = if k < (benchmark / 100) * 30
           0
-        elsif k < 1
+        elsif k < (benchmark / 100) * 60
           1
-        elsif k < 2
+        elsif k < (benchmark / 100) * 90
```
bilbof: I think you'd need `100.0` to be a float. Otherwise you might lose some precision, as `(4 / 100.0) * 60 = 2.4` and `(4 / 100) * 60 = 0`

bilbof (commented):
Looks great! One small comment.

Did you already try it with the CTR dataset?

lib/learn_to_rank/ctr_to_judgements.rb
Tetrino: Yeah I misremembered how ruby handles types. Fixed shortly.

barrucadu:
@Tetrino I'll close this because the target branch has been merged and this is now pretty out of date.

PR #1778, opened 2019-11-07
ADR for Learning To Rank

bilbof:
This is our Architecture Decision Record for Learning To Rank.

Rendered version: https://github.com/alphagov/search-api/blob/ad7c7841224e150b352889e2cf6617a06c4a4033/doc/arch/adr-010-learn-to-rank.md

https://trello.com/c/jdwVqvmg/1121-adr-for-learning-to-rank

### DIFF
--- a/doc/arch/adr-010-learn-to-rank.md
+++ b/doc/arch/adr-010-learn-to-rank.md
@@ -79,7 +79,7 @@ harry potter  3         Harry Potter
 However, after doing this, we would train a model with this data
 to rank the results better. The model optimises for nDCG.
 
-We train the model using using a combination of relevancy judgements
+We train the model using a combination of relevancy judgements
 and features: numeric measures which indicate how well we think the
 result matches the query.
 
@@ -95,12 +95,16 @@ harry potter  3         Harry Potter  0.2         1234        1234     1234     
 ...
 ```
 
-For example, take the SVM dataset below for the query 'harry potter' (qid:1)
+For example, take the training dataset below for the query 'harry potter' (qid:1)
 
 ```
 0 qid:1 popularity:0.3 title_score:0.2 description_score:0.3 recency:0.9 pagerank:5 ...
 3 qid:1 popularity:0.1 title_score:0.6 description_score:0.7 recency:0.5 pagerank:2 ...
 ```
+
+> **Note**: This format is frequently used for Learning To Rank training
+> datasets, including the TensorFlow ranking library.
+> [Details on the format](https://sourceforge.net/p/lemur/wiki/RankLib%20File%20Format).
 
 These graded query:document pairs and their features can be used
 to train a model.
@@ -216,72 +220,19 @@ custom machine learning code, as PyTorch don't currently have a ranking library.
 [TensorFlow's ranking library]: https://github.com/tensorflow/ranking
 [elasticsearch-learning-to-rank]: https://github.com/o19s/elasticsearch-learning-to-rank
 
-### Language
+### Further decisions
 
-We would like to keep the code for this project in Ruby as much as
-possible. Search API is written in Ruby and it is the primary language
-used at GOV.UK. Necessarily some machine learning code will need to be
-written in Python because TensorFlow is a Python library.
+We have further decisions to make around the following areas, which may
+be the subject of further ADRs:
 
-### Dockerised model
-
-We will use TensorFlow Serving to make a re-ranker model available
-to Search API on search machines.
-
-We will run a docker container (containing our pre-trained model) on
-search machines to which Search API can send requests.
-
-Should the docker container fail, Search API should be resilient to this,
-and fall back to the default ranking algorithm, with results not
-re-ranked by the model.
-
-### Training
-
-**Training the model** should happen on search machines on a schedule.
-An instance of Search API will train a model using the latest relevancy
-judgements stored in S3, and will then upload a new version of the model to S3.
-On a schedule all search machines will pull the latest version of the model
-and push the new version to the locally running TF Serving container.
-
-**Training data** is made up of relevancy judgements and document features.
-The judgements may be sourced from user behaviour (implicit signals
-of relevancy) perhaps with a Click Model; so far we have used explicit
-manual judgements.
-
-Acquiring relevancy judgements is a work in progress. Our next project
-will be to create a process by which we can acquire relevancy judgements
-on a rolling basis automatically.
-
-### Re-ranking requests
-
-**Using the model** to re-rank results will involve a user searching,
-Search API will perform the query for the top-k results (say 100),
-next Search API will make a request for new scores from the re-ranker
-for the results, finally Search API will provide the re-ranked top-n
-results to the user.
-
-We may see greater search query latency when using the model. Search
-API will need to pull more data out of elasticsearch and do more
-computation, in addition to making a request to the re-ranker.
-This is something we will monitor, as we don't want to provide a much
-slower search experience at a cost of improving relevance.
-
-### Monitoring
-
-We are monitoring nDCG and click-through rates in a search relevancy
-dashboard in grafana.
-
-We will set up more monitoring and alerts as we get further into the project.
-
-### Testing the model
-
-We hope to re-use our existing offline nDCG tests and online AB testing
-process for Learning To Rank.
-
-AB testing models should be possible as TF Serving supports
-versioning models and can serve multiple models from one container.
-We plan to use AB testing to validate that the model provides better
-results to users (measured by click-through rate).
+- language: divide between ruby and python
+- hosting the trained model (dockerised?)
+- training data pipeline architecture
+- where does compute-intensive training happen?
+- failure resilience (what happens when reranker process fails? should we backup datasets, models?)
+- monitoring and alerts
+- testing/validating the model
+- how would LTR change local development?
 
 ## What success will look like
 
--- a/doc/arch/adr-010-learn-to-rank.md
+++ b/doc/arch/adr-010-learn-to-rank.md
@@ -1,0 +1,297 @@
+# Decision record: Learning To Rank
+
+**Date:** 2019-11-07
+
+The search team have decided to implement [Learning To Rank][].
+
+1. What is Learning To Rank?
+1. Motivation
+1. Implementation
+1. What success will look like
+
+[Learning To Rank]: https://en.wikipedia.org/wiki/Learning_to_rank
+
+## What is Learning To Rank?
+
+Learning To Rank (LTR) is the application of machine learning to create a
+ranking model that can be used to improve search relevance.
+
+This is a common approach for advanced search applications.
+
+The ranking of search results on websites is frequently determined by
+machine learning models. For example, Airbnb, Yelp, Wikipedia, Bloomberg,
+and commercial search engines.
+
+There is a good explanation on what LTR is here:
+
+- https://elasticsearch-learning-to-rank.readthedocs.io/en/latest/core-concepts.html
+- https://opensourceconnections.com/blog/2017/02/24/what-is-learning-to-rank/
+
+### How does Learning To Rank work?
+
+It's easiest to explain with an example:
+
+Say you searched for 'Harry Potter'. Search API might return the following results:
+
+1. HMRC Sign in
+2. Harry Potter World
+3. Pottery barn
+4. Harry Potter
+
+These are not good results.
+
+To assess search quality, we can assign relevancy judgements to the
+results of a query in the range 0-3, with 0 being irrelevant and 3
+being perfect. For example, for these results:
+
+```
+query         rating    document
+harry potter  0         HMRC Sign in
+harry potter  2         Harry Potter World
+harry potter  0         Pottery barn
+harry potter  3         Harry Potter
+```
+
+We can then use a ranking quality metric, such as [normalised
+discounted cumulative gain (nDCG)](https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Normalized_DCG) to assign a score to
+this list of results.
+
+We can then use nDCG to tell us how good the results are.
+We might get a score of 0.3 for this result set (from 0 to 1).
+
+We could then investigate why the results are bad. Perhaps there is a
+best bet. Or popularity might be pushing HMRC docs up. We can fix these
+things, then check the new nDCG score, and we might get 0.7 for this query.
+
+In contrast, with Learning To Rank, search result ranks would be
+determined by a machine learning model without manually changing
+the boost of a field.
+
+With Learning To Rank we would go through the same process of
+rating the search results.
+
+```
+harry potter  0         Pottery barn
+harry potter  3         Harry Potter
+...
+```
+
+However, after doing this, we would train a model with this data
+to rank the results better. The model optimises for nDCG.
+
+We train the model using using a combination of relevancy judgements
+and features: numeric measures which indicate how well we think the
+result matches the query.
+
+The text similarity scores Elasticsearch assigns to fields in the
+document are examples of features. The model tries to find correlations
+between document features and relevancy judgements, which allows it to
+predict how to order documents we don't have relevancy judgements for.
+
+```
+query         rating    document      title_score view_count  recency  pagerank ...
+harry potter  0         Pottery barn  0.2         9876        0.12     9876     ...
+harry potter  3         Harry Potter  0.2         1234        1234     1234     ...
+...
+```
+
+For example, take the SVM dataset below for the query 'harry potter' (qid:1)
+
+```
+0 qid:1 popularity:0.3 title_score:0.2 description_score:0.3 recency:0.9 pagerank:5 ...
+3 qid:1 popularity:0.1 title_score:0.6 description_score:0.7 recency:0.5 pagerank:2 ...
+```
+
+These graded query:document pairs and their features can be used
+to train a model.
+
+The model might learn that title and description scores are slightly
+more useful that popularity and recency, and how much, in certain
+circumstances.
+
+Rather than manually tuning boosts, a model learns from relevancy
+judgements the role that features should play when providing relevant results.
+
+This is a high level overview of what Learning To Rank is for our use case.
+There has been a lot of research in this area, and within the fields of
+information retrieval & machine learning which Learning To Rank intersects.
+There is a lot more out there to learn!
+
+For more details about how we plan to implement Learning To Rank, see
+the implementation section below.
+
+## Motivation
+
+We have decided to try Learning To Rank because it has the potential
+to greatly improve search relevance in a way that is:
+
+- automatable
+- scalable
+- more maintainable
+- more successful
+
+**Automatable** because the model won't require human involvement to
+tune field boosts on the latest user data
+
+**Scalable** because automation; but also because the model should
+provide better results with more data.
+
+**Maintainable** because relevancy tuning can be automated to not become
+stale.
+
+**Successful** because results should achieve higher nDCG scores and
+click-through rates than we can achieve through manual tuning alone.
+
+### The problem
+
+Manual relevancy tuning can work, but doesn't scale well. Eventually
+the process becomes a game of whack-a-mole. If we downgrade popularity
+boosting, then this query might have better results, but other queries
+might have worse results.
+
+Hand-tuning relevance doesn't work very well because field boosts
+are complex, and difficult to maintain.
+
+Moreover field boosts might have dependencies: some document types might
+benefit from a stronger title boost than other document types, for
+instance. Expressing that just with Elasticsearch would be prohibitively
+complex, and be almost impossible to maintain.
+
+### Validating Learning To Rank
+
+We assumed that we would be able to get better results using Learning
+To Rank. To validate this assumption, we worked on a spike.
+
+In a [validation spike][] we found that we could build a model that provided
+a better nDCG score than our current search algorithm for the queries we
+have manually gathered relevancy judgements for.
+
+This validated for us that Learning To Rank is feasible and could
+lead to more relevant results for users.
+
+Though Learning To Rank won't make relevance tuning easier, it could
+make search results more relevant.
+
+### Final decision
+
+We feel confident about implementing Learning To Rank (LTR) for the
+following reasons:
+
+- We think that it will improve search relevancy for citizens, enabling
+  them to find what they need on GOV.UK
+- We have validated our assumptions with a spike
+- LTR is a well trodden path, and other organisations have written about
+  their implementations with varying levels of openness (e.g. Wikimedia)
+- GOV.UK has implemented other machine learning projects with success
+  (related links, topic taxonomy), and we have a data science team which
+  the search team has already worked with
+- TensorFlow provides an off-the-shelf ranking module, which solves
+  some of the machine learning problems of LTR for us
+- Manual relevancy tuning is not scalable, and won't get us to the same
+  level of relevancy as LTR
+
+## Implementation plan
+
+We have decided to implement Learning To Rank by following on from
+the work started in the [validation spike][].
+
+As we are at the beginning of the project, this section may not
+accurately reflect the final implementation.
+
+[validation spike]: https://github.com/alphagov/search-api/pull/1768
+
+### Machine learning library
+
+We have decided to use [TensorFlow's ranking library][] for our
+implementation of Learning To Rank.
+
+TensorFlow is well supported, and the ranking library is under active development.
+
+We considered using the [elasticsearch-learning-to-rank][] plugin.
+However, AWS Elasticsearch does not permit us to install custom plugins.
+
+We also considered PyTorch. However, this would require writing much more
+custom machine learning code, as PyTorch don't currently have a ranking library.
+
+[TensorFlow's ranking library]: https://github.com/tensorflow/ranking
+[elasticsearch-learning-to-rank]: https://github.com/o19s/elasticsearch-learning-to-rank
+
+### Language
+
+We would like to keep the code for this project in Ruby as much as
+possible. Search API is written in Ruby and it is the primary language
+used at GOV.UK. Necessarily some machine learning code will need to be
+written in Python because TensorFlow is a Python library.
+
+### Dockerised model
+
+We will use TensorFlow Serving to make a re-ranker model available
+to Search API on search machines.
+
+We will run a docker container (containing our pre-trained model) on
+search machines to which Search API can send requests.
+
+Should the docker container fail, Search API should be resilient to this,
+and fall back to the default ranking algorithm, with results not
+re-ranked by the model.
+
+### Training
+
+**Training the model** should happen on search machines on a schedule.
+An instance of Search API will train a model using the latest relevancy
+judgements stored in S3, and will then upload a new version of the model to S3.
+On a schedule all search machines will pull the latest version of the model
+and push the new version to the locally running TF Serving container.
+
+**Training data** is made up of relevancy judgements and document features.
+The judgements may be sourced from user behaviour (implicit signals
+of relevancy) perhaps with a Click Model; so far we have used explicit
+manual judgements.
+
+Acquiring relevancy judgements is a work in progress. Our next project
+will be to create a process by which we can acquire relevancy judgements
+on a rolling basis automatically.
+
+### Re-ranking requests
+
+**Using the model** to re-rank results will involve a user searching,
+Search API will perform the query for the top-k results (say 100),
+next Search API will make a request for new scores from the re-ranker
+for the results, finally Search API will provide the re-ranked top-n
+results to the user.
+
+We may see greater search query latency when using the model. Search
+API will need to pull more data out of elasticsearch and do more
+computation, in addition to making a request to the re-ranker.
+This is something we will monitor, as we don't want to provide a much
+slower search experience at a cost of improving relevance.
+
+### Monitoring
+
+We are monitoring nDCG and click-through rates in a search relevancy
+dashboard in grafana.
+
+We will set up more monitoring and alerts as we get further into the project.
+
+### Testing the model
+
+We hope to re-use our existing offline nDCG tests and online AB testing
+process for Learning To Rank.
+
+AB testing models should be possible as TF Serving supports
+versioning models and can serve multiple models from one container.
+We plan to use AB testing to validate that the model provides better
+results to users (measured by click-through rate).
+
+## What success will look like
+
+Success will look like:
+
+- Click-through rate on the top 3 and top 5 results will go up
+- nDCG will go up
+- Search latency will not increase much
+
+Before we can deploy the model it must provide a better CTR and nDCG.
+
+A further hope for this project is that we will be able to remove
+some complexity by deleting the boosting code in Search API.


laurentqro (approved):

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+manual judgements.
+
+Acquiring relevancy judgements is a work in progress. Our next project
+will be to create a process by which we can acquire relevancy judgements
+on a rolling basis automatically.
+
+### Re-ranking requests
+
+**Using the model** to re-rank results will involve a user searching,
+Search API will perform the query for the top-k results (say 100),
```
sihugh: We'll have to consider performance here and do appropriate testing so get confidence that we'll be able to support the extra load. I wonder if there's a way of measuring what the lowest ranked results that appear are, and adjust the window appropriately

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+next Search API will make a request for new scores from the re-ranker
+for the results, finally Search API will provide the re-ranked top-n
+results to the user.
+
+We may see greater search query latency when using the model. Search
+API will need to pull more data out of elasticsearch and do more
+computation, in addition to making a request to the re-ranker.
+This is something we will monitor, as we don't want to provide a much
+slower search experience at a cost of improving relevance.
```
sihugh: Yes! I'm wondering about trying to introduce performance budgets.  We could mitigate some backend slowdown by improving the experience elsewhere - for example by rendering less on the page, or doing some optimisation of assets.  It's important that we can keep a track of this though

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+Success will look like:
+
+- Click-through rate on the top 3 and top 5 results will go up
+- nDCG will go up
+- Search latency will not increase much
+
+Before we can deploy the model it must provide a better CTR and nDCG.
+
+A further hope for this project is that we will be able to remove
+some complexity by deleting the boosting code in Search API.
```
sihugh, line 248: A reduction in magic numbers that are hard to change would be a good thing.

sihugh (commented):
This is really good Bill.  I'm going to get wider feedback though, as we're doing some slightly unusual things (such as running docker containers on search boxes).  I wonder how this fits with the future tech-ops hosting strategy 🤔

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+
+We have decided to try Learning To Rank because it has the potential
+to greatly improve search relevance in a way that is:
+
+- automatable
+- scalable
+- more maintainable
+- more successful
+
+**Automatable** because the model won't require human involvement to
```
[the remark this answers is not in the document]
sihugh: (And figuring out how to react to them)

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+
+Should the docker container fail, Search API should be resilient to this,
+and fall back to the default ranking algorithm, with results not
+re-ranked by the model.
+
+### Training
+
+**Training the model** should happen on search machines on a schedule.
+An instance of Search API will train a model using the latest relevancy
+judgements stored in S3, and will then upload a new version of the model to S3.
```
sihugh: @karlbaker02 's approach was I think to spin up a temporary EC2 instance for generating models using a concourse pipeline.  Is this solving a different problem?

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+We will run a docker container (containing our pre-trained model) on
+search machines to which Search API can send requests.
+
+Should the docker container fail, Search API should be resilient to this,
+and fall back to the default ranking algorithm, with results not
+re-ranked by the model.
+
+### Training
+
+**Training the model** should happen on search machines on a schedule.
```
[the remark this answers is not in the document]
bilbof: > we fall back to non-LTR enhanced search until we've generated a new one?

Yes I thought this would work. It'll be degraded, but still will return results. Generating a new model should be straightforward provided we have backups of generated relevancy judgements / training data (or a way to regenerate those, too). Right now, our relevancy judgements are already stored in an S3 bucket and pulled by Search API for the rank evaluation job.

My thinking was that one machine would generate the model file, and upload it to S3. Other machines would then pull the latest version of the model from S3.

If we want additional redundancy we could backup the S3 bucket.

With the models in the S3 bucket we can have model versions available to pull without waiting for re-training, and could roll back to different models quickly.

TensorFlow Serving can serve multiple models from one container. They also have some versioning built in for switching over to new model versions (https://www.tensorflow.org/tfx/serving/architecture#version_policy).

doc/arch/adr-010-learn-to-rank.md
bilbof: That's interesting. Perhaps we could do some analyses of the es_score@k to work out what the top-k size should be, or figure out a way to do it at query time. I added a task to the LTR epic for this.

There's a section on improving reranking performance in the second half of the Bloomberg talk on LTR: https://youtu.be/eMuepJpjUjI?t=1235.

doc/arch/adr-010-learn-to-rank.md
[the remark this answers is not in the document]
bilbof: That's a good point, I think you're right. I've added a task to the epic to investigate this.

doc/arch/adr-010-learn-to-rank.md
bilbof: Yes definitely. It's nice that the performance bottlenecks seem to be within our control.

bilbof: Do you have any thoughts on the performance budgets we could place on Search API response time prior to implementing LTR?

doc/arch/adr-010-learn-to-rank.md
bilbof: Very good point. It might be nice to do have a separate instance, not responsible for responding to user requests, do the training.

It might be that the model doesn't need a long training time (e.g. 10 minutes) so doing that out-of-hours once a day won't require a specialised machine. On the other hand if it'll be a multi-hour highly compute intensive job I can see the value of using specialised instance.

At the moment I am not sure about the compute / time requirements of training the model, so we could delay this decision. What do you think?

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+query         rating    document
+harry potter  0         HMRC Sign in
+harry potter  2         Harry Potter World
+harry potter  0         Pottery barn
+harry potter  3         Harry Potter
+```
+
+These judgements are an expression of the ideal ordering.
+
+We can then use nDCG to tell us how good the results are.
```
barrucadu, line 59: It's not clear what nDCG is here, say something like

> We can then use a ranking quality metric, such as [normalised discounted cumulative gain (nDCG)](https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Normalized_DCG) to assign a score to this list of results.

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+1. HMRC Sign in
+2. Harry Potter World
+3. Pottery barn
+4. Harry Potter
+
+These are not good results.
+
+Our approach now (prior to implementing Learning To Rank) would be to
+manually rank these results, giving them a score of between 0-3 with
+0 being irrelevant and 3 being completely relevant:
```
barrucadu: This sounds like we determine the order of search results by manually ranking them.  Say something like

> To assess search quality, we can assign *relevancy judgements* to the results of a query in the range 0-3, with 0 being irrelevant and 3 being perfect.  For example, for these results:

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+
+```
+query         rating    document
+harry potter  0         HMRC Sign in
+harry potter  2         Harry Potter World
+harry potter  0         Pottery barn
+harry potter  3         Harry Potter
+```
+
+These judgements are an expression of the ideal ordering.
```
barrucadu: I think remove this.

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+
+We can then use nDCG to tell us how good the results are.
+We might get a score of 0.3 for this result set (from 0 to 1).
+
+We could then investigate why the results are bad. Perhaps there is a
+best bet. Or popularity might be pushing HMRC docs. We can fix these
```
barrucadu: Should be

> pushing HMRC docs up

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+things, then re-run nDCG, and we might get 0.7 for this query.
```
barrucadu: Change

> then re-run nDCG

To

> then check the new nDCG score

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+
+In contrast, with Learning To Rank, search result ranks would be
+determined by a model without manually changing the boost of a field.
```
barrucadu: Change to

>by a machine learning model

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+```
+harry potter  0         Pottery barn
+harry potter  3         Harry Potter
+...
+```
+
+However, after doing this, we would train a model with this data
+to rank the results better. The model optimises for nDCG.
+
+We provide the model with features, such as the fields we boost on.
```
barrucadu: Add more explanation of how features are used:

> We train the model using using a combination of relevancy judgements and *features*: numeric measures which indicate how well we think the result matches the query. The text similarity scores Elasticsearch assigns to fields in the document are examples of features.  The model tries to find correlations between document features and relevancy judgements, which allows it to predict how to order documents we *don't* have relevancy judgements for.

barrucadu: I see that this is explained a little further down, but I think it's good to put at least some of the explanation up front so the reader isn't left wondering what the point is.

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+instance. Expressing that just with Elasticsearch would be prohibitively
+complex, and be almost impossible to maintain.
+
+### Validating Learning To Rank
+
+We assumed that we would be able to get better results using Learning
+To Rank. To validate this assumption, we worked on a spike.
+
+In a [validation spike][] we found that we could build a model that provided
+a better nDCG score than our current search algorithm.
```
barrucadu: Add

> for the queries we have manually gathered relevancy judgements for.

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,286 @@
...
+accurately reflect the final implementation.
+
+[validation spike]: https://github.com/alphagov/search-api/pull/1768
+
+### Language
+
+We would like to keep the code for this project in Ruby as much as
+possible. Search API is written in Ruby and it is the primary language
+used at GOV.UK. Necessarily some machine learning code will need to be
+written in Python.
```
barrucadu: I think move this section below "Machine learning library" and add

> because TensorFlow is a Python library

bilbof:
Thanks @barrucadu I made those changes.

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+the work started in the [validation spike][].
+
+As we are at the beginning of the project, this section may not
+accurately reflect the final implementation.
+
+[validation spike]: https://github.com/alphagov/search-api/pull/1768
+
+### Machine learning library
+
+We have decided to [TensorFlow's ranking library][] for our
```
hannako: We have decided to use...

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+[Learning To Rank]: https://en.wikipedia.org/wiki/Learning_to_rank
+
+## What is Learning To Rank?
+
+Learning To Rank (LTR) is the application of machine learning to create a
+ranking model that can be used to improve search relevance.
+
+This is a common approach for advanced search applications.
+
+The ranking of search results on websites are frequently determined by
```
hannako: ...is frequently...

bilbof:
Thanks @hannako, I made those changes.

doc/arch/adr-010-learn-to-rank.md
sophiebrown01: Our model for related links originally took around 9hrs to complete, which has been reduced in half since through various improvements; this, combined with the need for a different instance type was the main reason we opted for a temporary EC2 instance managed via Concourse. Of course, this was a new project so there wasn't an existing box which might have made sense to put this on, something you have in this case as it's directly related to search.

If the model doesn't take long to generate then it could be run on an existing machine (maybe out-of-hours to make sure we don't compromise the machine's resources when search might be being used more)? 

Another option you could consider is AWS Lambda, so long as your process completes within 15 minutes (the max running time allowed).

doc/arch/adr-010-learn-to-rank.md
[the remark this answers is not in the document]
sophiebrown01: Is it fair to say that automatable is directly relevant to accuracy also? If we remove the human involvement then we hopefully make this more accurate, which is the point. This is quite similar to successful, but thought I'd add this anyway.

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+- automatable
+- scalable
+- more maintainable
+- more successful
+
+**Automatable** because the model won't require human involvement to
+tune field boosts on the latest user data
+
+**Scalable** because automation; but also because the model should
+provide better results with more data.
```
sophiebrown01, line 141: When you say scalable, do you mean in the sense that it removes the need for a human to find-tune everything to improve search? This feels somewhat similar to the previous _automatable_. We could also add here, if it's the case, that we're scalable (perhaps adaptable) to improving search for future versions of documents because training the model determines which of the fields (features) are most useful to improving search.

sophiebrown01: Do we still need to provide some human-annotated relevancy judgements to enable LTR to improve search results, or will we be figuring this out from other data (clickthrough data of the result(s) users choose, for example) ? (this fits better under the _maintainable_ heading)

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+complex, and be almost impossible to maintain.
+
+### Validating Learning To Rank
+
+We assumed that we would be able to get better results using Learning
+To Rank. To validate this assumption, we worked on a spike.
+
+In a [validation spike][] we found that we could build a model that provided
+a better nDCG score than our current search algorithm for the queries we
+have manually gathered relevancy judgements for.
```
sophiebrown01, line 171: I believe we tested this on regular site search? Have we tested how LTR performs on specialist finders, or other specialist areas?

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+the work started in the [validation spike][].
+
+As we are at the beginning of the project, this section may not
+accurately reflect the final implementation.
+
+[validation spike]: https://github.com/alphagov/search-api/pull/1768
+
+### Machine learning library
+
+We have decided to use [TensorFlow's ranking library][] for our
```
sophiebrown01, line 209: A small thing to note, but TensorFlow Ranking is currently at v0.2.0. Had a chat with @barrucadu yesterday on this topic in general - we should be careful about using pre 1.0.0 releases as they often contain breaking changes without updating the major version number to indicate this.

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+
+We also considered PyTorch. However, this would require writing much more
+custom machine learning code, as PyTorch don't currently have a ranking library.
+
+[TensorFlow's ranking library]: https://github.com/tensorflow/ranking
+[elasticsearch-learning-to-rank]: https://github.com/o19s/elasticsearch-learning-to-rank
+
+### Language
+
+We would like to keep the code for this project in Ruby as much as
```
sophiebrown01: Given the split between model generation and model consumption as two distinct processes this feels quite natural 👍

doc/arch/adr-010-learn-to-rank.md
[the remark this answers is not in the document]
sophiebrown01: Do we have and do we need the ability to fallback to the non-LTR enhanced search at any point should we need to?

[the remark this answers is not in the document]
sophiebrown01: In terms of versioning, do we also need to think about how iterating the algorithm will impact users and whether this is subject to further testing?

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Training
+
+**Training the model** should happen on search machines on a schedule.
+An instance of Search API will train a model using the latest relevancy
+judgements stored in S3, and will then upload a new version of the model to S3.
+On a schedule all search machines will pull the latest version of the model
+and push the new version to the locally running TF Serving container.
+
+**Training data** is made up of relevancy judgements and document features.
```
sophiebrown01: Is it useful to perform some spot-check validation on a sample of relevancy judgements over time? It's not practical to inspect everything and we assume that the results users are clicking on are the best results for the query they've entered, but if we check a selection of these from every so often we can help to validate that we're training the model appropriately. 

Touching on what @sihugh mentioned earlier on also about gaming the system, do we have any thoughts about how we can ensure the model isn't misled by relevancy judgements through the actions of a malicious user?

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+We plan to use AB testing to validate that the model provides better
+results to users (measured by click-through rate).
+
+## What success will look like
+
+Success will look like:
+
+- Click-through rate on the top 3 and top 5 results will go up
+- nDCG will go up
+- Search latency will not increase much
```
sophiebrown01, line 243: This is probably still to be defined, but it'd be good to specify how much `much` is here.

sophiebrown01 (commented):
Looking good Bill, a really clear explanation of what LTR is and why we're thinking of using it 💯

doc/arch/adr-010-learn-to-rank.md
MatMoore: I don't know what the infrastructure is like now, so take this with a grain of salt, but this part stood out to me as well. I think that GOV.UK's use of ML will only grow over time, and there's a risk of building up a lot of infrastructure that does similar things in different ways. The more data engineering there is on GOV.UK the more likely it is you will need to add new processing steps, consume data from different places, or reuse code for other purposes. 

In this case I don't think there's a need for model building to happen on the same machines that serve requests, and if you can isolate those two things (serving requests vs training the model) the processes will be more resiliant and you can scale them independently.

I haven't used concourse before, but if you have it set up already, and it works, that sounds much better than building it all from scratch. I'm imagining this will be a pipeline with multiple steps, to clean data, train the model, test it, and deploy it. There are lots of tools for this kind of thing (I know verify have built a bunch of data pipelines in luigi) but I think the most important thing is that it's easier to maintain than the current system of tweaking magic numbers and things!

I also strongly lean towards python for any data processing because you can count on data scientists and data engineers knowing it, and it comes with a massive community and ecosystem. So I think it's absolutely fine for search to stick with python despite using ruby for the API.

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Machine learning library
+
+We have decided to use [TensorFlow's ranking library][] for our
+implementation of Learning To Rank.
+
+TensorFlow is well supported, and the ranking library is under active development.
+
+We considered using the [elasticsearch-learning-to-rank][] plugin.
+However, AWS Elasticsearch does not permit us to install custom plugins.
```
MatMoore, line 215: That's a shame! Is using another hosted elasticsearch solution an option?

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+possible. Search API is written in Ruby and it is the primary language
+used at GOV.UK. Necessarily some machine learning code will need to be
+written in Python because TensorFlow is a Python library.
+
+### Dockerised model
+
+We will use TensorFlow Serving to make a re-ranker model available
+to Search API on search machines.
+
+We will run a docker container (containing our pre-trained model) on
```
MatMoore: Are you using any of the AWS managed services for running docker, like ECS? It depends on what the rest of GOV.UK is doing already but I wonder if that would be simpler/easier to operate than running it on the same machines.

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+search machines to which Search API can send requests.
+
+Should the docker container fail, Search API should be resilient to this,
+and fall back to the default ranking algorithm, with results not
+re-ranked by the model.
```
MatMoore: My concern here is that you will need to maintain both versions in order to provide this fallback, and the fallback solution will degrade over time. Is there a way to simplify the fallback, so it's good enough, but needs less ongoing maintenence? Or can you remove the fallback after some period of time, and say that if LTR is down, search is down?

doc/arch/adr-010-learn-to-rank.md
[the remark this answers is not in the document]
MatMoore: ☝️ What @karlbaker02 said! Do you need to be able to run multiple models in production at the same time, to do multivariate testing?

doc/arch/adr-010-learn-to-rank.md
MatMoore: 💯 Perhaps some of this information can be surfaced in the publishing tools as well? If publishers can see the judgement lists for specific queries, maybe they can flag where content is completely missing.

MatMoore:
This is a really well written ADR! Really great to hear that you've done a spike that demonstrates this can work, and I'm really interested to see how this goes. It would be a great thing to blog about!

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+
+However, after doing this, we would train a model with this data
+to rank the results better. The model optimises for nDCG.
+
+We train the model using using a combination of relevancy judgements
+and features: numeric measures which indicate how well we think the
+result matches the query.
+
+The text similarity scores Elasticsearch assigns to fields in the
+document are examples of features. The model tries to find correlations
```
MatMoore, line 87: I think it would be helpful to have a section that lists the features you are considering using, and exploring which ones are likely to have the biggest influences on the model.

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+predict how to order documents we don't have relevancy judgements for.
+
+```
+query         rating    document      title_score view_count  recency  pagerank ...
+harry potter  0         Pottery barn  0.2         9876        0.12     9876     ...
+harry potter  3         Harry Potter  0.2         1234        1234     1234     ...
+...
+```
+
+For example, take the SVM dataset below for the query 'harry potter' (qid:1)
```
MatMoore: SVM?

doc/arch/adr-010-learn-to-rank.md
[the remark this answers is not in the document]
bilbof: > Is it fair to say that automatable is directly relevant to accuracy also? 

Yes I think so. Rather than requiring a human to make a change, we can have the model use the latest user data and automatically adjust the way it ranks results. For example, boosting factors are arbitrary, can get out of date, and may need manually changing in future, requiring human involvement. But this wouldn't happen with LTR because the model will automatically compute the optimal weighting each field should have for a given query.

There are also some things that we won't be able to do manually to improve relevancy, but that the reranker model can, one example: https://github.com/alphagov/search-api/pull/1778/files#diff-5ea6bc45fb5fedbdaa06cec834586979R155

> ensuring the judgements it's trained on are up to date and representative

I agree!

> Maybe there are particular topics/search terms where you want more control/oversight

Yes, one thing we are thinking about is how we can allow editors to control the algorithm.

My thought at the beginning of the project was that we would use both manual and autogenerated relevancy judgements to train the model. This would let us manually influence the model - @hannako is working on a prototype for this.

But a month of user data can generate more than a million relevancy judgements. So the manual judgements might get drowned out. I was thinking perhaps we could use manual judgements in a final validation stage so we could spot any queries that are performing badly.

We could also use the model-generated scores as a multiplicative factor, so that best bets continue to work, and use something like best bets as a way to curate results.

> How do you account for bot activity elsewhere on GOV.UK?

We use javascript for collecting implicit judgements, which removes some bot activity. We could definitely look into this, there's likely to be monitoring around this that I'm not aware of.

Another note is that a bot viewing and clicking a lot on a favoured result for a particular query won't necessarily result in that result being ranked higher for that query by the model. That behaviour would produce a set of relevancy judgements for that query, which would be perhaps 20 judgements in a set of millions. I think gaming the model to lead to a particular outcome would be tricky, as it would require some sophistication and a lot of activity that we might notice.

> What automated quality checks could you run on newly trained models to prevent broken models getting into production?

I'm hoping we can re-use the approach we've had for manual changes for shipping new models (offline ranking evaluation metrics followed by online AB tests).

At the moment we've been using nDCG to compare models with the previous version (right now we just compare it to our existing search). The ranking model also records NDCG@1,3,5,10 and other metrics, so we can see if it is better or worse for the data we've given it. We can also run a set of manual relevancy judgements - that we know are good quality - against the model, to see if it gets better for those judgements (I've been doing this when working on the model).

> Are particular audiences over/under represented when you are producing the judgments? You can potentially weight clicks differently based on demographic or device type.

We normalise judgements to scores between 0-3, so judgements for highly popular queries have the same weighting for very low popularity pages (e.g. a query for "whitechapel building" that gets 20 hits has the same level of impact as the judgements for "brexit").

I think this is quite a nice feature of the model, since it treats the judgements of queries equally, regardless of popularity or how they were generated.

However, if we have a niche kind of query, with not many judgements for that type, then the results for that query type could get worse.

For example, searches for people are relatively not very popular. The model (arguably) struggles a bit with search for names at the moment. It favours articles by/about people rather than the canonical person pages.

We'll need to find a way to solve this problem, so maybe manually creating relevancy judgements for these underrepresented queries would be the answer. Weighting clicks/judgements is an interesting idea. I'm not sure I know how we'd do this at the moment - there's a card in our backlog to look into how we can react to underperforming queries.

doc/arch/adr-010-learn-to-rank.md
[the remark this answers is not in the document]
bilbof: We can fall back to the non-LTR search. Perhaps we could keep a feature flag on the reranker for debugging and rollback purposes.

Yes, I think we would need multiple models in production at the same time to do an AB test.

doc/arch/adr-010-learn-to-rank.md
bilbof: Generating a new model is definitely becoming slower since we started the project. Previously we could generate a new model in under an hour, because we were dealing with small datasets.

Now we spend a few minutes fetching data from big query, followed by 2-3 hours of generating relevancy judgements from that data, and then a further 1-2 hours training the model. The time it takes to generate training sets and train the model grows with the number of days of user data that we use. There are likely a lot of performance improvements that could be made in this code, but it's likely going to be fairly slow!

It might be that we could run this process once a week out of hours. But if we wanted to try out new models frequently in-hours it might be nice to have a dedicated machine for this.

Running it on concourse sounds good. @karlbaker02 let's chat about this! @MatMoore is probably right about us duplicating data pipeline stuff for this work, I'm interested in what the data labs team has already developed that we could re-use.

doc/arch/adr-010-learn-to-rank.md
bilbof: We can use both human-generated judgements and implicit judgements from user behaviour for training the model.

At the moment the judgements generated from user data dwarfs the manually created judgements (millions vs ~2000). Using the user generated data is easier, too, since they don't get stale. If the judgements are good quality then I'm not sure if we will need manual judgements, which makes the approach maintainable, which I suppose is part of it being scalable. 🤷‍♂

doc/arch/adr-010-learn-to-rank.md
bilbof: We've just been trying this out on site search, though when we run an AB test we should investigate the impact it has on other finders.

doc/arch/adr-010-learn-to-rank.md
bilbof: That's definitely worth investigating! I've added a task to our backlog to find a way for editors to spot check the autogenerated relevancy judgements.

> do we have any thoughts about how we can ensure the model isn't misled by relevancy judgements through the actions of a malicious user?

I added some thoughts on gaming this in a comment above. I think it's worth investigating - there's a card on our backlog for this too.

doc/arch/adr-010-learn-to-rank.md
bilbof: I think it's unlikely, since we're migrating most things into AWS at the moment. But it's worth asking the question - I'll see what our tech archs think :)

doc/arch/adr-010-learn-to-rank.md
bilbof: I don't think we are, at least within GOV.UK. It's worth looking into though. I wonder what latency is like if search-api needs to speak to a container on another machine compared to a locally running container.

doc/arch/adr-010-learn-to-rank.md
bilbof: I think the fallback being a degraded version is acceptable, if it is only a rare fallback. I'm assuming with this fallback that returning any response is better than no response. It could just be search-api as it is now, without any boosting code.

doc/arch/adr-010-learn-to-rank.md
bilbof: I think listing the features we're using in documentation would be a good thing. At the moment I don't want to add much more on the implementation to the ADR, since it will get out of date quite soon. 

The features we've been experimenting with are: popularity, format, public_timestamp, es_score, title_score, description_score, indexable_content_score, all_searchable_text_score, title_length, description_length, link_length.

doc/arch/adr-010-learn-to-rank.md
bilbof: The libsvm format (https://sourceforge.net/p/lemur/wiki/RankLib%20File%20Format/) which seems quite common in LTR datasets. We chose this because it's the format required by the TensorFlow ranking module.

doc/arch/adr-010-learn-to-rank.md
barrucadu: We need both versions anyway, to fetch a large results window for the LTR model to rerank.  We don't want to retrieve every document from the index on every search, so search-api needs to be kind-of-good anyway.

doc/arch/adr-010-learn-to-rank.md
MatMoore: Ah, I see. Might be useful to mention that in the ADR as well - the acronym threw me off a bit the first time I read it.

doc/arch/adr-010-learn-to-rank.md
vmogoreanu: I agree with everything @MatMoore said above

A few points from me:

- Choosing Concourse for Related Links was mostly because it was there and was supported by RE (and still is supported). Now that we have more of this type of work we have a chance to reconsider this particular choice for data pipelines. We are currently looking at various options including AWS Glue for ETL type of jobs
- It doesn’t feel right hosting a docker image with trained model on search boxes. And if we end up using docker containers  we need to get RE’s view on hosting them (Fargate?). We might even need to consider GOV.UK PaaS for this. Another option I am researching at the moment is AWS EMR, which is a serverless cluster that can run Tensorflow tasks among other things
- I am very sceptical about AWS Lambda for this as it also have RAM, storage space and other limitations

doc/arch/adr-010-learn-to-rank.md
vmogoreanu: I would consider the model a separate loosely coupled micro-service, all in Python with REST API interface. It is easier to find someone with Python skills to do/maintain ML in the future and we already have data scientists who all work in Python only

doc/arch/adr-010-learn-to-rank.md
vmogoreanu: No, not another ElasticSearch migration!

doc/arch/adr-010-learn-to-rank.md
vmogoreanu: We do need the fallback though, as it is new and complicated and we do not yet have a lot of experience running ML in Production. Related links for example just publishes content so nothing bad happens if it fails to run

kevindew (commented):
This looks great - really interesting concept that I know next to nothing about. Loads of great input from @MatMoore too ⭐️ 

Couple of things I wondered were:

Do we think there'll need to be a regular human involvement in this long term (eg to deal manipulation sanity checks) and if so do we need to highlight what expected regular maintenance is to be? I'm just wondering if we need to prepare for a scenario when there may not be a search team again, if this is a reasonable possibility.

Do you anticipate Python code / wider machine learning to be part of the Search API repo or to be a separate repo? Just flagging this based of the confusion levels we can hit when we have two similar languages in the same repo.

ChrisBAshton:
Looks good! Just tweaked the 'rendered version' URL to reflect the 'latest' version of the branch. I was about to comment about the lack of explanation for nDCG but the latest version links off to Wikipedia's explanation, I was viewing an out of date copy!

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+[elasticsearch-learning-to-rank]: https://github.com/o19s/elasticsearch-learning-to-rank
+
+### Language
+
+We would like to keep the code for this project in Ruby as much as
+possible. Search API is written in Ruby and it is the primary language
+used at GOV.UK. Necessarily some machine learning code will need to be
+written in Python because TensorFlow is a Python library.
+
+### Dockerised model
```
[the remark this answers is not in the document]
bilbof: @kevindew sorry, that part is lacking some detail. Docker is suggested here because we're using [TensorFlow Serving](https://github.com/tensorflow/serving) which recommends using Docker.

doc/arch/adr-010-learn-to-rank.md
```diff
@@ -0,0 +1,297 @@
...
+```
+harry potter  0         Pottery barn
+harry potter  3         Harry Potter
+...
+```
+
+However, after doing this, we would train a model with this data
+to rank the results better. The model optimises for nDCG.
+
+We train the model using using a combination of relevancy judgements
```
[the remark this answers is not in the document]
bilbof: Thanks!

kevindew:
Another thought I had, have we considered and/or need to express the impact for local development and testing. I guess we've got scenarios such as local dev on search API, people working on apps that use search API as a dependency, using govuk-docker, the old dev vm and publishing-e2e-tests.

doc/arch/adr-010-learn-to-rank.md
bilbof: Thanks, I added a note there :)

bilbof:
Thanks everyone for your comments so far!

I've removed some parts of the implementation section from the decision record that are to be confirmed, because they could become out of date quite quickly. E.g. we haven't reached a decision on how we'll host the model, so it makes sense to wait before we make a record of this. I can see the need for an RFC on the implementation details (and an additional ADR) once we've tried out some approaches and know a bit more.

bilbof:
@kevindew local development of search-api should be possible without the learning to rank model running locally. The default behaviour is that the model is hidden by behind a feature flag; if the feature is enabled search-api should be able to respond to requests as normal even if the model isn't running.

sihugh (approved):

PR #1786, opened 2019-11-12
Add a flag to request highlighted spelling suggestions

barrucadu:
[Trello card](https://trello.com/c/J0OF5keV/1134-bold-only-the-corrected-words-in-spelling-suggestions)

### DIFF
--- a/lib/search/presenters/spell_check_presenter.rb
+++ b/lib/search/presenters/spell_check_presenter.rb
@@ -3,13 +3,27 @@ module Search
     def present
       return [] unless any_suggestions?
 
-      es_response["suggest"]["spelling_suggestions"].first["options"].map { |option| option["text"] }
+      suggestions.map do |option|
+        if highlighted_suggestions?
+          { text: option["text"], highlighted: option["highlighted"] }
+        else
+          option["text"]
+        end
+      end
     end
 
   private
 
     def any_suggestions?
       es_response["suggest"] && es_response["suggest"].any?
+    end
+
+    def highlighted_suggestions?
+      suggestions.detect { |option| option.has_key? "highlighted" }
+    end
+
+    def suggestions
+      es_response["suggest"]["spelling_suggestions"].first["options"]
     end
   end
 end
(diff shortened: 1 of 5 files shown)


lib/search/presenters/spell_check_presenter.rb
```diff
@@ -3,13 +3,27 @@ module Search
...
     def any_suggestions?
       es_response["suggest"] && es_response["suggest"].any?
     end
+
+    def highlighted_suggestions?
+      suggestions.detect { |option| option.has_key? "highlighted" }
+    end
+
+    def suggestions
+      es_response["suggest"]["spelling_suggestions"].first["options"]
```
bilbof, line 26: Is it possible `spelling_suggestions` won't exist or could be empty? If so I guess this would raise an error, but the logic here hasn't changed so it looks like it is fine?

lib/search/presenters/spell_check_presenter.rb
```diff
@@ -3,13 +3,27 @@ module Search
     def present
       return [] unless any_suggestions?
 
-      es_response["suggest"]["spelling_suggestions"].first["options"].map { |option| option["text"] }
+      suggestions.map do |option|
+        if highlighted_suggestions?
+          { text: option["text"], highlighted: option["highlighted"] }
```
bilbof, line 8: Is this logic so we AB test this?

bilbof (approved):

lib/search/presenters/spell_check_presenter.rb
barrucadu: It's so that I don't need to do something like `if suggestion.is_a? String` in finder-frontend, as a change will also be needed for that.  When finder-frontend is deployed this can become the default behaviour and the `spelling_with_highlighting` removed.

lib/search/presenters/spell_check_presenter.rb
barrucadu: It would raise an error if evaluated when there are no suggestions, but `return [] unless any_suggestions?` happens before `suggestions` is used.

PR #1789, opened 2019-11-14
Enable AB testing for Levenshtein algorithm for suggestions

Tetrino:
This PR allows us to use the Levenshtein algorithm for search
suggestions over the default Elasticsearch algorithm, which relies on
an "optimised" system that may not be too accurate for our purposes.

Trello card: https://trello.com/c/c2YUDVg3/1117-try-different-ways-of-doing-spelling-suggestions

### DIFF
--- a/spec/unit/query_components/suggest_spec.rb
+++ b/spec/unit/query_components/suggest_spec.rb
@@ -1,0 +1,35 @@
+require "spec_helper"
+
+RSpec.describe QueryComponents::Suggest do
+  context "with b variant of suggestions" do
+    it "makes the suggestion algorithm levenshtein" do
+      SPELLING_FIELD = "spelling_text".freeze
+
+      builder = described_class.new(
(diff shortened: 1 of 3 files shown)


laurentqro (approved):

spec/unit/query_components/suggest_spec.rb
```diff
@@ -0,0 +1,35 @@
+require "spec_helper"
+
+RSpec.describe QueryComponents::Suggest do
+  context "with b variant of suggestions" do
+    it "makes the suggestion algorithm levenshtein" do
+      SPELLING_FIELD = "spelling_text".freeze
```
bilbof, line 6: Here `SPELLING_FIELD` could be `described_class::SPELLING_FIELD` so you don't need to change the test when you change the class.

bilbof (approved):
Great one 👍

spec/unit/query_components/suggest_spec.rb
Tetrino: Tried this, wouldn't run, so leaving it for now.

PR #1794, opened 2019-11-20
Fix a couple of small LTR doc issues

barrucadu:


bilbof (approved):

barrucadu:
@bilbof Do you think adding `govuk-docker` instructions to the doc like I did is useful?

bilbof (approved):
🚀

PR #1802, opened 2019-11-21
Don't send viewcounts and rank_eval for individual queries to graphite

bilbof:
This is taking up a fair amount of disk space on graphite. For example viewcounts took up 4GB.

We aren't using viewcounts. rank_eval might be useful in future, because we'd be able to see how relevancy for a query had changed over time, but it's not useful at the moment.

https://trello.com/c/3e0NSK06/1162-stop-logging-view-count-and-rank-eval-for-individual-queries

### DIFF
--- a/lib/tasks/relevancy.rake
+++ b/lib/tasks/relevancy.rake
@@ -48,8 +48,6 @@ namespace :relevancy do
   task :send_ga_data_to_graphite do
     puts "Sending overall CTR to graphite"
     report_overall_ctr
-    puts "Sending viewcounts to graphite"
-    report_popular_queries
     puts "Sending query click-through-rates to graphite"
     report_query_ctr
     puts "Finished"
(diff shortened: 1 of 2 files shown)


lib/tasks/relevancy.rake
```diff
@@ -48,8 +48,6 @@ namespace :relevancy do
   task :send_ga_data_to_graphite do
     puts "Sending overall CTR to graphite"
     report_overall_ctr
-    puts "Sending viewcounts to graphite"
-    report_popular_queries
```
barrucadu, line 52: Should this method still be called?  If not, can it be deleted?

bilbof: The `report_popular_queries` method is still used by the `relevancy:show_top_queries` task. Should we delete that task?

barrucadu: Ah, it's still in use.  No, that's fine.

barrucadu (approved):

PR #1806, opened 2019-11-26
Add flag to response indicating if reranking happened

barrucadu:
When A/B testing LTR, it's not enough to know that someone is in the B bucket - they have to be in the B bucket with the reranker working.  To do that, we need search-api to report if any reranking happened.

B bucket with reranker docker container not running:

Screenshot 2019-11-26 at 11 50 18

B bucket with reranker docker container running:

Screenshot 2019-11-26 at 11 50 44

B bucket with reranker docker container running but no results found:

Screenshot 2019-11-26 at 11 50 49

A bucket:

Screenshot 2019-11-26 at 11 50 57

---

[Trello card](https://trello.com/c/KuIjFRQT/1172-include-flag-in-search-api-response-if-results-are-ranked)

### DIFF
--- a/lib/learn_to_rank/ranker.rb
+++ b/lib/learn_to_rank/ranker.rb
@@ -9,7 +9,7 @@ module LearnToRank
     end
 
     def ranks
-      return [] unless feature_sets.any?
+      return nil unless feature_sets.any?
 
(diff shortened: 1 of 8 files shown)


Tetrino (approved):
Seems solid, let's see what happens.

PR #1804, opened 2019-11-25
Make people indexable

thomasleese:
This is the first step to migrating this format out of Whitehall and into the Search API.

It makes the format available for the Search API to parse and index in Elasticsearch.

I followed the instructions here for how to do this: https://github.com/alphagov/search-api/blob/master/doc/new-indexing-process.md

Depends on https://github.com/alphagov/govuk-content-schemas/pull/937 and https://github.com/alphagov/whitehall/pull/5182.

[Trello Card](https://trello.com/c/6YMqZnlo/1575-start-indexing-people-in-the-search-api)

PR #1809, opened 2019-11-27
Add person elasticsearch type

thomasleese:
This was missed in https://github.com/alphagov/search-api/pull/1804 and we need it to support indexing people from the messaging queue.

This change will also require reindexing the schemas.


cbaines (approved):
I don't know much about the Search API, but this sounds good to me :+1:

PR #1818, opened 2019-12-02
Group svm rows by qid

bilbof:
Currently we randomly shuffle SVM rows (query document pairs)
and put them into test, train and validation sets.

This has the effect that not all of the results for a query
appear in the same set (e.g. only one or two results - out
of 20 - might be in the validation set).

However, this means that NDCG reported by the model may be
inaccurate, since NDCG@10 for 1 document will be always be
1 (ideal score).

This changes the way we divide SVM rows by test/train/validation
sets, so that we first group by query id (qid) so that all
of the results for a query are available in a set.

Trello: https://trello.com/c/GYPuBRUQ


Tetrino (approved):
👍

PR #1835, opened 2019-12-10
Enable LTR using string env variables

sihugh:
We were previously relying on the truthiness of `ENV["ENABLE_LTR"]` to control enabling learn to rank.  We were trying to set the env var to a boolean in puppet, but that's not possible.

We'll do a check for the string "true" instead (to avoid an env var of "false" accidentally passing the truthiness rules) and convert puppet to setting strings instead.

https://trello.com/c/YL1JhAcm/1190-enable-ltr-in-staging-and-integration

### DIFF
--- a/doc/learning-to-rank.md
+++ b/doc/learning-to-rank.md
@@ -30,7 +30,7 @@ database in the `venv` directory.  If you close the shell, you can run
 Using LTR
 ---------
 
-**Set the `ENABLE_LTR` environment variable, or all of this is disabled.**
+**Set the `ENABLE_LTR` environment variable to "true", or all of this is disabled.**
 
 There are several rake tasks for training and serving a TensorFlow
 model in the `learn_to_rank` namespace.
@@ -58,7 +58,7 @@ This task needs to be run with access to Elasticsearch.  If you're
 using govuk-docker the full command will be:
 
 ```sh
-govuk-docker run -e ENABLE_LTR=1 search-api-lite bundle exec rake 'learn_to_rank:generate_training_dataset[judgements.csv]'
+govuk-docker run -e ENABLE_LTR=true search-api-lite bundle exec rake 'learn_to_rank:generate_training_dataset[judgements.csv]'
 ```
 
 Once you have the training dataset you can train and serve a model:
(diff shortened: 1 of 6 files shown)


lib/tasks/learn_to_rank.rake
```diff
@@ -98,7 +99,7 @@ namespace :learn_to_rank do
   end
 
   def assert_ltr!
-    raise "set $ENABLE_LTR to use learn_to_rank" if ENV["ENABLE_LTR"].nil?
+    raise "set $ENABLE_LTR to use learn_to_rank" if !Search::RelevanceHelpers.ltr_enabled?
```
barrucadu: This should give a more informative message like `set $ENABLE_LTR to "true"`, as any arbitrary string won't work now.

sihugh: Good point, thanks

barrucadu (commented):
`doc/learning-to-rank.md` also needs changing, as it gives the example of `ENABLE_LTR=1`

doc/learning-to-rank.md
```diff
@@ -58,7 +58,7 @@ This task needs to be run with access to Elasticsearch.  If you're
 using govuk-docker the full command will be:
 
 ```sh
-govuk-docker run -e ENABLE_LTR=1 search-api-lite bundle exec rake 'learn_to_rank:generate_training_dataset[judgements.csv]'
```
barrucadu, line 61: Line 78 of this file also needs changing, and the bold warning at line 33 could be clarified too.

lib/tasks/learn_to_rank.rake
```diff
@@ -98,7 +99,7 @@ namespace :learn_to_rank do
   end
 
   def assert_ltr!
-    raise "set $ENABLE_LTR to use learn_to_rank" if ENV["ENABLE_LTR"].nil?
+    raise 'set $ENABLE_LTR to "true" to use learn_to_rank' if !Search::RelevanceHelpers.ltr_enabled?
```
barrucadu: Surprised the linter didn't complain about using `if !` instead of `unless`

doc/learning-to-rank.md
sihugh: I'm getting there! :-D

lib/tasks/learn_to_rank.rake
sihugh: Yeah, weird... is the Rakefile not included.  Seems like a reasonable adjustment.  Incoming

barrucadu (approved):

sihugh:
Thanks @barrucadu

PR #1836, opened 2019-12-10
Allow filtering on documents based on attached roles

thomasleese:
This allows users to filter on documents based on any attached roles. Currently you can do https://www.gov.uk/api/search.json?filter_people=boris-johnson to filter all documents tagged to the person "Boris Johnson". This PR should make it possible to filter on all documents tagged to the role "Prime Minister" with https://www.gov.uk/api/search.json?filter_roles=prime-minister.

This is required to support rendering the "Announcements" section here https://www.gov.uk/government/ministers/prime-minister#announcements from Search API rather than from Whitehall's own index.

[Trello Card](https://trello.com/c/6r6CjemK/1404-investigate-how-we-can-support-filtering-on-roles-in-the-search-api)

### DIFF
--- a/lib/govuk_index/presenters/expanded_links_presenter.rb
+++ b/lib/govuk_index/presenters/expanded_links_presenter.rb
@@ -20,6 +20,10 @@ module GovukIndex
       slugs("people", "/government/people/")
     end
 
+    def roles
+      slugs("roles", "/government/ministers/")
+    end
+
     def policy_groups
       slugs("working_groups", "/government/groups/")
     end
--- a/config/schema/elasticsearch_types/edition.json
+++ b/config/schema/elasticsearch_types/edition.json
@@ -35,6 +35,7 @@
     "organisations",
     "parent_organisations",
     "people",
+    "roles",
     "personal_data",
     "policies",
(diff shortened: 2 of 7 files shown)


lib/govuk_index/presenters/expanded_links_presenter.rb
```diff
@@ -20,6 +20,10 @@ def people
       slugs("people", "/government/people/")
     end
 
+    def roles
+      slugs("roles", "/government/ministers/")
```
thomasleese, line 24: For now, only ministerial roles get base paths. In the future we might have to support more types of roles, although that's not necessary at the moment.

bevanloon (approved):
Look fine to me. 
Be good to have a search person to give it the once over as well - @barrucadu if you have a minute please?

barrucadu (approved):
Looks good to me.  if you wanted to be really sure, you could add another integration test like the taxon filtering ones: https://github.com/alphagov/search-api/blob/master/spec/integration/search/search_spec.rb#L560-L576

But there are lots of filters (eg people) which don't have such tests, so it's up to you.

PR #1840, opened 2019-12-11
Skip model download if model has already been downloaded

laurentqro:
As this task will be run on each deploy, we want to avoid pulling the model from S3 if the model has already been downloaded, to avoid slowing down deployment unnecessarily.

--

https://trello.com/c/Azj599Sw/1213-improve-resilience-of-learn-to-rank

### DIFF
--- a/lib/tasks/learn_to_rank.rake
+++ b/lib/tasks/learn_to_rank.rake
@@ -41,6 +41,30 @@ namespace :learn_to_rank do
     end
   end
 
+  desc "Pull learn to rank model from S3"
+  task :pull_model, [:model_filename] do |_, args|
+    bucket_name = ENV["AWS_S3_RELEVANCY_BUCKET_NAME"]
+    raise "Missing required AWS_S3_RELEVANCY_BUCKET_NAME" if bucket_name.blank?
+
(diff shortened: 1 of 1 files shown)


lib/tasks/learn_to_rank.rake
```diff
@@ -41,48 +41,56 @@ namespace :learn_to_rank do
...
+    raise "Missing required AWS_S3_RELEVANCY_BUCKET_NAME" if bucket_name.blank?
+
+    models_dir = ENV["TENSORFLOW_MODELS_DIRECTORY"]
+    raise "Please specify the Tensorflow models directory" if models_dir.blank?
+
+    prefix            = "ltr"
+    s3_objects        = Aws::S3::Bucket.new(bucket_name).objects(prefix: prefix)
+    model_files       = s3_objects.map { |object| object.key.delete("#{prefix}/") }
+    latest_model      = model_files.max_by(&:to_i)
+    model_filename    = args.model_filename || latest_model
```
bilbof: This isn't introduced by your change, but might be nice to skip the lookup of the latest_model if `args.model_filename` is already present.

Something like

```ruby
args.model_filename || begin
    # fetch latest_model
end
```

bilbof (approved):

bilbof (approved):

PR #1841, opened 2019-12-12
Report ndcg to graphite

bilbof:
https://trello.com/c/eLOxX370/1215

This adds a new rake task `relevancy:send_metrics_to_graphite` that will run on a schedule. It will send NDCG to graphite, to populate a new metric.

This is necessary since it uses the new Evaluate::NDCG class that computes NDCG manually (rather than using the rank_eval endpoint provided by Elasticsearch), and that we can use to compare the Learning To Rank model to our existing search relevancy performance.

The Elasticsearch rank eval endpoint computes NDCG internally, so we can't use it with the dockerised Reranker.


hannako (approved):
Looks ok to me Bill!

PR #1844, opened 2019-12-13
Move history mode fields to base document type

brucebolt:
The three fields associated with historic mode (is_historic, is_political and government_name) are currently associated with the edition type (which is used only for the government index, i.e. Whitehall documents).  Since history mode is now being implemented in Content Publisher, these fields are being moved to the base type so all indices can use them (i.e. both Whitehall and Content Publisher documents).

Trello card: https://trello.com/c/RZwXj2KH


JonathanHallam (approved):
👍

kevindew (approved):
Looks good to me 👍

PR #1845, opened 2019-12-13
Replace require for gds_api/rummager

brucebolt:
`gds_api/rummager` was removed in a breaking change in https://github.com/alphagov/gds-api-adapters/pull/966.  This was preventing search-api from being built with govuk-docker.

Therefore updating to use `gds_search` to keep this task working.

### DIFF
--- a/lib/tasks/report_inconsistent_aggregate_values.rake
+++ b/lib/tasks/report_inconsistent_aggregate_values.rake
@@ -1,4 +1,4 @@
-require "gds_api/rummager"
+require "gds_api/search"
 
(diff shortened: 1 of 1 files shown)


lib/tasks/report_inconsistent_aggregate_values.rake
```diff
@@ -1,4 +1,4 @@
-require "gds_api/rummager"
+require "gds_api/search"
```
barrucadu, line 1: You also need to use `GdsApi::Search` lower down

barrucadu (approved):

PR #1850, opened 2019-12-16
Present political data to the govuk index

brucebolt:
Content Publisher now marks documents as political for use in history mode.  We need to present the relevant values to the govuk index to enable this information to be shown and filtered on for Content Publisher documents.

Not to be merged until https://github.com/alphagov/search-api/pull/1844 is merged and Elasticsearch has been reindexed.

Trello card: https://trello.com/c/RZwXj2KH


lib/govuk_index/presenters/common_fields_presenter.rb
```diff
@@ -81,6 +81,18 @@ def section_id
       @_section_id ||= payload.dig("details", "section_id") if format == "hmrc_manual_section"
     end
 
+    def is_historic
+      is_political && payload.dig("expanded_links", "government", 0, "details", "current") == false
+    end
+
+    def is_political
+      payload.dig("details", "political") || false
```
JonathanHallam, line 89: I'm assuming that the use of ` || false` here means that payload["details"]["political"] isn't always set?

brucebolt: Yes, `dig` returns `nil` if any of the keys are not present.  I'm working on an assumption that some applications may send a payload that doesn't contain a political status.

JonathanHallam (approved):
Looks good to me. Not a massive fan of how far we have to dig down to get the fields we need but that says more about the data than the code!

lib/govuk_index/presenters/common_fields_presenter.rb
JonathanHallam: 👍

lib/govuk_index/presenters/common_fields_presenter.rb
```diff
@@ -81,6 +81,18 @@ def section_id
...
+    end
+
+    def government_name
+      payload.dig("expanded_links", "government", 0, "title")
+    end
```
kevindew, line 94: I wonder if you want a helper method of government that can return: `payload.dig("expanded_links", "government").first`

and then government_name could be `government && government.dig("details", "title")` and is historic could be `is_political && government && government["current"]`

lib/govuk_index/presenters/common_fields_presenter.rb
```diff
@@ -81,6 +81,18 @@ def section_id
       @_section_id ||= payload.dig("details", "section_id") if format == "hmrc_manual_section"
     end
 
+    def is_historic
```
kevindew: I wonder if this and is_political should be `historic?` and `political?`. This would also mean a change to `is_withdrawn` for consistency.

This would be more idiomatic to Ruby as prefixing methods with `is_`, `has_` and `can_` are [regularly discouraged](https://docs.rubocop.org/en/latest/cops_naming/#namingpredicatename). Doing a search in search-api codebase shows that the only 'def is_' method we have like this is `is_withdrawn`

spec/unit/govuk_index/presenters/common_fields_presenter_spec.rb
```diff
@@ -124,6 +124,148 @@
...
+        document_type: "news_story",
+      },
+      details: {
+        political: true,
+      },
+    )
+
+    presenter = common_fields_presenter(payload)
+
+    expect(presenter.is_political).to eq(true)
```
kevindew: Minor but you switch from `eq(true)` to `be true` later in this test: https://github.com/alphagov/search-api/pull/1850/files#diff-3d91464b869a9b45ca7c4ac752b0e927R208

kevindew (commented):
Looks great - just a few minor suggestions.

spec/unit/govuk_index/presenters/common_fields_presenter_spec.rb
brucebolt: Ah, the problem of writing code over several days!

lib/govuk_index/presenters/common_fields_presenter.rb
brucebolt: Good point.  I didn't include the `?` to maintain consistency with `is_withdrawn`, but it's definitely a better idea to update the name of `is_withdrawn` to `withdrawn?`.

kevindew (approved):
Aces, thanks.

Might want to flag this as DNM until we've got the index re-done. I'm not too sure of the consequences if we merge this before then.

PR #1853, opened 2019-12-17
Show role search results from govuk index

thomasleese:
We've migrated person pages so now they are being picked up by search-api from the message queue. This PR changes it so search results are now coming from that new index rather than the old government index.

### DIFF
--- a/config/govuk_index/migrated_formats.yaml
+++ b/config/govuk_index/migrated_formats.yaml
@@ -52,6 +52,8 @@ migrated:
 - hmrc_manual_section
 - manual
 - manual_section
+- minister # from Whitehall, should be `ministerial_role`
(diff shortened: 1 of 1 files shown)


config/govuk_index/migrated_formats.yaml
```diff
@@ -52,6 +52,8 @@ migrated:
 - hmrc_manual_section
 - manual
 - manual_section
+- minister # from Whitehall, should be `ministerial_role`
```
thomasleese, line 55: This is a bit of a hack to avoid duplicate results because Whitehall sends these as document_type of `minister` instead of `ministerial_role`.

bevanloon (approved):

PR #1864, opened 2019-12-19
Move stats radio button facet to the top

sihugh:
At present, it looks like the document type field is under the "Topic" heading
in the filters section.  This fixes that, and also makes it nearer the top
on mobile.  It's more used than Topic on the stats finder, so this makes sense.

We'd originally kept topic as the top filter to keep things broadly consistent
between supergroup finders as we expected to build some navigation to allow
users to flip between them.  This hasn't happened, so let's do a small
thing to make life easier for people.

## Before 

Screenshot 2019-12-18 16 35 20

## After

Screenshot 2019-12-18 16 35 28

### DIFF
--- a/lib/learn_to_rank/ranker.rb
+++ b/lib/learn_to_rank/ranker.rb
@@ -60,7 +60,7 @@ module LearnToRank
     end
 
     def ranker_error(response)
-      response.nil? || response.code != 200
+      response.nil? || response.body.blank? || response.code != 200
     end
 
     def log_response(response)
--- a/config/finders/statistics_finder.yml
+++ b/config/finders/statistics_finder.yml
@@ -44,6 +44,13 @@ details:
     type: text
     display_as_result_metadata: true
     filterable: false
+  - key: content_store_document_type
+    name: Statistics
+    type: research_and_statistics
+    display_as_result_metadata: false
+    filterable: true
+    hide_facet_tag: true
+    preposition: that are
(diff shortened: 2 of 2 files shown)


lib/learn_to_rank/ranker.rb
```diff
@@ -60,7 +60,7 @@ def tensorflow_serving_ip
     end
 
     def ranker_error(response)
-      response.nil? || response.code != 200
+      response.body.nil? || response.body.empty? || response.code != 200
```
laurentqro: I think `response.body.nil? || response.body.empty?` could be replaced with `response.body.blank?`

```
Loading development environment (Rails 5.2.4)
[1] pry(main)> nil.blank?
=> true
[2] pry(main)> " ".blank?
=> true
[3] pry(main)> "".blank?
=> true
```

laurentqro:
@sihugh small comment, otherwise 🚢

laurentqro (approved):

lib/learn_to_rank/ranker.rb
sihugh: I'd thought that as search-api is Sinatra, it wouldn't have `blank?`.  Turns out we have ActiveSupport added so I can use it.  Thanks

laurentqro: Ah yes, actually hadn't realised that. Great!

PR #1867, opened 2019-12-19
Add parts to search index

bilbof:
Currently guidance parts (sections of guides, such as
'How to claim' in the Universal Credit guide) are indexed
into the blob of indexable_content, and aren't available
to frontend applications for rendering.

This indexes parts separately and makes them available for
returning in responses.

https://trello.com/c/jRP2uUla/1239-display-results-in-parts-be-work

### DIFF
--- a/lib/govuk_index/presenters/details_presenter.rb
+++ b/lib/govuk_index/presenters/details_presenter.rb
@@ -39,6 +39,21 @@ module GovukIndex
       details.dig("image", "url")
     end
 
+    def parts
+      parts = details.fetch("parts", [])
+      return unless parts && parts.any?
+
+      parts.map do |part|
+        body = part.fetch("body", [{}]).first
+        body_content = body["content"]
+        {
+          "slug" => part["slug"],
+          "title" => part["title"],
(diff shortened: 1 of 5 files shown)


lib/govuk_index/presenters/details_presenter.rb
```diff
@@ -39,6 +39,21 @@ def image_url
       details.dig("image", "url")
     end
 
+    def parts
+      parts = details.fetch("parts", [])
+      return unless parts && parts.any?
+
+      parts.map do |part|
+        body = part.fetch("body", [{}]).first
```
barrucadu, line 47: I looked at [the universal credit content item](https://www.gov.uk/api/content/universal-credit), and its parts have the schema `{"title": "...", "slug": "...", "body": "..."}`, not `"body": [{"content": "..."}]`.

bilbof: Yeah that tripped me up, too. The way these get represented to search api (`body: [{ content ...`] is different to how they look in the content store.

barrucadu (approved):

PR #1868, opened 2019-12-19
Fix HTTParty response check

bilbof:
This resolves a deprecation warning from the HTTParty gem.

> HTTParty will no longer override `response#nil?`. This functionality will be
> removed in future versions. Please, add explicit check `response.body.nil? || response.body.empty?`.
> For more info refer to: https://github.com/jnunemaker/httparty/issues/568.


bilbof:
Simon got there first!

PR #1881, opened 2019-12-30
Add HTTP endpoint to fetch bigquery data, generate SVM files, and upload to S3

barrucadu:
Tested on integration by running the rake task.  Output here: https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-search-relevancy/data/2019-12-30/?region=eu-west-1&tab=overview

---

`GET /ltr/train` - returns the status of the current LTR job.

Response codes:
- 200, if in `ready` state (the last executed job succeeded).
- 500, if in `error` state (the last executed job failed).
- 202, if the last triggered job has not finished.

`POST /ltr/train` - triggers a new LTR job.

Response codes:
- 202, if job successfully triggered.
- 409, if a job is already running.

These endpoints need the `ltr_training` permission, and for LTR to be
enabled on the host.  Failing either of those conditions produces a
403 error.

---

This will be used by something like this:

```python
import os, requests, sys, time

headers = {"Authorization": f"Bearer {os.environ['SEARCH_API_TOKEN']"}

trigger = requests.post("http://search-api/ltr-train", headers=headers)
print(f"POST http://search-api/ltr-train: {trigger.status_code}, {trigger.text}")
if trigger.status_code != 202:
    sys.exit(1)

status_code = 202
while status_code == 202:
    check = requests.get("http://search-api/ltr-train", headers=headers)
    print(f"GET http://search-api/ltr-train: {check.status_code}, {check.text}")
    status_code = check.status_code
    time.sleep(60)

if status_code != 200:
    sys.exit(1)
```

RE think this is the better approach than having Concourse or Jenkins SSH into a search machine and run a script, as the eventual goal is to move away from machines entirely.

---

[Trello card](https://trello.com/c/ZSkcaF6E/1243-add-api-endpoint-in-search-api-to-generate-training-data)


Tetrino (dismissed):

Tetrino (approved):

lib/learn_to_rank/data_pipeline.rb
```diff
@@ -0,0 +1,122 @@
...
+          now = Time.now.strftime("%Y-%m-%d")
+          Aws::S3::Object.new(bucket_name: s3_bucket, key: "data/#{now}/train.txt").put(body: svm[:train])
+          Aws::S3::Object.new(bucket_name: s3_bucket, key: "data/#{now}/test.txt").put(body: svm[:test])
+          Aws::S3::Object.new(bucket_name: s3_bucket, key: "data/#{now}/validate.txt").put(body: svm[:validate])
+        end
+
+        LearnToRank::DataPipeline.set_status(LearnToRank::DataPipeline::READY_STATUS)
+      rescue StandardError => e
+        LearnToRank::DataPipeline.set_status(LearnToRank::DataPipeline::ERROR_STATUS)
+        GovukError.notify(e)
```
bilbof: Once the pipeline enters an error state, what would one do to reset the state to ready?

barrucadu: The redis key expires after an hour, but it can also be manually deleted if you can connect to redis (redis-cli or govuk_app_console)

lib/rummager/app.rb
```diff
@@ -274,6 +274,38 @@ def get_type_from_request_body(request_body)
...
+  get "/ltr/train" do
+    require_authentication "ltr_training"
+    halt(403, "LTR is not enabled") unless Search::RelevanceHelpers.ltr_enabled?
+
+    status = LearnToRank::DataPipeline.get_status
+    code = case status
+           when LearnToRank::DataPipeline::READY_STATUS
+             200
+           when LearnToRank::DataPipeline::ERROR_STATUS
+             500
```
bilbof, line 286: I'm not sure we should return a 500 here. A 200 would probably be more accurate for successful status requests, since the request succeeds. We can add more details about the status of the object/job within the JSON response, e.g.

```json
200
{ "job": { "status": { "code": "FAILED", "message": "The job failed" } } }
```

barrucadu: I'm also not sure what the appropriate code to return here is but, by that reasoning, all three cases should return 200 - the status got successfully returned.  I wanted to use different status codes so that whatever is consuming this API doesn't need to recognise the error text we return.

Given that returning a 202 for the "in progress" state seemed fine to me, I wanted to use an error code for this case.  And it's definitely not a client error.

lib/learn_to_rank/data_pipeline.rb
```diff
@@ -0,0 +1,122 @@
...
+    def self.set_status(status)
+      redis = Redis.new(
+        host: ENV.fetch("REDIS_HOST", "127.0.0.1"),
+        port: ENV.fetch("REDIS_PORT", 6379),
+        namespace: "search-api-ltr",
+      )
+      redis.setex("ltr-status", REDIS_EXPIRATION_PERIOD, status)
+    end
+
+    class Worker
```
bilbof: Bit of a pedantic point, but I'd prefer if worker went in a separate file, and was called something more specific (we may end up with multiple workers for this). I find it easier to unit test small parts when they're clearly separated.

lib/learn_to_rank/data_pipeline.rb
bilbof: What are the benefits of having the worker in an error state for an hour?

barrucadu: We want the API consumer to be able to tell that there's been an error, but I don't think we want to have to manually intervene to reset the state.

lib/learn_to_rank/data_pipeline.rb
barrucadu: I've moved it to a separate file but left it called "Worker" for now, as I don't think there's a useful more specific name yet.

PR #1894, opened 2020-01-08
Pull relevancy judgements from S3 for learn_to_rank:reranker:evaluate

bilbof:
The NDCG evaluation task will pull the latest CSV of judgments from S3 should no CSV be provided. This is handy, as it lets us run this task remotely.


barrucadu (approved):
LGTM

Do we want [the data pipeline worker](https://github.com/alphagov/search-api/blob/master/lib/learn_to_rank/data_pipeline/worker.rb) to upload the judgements it generates to S3 (and possibly all the intermediate state)?

Also, I'm testing the pipeline again, so can you hold off on merging this [until the jobs have stopped flashing](https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/search-learn-to-rank)

barrucadu:
Actually, never mind on waiting, a depenabot PR got deployed, which I expect will have killed the sidekiq process.  I'll cancel the job and just see what the overnight run does.

bilbof:
@barrucadu oops sorry that was me :/

I think it would be useful for the generated data to be uploaded somewhere. It'll give us a backup if something goes wrong with the data pipeline and/or the model and we need to re-train a model.

We can add this later though, since it'll requires some extra thinking about if/how we want to version the data.

barrucadu:
Re-training will work at the moment, because it's the svm data that gets uploaded, and the date is put in the filename.  So we can train using any day's data.  It's the intermediate steps which ultimately feed into the svm data (bigquery export, relevancy judgements, augmented judgements) which aren't preserved.

PR #1897, opened 2020-01-10
Fix up some issues with generating LTR training data

barrucadu:
See commit messages.

---

[Trello card](https://trello.com/c/ZfXrYH46/1234-host-ltr-model-training-in-sagemaker-glue)

### DIFF
--- a/lib/learn_to_rank/data_pipeline.rb
+++ b/lib/learn_to_rank/data_pipeline.rb
@@ -4,7 +4,10 @@ module LearnToRank
     READY_STATUS = "Ready".freeze
     ERROR_STATUS = "Error".freeze
 
+    REDIS_NAMESPACE = "search-api-ltr".freeze
     REDIS_EXPIRATION_PERIOD = 60 * 60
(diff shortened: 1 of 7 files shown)


lib/learn_to_rank/data_pipeline.rb
```diff
@@ -51,21 +54,27 @@ def self.can_run?
...
+    def self.set_latest_data(data)
+      redis.set(REDIS_LATEST_DATA_KEY, data)
+    end
+
+    def self.redis
+      Redis.new(
         host: ENV.fetch("REDIS_HOST", "127.0.0.1"),
         port: ENV.fetch("REDIS_PORT", 6379),
-        namespace: "search-api-ltr",
+        namespace: REDIS_NAMESPACE,
```
bilbof, line 76: Do the search machines share a single redis instance, or do they have one instance per machine?

bilbof (approved):

lib/learn_to_rank/data_pipeline.rb
barrucadu: I believe they share a redis instance

PR #1898, opened 2020-01-10
Add summary to parts

bilbof:
This stores a truncated plain text summary of the body content under the body field on parts, ready for displaying to users in search results.

This will require re-representing guides in Publisher, but won't require re-indexing.

https://trello.com/c/szgqkAGr/1263-generate-part-summaries-at-index-time


barrucadu:
General question: do we have HTML in elasticsearch for part bodies at the moment?  Could that cause problems (search queries matching against HTML tags and attributes, eg)?

bilbof:
Part bodies aren't HTML yet, but this would make them HTML.

That's a good point, I didn't consider it. Do you think we should strip the HTML tags from the body content too?

barrucadu:
Yeah, I'm wary about storing HTML in a searchable field.  What was the rationale for getting the `text/html` body, rather than the govspeak body (which also exists?)

bilbof:
I made that change 👍 

I copied the parser we use for indexable content, which also uses the HTML content rather than govspeak.

lib/govuk_index/presenters/parts_presenter.rb
```diff
@@ -0,0 +1,58 @@
...
+      html = html_content(part_bodies)
+      return unless html
+
+      strip_html_tags(html)
+    end
+
+    def summarise(body_content)
+      return unless body_content.present?
+
+      body_content.truncate(75, omission: ELLIPSIS, separator: SEPARATOR)
```
barrucadu: Now that this is just a method call on the body content, which is going into the search index, I'm wondering if indexing the summary is useful - could we just truncate the body when presenting search results?  What do you think?

bilbof: Yeah that's interesting. Perhaps we could store the truncated body as the body, and not have to pass a lot of extra data over in search responses? Added bonus that we don't need to re-index!

lib/govuk_index/presenters/parts_presenter.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    def presented_parts
+      return unless parts && parts.any?
+
+      parts.map do |part|
+        {
+          "slug" => part["slug"],
+          "title" => part["title"],
+          "body" => summarise(part.fetch("body", [{}])),
```
barrucadu, line 14: I still think we should have the full body in the search index, if we did before.  The `summarize` part could happen when presenting search results.

bilbof:
Aside from generating some clickbait-y digests, I think this is ready for re-review :)

> You must pay the charge if you use the Dartford Crossing between 6am and…

https://www-origin.staging.publishing.service.gov.uk/api/search.json?filter_link=/pay-dartford-crossing-charge&fields=parts

lib/govuk_index/presenters/parts_presenter.rb
bilbof: As discussed offline we'll re-use the body field for the truncated body, and not add a summary field.

barrucadu (approved):

PR #1899, opened 2020-01-10
Add summaries to parts

bilbof:
Ahead of merging 2432596 we need to re-index with the new schema.

This should be merged and deployed, then we'll need to re-index, before starting to index parts with summaries.

https://trello.com/c/szgqkAGr/1263-generate-part-summaries-at-index-time


barrucadu (approved):

bilbof:
Decided not to add a new field.

PR #1903, opened 2020-01-13
Use lazy enumerators in LTR training data generation

barrucadu:
The current training data generation uses a lot of memory, as it
holids in memory all at the same time:

- Bigquery results
- Relevancy judgements
- Augmented relevancy judgements
- SVM data

And there's other things (like SVM data in another form) which
temporarily live in memory, before being garbage collected.  That's a
lot to hold in memory at once.

This commit changes the data generation to use lazy enumerators, so
everything other than the bigquery data (which is still held in memory
in full) is only computed when needed.

Unfortunately, we can no longer print out a progress metric when
computing the augmented relevancy judgements, because to get the total
number of judgements would require forcing that iterator, which
defeats the point.

Finally, I've reduced the number of queries cached in the
judgement-augmentor to 25, as we're getting quite a lot of data from
elasticsearch at a time.

**Machine metrics of the pipeline rake task with this commit:**

Screenshot 2020-01-13 at 14 50 03

The deploy of this branch corresponds with the sharp dip between 14:28 and 14:30.  Peak memory usage is about 0.524GB: while running the task memory usage reaches 2.386GB, and when the task completes it drops to 1.861GB.

**Machine metrics of the pipeline rake task with `master`:**

Screenshot 2020-01-13 at 15 13 57

The deploy of this branch corresponds with the sharp dip between 14:52 and 14:54.  Peak memory usage is about 0.701GB: while running the task memory usage reaches 2.535GB, and when the task completes it drops to 1.834GB.

---

[Trello card](https://trello.com/c/ZfXrYH46/1234-host-ltr-model-training-in-sagemaker-glue)


Tetrino (approved):

barrucadu:
~I've just spotted an opportunity for earlier GC: `queries` can be in a narrower scope.  I'll try that on staging first.~

This turned out not to work, memory usage went *up*.

lib/learn_to_rank/data_pipeline/embed_features.rb
```diff
@@ -1,33 +1,33 @@
...
-        logger.info "#{i}/#{count}: #{(i / count) * 100}%"
-        feats = features(judgement)
+      cached_queries = {}
+      augmented = judgements.lazy.map do |judgement|
+        next nil unless judgement
+
+        feats = features(judgement, cached_queries)
         next nil unless feats
 
+        cached_queries = {} if cached_queries.keys.count > 25
```
bilbof: Are judgements already sorted by query? Just wondering since it'd help the caching here.

lib/learn_to_rank/data_pipeline/embed_features.rb
```diff
@@ -58,11 +58,9 @@ def fetch_document(judgement)
           "fields" => %w(popularity content_id title format description link public_timestamp organisation_content_ids updated_at indexable_content),
           "count" => %w[20],
         }
-        @cached_queries[judgement[:query]] ||= do_fetch(query)
-        results = @cached_queries[judgement[:query]]
-        if @cached_queries.keys.count > 50
-          flush_cached_queries
-        end
+        cached_queries[judgement[:query]] ||= do_fetch(query)
+        results = cached_queries[judgement[:query]]
```
bilbof: Since `cached_queries` is already being treated like an instance variable here, rather than passing it to methods, should we keep it as `@cached_queries`?

lib/learn_to_rank/data_pipeline/embed_features.rb
barrucadu: They should be, as the bigquery groups on `searchTerm`

lib/learn_to_rank/data_pipeline/embed_features.rb
barrucadu: That makes sense.  As the judgements are grouped by queries, I'll replace this caching with two instance variables: `@last_search_term` and `@last_query`

bilbof (approved):
Great stuff 💯

lib/learn_to_rank/data_pipeline/embed_features.rb
```diff
@@ -48,20 +47,30 @@ def features(judgement, cached_queries)
...
 
-    def fetch_document(judgement, cached_queries)
+    def fetch_document(judgement)
+      unless judgement[:query] == @last_query
+        @last_query = judgement[:query]
+        @last_results = do_fetch(judgement)
+      end
+
+      unless @last_results.nil?
+        @last_results.find { |doc| doc["link"] == judgement[:link] }
```
bilbof, line 57: ~A mini optimization here would be to create a new hash for the results indexed by link, so you don't need to do a find.~ Ignore this - not worth it :)

bilbof (approved):
Looks good 👍

PR #1906, opened 2020-01-14
Wait for model deployment to complete before deleting old artefacts

barrucadu:
Example: https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/search-learn-to-rank/jobs/staging-deploy/builds/6

---

The deploy method has a parameter called "wait":

> wait (bool) – Whether the call should wait until the deployment of
> this model completes (default: True).

We're not setting it to false, but this script is proceeding to delete
the old endpoint config and model before the endpoint has updated to
the new one.

### DIFF
--- a/ltr/concourse/deploy.py
+++ b/ltr/concourse/deploy.py
@@ -3,6 +3,8 @@
 import os
 import sagemaker
 import sagemaker.tensorflow
+import sys
+import time
 
 from datetime import date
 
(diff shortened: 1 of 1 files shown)


ltr/concourse/deploy.py
```diff
@@ -43,6 +45,19 @@
     update_endpoint=True,
 )
 
+# wait for the deployment to complete
+endpoint_status = "Updating"
+while endpoint_status == "Updating":
```
bilbof, line 50: Do you think checking if `endpoint_status != OK` (replacing OK with whatever the OK status is) would help? I'm wondering if there are other statuses that aren't OK during which we'd also want to wait.

bilbof (approved):
LGTM!

ltr/concourse/deploy.py
barrucadu: I think we want to bail out when it's not `Updating`.  There's a list of them here: https://docs.aws.amazon.com/cli/latest/reference/sagemaker/describe-endpoint.html

`Creating` won't happen (outside the bootstrap job), `SystemUpdating` means AWS maintenance is happening and we don't know how long it'll take (and if we're checking for that we should do it before even trying to deploy the model), and `RollingBack` will get back to the old working configuration successfully if we don't delete it.

bilbof: OK, thanks for that detail!

PR #1912, opened 2020-01-15
Add healthcheck for dockerised LTR model

bilbof:
This adds a connectivity and model status healthcheck for the model served by the tensorflow serving docker container on search machines.

Still to add: AWS Sagemaker status check

https://trello.com/c/Jgbd0TrY/1255-add-model-status-to-search-api-healthcheck

### DIFF
--- a/lib/learn_to_rank/ranker.rb
+++ b/lib/learn_to_rank/ranker.rb
@@ -4,6 +4,8 @@ require "httparty"
 module LearnToRank
   class Ranker
     include Errors
+    include RankerApiHelper
+
     # Ranker takes feature sets and requests new scores for them
     # from a pre-trained model.
     def initialize(feature_sets = [])
@@ -60,7 +62,7 @@ module LearnToRank
     end
 
     def fetch_new_scores_from_serving(examples)
-      url = "http://#{tensorflow_serving_ip}:8501/v1/models/ltr:regress"
+      url = "#{tensorflow_container_url}:regress"
       options = {
         method: "POST",
         body: {
@@ -87,16 +89,6 @@ module LearnToRank
       end
 
       JSON.parse(response.body).fetch("results")
-    end
-
-    def tensorflow_serving_ip
-      if ENV["TENSORFLOW_SERVING_IP"].present?
-        ENV["TENSORFLOW_SERVING_IP"]
-      elsif %w(development).include? ENV["RACK_ENV"]
-        "reranker"
-      else
-        "0.0.0.0"
-      end
     end
 
     def logger
--- a/lib/learn_to_rank/ranker_status.rb
+++ b/lib/learn_to_rank/ranker_status.rb
@@ -1,0 +1,96 @@
+require "aws-sdk-sagemaker"
+require "httparty"
+
+module LearnToRank
+  class RankerStatus
+    include RankerApiHelper
+
+    attr_reader :errors
+
+    def initialize(timeout: 0.1)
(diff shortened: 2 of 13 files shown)


lib/learn_to_rank/ranker_status.rb
```diff
@@ -0,0 +1,103 @@
...
+      response = HTTParty.get(url, options)
+      validate_response!(response)
+      model = JSON.parse(response.body).fetch("model_version_status", []).last
+      validate_model_healthy!(model)
+      true
+    rescue HTTParty::Error, SocketError, Timeout::Error => e
+      raise RankerServerError.new(message: e.message)
+    end
+
+    def tensorflow_serving_ip
```
barrucadu: Can we re-use the method from `LearnToRank::Ranker` here?

lib/learn_to_rank/ranker_status.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    def healthy?
+      @healthy && errors.none?
+    end
+
+  private
+
+    attr_reader :timeout
+
+    GOOD_STATES = %w(AVAILABLE InService Updating Creating SystemUpdating RollingBack).freeze
```
barrucadu: We might want to think more about the sagemaker states - eg, `RollingBack` could be cause for a warning, as it means something's gone wrong with the current model.

lib/learn_to_rank/ranker_status.rb
bilbof: Yep, I've moved this into a helper.

lib/learn_to_rank/ranker_status.rb
bilbof: OK. I've removed those AWS states for now, I'll re-add some later when I check Sagemaker health.

lib/learn_to_rank/ranker.rb
```diff
@@ -89,14 +91,8 @@ def fetch_new_scores_from_serving(examples)
...
 
-    def tensorflow_serving_ip
-      if ENV["TENSORFLOW_SERVING_IP"].present?
-        ENV["TENSORFLOW_SERVING_IP"]
-      elsif %w(development).include? ENV["RACK_ENV"]
-        "reranker"
-      else
-        "0.0.0.0"
-      end
+    def log_error(error)
```
barrucadu: I think you've gained an unused function through a rebase

bilbof: Thanks for spotting that - I've removed it!

spec/integration/app/healthcheck_spec.rb
```diff
@@ -27,6 +31,28 @@
...
   end
 
+  describe "#reranker_healthcheck check" do
+    # We only check for cannot connect because govuk_app_config has tests for this
+    context "when reranker healthcheck fails" do
+      before do
+        stub_ranker_container_doesnt_exist
+      end
+
+      it "returns a critical status" do
```
barrucadu: "returns a warning status" - or should the healthcheck actually be critical?

I'm leaning towards a warning being the right status, as search isn't *broken* when the reranker is down, users just get a degraded experience.

barrucadu (approved):
Looks good, there's just one misnamed test

spec/integration/app/healthcheck_spec.rb
bilbof: Thanks, I've fixed that.

PR #1916, opened 2020-01-16
Report LTR errors to Sentry

bilbof:
This will send Learning To Rank reranker errors to staging. Errors will continue to be rescued before reaching users.


lib/learn_to_rank/reranker.rb
```diff
@@ -20,6 +21,7 @@ def rerank(query: "", es_results: [])
       reorder_results(es_results, new_scores)
     rescue StandardError => e
       GovukError.notify(e, extra: { query: query })
```
barrucadu: This `GovukError.notify` needs to go away now, or we'll notify sentry twice.

bilbof: 🤔 I might be wrong, but I think the other cases where we call `GovukError.notify` in `Ranker` are also in rescue blocks, so this call in the `rescue StandardError` block won't get called, but I'd still like to keep it as a catch all for those that don't get rescued.

barrucadu: Yeah, I suspect this block is redundant, but in case it's not we now have

```
rescue StandardError => e
  GovukError.notify(e, extra: { query: query })
  report_error(e, extra: { query: query })
```

And `report_error` also calls `GovukError.notify`

bilbof: Ah sorry you are totally right, I've removed that call.

barrucadu (approved):

PR #1925, opened 2020-01-21
Add timing metrics to all top-level stages of reranking

barrucadu:
Logging only the time to fetch new scores from TensorFlow doesn't give us the whole picture.

### DIFF
--- a/lib/learn_to_rank/reranker.rb
+++ b/lib/learn_to_rank/reranker.rb
@@ -11,9 +11,8 @@ module LearnToRank
     include Errors
     # Reranker re-orders elasticsearch results using a pre-trained model
     def rerank(query: "", es_results: [])
-      feature_sets = FeatureSets.new.call(query, es_results)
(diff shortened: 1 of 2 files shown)


bilbof (approved):
Nice one 👍

lib/learn_to_rank/reranker.rb
```diff
@@ -11,9 +11,8 @@ class Reranker
     include Errors
     # Reranker re-orders elasticsearch results using a pre-trained model
     def rerank(query: "", es_results: [])
-      feature_sets = FeatureSets.new.call(query, es_results)
-      new_scores   = Ranker.new(feature_sets).ranks
-
+      feature_sets = fetch_feature_sets(query, es_results)
```
barrucadu, line 14: Typo this as `es_result` and all the tests return `55` rather than `nil`, that was strange

bilbof: 🤔"all else being **55**, what should rank should this result have?"

PR #1929, opened 2020-01-22
Update internal search for transition

vanitabarrett:


kr8n3r:
does this need updating as well @vanitabarrett 
https://github.com/alphagov/search-api/blob/master/config/finders/news_and_communications_finder.yml#L34 ?

vanitabarrett:
It will do, yep. I can't remember, is that out of scope for our Friday changes @koetsier ?

huwd (approved):

PR #1930, opened 2020-01-22
Display parts on top 3 results only

bilbof:
This will only show parts on the top 3 results, and enforces a limit of 10 parts per result.

The limit could be higher, it was chosen arbitrarily since the average number of parts is 4.

This could also be implemented in `finder-frontend` - doing this in search-api means we can reduce the amount of data unnecessarily returned to finder-frontend.

https://trello.com/c/EMaEQplm/1302-update-display-parts-be-work

### DIFF
--- a/lib/search/presenters/result_presenter.rb
+++ b/lib/search/presenters/result_presenter.rb
@@ -2,11 +2,12 @@ module Search
   class ResultPresenter
     attr_reader :raw_result, :registries, :schema, :search_params
 
-    def initialize(raw_result, registries, schema, search_params)
+    def initialize(raw_result, registries, schema, search_params, result_rank:)
       @raw_result = raw_result
       @registries = registries
       @schema = schema
       @search_params = search_params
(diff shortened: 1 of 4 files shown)


lib/search/presenters/result_presenter.rb
```diff
@@ -2,11 +2,12 @@ module Search
   class ResultPresenter
     attr_reader :raw_result, :registries, :schema, :search_params
 
-    def initialize(raw_result, registries, schema, search_params)
+    def initialize(raw_result, registries, schema, search_params, result_rank:)
```
barrucadu, line 5: Since things will break if this isn't set (comparing `nil` to a number), I think this should be a mandatory argument

bilbof: I think `result_rank` doesn't have a default value here so it's required - it will raise `ArgumentError (missing keyword: result_rank)`

barrucadu (approved):

PR #1934, opened 2020-01-24
Introduce smarter autocomplete results through popularity weighting

Tetrino:
Unlike the actual search functionality, Elasticsearch suggest API does
not allow automatic results ranking on any "suggest" fields used in
the Elasticsearch schema for autocomplete functionality. This combined
with a need for speed resulted in a "dumb" original implementation.

This pull request improves the previous implementation in the following
key ways:

- Allows us to use the suggestion API as provided by Elasticsearch
- Allows us to use the popularity of a given document to weigh
  autocomplete results
- Allows us an easy means to keep popularity of autocomplete results
  concurrent with the base popularity used for ranking.
- Opens the door for potential improvements in using context-based
  suggestion ranking.

The autocomplete field is updated when the update_popularity script is
called, ensuring concurrent values in both fields.

Servers will need to be reindexed and the update_popularity rake task run for these changes to be fully in effect.

Trello: https://trello.com/c/RRpQJpF9/1323-improve-autocomplete-results

### DIFF
--- a/config/schema/elasticsearch_schema.yml
+++ b/config/schema/elasticsearch_schema.yml
@@ -85,19 +85,7 @@ index:
           filter: [trim, lowercase]
           char_filter: [normalize_quotes, strip_quotes]
 
-        edge_ngram_analyzer:
-          filter: [lowercase]
-          tokenizer: edge_ngram_tokenizer
-
-        edge_ngram_search_analyzer:
-          tokenizer: lowercase
-
       tokenizer:
-        edge_ngram_tokenizer:
-            type: edge_ngram
-            min_gram: 1
-            max_gram: 20
-            token_chars: [letter]
 
       char_filter:
         strip_quotes:
--- a/config/schema/field_types.json
+++ b/config/schema/field_types.json
@@ -111,11 +111,6 @@
           "index": true,
           "analyzer": "with_index_synonyms",
           "search_analyzer": "with_search_synonyms"
-        },
-        "edgengram": {
-          "type": "text",
-          "analyzer": "edge_ngram_analyzer",
-          "search_analyzer": "edge_ngram_search_analyzer"
         }
       }
     }
@@ -153,6 +148,18 @@
       "type": "text",
       "index": true,
       "analyzer": "best_bet_stemmed_match"
+    }
+  },
+
+  "autocomplete_result" : {
+    "description": "Type used to apply autocomplete suggestions to a given document based on title headings",
+    "es_config": {
+      "type": "completion",
+      "preserve_separators": false,
+      "preserve_position_increments": false,
+      "max_input_length": 20,
+      "analyzer": "searchable_text",
+      "search_analyzer": "searchable_text"
     }
   },
 
--- a/lib/govuk_index/popularity_worker.rb
+++ b/lib/govuk_index/popularity_worker.rb
@@ -19,12 +19,17 @@ module GovukIndex
 
     def process_record(record, popularities)
       base_path = record["identifier"]["_id"]
+      title = record["document"]["title"]
       OpenStruct.new(
         identifier: record["identifier"].merge("version_type" => "external_gte", "_type" => "generic-document"),
-        document: record["document"].merge(
+        document: record["document"].merge!(
           "popularity" => popularities.dig(base_path, :popularity_score),
           "popularity_b" => popularities.dig(base_path, :popularity_rank),
           "view_count" => popularities.dig(base_path, :view_count),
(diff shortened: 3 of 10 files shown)


config/schema/elasticsearch_schema.yml
```diff
@@ -85,19 +85,7 @@ index:
           filter: [trim, lowercase]
           char_filter: [normalize_quotes, strip_quotes]
 
-        edge_ngram_analyzer:
```
Tetrino, line 88: These were part of the original implementation and are no longer needed.

config/schema/field_types.json
```diff
@@ -111,11 +111,6 @@
           "index": true,
           "analyzer": "with_index_synonyms",
           "search_analyzer": "with_search_synonyms"
-        },
```
Tetrino, line 114: Part of original implementation, no longer needed.

lib/search/query_components/autocomplete.rb
```diff
@@ -1,30 +1,15 @@
 module QueryComponents
   class Autocomplete < BaseComponent
-    AUTOCOMPLETE_FIELD = "title.edgengram".freeze
+    AUTOCOMPLETE_FIELD = "autocomplete".freeze
 
     def payload
       {
-        "bool" => {
```
Tetrino, line 7: Suggest query does not make use of the bool functionality in Elasticsearch. Refined method shouldn't require these  exceptions to avoid certain types of document, however we should be aware of the need to enable context-based suggestions in the future.

lib/govuk_index/popularity_worker.rb
```diff
@@ -19,12 +19,17 @@ def perform(records, destination_index)
...
       OpenStruct.new(
         identifier: record["identifier"].merge("version_type" => "external_gte", "_type" => "generic-document"),
-        document: record["document"].merge(
+        document: record["document"].merge!(
           "popularity" => popularities.dig(base_path, :popularity_score),
           "popularity_b" => popularities.dig(base_path, :popularity_rank),
           "view_count" => popularities.dig(base_path, :view_count),
+          "autocomplete" => { # Relies on updated popularity. Title is for new records.
+            "input" => title,
+            "weight" => popularities.dig(base_path, :popularity_rank),
```
barrucadu, line 31: The lower the rank the more popular the document.  The most popular document is ranked 1, the least popular has rank N (where N is the number of documents on GOV.UK).  So by using the rank as the weight, are we sorting suggestions in *reverse* popularity order?

Should this be `-popularities.dig(base_path, :popularity_rank)` (with some appropriate handling of docs with missing ranks)?

Tetrino: @barrucadu If it should be in reverse, this is not reflected in the results we use server side, see https://www.gov.uk/api/search.json?q=driving&count=10&suggest=autocomplete&fields=popularity_b

It appears if the rank is supposed to be lower the more popular it is, our ranking system has fundamentally failed somewhere. Alternatively this is just poor naming internally and we're getting ranks mixed.

barrucadu: Ah, I remember now, we "fix" the ranking order here: https://github.com/alphagov/search-api/blob/3f4d72c5b1fe97a51fb7e12df34b0773a07578cb/lib/indexer/popularity_lookup.rb#L46

I think this was to avoid exactly the same problem when we were trying out using ranks in popularity boosting.

Tetrino: I'm currently using the final popularity result post fix. Is this unwise if there are changes coming, or is this fine?

barrucadu: I think it's fine

lib/search/query_components/autocomplete.rb
```diff
@@ -1,30 +1,15 @@
...
-              },
-            },
-          },
-          "must_not" => {
-            # The below are excluded from any autocomplete suggestions
-            # If modified remember to update autocomplete_spec!
-            "terms" => {
-              "format" => [
-                #As referenced from config/govuk_index/mapped_document_types.yaml
-                "hmrc_manual_section",
```
sihugh, line 22: Is it worth us trying to still exclude these types from the autocomplete?

Tetrino: @sihugh in testing they simply didn't come up, our scoring system seemingly excludes them from all site search, and it would be beneficial to allow them in the system for future contextual searches (which is apparently supported by the ES API: https://www.elastic.co/guide/en/elasticsearch/reference/6.7/suggester-context.html )

sihugh (approved):

PR #1936, opened 2020-01-24
Enable LTR as default for NDCG

bilbof:
This will use the reranker when computing NDCG, since it is turned on by default now.

The intent is to make the NDCG metric reflect the effect LTR has: https://grafana.blue.production.govuk.digital/dashboard/file/search_relevancy.json?panelId=7&fullscreen

### DIFF
--- a/lib/evaluate/ndcg.rb
+++ b/lib/evaluate/ndcg.rb
@@ -28,7 +28,7 @@ module Evaluate
 
   private
 
-    DEFAULT_PARAMS = { "count" => %w[20], "fields" => %w[link content_id] }.freeze
+    DEFAULT_PARAMS = { "count" => %w[20], "fields" => %w[link content_id], "ab_tests" => %w(relevance:B) }.freeze
 
     attr_reader :data, :field
 


lib/evaluate/ndcg.rb
```diff
@@ -28,7 +28,7 @@ def compute_ndcg
 
   private
 
-    DEFAULT_PARAMS = { "count" => %w[20], "fields" => %w[link content_id] }.freeze
+    DEFAULT_PARAMS = { "count" => %w[20], "fields" => %w[link content_id], "ab_tests" => %w(relevance:B) }.freeze
```
barrucadu, line 31: On line 14 we set `"ab_tests" => [ab_tests]` if any are passed in to `initialize`.  Do we want `relevance:B` to apply in that case as well?  eg, change this line to

```ruby
DEFAULT_AB_TESTS = %w(relevance:B)
DEFAULT_PARAMS = { "count" => %w[20], "fields" => %w[link content_id], "ab_tests" => DEFAULT_AB_TESTS }.freeze
```

And in line 14 merge `ab_tests` and `DEFAULT_AB_TESTS`?

bilbof: This way one can still run NDCG without LTR enabled, do you think it'll be confusing in future when we run NDCG to check other ab tests?

I guess another solution would be to remove the AB test, but it's quite useful for the feature toggle.

barrucadu: I think this is fine for now.  I imagine we'll be removing the feature toggle at some point.

barrucadu (approved):

PR #1939, opened 2020-01-27
Enable fuzzy queries in autocomplete search

Tetrino:
Fuzzy queries allow for users to misspell their searches and still
retrieve meaningful autocomplete results. See ES Api:
https://www.elastic.co/guide/en/elasticsearch/reference/6.7/search-suggesters-completion.html#fuzzy

This pull request also reduces the maximum return count to that agreed
to in design.

Trello: https://trello.com/c/RRpQJpF9/1323-improve-autocomplete-results

### DIFF
--- a/lib/search/query_components/autocomplete.rb
+++ b/lib/search/query_components/autocomplete.rb
@@ -8,8 +8,12 @@ module QueryComponents
           "prefix" => search_term,
           "completion" => {
             "field" => AUTOCOMPLETE_FIELD,
-            "size" => 10,
+            "size" => 8,
             "skip_duplicates" => true,
+            "fuzzy" => {
+              # For completion API we have to explicitly state this
+              "fuzziness" => "AUTO",
+            },
           },
         },
       }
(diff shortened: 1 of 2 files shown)


lib/search/query_components/autocomplete.rb
```diff
@@ -8,8 +8,11 @@ def payload
           "prefix" => search_term,
           "completion" => {
             "field" => AUTOCOMPLETE_FIELD,
-            "size" => 10,
+            "size" => 8,
             "skip_duplicates" => true,
+            "fuzzy" => {
+              "fuzziness" => 2,
```
sihugh: Is there a reason that we're not using `AUTO` for the fuzziness factor?  The docs seem to imply that it's generally advised? https://www.elastic.co/guide/en/elasticsearch/reference/6.7/common-options.html#fuzziness

Tetrino: Good catch, amended.

Tetrino: (For completion we have to explicitly state the factor, even if it's AUTO, so the line itself remains)

sihugh (approved):

lib/search/query_components/autocomplete.rb
```diff
@@ -8,8 +8,12 @@ def payload
           "prefix" => search_term,
           "completion" => {
             "field" => AUTOCOMPLETE_FIELD,
-            "size" => 10,
+            "size" => 8,
             "skip_duplicates" => true,
+            "fuzzy" => {
+              # For completion API we have to explicitly state this even if we leave it at default AUTO
```
barrucadu: If we have to explicitly state it, it's not the default

PR #1941, opened 2020-01-28
Use govuk index for people registry

thomasleese:
This changes the registries to use the `govuk` index when finding people as they have been migrated over. This was missed in https://github.com/alphagov/search-api/pull/1852.

The main change is 2ad83d4 but I also did some refactoring in the earlier commits. 

[Trello Card](https://trello.com/c/yTYIqjiA/1699-5-support-searching-for-documents-tagged-to-roles)

### DIFF
--- a/lib/search/registries.rb
+++ b/lib/search/registries.rb
@@ -21,7 +21,7 @@ module Search
         document_series: registry_for_document_format("document_series"),
         document_collections: registry_for_document_format("document_collection"),
         world_locations: registry_for_document_format("world_location"),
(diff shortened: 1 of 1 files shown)


issyl0 (approved):
The refactoring makes sense. But for the uninitiated, what's a govuk-registry vs. a normal registry?

thomasleese:
The `govuk` index contains documents that were indexed from the message queue, whereas the normal `government` index contains documents that were sent from Whitehall. We're trying to move to a situation where all content comes from the message queue.

PR #1948, opened 2020-01-30
Deployment stability improvements

bilbof:
This makes some changes to how search-api is deployed and started up, to reduce timeout errors that we occasionally see.

Currently when we deploy Nginx sees timeouts when making requests to the new unicorn workers that get spun up.

In order to increase application startup times this makes use of [bootsnap](https://github.com/Shopify/bootsnap) (staging environment):

bootsnap-error-rate

Bootsnap doesn't entirely resolve the issue. The new workers are still running cold - Search API caches configuration files and has big singletons like search config that need instantiating when the first requests come in - this doesn't happen at boot time.

Optimisations included:

- throttle the rate at which workers are spawned. This reduces the likelihood of a user hitting a cold worker, which improves performance for 90% of requests (thanks @barrucadu for the suggestion!).
- Use `preload_app` to speed up server boot times.
- Reduce the timeout of unicorn workers to 15s to match the nginx read timeout, so when nginx abandons a request (e.g. for a retry) the unicorn workers won't keep doing work for that request - done here https://github.com/alphagov/govuk-puppet/pull/10108.

Gradually spawning workers had the biggest impact. For example:

Screenshot 2020-01-30 at 14 42 03

We could also do some refactoring so that we fill the caches and start up all the required singleton classes at boot time. However, I'm not sure those are entirely to blame for the latency of initial requests, since reranker requests are also slower at deploy time. This'd need further investigation if it becomes a bigger problem.

https://trello.com/c/CVrZaaPv/1320


config/unicorn.rb
```diff
@@ -2,4 +2,16 @@
...
+# Preload the entire app. By preloading an application you can save some
+# RAM resources as well as speed up server boot times.
+preload_app true
+
+after_fork do
+  # Throttles the master from starting new workers too quickly.
+  # Creates a more gradual roll out of workers so that users are
+  # less likely to have their request handled by a worker that has
+  # just booted, and may respond to requests slower.
+  sleep 5 unless ENV["RACK_ENV"] == "development"
```
barrucadu, line 16: This definitely seems like a good improvement, my only concern is that unicornherder has a fixed amount of time it waits before it kills off the old processes.  So if we add more workers (or increase this timeout), we run the risk of not having all the new workers up before the old ones go away.

Not a problem now, but something to keep an eye on in the future.

barrucadu (approved):

config/unicorn.rb
bilbof: Yeah I found 10 seconds was too long, because very shortly after the final new worker had started all of the previous workers quit.

PR #1949, opened 2020-01-30
Add a roles facet

thomasleese:
This adds support for being able to search for documents tagged to roles so we can display the announcements list.

So that we can fully migrate roles away from Whitehall (we need to be able to display a list of announcements on each role page).

[Trello Card](https://trello.com/c/yTYIqjiA/1699-5-support-searching-for-documents-tagged-to-roles)


sihugh (approved):

sihugh:
🎢 🙄 🗞

PR #1961, opened 2020-02-04
Update news and coms finder from transition to brexit

huwd:
Trello: https://trello.com/c/Uanimli2/415-change-brexit-to-transition-period-on-search-checkboxes

News and communications finders have a special checkbox to limit all searches to brexit materials.
The brexit taxon has now been moved to transition. Whilst the check continues to work it is now mis-labelled.

This work re-lables the checkbox in search API so that it can be displayed with the new text in finder-frontend.


koetsier (approved):
looks good

koetsier (approved):
👍

PR #1967, opened 2020-02-06
Remove business finder tech debt

huwd:
Trello: https://trello.com/c/qBM5QLID/482-business-finder-tech-debt

Part of the work to clear up behind ourselves - this PR removes all references to the business finder.

Business finder is currently unpublished for users, but a bunch of config, tests and docs still remained.


spec/unit/publishing_api_finder_publisher_spec.rb
```diff
@@ -1,127 +0,0 @@
-require "spec_helper"
-require "publishing_api_finder_publisher"
-
-RSpec.describe PublishingApiFinderPublisher do
```
vanitabarrett, line 4: Looks like we're deleting tests for something which still exists here - should we be deleting the `PublishingApiFinderPublisher` class itself if it's no longer needed?

vanitabarrett (changes requested):
There are quite a few instances of `appear_in_find_eu_exit_guidance_business_finder` still in the search-api codebase. I'm not sure what they were added for - we should confirm that it's ok to remove and delete if so.

Just one other comment about deleting the class associated with the tests we're removing.

laurentqro:
> There are quite a few instances of `appear_in_find_eu_exit_guidance_business_finder` still in the search-api codebase. I'm not sure what they were added for - we should confirm that it's ok to remove and delete if so.

It looks like the `appear_in_find_eu_exit_guidance_business_finder` checkbox approach was replaced the `facet_groups` approach:

* https://github.com/alphagov/finder-frontend/commit/6f484a634
* https://github.com/alphagov/search-api/pull/1535/files

The finder content schema needs to have `appear_in_find_eu_exit_guidance_business_finder` removed too: https://github.com/alphagov/govuk-content-schemas/blob/master/formats/shared/definitions/finder.jsonnet#L51.

laurentqro:
For some additional context about `facet_groups`: https://github.com/alphagov/govuk-content-schemas/pull/878.

spec/unit/publishing_api_finder_publisher_spec.rb
huwd: > There are quite a few instances of `appear_in_find_eu_exit_guidance_business_finder` still in the search-api codebase. I'm not sure what they were added for - we should confirm that it's ok to remove and delete if so.
> 
> Just one other comment about deleting the class associated with the tests we're removing.

@koetsier, humm - you know a bit more about this than I do i think. Can i double check [removing this stuff](https://github.com/alphagov/search-api/blob/58ce0467112f826d73110eaa0bb967d82ebaa7f5/config/schema/field_definitions.json#L819) (and all it's references in schema configs isn't related to [the transition checkbox](https://github.com/alphagov/search-api/blob/0259347bc22882f658c130e40989ab259d97d58d/config/finders/news_and_communications_email_signup.yml#L36)?

... I can't see any overlap, but would like a second opinion?

huwd:
> There are quite a few instances of `appear_in_find_eu_exit_guidance_business_finder` still in the search-api codebase. I'm not sure what they were added for - we should confirm that it's ok to remove and delete if so.
> 
> Just one other comment about deleting the class associated with the tests we're removing.

@koetsier, humm - you know a bit more about this than I do i think. Can i double check [removing this stuff](https://github.com/alphagov/search-api/blob/58ce0467112f826d73110eaa0bb967d82ebaa7f5/config/schema/field_definitions.json#L819) (and all it's references in schema configs isn't related to [the transition checkbox](https://github.com/alphagov/search-api/blob/0259347bc22882f658c130e40989ab259d97d58d/config/finders/news_and_communications_email_signup.yml#L36)?

... I can't see any overlap, but would like a second opinion?

spec/unit/publishing_api_finder_publisher_spec.rb
huwd: So i understand the class is needed. But was only tested through the medium of brexit business finder tests and advanced search tests...

... with both those features deprecated and removed, there are no tests left 😬...
We could try and work through what _is_ left using it and write some new tests for that?

vanitabarrett: @huwd Do you know where this class is used/needed?

huwd:
> > There are quite a few instances of `appear_in_find_eu_exit_guidance_business_finder` still in the search-api codebase. I'm not sure what they were added for - we should confirm that it's ok to remove and delete if so.
> 
> It looks like the `appear_in_find_eu_exit_guidance_business_finder` checkbox approach was replaced the `facet_groups` approach:
> 
>     * [alphagov/finder-frontend@6f484a6](https://github.com/alphagov/finder-frontend/commit/6f484a634)
> 
>     * https://github.com/alphagov/search-api/pull/1535/files
> 
> 
> The finder content schema needs to have `appear_in_find_eu_exit_guidance_business_finder` removed too: https://github.com/alphagov/govuk-content-schemas/blob/master/formats/shared/definitions/finder.jsonnet#L51.

Addressed these:
- [here govuk-content-schemas#961](https://github.com/alphagov/govuk-content-schemas/pull/961) and
- [here search-api#1978](https://github.com/alphagov/search-api/pull/1978)

spec/unit/publishing_api_finder_publisher_spec.rb
huwd: Humm i just checked again, I could have sworn i'd seen something else using the `PublishingApiFinderPublisher` class... but currently on the branch there's nothing.

I'll try removing it and re-running the tests.

huwd: Turns out tests pass... gonna put it on integration and click around a bunch... check nothing is weird... I'm still super suspicious

vanitabarrett (approved):

PR #1972, opened 2020-02-10
Turn /search/advanced into a special route

barrucadu:
We don't use this finder any more, finder-frontend redirects it to
some other finder:
https://github.com/alphagov/finder-frontend/blob/master/app/controllers/redirection_controller.rb#L33

We still need the route, so put it in the SpecialRoutePublisher
instead.


huwd:
Guess these can go too?
- ~~https://github.com/alphagov/search-api/blob/master/doc/advanced-search.md~~ ah you got that one
- https://github.com/alphagov/search-api/blob/master/README.md#additional-docs (advanced search doc is linked?)

huwd:
Can't work out if this is still relevant?
https://github.com/alphagov/search-api/blob/master/doc/publishing-finders.md

barrucadu:
I'll remove the line about advanced search there, but I'll leave the brexit one as the config file is still there (even though it does redirect at the moment)

huwd (approved):

PR #1985, opened 2020-02-17
Better manage fields from finder schemas

Tetrino:
Removes the requirement for a blacklist in the email signup content
presenter by ensuring that we always filter out any content that does
not exist in the current schema through direct reference to the
schema via the govuk_schemas gem.

https://trello.com/c/U9lY0LEt/1376-better-manage-fields-from-finder-schemas-to-email-signup-content-item-presenter


lib/publishing_api_finder_publisher.rb
```diff
@@ -98,12 +99,11 @@ def initialize(schema, timestamp)
 
   def details
     details = schema.fetch("details", {})
-    details.merge(
+    details.slice(GovukSchemas::Schema.find(publisher_schema: "finder_email_signup"))
```
bilbof, line 104: It might be nice to put this part in a separate method, so that you can explain what it does in a method name. Does `details.slice(schema)` give you the details hash with unpermitted keys removed?

lib/publishing_api_finder_publisher.rb
```diff
@@ -1,4 +1,5 @@
 require "gds_api/publishing_api_v2"
+require "govuk_schemas"
```
bilbof, line 2: You'll also need to move `govuk_schemas` out of the test group in the Gemfile https://github.com/alphagov/search-api/blob/7f9346010101255e6ebd43bbf948a972b68ef1b2/Gemfile#L41-L45

bilbof (commented):
Looks good just a couple comments

lib/publishing_api_finder_publisher.rb
Tetrino: Added a comment to better explain what's happening here.

bilbof (approved):

PR #1987, opened 2020-02-18
Serve model variant b

bilbof:
This enables deploying a model using a model's tag name, which will create a separate new sagemaker endpoint, keeping the others intact.

It doesn't handle removing old sagemaker endpoints though. So we still need to figure out how that'd happen.

I see a deploy of a new candidate model & graceful migration to the new model working like this:

1. Make necessary changes to the search api application.
2. Train a new model (ideally with the data pipeline) and upload it to s3.
3. Add a new concourse deploy job to the pipeline that takes a `model_tag` for your model
4. Run the concourse job to deploy your model.
5. Once an AB test is completed confirming we want to migrate to the candidate model, send all traffic to this model, remove the old endpoint, and amend the concourse pipeline to train the new model on a schedule.

Some extra work needs to be done in Search API to permit training/serving different models e.g. those with different features (possibly by passing a model_variant param in the training / query requests).

This change should suffice for comparing two pre-trained models we have ready to AB test.

https://trello.com/c/1EjO05Dj/1333

### DIFF
--- a/ltr/concourse/deploy.py
+++ b/ltr/concourse/deploy.py
@@ -8,13 +8,17 @@ import time
 
 govuk_environment = os.environ["GOVUK_ENVIRONMENT"]
 role = os.environ["ROLE_ARN"]
-model_name = os.environ["SCRIPT_INPUT_DATA"].strip()
+model_name = os.getenv("SCRIPT_INPUT_DATA")
+model_tag = os.getenv("MODEL_TAG")
+if not model_name and not model_tag:
+    raise Exception("Require model_name or model_tag to be set.")
 
+endpoint_version = f"-{model_tag}" if model_tag else ""
 s3_bucket = os.getenv("S3_BUCKET", f"govuk-{govuk_environment}-search-relevancy")
 default_initial_instance_count = int(os.getenv("INITIAL_INSTANCE_COUNT", 3))
 default_instance_type = os.getenv("INSTANCE_TYPE", "ml.t2.medium")
 endpoint_name = os.getenv(
-    "ENDPOINT_NAME", f"govuk-{govuk_environment}-search-ltr-endpoint"
+    "ENDPOINT_NAME", f"govuk-{govuk_environment}-search-ltr-endpoint{endpoint_version}"
 )
 
 session = sagemaker.Session()
@@ -40,12 +44,27 @@ except Exception:
     print("could not find current endpoint, will create a new one", file=sys.stderr)
 
 # find the model
-model_keys = session.boto_session.client("s3").list_objects_v2(
-    Bucket=s3_bucket, Prefix=f"model/{model_name}"
-)["Contents"]
+
+model_location_prefix = f"model/{(model_tag or model_name).strip()}"
+print(f"looking with prefix {model_location_prefix}", file=sys.stderr)
+
+try:
+    model_keys = session.boto_session.client("s3").list_objects_v2(
+        Bucket=s3_bucket, Prefix=model_location_prefix
+    )["Contents"]
+except KeyError:
+    print(f"Couldn't find the model in {model_location_prefix}", file=sys.stderr)
+    sys.exit(1)
+
 if len(model_keys) != 1:
-    raise Exception("Expected exactly 1 model key")
+    print(f"Found too many models at {model_location_prefix}! Found models:", file=sys.stderr)
+    for key in model_keys:
+        print(f"- {key['Key']}", file=sys.stderr)
+    sys.exit(1)
+
 model_key = model_keys[0]["Key"]
+
+print(f"Deploying model {model_key} to Sagemaker...", file=sys.stderr)
 
 # deploy the model by creating a new endpoint config and updating the
 # existing endpoint (or creating a new one)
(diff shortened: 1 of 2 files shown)


ltr/concourse/deploy.py
```diff
@@ -10,11 +10,17 @@
 role = os.environ["ROLE_ARN"]
 model_name = os.environ["SCRIPT_INPUT_DATA"].strip()
 
+try:
+    model_tag = os.environ["MODEL_TAG"]
+except KeyError:
+    model_tag = ""
```
barrucadu: This can be `os.getenv("MODEL_TAG", "")`

More generally, you can do `dict.get(key, default)` - but there's a specific function in `os` for environment variables, so it's best to use that one here.

barrucadu: Even better, you can just use `os.getenv("MODEL_TAG")` here, and then change line 49 to `model_location = model_tag if model_tag else model_name` - `None` is falsey.

ltr/concourse/deploy.py
```diff
@@ -40,13 +46,16 @@
     print("could not find current endpoint, will create a new one", file=sys.stderr)
 
 # find the model
+model_location = model_tag if model_tag != "" else model_name
 model_keys = session.boto_session.client("s3").list_objects_v2(
-    Bucket=s3_bucket, Prefix=f"model/{model_name}"
+    Bucket=s3_bucket, Prefix=f"model/{model_location}"
```
barrucadu: You could even put the `model_location` definition in here if you wanted, making this expression:

```
model_keys = session.boto_session.client("s3").list_objects_v2(
    Bucket=s3_bucket, Prefix=f"model/{model_tag if model_tag else model_name}"
)["Contents"]
```

ltr/concourse/deploy.py
```diff
@@ -40,13 +46,16 @@
...
 )["Contents"]
 if len(model_keys) != 1:
     raise Exception("Expected exactly 1 model key")
 model_key = model_keys[0]["Key"]
 
+print(f"Deploying model {model_key} (found at model/{model_location}) to Sagemaker...")
```
barrucadu: Can this print to `sys.stderr` like the other ones?

ltr/concourse/pipeline.yaml
```diff
@@ -75,6 +75,13 @@ resources:
     bucket: ((readonly_private_bucket_name))
     region_name: eu-west-2
     regexp: search-learn-to-rank/integration-model-(.*).txt
+# AB Test variant B (list_mle_model)
+- name: integration-model-variant-b
+  type: s3-iam
+  source:
+    bucket: ((readonly_private_bucket_name))
+    region_name: eu-west-2
+    regexp: search-learn-to-rank/integration-model-variant-b-(.*).txt
```
barrucadu: I think this regex would clash with the `search-learn-to-rank/integration-model-(.*).txt` one - but since the `(.*)` part is always a UNIX timestamp (see https://github.com/alphagov/search-api/blob/master/ltr/concourse/task.sh#L39), these `(.*)`s could be replaced with `([0-9]*)`s

ltr/concourse/deploy.py
bilbof: I want to print the location later so will keep it separate for now.

bilbof:
@barrucadu thanks for your review, I made those changes

ltr/concourse/pipeline.yaml
```diff
@@ -75,6 +75,13 @@ resources:
     bucket: ((readonly_private_bucket_name))
     region_name: eu-west-2
     regexp: search-learn-to-rank/integration-model-(.*).txt
+# AB Test variant B (list_mle_model)
+- name: integration-model-variant-b
+  type: s3-iam
+  source:
+    bucket: ((readonly_private_bucket_name))
+    region_name: eu-west-2
+    regexp: search-learn-to-rank/integration-model-variant-b-([0-9]*).txt
```
barrucadu: I think the other regexes will need to change too (or at least the `search-learn-to-rank/integration-model-(.*).txt` one)

ltr/concourse/deploy.py
```diff
@@ -40,13 +42,16 @@
...
+model_location = model_tag or model_name
 model_keys = session.boto_session.client("s3").list_objects_v2(
-    Bucket=s3_bucket, Prefix=f"model/{model_name}"
+    Bucket=s3_bucket, Prefix=f"model/{model_location}"
 )["Contents"]
 if len(model_keys) != 1:
     raise Exception("Expected exactly 1 model key")
 model_key = model_keys[0]["Key"]
 
+print(f"Deploying model {model_key} (found at model/{model_location}) to Sagemaker...", file=sys.stderr)
```
barrucadu: `{model_key}` will actually have the full s3 path of the thing

bilbof: Ah cool, I don't need to set `model_location` :)

barrucadu (approved):

ltr/concourse/deploy.py
```diff
@@ -42,11 +44,24 @@
     print("could not find current endpoint, will create a new one", file=sys.stderr)
 
 # find the model
-model_keys = session.boto_session.client("s3").list_objects_v2(
-    Bucket=s3_bucket, Prefix=f"model/{model_tag or model_name}"
-)["Contents"]
+print(f"looking with prefix model/{(model_tag or model_name).strip()}", file=sys.stderr)
+
+# TODO: exception handling for no Contents
```
barrucadu: TODO: remove todo

ltr/concourse/pipeline.yaml
```diff
@@ -47,6 +47,7 @@ resources:
   type: git
   source:
     uri: https://github.com/alphagov/search-api.git
+    branch: serve-model-variant-b
```
barrucadu: Don't commit this

bilbof:
Looks like this was already merged, but that isn't showing up here.

PR #1990, opened 2020-02-19
Add Healthcheck for Elasticsearch node drive space

Tetrino:
Typically we only discover that Elasticsearch has run out of diskspace
when we start to see errors. This usually happens because there are old
Elasticsearch indices that have not been removed with the relevant
rake task clogging up the system.

This healthcheck will test the overall space available and cause an
alert if we have less than 20% free on the disks. A percentage decided upon based on current usage trends.

https://trello.com/c/NREu7dvS/1380-alerts-for-duplicate-elasticsearch-indices

### DIFF
--- a/lib/healthcheck/elasticsearch_index_diskspace_check.rb
+++ b/lib/healthcheck/elasticsearch_index_diskspace_check.rb
@@ -1,0 +1,57 @@
+require "govuk_app_config"
+
+module Healthcheck
+  # This is a custom check that is called by GovukHealthcheck
+  # See GovukHealthcheck (govuk_app_config/docs/healthchecks.md) for usage info
+  class ElasticsearchIndexDiskspaceCheck
+    def name
+      :elasticsearch_diskspace
+    end
+
+    def status
+      low_diskspace? ? :critical : :ok
+    end
+
+    def message
+      if low_diskspace?
+        "there is not enough diskspace for the elasticsearch indices! Consider running index cleanup rake task"
+      else
+        "there is enough diskspace for the elasticsearch indices"
+      end
+    end
+
+    def details
+      low_diskspace?
+      { "extra" => "total free space remaining on nodes: #{percentage_free}%" }
+    end
+
+    # Optional
+    def enabled?
+      true # false if the check is not relevant at this time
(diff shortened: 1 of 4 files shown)


lib/healthcheck/elasticsearch_index_diskspace_check.rb
```diff
@@ -0,0 +1,60 @@
...
+      if diskspace_free?
+        "there is enough diskspace for the elasticsearch indices"
+      else
+        "there is not enough diskspace for the elasticsearch indices! Consider running index cleanup rake task"
+      end
+    end
+
+    def details
+      diskspace_free?
+      { "extra" => "total free space remaining on nodes: #{@percentage}%" }
```
bilbof: It looks like the method `diskspace_free?` could be split up. Rather than set instance variables in the initialize method, you could memoize the methods that provide the data for those variables. This would mean your methods don't have side effects, and make it the calls a bit more transparent. For example:

```ruby
LOW_SPACE_THRESHOLD = 20 # percent

def cluster_stats
  @cluster_stats ||= begin
    client = Services.elasticsearch(cluster: Clusters.default_cluster)
    es_stats = client.perform_request "GET", "_nodes/stats/fs?pretty=true"
    es_stats.body.dig("nodes").each_with_object({ total: 0, avail: 0 }) do |(_, node_stat), hsh|
      hsh[:total] += node_stat.dig("fs", "total", "total_in_bytes")
      hsh[:avail] += node_stat.dig("fs", "total", "available_in_bytes")
    end
  end
end

def percentage_free
  cluster_stats[:avail] / (cluster_stats[:total] / 100)
end

def low_diskpace?
  percentage_free < LOW_SPACE_THRESHOLD
end
```

bilbof (approved):
Looks good 👍one comment on the methods in the healthcheck class

lib/healthcheck/elasticsearch_index_diskspace_check.rb
Tetrino: Good call, I'm still getting my head around some aspects of Ruby's function handling, took the easy route!

PR #1991, opened 2020-02-19
Allow users to choose Sagemaker endpoint that handles their request

bilbof:
This enables Learning to rank for all queries. The query parameter `ab_tests=relevance:B` will no longer have an effect. One can disable reranking by using the parameter `ab_tests=relevance:disable`.

This adds new functionality: users of search api can set `ab_tests=mv:<model variant>`, which will be used by search-api to determine which model should handle the request.

For example, if you have added a model endpoint `B` to Sagemaker, then you can enable `ab_tests=mv:B` by adding 'B' to the permitted model variants in Search API.

https://trello.com/c/1EjO05Dj/1333


Tetrino (approved):

bilbof:
Sorry @Tetrino I added a small commit.

PR #2001, opened 2020-02-25
Automate deletion of old unused indexes

Tetrino:
We were having issues wherein the reliance on manual deletion of old
indexes from the Elasticsearch system was clogging up the storage space
of our cluster. This commit adds a Rake task designed to mitigate this
problem.

The task is designed to be run on schedule in Jenkins. It relies on the
ability to source the "updated_at" value from the last edited record on
a given index. With this in mind, we have two steps.

**1. We gather all indices under the given index name, ignoring the last
two that were created, into a map. The last two that were created are
almost universally guaranteed to be the currently used index, and
the index it replaced.**

**2. Based on whether we have access to an "updated_at" field on the given
index, we either:**

- Check if the age of the index as given by the most recent "updated_at"
 value is older than or equal to N days as given by the environment variable
 MAX_INDEX_AGE and if it is, delete it. Current indexes: government,
 govuk, detailed

- If we do not have access to the field, we simply remove the extra
wasted indexes as we have no way to accurately gauge the age of them.
Current indexes: metasearch, page-traffic

This should effectively manage overall storage use on our cluster, once
the relevant Jenkins job is created.

https://trello.com/c/spNtyEbO/1403-automate-deletion-of-old-unused-indexes

### DIFF
--- a/lib/index_group.rb
+++ b/lib/index_group.rb
@@ -11,9 +11,9 @@ module SearchIndices
     def initialize(base_uri, name, schema, search_config)
       @base_uri = base_uri
       @client = Services::elasticsearch(hosts: base_uri)
-      # Index creation can take longer than other requests, so create a separate
+      # Index creation/deletion can take longer than other requests, so create a separate
       # client with a longer timeout
-      @index_creation_client = Services::elasticsearch(hosts: base_uri, timeout: 10)
+      @long_timeout_client = Services::elasticsearch(hosts: base_uri, timeout: 30)
       @name = name
       @schema = schema
       @search_config = search_config
@@ -31,7 +31,7 @@ module SearchIndices
         "settings" => settings,
         "mappings" => mappings,
       }
-      @index_creation_client.indices.create(
+      @long_timeout_client.indices.create(
         index: index_name,
         body: index_payload,
       )
@@ -104,7 +104,21 @@ module SearchIndices
       end
     end
 
+    def timed_clean(day_limit)
+      cleaned_aliases.each do |name, details|
+        last_update = find_last_update(name)
+        if last_update.nil?
+          logger.info "Auto-cleaning index #{name} which is unused and can not have its age verified"
+          delete(name) # Some indexes will never have this, we already filter them in clean step so remove spares
+          next
+        end
+
+        delete_based_on_age_check(name, details, day_limit, last_update)
+      end
+    end
+
   private
+
 
     def logger
       Logging.logger[self]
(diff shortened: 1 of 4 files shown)


lib/index_group.rb
```diff
@@ -152,7 +166,48 @@ def name_pattern
 
     def delete(index_name)
       logger.info "Deleting index #{index_name}"
-      @client.indices.delete(index: index_name)
+      @long_timeout_client.indices.delete(index: index_name)
+    end
+
+    # Sort the index map by date, remove the most recent two (active and latest inactive) and return hash
+    def cleaned_aliases
```
sihugh, line 173: I'd love some tests for this in general, but especially around ensuring that we discard the correct indices from the list of those that we want to delete 😱

lib/index_group.rb
```diff
@@ -152,7 +166,48 @@ def name_pattern
...
+    end
+
+    def delete_based_on_age_check(name, day_limit, last_update)
+      age = Time.now.to_i - last_update.to_datetime.to_f # Everything is handled in seconds, thanks Ruby!
+      days = age.to_i / 86_400 # One day in seconds
+
+      return unless days >= day_limit.to_i
+
+      logger.info "Auto-cleaning index #{name} which is unused and #{days} days old"
+      delete(name)
```
sihugh: The other `clean` method has a guard clause on the delete:
```
delete(name) if details.fetch("aliases", {}).empty?
```
Is this something we want to consider?

Tetrino: Done

lib/index_group.rb
Tetrino: Honestly I wasn't too happy with this logic myself the more I thought about it, so was holding back on tests until I had some comment one way or the other as it could change how the test logic is written.

I'm going to put some guards in here. It'll make everything so much safer!

lib/index_group.rb
```diff
@@ -11,9 +11,9 @@ class IndexGroup
     def initialize(base_uri, name, schema, search_config)
       @base_uri = base_uri
       @client = Services::elasticsearch(hosts: base_uri)
-      # Index creation can take longer than other requests, so create a separate
+      # Index creation/deletion can take longer than other requests, so create a separate
       # client with a longer timeout
-      @index_creation_client = Services::elasticsearch(hosts: base_uri, timeout: 10)
+      @long_timeout_client = Services::elasticsearch(hosts: base_uri, timeout: 30)
```
Tetrino, line 16: Had to extend this as the deletion calls could time out even with a 10 second leeway.

lib/tasks/indices.rake
```diff
@@ -186,6 +186,15 @@ this task will run against all active clusters.
     end
   end
 
+  desc "Cleans out unused indices older than the number of days given by the MAX_INDEX_AGE flag"
+  task :timed_clean do
+    Clusters.active.each do |cluster|
+      index_names.each do |index_name|
+        search_server(cluster: cluster).index_group(index_name).timed_clean(max_index_age)
```
sihugh, line 193: Does this work?  I'm wondering how `max_index_age` is assigned

Tetrino: It's a flag the same way SEARCH_INDEX is. So example usage:

`rake search:timed_clean SEARCH_INDEX=all MAX_INDEX_AGE=7`

Works as designed locally, complete with yelling at you if you don't use that flag. 👍

sihugh (approved):

PR #2010, opened 2020-03-04
Remove tensorflow/serving stuff for production

barrucadu:
Previously, when sagemaker isn't configured, search-api would assume it can talk to a tensorflow serving container, on 0.0.0.0 if all else fails.  This PR makes it instead default to not reranking at all
if neither sagemaker nor serving have their env vars set.

Also removes a now-redundant rake task.

---

[Trello card](https://trello.com/c/iT2Mh8u6/1413-remove-dockerised-model-and-code)


sihugh (approved):
Very neat 👍

PR #2024, opened 2020-03-17
Create a new database index for search-suggestions

Tetrino:
Creates an index designed to store metrics for search suggestions,
with initial design relying on search terms, unique search occurrences
and search exit percentage.

`./spec/integration/search/batch_search_spec.rb` tests fail currently for an unknown reason, likely not registering something correctly as it's not seeing the new index.

PR made purely for feedback and suggestions.


sihugh:
Closing for clarity as suggestions are being worked on separately by another team at the moment

PR #2033, opened 2020-03-23
Do LTR data generation by scaling up an ASG, SSHing in, and running the rake tasks

barrucadu:
See also https://github.com/alphagov/govuk-aws/pull/1278

---

Currently training data is generated by sending an HTTP request to
search-api and waiting. But we want to now start producing more
training data, which will significantly slow that down. We don't want
to put lots of extra strain on a live search-api instance, so instead
here's how it's going to work:

1. Concourse will scale up the ASG to 1 member
2. Concourse will SSH in to the new instance, then:
    1. Clone search-api from github
    2. Install dependencies
    3. Run the training data generation rake task
3. Concourse will scale down the ASG to 0 members

---

[Trello card](https://trello.com/c/IsptO4Kk/1451-implement-elephant-model-changes)


Tetrino (approved):
Moving stuff from Ruby land to Python land is always a bit terrifying when you see the diff, but I can't see anything off here. 👍

PR #2036, opened 2020-03-24
Add coronavirus page format

sihugh:
This is the coronavirus landing page which we're working on publishing.

If format boosting is used then it should be similar to other navigation pages.

https://trello.com/c/0uZVAYYt/35-look-at-options-for-publishers-to-update-page


hannako (approved):
🤞

PR #2040, opened 2020-03-27
Cache index names for 30s

barrucadu:
This should reduce load on elasticsearch, as we don't need to fetch
the real name of every alias on every query.  On the other hand,
getting the index name of an alias is a cheap operation, so it may not
save much.

---

[Trello card](https://trello.com/c/5zpvZfi9/1455-see-if-there-are-any-low-hanging-fruit-to-reduce-load-on-elasticsearch)

### DIFF
--- a/lib/cache.rb
+++ b/lib/cache.rb
@@ -8,6 +8,12 @@ class Cache
 
   @cache = {}
 
+  @redis_cache = Redis.new(
+    host: ENV.fetch("REDIS_HOST", "127.0.0.1"),
+    port: ENV.fetch("REDIS_PORT", 6379),
+    namespace: "search-api-cache",
+  )
+
   def self.get(key)
     if @cache.has_key?(key)
       @cache[key]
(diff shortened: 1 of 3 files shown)


bilbof (commented):
Could there be a situation (e.g. during reindexing) where queries would fail because the index aliases are out of date?

barrucadu:
I don't think indexing will be affected (because that doesn't use this code), but querying will temporarily break if you delete the old indices under 30 seconds after the alias switches over, yes.

Tetrino:
> I don't think indexing will be affected (because that doesn't use this code), but querying will temporarily break if you delete the old indices under 30 seconds after the alias switches over, yes.

How temporary is the break? I think it worth the small risk if it's just that 30 seconds, or is the break longer?

barrucadu:
It's just that 30s, as after that the redis key will expire and it'll be fetched by querying elasticsearch again.

bilbof:
> querying will temporarily break if you delete the old indices under 30 seconds after the alias switches over

Perhaps we could update the cached index aliases as part of the reindexing job. It would be ideal if we could prevent user requests from failing.

barrucadu:
Going to close this for now, because it would be good to do some performance testing, which I don't have time for at the moment.

PR #2047, opened 2020-03-31
Don't fetch content for Welsh attachments

barrucadu:
We don't index non-English content in search, as we have no way of
scoping a search by language.


Tetrino (approved):

barrucadu:
I'll add a test for this before merging it

PR #2050, opened 2020-04-01
Add new synonyms for misspellings and refining COVID-19 synonyms

Tetrino:


barrucadu (approved):
So many misspellings!

Tetrino:
> So many misspellings!

So many that I had to remove some duplicates!

PR #2051, opened 2020-04-03
Change covid synonyms

barrucadu:
Giving "covid-19" as an option means that all covid-related searches
match documents which contain the string "19" in them, which isn't
great.  Since "covid" matches documents that contain "covid",
"covid-19", and "covid 19", I think we should use that instead.


sihugh (approved):
Terrific, thanks Michael

PR #2065, opened 2020-04-23
Only turn attachments into parts if the URL matches the parent

barrucadu:
We don't index part URLs, only slugs, and assume that all parts have a
URL of the form `{parent_url}/{slug}`.

This is usually correct, but not for HTML attachments on statistics
pages.  There the parent page is under /government/statistics/..., but
the attachment is under /government/publications/....  We could fix
that by storing URLs, but that'd require reindexing a lot of stuff, so
for now let's just go for the simpler solution of rejecting those
attachments.

We'll need to reindex after deploying this.  I think this is the task we want: https://github.com/alphagov/whitehall/blob/master/lib/tasks/search.rake#L80

### DIFF
--- a/lib/govuk_index/payload_preparer.rb
+++ b/lib/govuk_index/payload_preparer.rb
@@ -24,7 +24,10 @@ module GovukIndex
     def prepare_parts(payload)
       return payload unless payload["details"].fetch("parts", []).empty?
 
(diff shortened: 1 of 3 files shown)


thomasleese (approved):
Thanks!

PR #2066, opened 2020-04-23
Fix errors introduced in 96ad601

barrucadu:
https://deploy.blue.production.govuk.digital/job/run-rake-task/65917/console
https://deploy.blue.production.govuk.digital/job/run-rake-task/65916/console


benthorner (approved):
Ah, the foibles of top-level methods (and untested rake tasks) 😉.

Thanks for spotting and fixing!

barrucadu:
I introduced a new linting error 😱

PR #2071, opened 2020-04-24
Cache index names for 1 hour

barrucadu:
This should reduce load on elasticsearch, as we don't need to fetch
the real name of every alias on every query.  On the other hand,
getting the index name of an alias is a cheap operation, so it may not
save much.

---

#2040 but rebased and with a 1-hour timeout.


sihugh:
Closing for now.

PR #2097, opened 2020-05-19
Boost content tagged to the coronavirus taxon

sihugh:
We're gradually moving away from the topical event, but we need to retain
the search strategy.

We'll remove the topical event boost when we actually retire it.

Related to https://github.com/alphagov/search-api/pull/2035

### DIFF
--- a/config/query/boosting.yml
+++ b/config/query/boosting.yml
@@ -34,8 +34,6 @@ base:
     devolved: 0.3
   is_historic:
     true: 0.5
-  topical_events:
-    coronavirus-covid-19-uk-government-response: 5
   part_of_taxonomy_tree:
     5b7b9532-a775-4bd2-a3aa-6ce380184b6c: 5 # /coronavirus-taxon
 # Overrides to apply to the sitemap priorities in external search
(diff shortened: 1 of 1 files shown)


config/query/boosting.yml
```diff
@@ -36,6 +36,8 @@ base:
     true: 0.5
   topical_events:
     coronavirus-covid-19-uk-government-response: 5
+  part_of_taxonomy_tree:
+    5b7b9532-a775-4bd2-a3aa-6ce380184b6c: 5 # /coronavirus-taxon
```
barrucadu, line 38: This means things tagged to both will be double-boosted

barrucadu: But I don't think we have a way around that easily

barrucadu (approved):

config/query/boosting.yml
sihugh: I had just about worked that out, and was trying to decide whether that was a problem 🤔

sihugh: They're close enough that I can retire the topical event one I think.

sihugh: I've now dropped the topical event boost

barrucadu (approved):

sihugh:
Thanks @barrucadu

PR #2098, opened 2020-05-19
Increase Sagemaker instance count

bilbof:
We've seen higher traffic volumes affecting reranker request latency so this is a reaction to that.

### DIFF
--- a/ltr/concourse/deploy.py
+++ b/ltr/concourse/deploy.py
@@ -15,7 +15,7 @@ if not model_name and not model_tag:
 
 endpoint_version = f"-{model_tag}" if model_tag else ""
 s3_bucket = os.getenv("S3_BUCKET", f"govuk-{govuk_environment}-search-relevancy")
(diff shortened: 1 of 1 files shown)


sihugh (approved):

ltr/concourse/deploy.py
```diff
@@ -15,7 +15,7 @@
 
 endpoint_version = f"-{model_tag}" if model_tag else ""
 s3_bucket = os.getenv("S3_BUCKET", f"govuk-{govuk_environment}-search-relevancy")
-default_initial_instance_count = int(os.getenv("INITIAL_INSTANCE_COUNT", 3))
+default_initial_instance_count = int(os.getenv("INITIAL_INSTANCE_COUNT", 4))
```
barrucadu, line 18: This won't actually scale up the number of instances we currently have - `default_initial_instance_count` is only used if an endpoint doesn't currently exist (see line 38)

bilbof: Thanks @barrucadu, I've made the change through the AWS UI, too.

PR #2109, opened 2020-06-01
Remove redundant RuboCop config

benthorner:
https://github.com/alphagov/rubocop-govuk/issues/73


lib/search/query_parameters.rb
```diff
@@ -95,11 +95,11 @@ def model_variants
   private
 
     def determine_if_quoted_phrase
-      if @query =~ QUOTED_STRING_REGEX
-        @is_quoted_phrase = true
-      else
-        @is_quoted_phrase = false
-      end
+      @is_quoted_phrase = if @query =~ QUOTED_STRING_REGEX
```
thomasleese:
```suggestion
      @is_quoted_phrase = @query =~ QUOTED_STRING_REGEX
```

benthorner: Oh good spot - updated.

benthorner: Turns out if was more complicated than I thought. Should _actually_ be fixed now.

thomasleese (approved):

PR #2131, opened 2020-06-19
Add document parts to sitemap

sihugh:
We want to be able to include "parts" in our sitemaps because they are important (and when viewed externally, separate) pages from the pages we present as results in site search.

We also want to be able to give them a priority, so that external searches will hopefully understand how our site is structured better than before.  At present, HTML publications (treated as a `part` within the search index) are not in the sitemap, so when crawled organically (by a crawler following links), they will probably get the default weight applied to them.

This may sometimes be higher than the document they've been linked from (for example, if the "parent" publication is withdrawn, then it is down-weighted).  We want to address this, and ensure that any weighting boost/suppression of a parent doc is also applied to its parts.

There are a couple of performance tweaks in here as well, which (on integration at least), reduce the time to generate the sitemaps from ~9 minutes to less than 4 minutes.

https://trello.com/c/RCkUOYwp/343-check-html-publication-weighting-in-sitemaps

### DIFF
--- a/lib/sitemap/generator.rb
+++ b/lib/sitemap/generator.rb
@@ -43,9 +43,9 @@ module Sitemap
         xml.urlset(xmlns: "http://www.sitemaps.org/schemas/sitemap/0.9") do
           documents.each do |document|
             xml.url do
-              xml.loc document.url
-              xml.lastmod document.last_updated if document.last_updated
-              xml.priority document.priority
+              xml.loc document[:url]
+              xml.lastmod document[:last_updated] if document[:last_updated]
+              xml.priority document[:priority]
             end
           end
         end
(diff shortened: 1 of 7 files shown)


lib/sitemap/generator.rb
```diff
@@ -83,7 +83,23 @@ def get_all_documents
...
+
+          parts.each do |part|
+            part_document = document["_source"].merge(
+              "is_part" => true,
+              "link" => document["_source"]["link"] + "/" + part["slug"],
+            )
+
+            chunk << SitemapPresenter.new(part_document, property_boost_calculator)
+
+            if chunk.size >= SITEMAP_LIMIT
```
hannako, line 100: Not sure I follow why it was `if chunk.size == SITEMAP_LIMIT` before but `>=` now

hannako (approved):
Just one comment where maybe I've not followed the logic. But other than that looks good :+1

lib/sitemap/generator.rb
sihugh: Ah yes, I think I was worried at one point that I might miscount somehow and potentially end up exceeding the limit (by adding more than one part before checking the chunk size for example). 
It felt safer (for this, and future changes) to make the check do `>=` so that it would be able to work even if I'd cocked up the count somewhere.

PR #2144, opened 2020-06-30
Increase training data for LTR model

bilbof:
This will train the model on substantially more data (90 days), which
we believe will make it more accurate at reranking results.
We ran an AB test (Elephant) which increased click through
rate on the top 1,3,5 results, so this ships that model.

This intends to train the model on Sundays at 10pm.

Since we are now training the model with substantially
more data, this takes a lot longer so it's better to
do this at the weekend.

### DIFF
--- a/lib/learn_to_rank/data_pipeline/bigquery.rb
+++ b/lib/learn_to_rank/data_pipeline/bigquery.rb
@@ -4,7 +4,7 @@ module LearnToRank::DataPipeline
   module Bigquery
     def self.fetch(credentials)
       now = Time.now
-      before = now - 2 * 7 * 24 * 60 * 60
+      before = now - 90 * 24 * 60 * 60
       sql = "SELECT * FROM (
   SELECT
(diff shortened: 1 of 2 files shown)


lib/learn_to_rank/data_pipeline/bigquery.rb
```diff
@@ -4,7 +4,7 @@ module LearnToRank::DataPipeline
   module Bigquery
     def self.fetch(credentials)
       now = Time.now
-      before = now - 2 * 7 * 24 * 60 * 60
+      before = now - 3 * 30 * 24 * 60 * 60
```
barrucadu: I think `90 * 24 * 60 * 60` would be marginally clearer here.

bilbof: Thanks, I've made that change!

barrucadu (approved):

sihugh (approved):

PR #2161, opened 2020-07-24
Train LTR model on 300 days

bilbof:
This uses 300 days of data to train the model on. The assumption is that more data is better, and that data between August (when we began collecting this data) and December was of greater volume and potentially higher quality. We'll monitor our online and offline metrics to see if this change has an impact.


sihugh (approved):

sihugh:
Only if we call the model "Sparta"

PR #2169, opened 2020-08-04
Revert "Combine results at the same rank for relevancy scoring"

bilbof:
I'm investigating why Search CTR and NDCG declined in March 2020. This coincided with changes in user behaviour in response to COVID-19, so changes may be required to support the new behaviour. However, as a first pass I'm investigating whether any releases correlated with the decrease in our offline and online metrics, starting with PRs #2041 and #2044. 

We found that reverting commit c3f14e86205f67895a7810faa052559f2b08ac81 (PR #2041) improves NDCG@1,10,20 by 1-2% and has a neutral effect on NDCG@3,5.

As we don't have a search team at the moment we're not running proper AB tests so will monitor the impact of this on CTR on 100% of traffic.

Graphs ([view in Grafana](https://grafana.integration.publishing.service.gov.uk/dashboard/file/search_relevancy.json?panelId=7&fullscreen&orgId=1&from=now-30d&to=now&edit))

Screenshot 2020-08-04 at 15 10 05
Screenshot 2020-08-04 at 15 10 23

### DIFF
--- a/lib/learn_to_rank/data_pipeline/relevancy_judgements.rb
+++ b/lib/learn_to_rank/data_pipeline/relevancy_judgements.rb
@@ -36,7 +36,8 @@ module LearnToRank::DataPipeline
     # a set of judgements between 0 and 3.
     def judgements(query, documents)
       clicks_above_position = 0
-      preprocess(documents).lazy.map do |document|
+      ranked_documents = documents.sort_by { |doc| doc[:rank] }
+      ranked_documents.lazy.map do |document|
         views = Float(document[:views]) - clicks_above_position
         clicks = Float(document[:clicks])
         views = clicks unless views >= clicks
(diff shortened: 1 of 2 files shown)


bilbof:
I should add that PR 2041 is a perfectly valid thing to have done. If I had more time I would spend some time trying to work out why NDCG is decreasing without removing this.

barrucadu (commented):
Is this sensible to revert by itself?  It was done because we might have multiple results at the same rank where only some of them exist in the Elasticsearch index.  Should we discount these additional results?  Just reverting this commit means that `clicks_above_position` is a bit weird for result sets that include parts.

bilbof:
@barrucadu I agree it's weird, and it is strange this had a negative impact on NDCG. If I had time to look into this I'd consider other options other than reverting.

barrucadu (approved):

PR #2181, opened 2020-09-17
Upgrade to Ruby 2.7.1

cdccollins:
These are the necessary changes for Ruby 2.7.1

I've left the version as 2.6.6 since there is an automatic way to update all of the apps' Ruby versions


Rhian (approved):
This upgrade looks good to me.

PR #2183, opened 2020-09-21
Stemming overrides

tarastockford:
Improve search results about importing by not returning content matching 'important'.
Improve search results about hospitality by not returning content about hospitals, and vice versa.
Don't mix results about Downing Street with content matching 'down'.


sihugh (approved):
👍

PR #2192, opened 2020-10-15
Make LTR Sagemaker instance types larger

sihugh:
We could do with a little more headroom in production. Moving to a C5 type removes the CPU credit quota. LTR is more compute heavy than anything else. We'll monitor for a bit, and move back down to C5.large if it seems overkill.

The env var is used in the [deploy script](https://github.com/alphagov/search-api/blob/master/ltr/concourse/deploy.py#L19)

I've changed staging too so that I can test the deploy.


hannako (approved):
With the caveat that I'm not at all familiar with Concourse! But seems like a sensible change  👍

PR #2201, opened 2020-10-21
Pass in BigQuery credentials as environment variable

bilbof:
This will reduce the risk of the credential being logged, and makes the task a bit more '12factor'.


ltr/concourse/fetch.sh
```diff
@@ -59,8 +59,8 @@ ssh -i /tmp/concourse_ssh_key -o StrictHostKeyChecking=no "ubuntu@${instance_ip}
 
   mkdir tmp
   set +x
-  echo "running bundle exec rake learn_to_rank:fetch_bigquery_export[BIGQUERY_CREDENTIALS,bigquery]"
-  \$docker bundle exec rake 'learn_to_rank:fetch_bigquery_export[$BIGQUERY_CREDENTIALS,bigquery]'
+  echo "running bundle exec rake learn_to_rank:fetch_bigquery_export[bigquery]"
```
ChrisBAshton, line 62: Is it safe to remove the `set +x`/`set -x` lines now?

ChrisBAshton (approved):
LGTM 👍

ltr/concourse/fetch.sh
bilbof: Yes it should be, I'll do it in a followup PR 🙂

PR #2209, opened 2020-11-02
Auto-update Concourse pipeline

bilbof:
This adds a new job that will watch GitHub for changes to the search-api repo and automatically set any pipeline changes that have been made, so we don't need to manually make those changes.


fredericfran-gds (approved):
looks good

PR #2241, opened 2021-01-12
Enable Fields for Get Involved Migration

Tetrino:
Without these fields we need to rely on the Publishing API for retrieving information about the Take Part pages and Consultations for use in the new Get Involved page post-migration from Whitehall to Government Frontend.

Once these fields are in and reindexed we'll be able to use this directly and save a lot of overhead. https://trello.com/c/7BjH9PyC/2083-8-migrate-government-get-involved-to-government-frontend


Tetrino:
Do not merge something went wrong somewhere, think I missed a file.

PR #2244, opened 2021-01-15
Remove need for REDIS_HOST and REDIS_PORT env vars

kevindew:
Trello: https://trello.com/c/dXud4WIt/206-inconsistent-redisurl-config

This is one of the few GOV.UK projects that uses this old approach to
configuring Redis - whereas most of the other ones use the approach
supported natively by the Redis gem [1] of using REDIS_URL. Removing
this will allow us to remove the need for REDIS_HOST and REDIS_PORT env
vars in our architecture and also allows this code to be simplified.

[1]: https://github.com/redis/redis-rb

It also applies a couple of minor tweaks to the Dockerfile and Rakefile.

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,24 +1,6 @@
 require "govuk_sidekiq/sidekiq_initializer"
 
-if ENV["RACK_ENV"] == "test"
-  redis_config = {
-    host: ENV.fetch("REDIS_HOST", "127.0.0.1"),
-    port: ENV.fetch("REDIS_PORT", 6379),
-    namespace: "search-api-test",
-  }
-
-  Sidekiq.configure_server do |config|
-    config.redis = redis_config
-  end
-
-  Sidekiq.configure_client do |config|
-    config.redis = redis_config
-  end
-else
-  redis_config = {
-    host: ENV.fetch("REDIS_HOST", "127.0.0.1"),
-    port: ENV.fetch("REDIS_PORT", 6379),
-  }
-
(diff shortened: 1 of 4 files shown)


config/initializers/sidekiq.rb
```diff
@@ -1,24 +1,6 @@
...
-  redis_config = {
-    host: ENV.fetch("REDIS_HOST", "127.0.0.1"),
-    port: ENV.fetch("REDIS_PORT", 6379),
-  }
-
-  GovukSidekiq::SidekiqInitializer.setup_sidekiq("search-api", redis_config)
-end
+GovukSidekiq::SidekiqInitializer.setup_sidekiq(
+  ENV.fetch("GOVUK_APP_NAME", "search-api"),
+  {},
```
thomasleese, line 5: Feels like `{}` might be quite a common value here now, and without knowing the `govuk_sidekiq` library it's very unclear what `{}` is referring to.

Before changing lots of apps, could we change `govuk_sidekiq` to be:

```rb
def setup_sidekiq(govuk_app_name, redis_config: {})
```

kevindew: That's a good point. I think it's only this app that is affected by this - as all the others use the Railtie. In a way this app is using the backdoor.

I have to make a breaking change for govuk_sidekiq to kill the REDIS_HOST and REDIS_PORT env vars so I can include something to make this cleaner as part of that.

thomasleese: Makes sense 👍🏻

thomasleese (approved):

PR #2245, opened 2021-01-15
Revert REDIS_URL changes

kevindew:
Something in https://github.com/alphagov/search-api/pull/2244 has caused a problem with the publishing-e2e-tests and I'm not sure what. It turns out the e2e tests haven't been running against Search API correctly since it was renamed from Rummager so this problem wasn't caught until the app was deployed to production.

This is a rather hasty revert to get the tests working again before the weekend.


sophiebrown01 (approved):
Looks good 👍

PR #2248, opened 2021-01-19
Remove redundant override for Brexit taxon

benthorner:
https://trello.com/c/maZEDGMH/775-update-brexit-taxon-title-and-description-in-content-tagger

This now matches the title and description set in Content Tagger,
so is no longer necessary.


sihugh (approved):
😌

PR #2256, opened 2021-02-01
Make 'standard' document type non-indexable

1pretz1:
Doesn't index the new 'standard' document type in the govuk index as,
like all other Whitehall publications, it will already be indexed in the
government index.

We're planning to migrate all publications over in one go to the
govuk-index in a future piece of work.

Trello:
https://trello.com/c/ugeapNY0/2318-5-create-new-standard-document-sub-type-in-whitehall

### DIFF
--- a/config/govuk_index/migrated_formats.yaml
+++ b/config/govuk_index/migrated_formats.yaml
@@ -420,6 +420,7 @@ non_indexable:
 - social_media_use
 - speech
 - staff_update
+- standard
 - statistical_data_set
 - statistics
 - statutory_guidance


brucebolt (dismissed):

config/govuk_index/migrated_formats.yaml
```diff
@@ -420,6 +420,7 @@ non_indexable:
 - social_media_use
 - speech
 - staff_update
+- standard
```
brucebolt, line 423: This is in the `non_indexable` section.  Is that intentional?

1pretz1: Originally that was not intentional but actually, on further inspection the other publication types don't seem to be indexed so we'll go with my original approach. 😄

1pretz1:
Thanks for the review @brucebolt and for catching the inconsistency! I've updated PR accordingly.

thomasleese:
How difficult would it be to get Whitehall to not index this new document type and instead get it into the Search API via the Publishing API queue? I'm just thinking since we've got a new document type it might be an opportunity to start it properly migrated (since we're going to have to do the work to migration this later anyway with Content Publisher).

1pretz1:
Sure, that makes sense @thomasleese. I'll have a look into this and report back.

1pretz1:
Just had a gander @thomasleese and it doesn't seem like we'd have to add too much bespoke code, it actually appears to somewhat simple. Just an extra query and boolean method in the [SearchIndex](https://github.com/alphagov/whitehall/blob/master/lib/whitehall/search_index.rb) class which would skip the doc type when deleting / adding to the government index.

Would it be much more effort to migrate the other Publication types too? The extra work that comes to mind is we'd have to do a bulk index and them remove the migrated publications from the government index. Probably out of scope for this card however, although it would be nice to have some uniformity between publications.

thomasleese:
@1pretz1 Looking at the code, it does seem like migrating the other publication types wouldn't be very complicated as they all share the same base class. I agree if we're going to have to do a bulk index we may as well do it to all publications in one go.

Since this probably is out of scope as you say, perhaps the best option then is to leave this PR as it is and have some follow up work to migrate all publications over to the `govuk` index in one go.

1pretz1:
Think that sounds reasonable @thomasleese. Shall I write up a card and pop it in the proposed stories column?

thomasleese:
@1pretz1 Yes, I think that's a good idea!

1pretz1:
Created the card [here](https://trello.com/c/ohSBuRpn/2055-migrate-all-publications-over-to-the-govuk-index).

Could I trouble either @thomasleese or @brucebolt for a re-review? 🥺

thomasleese (approved):

PR #2263, opened 2021-02-19
Retry when `bulk_index` times out

brucebolt:
We have logged over 425k events for `Faraday::TimeoutError` which come from the `bulk_index` method timing out. There appears to be no pattern as to what causes this, so we will retry the indexing 10 times if one attempt times out.  This should reduce the number of errors logged to Sentry and ensure the document is still correctly indexed.

If the error gets raised 10 consecutive times for one bulk index, we will still raise the error.

Trello card: https://trello.com/c/YDrY6zRx


ChrisBAshton (approved):
Hopefully this fixes it!🤞

PR #2264, opened 2021-02-22
Increase timeout for bulk indexing

brucebolt:
We currently specify a default 5s timeout for all indexing queries. Bulk indexing may take longer than this (evidenced by `Faraday::TimeoutError` being raised, even after multiple retries), so we should set a higher timeout limit.

10s is an arbitrary choice, based on doubling the current timeout.

Trello card: https://trello.com/c/YDrY6zRx


thomasleese (approved):
Sounds reasonable to me.

PR #2266, opened 2021-02-26
Update rubocop-govuk to 4.0.0.pre.1

kevindew:
Trello: https://trello.com/c/pqXbKZiC/215-test-rubocop-govuk-400pre1-prior-to-400-release

This bumps to the 4.0.0 pre release of rubocop-govuk which is part of [trialling the new rubocop-govuk release](https://github.com/alphagov/rubocop-govuk/issues/129). It seems to have gone into this app with some annoyances in the 
`Style/CommentAnnotation` cop and `Lint/MissingSuper` catching a few oddities.

As part of this I've also made the Gemfile consistent with GOV.UK Rails app conventions


lib/debug/rank_eval.rb
```diff
@@ -53,8 +53,8 @@ def evaluate
 
       {
         score: result["metric_score"],
-        query_scores: result["details"].each_with_object({}) do |(query, detail), acc|
-          acc[query] = detail["metric_score"]
+        query_scores: result["details"].transform_values do |detail|
```
leenagupte, line 56: This is much nicer, but I can't believe I've never seen it used before.

lib/indexer/comparer.rb
```diff
@@ -21,8 +21,8 @@ def run
       log "processing: #{@old_index_name}/#{@new_index_name}"
 
       search_body = {}
-      if @filtered_format
-        search_body[:post_filter] = { term: { format: @filtered_format } } if @filtered_format
+      if @filtered_format && @filtered_format
```
leenagupte: This is checking `@filtered_format` twice, so I think the second clause can be removed.

lib/tasks/export.rake
```diff
@@ -23,9 +23,10 @@ namespace :export do
         csv << fields.map do |f|
           value = hit["_source"][f]
 
-          if value.is_a? Hash
+          case value
+          when Hash
```
leenagupte, line 27: This is nice!

lib/tasks/debug.rake
```diff
@@ -75,7 +75,7 @@ namespace :debug do
 
       maxlen = results[:query_scores].map { |query, _| query.length }.max
       results[:query_scores].each do |query, score|
-        puts "#{(query + ':').ljust(maxlen + 1)} #{score}"
+        puts "#{("#{query}:").ljust(maxlen + 1)} #{score}"
```
leenagupte: This is a bit of nitpick, but the nested `"` look a bit odd. Can they be replaced with single-quotes?
```suggestion
        puts "#{('#{query}:').ljust(maxlen + 1)} #{score}"
```

lib/indexer/comparer.rb
```diff
@@ -21,7 +21,7 @@ def run
       log "processing: #{@old_index_name}/#{@new_index_name}"
 
       search_body = {}
-      if @filtered_format && @filtered_format
+      if @filtered_format
```
leenagupte, line 24: Ah! you caught it!

lib/search/escaping.rb
```diff
@@ -1,8 +1,10 @@
 module Search
   module Escaping
-    LUCENE_SPECIAL_CHARACTERS = Regexp.new("(" + %w[
-      + - && || ! ( ) { } [ ] ^ " ~ * ? : \\ /
-    ].map { |s| Regexp.escape(s) }.join("|") + ")")
+    LUCENE_SPECIAL_CHARACTERS = begin
+      chars = %w[+ - && || ! ( ) { } [ ] ^ " ~ * ? : \\ /]
+      escaped = chars.map { |s| Regexp.escape(s) }.join("|")
+      Regexp.new("(#{escaped})")
+    end
```
leenagupte, line 7: This is much much nicer!

leenagupte (approved):
There's one comment about nested double-quotes, but it's not a blocker.  LGTM 👍

lib/tasks/debug.rake
kevindew: Thanks for the suggestion here, I'd agree that this looks a bit odd. 

I felt that the underlying code was already rather cryptic so I've gone with an approach of breaking that out instead in https://github.com/alphagov/search-api/pull/2266/commits/97f56af45cbc3e27796279ded2ed4132f4e904e1

leenagupte: Splitting them out is better. The approval still stands 👍

PR #2271, opened 2021-03-19
Ensure that we scale down the LTR EC2 ASG

sihugh:
The [concourse job](https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/search-learn-to-rank) that runs this script fails occasionally. If this happens at an inopportune moment, then any subsequent jobs will also fail unless someone manually kills the running "search-ltr-generation" EC2 instance.

By using a trap, we can hopefully make it more reliable by ensuring we kill off the failed instance.

### DIFF
--- a/ltr/concourse/fetch.sh
+++ b/ltr/concourse/fetch.sh
@@ -77,14 +77,4 @@ ssh -i /tmp/concourse_ssh_key -o StrictHostKeyChecking=no "ubuntu@${instance_ip}
   aws s3 cp svm/validate.txt s3://$S3_BUCKET/data/$NOW/validate.txt
 EOF
 
-# the previous assumption may have timed out, so assume again before
-# scaling down the ASG
-assume_role
-
-echo "Scaling down ASG..."
-aws autoscaling set-desired-capacity \
-    --region "$AWS_REGION" \
-    --auto-scaling-group-name "$EC2_NAME" \
-    --desired-capacity 0
-
 echo "$NOW" > "out/${GOVUK_ENVIRONMENT}-${OUTPUT_FILE_NAME}-$(date +%s).txt"
--- a/ltr/concourse/pipeline.yaml
+++ b/ltr/concourse/pipeline.yaml
@@ -109,6 +109,13 @@ jobs:
       trigger: true
     - get: search-api-git
     - task: fetch
+      ensure:
+        params:
+          GOVUK_ENVIRONMENT: integration
+          ROLE_ARN: ((integration-role-arn))
+        run:
+          path: bash
+          args: ["search-api-git/ltr/concourse/after-fetch.sh"]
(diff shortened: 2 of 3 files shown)


ltr/concourse/fetch.sh
```diff
@@ -11,6 +11,18 @@ S3_BUCKET="govuk-${GOVUK_ENVIRONMENT}-search-relevancy"
...
+  # any previous assumption may have timed out, so assume again before scaling down the ASG
+  assume_role
+
+  echo "Scaling down ASG..."
+  aws autoscaling set-desired-capacity \
+      --region "$AWS_REGION" \
+      --auto-scaling-group-name "$EC2_NAME" \
+      --desired-capacity 0
+}
+trap scale_down_asg EXIT
```
bilbof: It might be worth moving the cleanup step into a separate [`ensure` step](https://concourse-ci.org/jobs.html#schema.step.ensure) in the Concourse fetch job.

The benefit would be that we'd still capture the status of the job (so the train job would get triggered correctly depending on the exit code of the fetch job), and the turn-down of the asg would always run.

```yaml
plan:
- task: fetch
  file: fetch-job.yml
  ensure:
    task: turn-down-fetch-machine
    file: turn-down.yml
```

sihugh: ooh, nice idea 👍

sihugh: @bilbof would you mind having another look at this please?

ltr/concourse/pipeline.yaml
```diff
@@ -109,6 +109,13 @@ jobs:
       trigger: true
     - get: search-api-git
     - task: fetch
+      ensure:
+        params:
+          GOVUK_ENVIRONMENT: integration
+          ROLE_ARN: ((integration-role-arn))
+        run:
+          path: bash
+          args: ["search-api-git/ltr/concourse/after-fetch.sh"]
```
bilbof, line 118: Looks good. I think you need to nest the `params` and `run` bits under a `config:` (or put it all in a `file:` if you don't want to duplicate stuff). You'll also need to define the `platform` and `image_resource` you want the task to use, and the step should be `task` here.

This should work:

```yaml
      ensure:
        task: teardown
        config:
          <<: *task-config
          inputs:
          - name: search-api-git
          params:
            GOVUK_ENVIRONMENT: integration
            ROLE_ARN: ((integration-role-arn))
          run:
            path: bash
            args: ["search-api-git/ltr/concourse/after-fetch.sh"]
```

You can use `fly validate-pipeline -c ltr/concourse/pipeline.yaml` to check it - but I should really put that in a GitHub Action!

theseanything:
Closing this as we are no longer using concourse - this functionality is no covered by #2348

PR #2282, opened 2021-04-09
Make smart answer schema indexable

theseanything:
Previously, Smart Answer start pages were represented by the transaction content schema. We have since created a new schema `smart_answer` to represent Smart Answers. This adds in support for this new schema so search results behave when they were transaction pages. The start page body content is no longer in the separate fields `introduction` and `more_information` and will now be included within the `hidden_search_terms` section.

This should result in no user facing changes.

### DIFF
--- a/spec/unit/govuk_index/presenters/indexable_content_presenter_spec.rb
+++ b/spec/unit/govuk_index/presenters/indexable_content_presenter_spec.rb
@@ -115,36 +115,16 @@ RSpec.describe GovukIndex::IndexableContentPresenter do
     end
   end
 
-  context "transaction with hidden_search_terms (smart answer start page)" do
-    let(:format) { "transaction" }
+  context "smart_answer format" do
+    let(:format) { "smart_answer" }
     let(:details) do
       {
-        "introductory_paragraph" => [
-          {
-            "content" => "intro\n",
-            "content_type" => "text/govspeak",
-          },
-          {
-            "content_type" => "text/html",
-            "content" => "<p>intro</p>\n",
-          },
-        ],
-        "more_information" => [
-          {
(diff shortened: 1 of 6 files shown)


barrucadu (commented):
https://github.com/alphagov/search-api/blob/master/spec/integration/govuk_index/publishing_event_processor_spec.rb#L116-L133 is failing because it uses `smart_answer` as an example of a format which should be blocked.  Other than that, looks ok to me.

spec/unit/govuk_index/presenters/indexable_content_presenter_spec.rb
```diff
@@ -115,36 +115,16 @@
...
-            "content" => "<p>more</p>\n",
-          },
-        ],
         "hidden_search_terms" => ["hidden 1", "hidden 2"],
       }
     end
 
     it "hidden_search_terms is correctly indexed" do
-      expect(subject.indexable_content).to eq("hidden 1\n\n\nhidden 2\n\n\nintro\n\n\n\nmore")
+      expect(subject.indexable_content).to eq("hidden 1\n\n\nhidden 2")
```
huwd, line 127: Is there any value adding that body and more_information data in here as an explicit example or is it enough to just keep testing that there's something in there?

theseanything: These tests seem to be testing that **_at least_** the correct information is been included in the `indexable_content` method and assume that `details` represented are accurate for the content schema. I guess there might be value in making sure "other" fields from from details aren't being included, but I've just written this to be consistent with the other tests.

barrucadu (approved):

PR #2284, opened 2021-04-12
Replace deprecated supertypes in boosting

sihugh:
[navigation_document_supertype](https://github.com/alphagov/govuk_document_types/blob/fc4e1110b24ba5aebbf84b7fec245d4cf27e9ef5/data/supertypes.yml#L1-L26) and [search_user_need_document_supertype](https://github.com/alphagov/govuk_document_types/blob/fc4e1110b24ba5aebbf84b7fec245d4cf27e9ef5/data/supertypes.yml#L1-L26)  have not been updated much in recent years and several new content types have arrived since. They are not really used around the system so it makes sense to replace them with equivalent content_purpose_supertype references.

I've updated the spec too. Guidance content is still boosted, but I think this spec was more about asserting that _something_ was boosted correctly rather than worrying about exactly what it was. With that in mind, I've switched the spec to checking that services get a bump.

Also removed a format that no longer exists on the system

https://trello.com/c/hp5BJD2i/117-unused-incorrect-supertypes-in-content-items

### DIFF
--- a/config/query/boosting.yml
+++ b/config/query/boosting.yml
@@ -14,7 +14,6 @@ base:
     # Inside Gov formats
     contact: 0.3
     document_collection: 1.3
-    document_series: 1.3
     minister: 1.7
     operational_field: 1.5
     organisation: 2.5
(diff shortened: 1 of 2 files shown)


hannako (approved):
Looks good 👍

PR #2285, opened 2021-04-13
Deprecated supertype removal

sihugh:
We don't query against any of these supertypes, and they're not really updated.

This PR removes the code that uses the supertypes, though they're [still defined in the schemas](https://github.com/alphagov/search-api/blob/master/config/schema/field_definitions.json#L47-L50). I'll remove them from there once we're all clear with these changes here.

I've cleared the change in `lib/analytics/extract.rb` with Tara. She say's it's duplicated, and agrees that it's not really used.

https://trello.com/c/hp5BJD2i/117-unused-incorrect-supertypes-in-content-items


bilbof (approved):
LGTM!

sihugh:
Thanks @bilbof!

PR #2288, opened 2021-04-14
Remove coronavirus taxon boost

sihugh:
Content shouldn't need this arbitrary boost. We should allow elasticsearch relevance a bit more influence on what to return.


kevindew (approved):
This sounds like a good idea to remove or at least try out. We've got other topical events at the mo (DoE) which we wouldn't want to end up lower in search because of this boost.

PR #2290, opened 2021-04-15
Dedupe boosting further

sihugh:
In https://github.com/alphagov/search-api/pull/2284 I replaced a couple of out of date supertype groups with more current ones, however it wasn't quite equivalent.

In this commit, I remove the content_purpose_subgroups I added, and replace them with direct additions of the content types that were not already present in the content_purpose_supergroup services group, or elsewhere tracked in this file.

I've dropped the boosting of `contact` entirely because it was boosted by 2.5 and also suppressed by 0.3.

I've removed step_by_step_nav and smart_answers as direct references because they're picked up in the boosting of content_purpose_supergroup services group.

I've left transaction as being boosted, as it always was prioritised over other content.

document_collection was duplicated, so I've given it a lift to be equivalent to the others, and removed the duplication.

The other additions are direct copies from the original configuration.

### DIFF
--- a/config/query/boosting.yml
+++ b/config/query/boosting.yml
@@ -3,17 +3,13 @@ base:
     coronavirus_landing_page: 2.5
     service_manual_guide: 0.3
     service_manual_topic: 0.3
-    step_by_step_nav: 2.5
     transaction: 1.5
-    smart_answer: 1.5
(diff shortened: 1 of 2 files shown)


bilbof (approved):
Sounds good! :shipit:

PR #2310, opened 2021-06-28
Allow pages to be found in search

sihugh:
These pages are findable by Google search, so we should allow them to be findable internally too.


maxgds (approved):

config/govuk_index/migrated_formats.yaml
```diff
@@ -342,7 +341,6 @@ non_indexable:
   - "/googlef35857dca8b812e7.html"
   - "/government"
   - "/government/get-involved"
```
chris-gds, line 343: 🤔 Would I expect to see the "[get involved](https://www.gov.uk/government/get-involved)" page from search such as this? (this is coming from the homepage)

image

maxgds: Confirmed that this appears when searched with quotes.

PR #2319, opened 2021-08-05
Remove /uk-nationals-living-eu

sihugh:
This is a custom page that will be redirected to a collection.

We're going to override the content item and route using short-url-manager.

https://trello.com/c/rTwoXCox/1695-redirect-where-do-you-live-tool-to-fcdo-living-in-collection-page


hannako (approved):
👍

PR #2361, opened 2022-02-07
Create DRCF digital markets research finder

baisa:
Following CMA request to create a specialist finder to bring together research from several organisations.

[Trello card](https://trello.com/c/0LHrDag7/157-create-new-specialist-finder-for-competition-and-markets-authority)


beccapearce (approved):

config/schema/field_definitions.json
```diff
@@ -993,5 +993,20 @@
   "product_recall_alert_date":{
     "description": "Date of the recall/alert",
     "type": "date"
+  },
+  "category":{
```
sihugh: I wonder if it's useful to disambiguate these somewhat? I can see a future person coming along and assuming that category, publisher and date_published are generically used and are appropriate for co-opting into some other project? I think I'd have to do some digging as a new person to work out that date_published is specific to a particular document type.

sihugh: (And it's easier to do this now than once the thing is published!)

I suggest perhaps:
- digital_market_research_category
- digital_market_research_publisher
- digital_market_research_area
- digital_market_research_topic
- digital_market_research_publish_date (or see whether you can reuse another existing date field?)

sihugh (dismissed):

config/schema/field_definitions.json
benjamineskola: I've renamed all of these as suggested. As far as the date is concerned it doesn't look like we have a generic publication date field at all, various date fields exist but none seem directly equivalent. (Although maybe it might make sense to create one that future types could use.)

config/schema/field_definitions.json
```diff
@@ -993,5 +993,20 @@
   "product_recall_alert_date":{
     "description": "Date of the recall/alert",
     "type": "date"
+  },
+  "digital_market_research_category": {
+    "type": "identifiers"
```
kevindew: Because there is only one of these should it be "identifier" singular as per https://github.com/alphagov/search-api/blob/348f433610531b9d4254349d18840028df73ad68/config/schema/field_definitions.json#L117-L120 ?

kevindew: Because there is only one of these should it be "identifier" singular as per https://github.com/alphagov/search-api/blob/348f433610531b9d4254349d18840028df73ad68/config/schema/field_definitions.json#L117-L120 ?

kevindew (approved):
This looks good to me now - thanks for all the work here.

PR #2371, opened 2022-03-09
Use env `LOG_TO_STDOUT` to enable stdout logging

fredericfran-gds:
In EKS, we want logging to go to stdout rather than to a logging
file. This commit introduce a new `LOG_TO_STDOUT` env which
when set will log to stdout.

By default, stdout logging is enable in the Docker image as decided
[here](https://github.com/alphagov/govuk-helm-charts/pull/23).
We need to disable that for publishing-e2e.

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -20,15 +20,18 @@ end
 
 if log_path
   enable :logging
-  log = File.new(log_path, "a")
-  log.sync = true
-  STDOUT.reopen(log)
-  STDERR.reopen(log)
+  unless ENV["LOG_TO_STDOUT"].present?
+    log = File.new(log_path, "a")
+    log.sync = true
+    STDOUT.reopen(log)
+    STDERR.reopen(log)
+  end
 end
 
 unless in_development
+  logger = ENV["LOG_TO_STDOUT"].present? ? Logger.new($stdout) : Logger.new("log/production.json.log") 
   use Rack::Logstasher::Logger,
-      Logger.new("log/production.json.log"),
+      logger,
       extra_request_headers: { "GOVUK-Request-Id" => "govuk_request_id", "x-varnish" => "varnish_id" }
 end
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ COPY . /app
 
 FROM $base_image
 
-ENV RAILS_ENV=production GOVUK_APP_NAME=search-api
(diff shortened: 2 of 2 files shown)


config.ru
```diff
@@ -20,16 +20,24 @@ end
 
 if log_path
   enable :logging
-  log = File.new(log_path, "a")
-  log.sync = true
-  STDOUT.reopen(log)
-  STDERR.reopen(log)
+  unless ENV["LOG_TO_STDOUT"].present?
```
bilbof, line 23: I find `unless ENV["LOG_TO_STDOUT"].present?` a little confusing, is `if !ENV.key?("LOG_TO_STDOUT")` clearer?

config.ru
```diff
@@ -20,16 +20,24 @@ end
...
-      extra_request_headers: { "GOVUK-Request-Id" => "govuk_request_id", "x-varnish" => "varnish_id" }
+  unless ENV["LOG_TO_STDOUT"].present?
+    use Rack::Logstasher::Logger,
+        Logger.new("log/production.json.log"),
+        extra_request_headers: { "GOVUK-Request-Id" => "govuk_request_id", "x-varnish" => "varnish_id" }
+  else
+    use Rack::Logstasher::Logger,
+        Logger.new($stdout),
+        extra_request_headers: { "GOVUK-Request-Id" => "govuk_request_id", "x-varnish" => "varnish_id" }
+  end
```
bilbof: Here perhaps you can do

```ruby
logger = ENV.key?("LOG_TO_STDOUT") ? Logger.new($stdout) : Logger.new("log/production.json.log")
use Rack::Logstasher::Logger,
        logger,
        extra_request_headers: { "GOVUK-Request-Id" => "govuk_request_id", "x-varnish" => "varnish_id" }
```

to avoid duplication

fredericfran-gds: looks good

config.ru
fredericfran-gds: I guess there are 2 changes now:
1. replace `unless` with `if not`: in this case, it is a simple one with no `else`, I guess we can keep it this way. (https://github.com/rubocop/ruby-style-guide#if-vs-unless)
2. using ENV.key vs ENV.present, I chose the latter to be the same as for RAILS_LOG_TO_STDOUT in rails.

smford: I dislike unless, but since it's the style we'd best roll with it.

smford (approved):
lgtm!

PR #2372, opened 2022-03-11
[spike] Increase top-k size to 40 by default for relevance queries

bilbof:
Queries will by default fetch 40 documents from Elasticsearch,
or the K documents specifies by the `k` query parameter (max: 40).
The reranker will re-rank K documents, and return only the top-k
specified through the `count` parameter.

This will put greater load on the Elasticsearch cluster but should
improve search quality for relevance queries.

### DIFF
--- a/lib/search/query.rb
+++ b/lib/search/query.rb
@@ -94,7 +94,7 @@ module Search
 
       return { reranked: false, es_response: es_response } if reranked.nil?
 
-      es_response["hits"]["hits"] = reranked
+      es_response["hits"]["hits"] = reranked.first(search_params.count)
       { reranked: true, es_response: es_response }
     end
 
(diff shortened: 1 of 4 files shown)


lib/search/query.rb
```diff
@@ -94,7 +94,7 @@ def rerank(es_response, search_params)
 
       return { reranked: false, es_response: es_response } if reranked.nil?
 
-      es_response["hits"]["hits"] = reranked
+      es_response["hits"]["hits"] = reranked.first(search_params.count)
```
sihugh, line 97: If `k` is `2 * count` then might we need to flipflop between first(count) and last(count) depending if `start % k` is 0 or `count`? We could end up reranking results out of existence otherwise?

bilbof: We could introduce some jitter here to give some results a chance. Maybe @hannako has some ideas on how we might occassionally swap downranked results into the top-k. 🤔 

We could ocassionally return the bottom-k to some users, might be fairer to spread it out a bit among users, since we can assume most of the time the top-k is going to be better.

sihugh:
We're no longer using this fancy stuff 😞

PR #2387, opened 2022-04-29
Update cacher_dir location

nsabri1:
While Replatforming search-api and disabling root containers from running in EKS - we detected an error:

**error message:**
_Errno::EACCES: Permission denied @ dir_s_mkdir - tmp
/usr/local/lib/ruby/2.7.0/fileutils.rb:250:in mkdir'_

To resolve this issue i have updated to use `/tmp` rather than `tmp`.

https://trello.com/c/Jef1t4xU/903-fix-app-permission-errors-due-to-nonroot

### DIFF
--- a/lib/rummager/app.rb
+++ b/lib/rummager/app.rb
@@ -6,7 +6,7 @@ require "sinatra"
 set :root, File.dirname(__FILE__)
 
 Bootsnap.setup(
-  cache_dir: "tmp/cache",
+  cache_dir: "/tmp/cache",
   development_mode: ENV["RACK_ENV"] == "development",
   load_path_cache: true,
   autoload_paths_cache: true,


sengi (approved):
LGTM - prob best to get a review from someone on #find-and-view-tech as well.

kevindew:
I'm not sure that this won't cause problems for some people on local machines as I'm not sure if the entries in /tmp/cache will conflict with other things - whereas having that directory embedded in the app cuts that issue out (and is more of an idiomatic approach for Ruby apps).

In general when we've had this sort of problem in the past we've created a [`tmp/.keep`](https://github.com/alphagov/content-publisher/tree/main/tmp) file in the repo which will ensure that tmp exists. If that would solve your problem I'd suggest taking that route. I am thinking it might not because it may just replace your `Errno::EACCES: Permission denied @ dir_s_mkdir - tmp` with a `Errno::EACCES: Permission denied @ dir_s_mkdir - tmp/cache` but I'm not sure quite what the file ownership would be.

sengi:
> I'm not sure that this won't cause problems for some people on local machines as I'm not sure if the entries in /tmp/cache will conflict with other things - whereas having that directory embedded in the app cuts that issue out (and is more of an idiomatic approach for Ruby apps).
> 
> In general when we've had this sort of problem in the past we've created a [`tmp/.keep`](https://github.com/alphagov/content-publisher/tree/main/tmp) file in the repo which will ensure that tmp exists. If that would solve your problem I'd suggest taking that route. I am thinking it might not because it may just replace your `Errno::EACCES: Permission denied @ dir_s_mkdir - tmp` with a `Errno::EACCES: Permission denied @ dir_s_mkdir - tmp/cache` but I'm not sure quite what the file ownership would be.

Yeah unfortunately that won't solve it, because we need the app not to write anywhere where it's not supposed to (and anything under the app's directory definitely falls into that category).

Is the issue that the cached bytecode files can have name clashes between versions? If so, how about if we name the directory something specific to the app?

kevindew:
> > I'm not sure that this won't cause problems for some people on local machines as I'm not sure if the entries in /tmp/cache will conflict with other things - whereas having that directory embedded in the app cuts that issue out (and is more of an idiomatic approach for Ruby apps).
> > In general when we've had this sort of problem in the past we've created a [`tmp/.keep`](https://github.com/alphagov/content-publisher/tree/main/tmp) file in the repo which will ensure that tmp exists. If that would solve your problem I'd suggest taking that route. I am thinking it might not because it may just replace your `Errno::EACCES: Permission denied @ dir_s_mkdir - tmp` with a `Errno::EACCES: Permission denied @ dir_s_mkdir - tmp/cache` but I'm not sure quite what the file ownership would be.
> 
> Yeah unfortunately that won't solve it, because we need the app not to write anywhere where it's not supposed to (and anything under the app's directory definitely falls into that category).
> 
> Is the issue that the cached bytecode files can have name clashes between versions? If so, how about if we name the directory something specific to the app?

Sure, that makes sense. I was just wondering locally whether you'd have a problem if you ran two bootsnap services at the same time that both were writing to `/tmp/cache` - I've not looked to see how much of a problem it is, just seems an easy situation for something that once had the safety of being local being moved to global. 

A specific directory would definitely improve that. I'm also wondering if you want to go down an env var route so you can leave things as per bootsnap recommended config for local dev/test environments and then configure it specifically for your container.

But what I did wonder is, all the Rails apps are also writing to a local tmp by default - do you have a solution for them already? Since this app is a bit of an odd-one-out being Sinatra I wonder if there's a risk of establishing an approach here that may need tinkering again for Rails.

nsabri1:
> > > I'm not sure that this won't cause problems for some people on local machines as I'm not sure if the entries in /tmp/cache will conflict with other things - whereas having that directory embedded in the app cuts that issue out (and is more of an idiomatic approach for Ruby apps).
> > > In general when we've had this sort of problem in the past we've created a [`tmp/.keep`](https://github.com/alphagov/content-publisher/tree/main/tmp) file in the repo which will ensure that tmp exists. If that would solve your problem I'd suggest taking that route. I am thinking it might not because it may just replace your `Errno::EACCES: Permission denied @ dir_s_mkdir - tmp` with a `Errno::EACCES: Permission denied @ dir_s_mkdir - tmp/cache` but I'm not sure quite what the file ownership would be.
> > 
> > 
> > Yeah unfortunately that won't solve it, because we need the app not to write anywhere where it's not supposed to (and anything under the app's directory definitely falls into that category).
> > Is the issue that the cached bytecode files can have name clashes between versions? If so, how about if we name the directory something specific to the app?
> 
> Sure, that makes sense. I was just wondering locally whether you'd have a problem if you ran two bootsnap services at the same time that both were writing to `/tmp/cache` - I've not looked to see how much of a problem it is, just seems an easy situation for something that once had the safety of being local being moved to global.
> 
> A specific directory would definitely improve that. I'm also wondering if you want to go down an env var route so you can leave things as per bootsnap recommended config for local dev/test environments and then configure it specifically for your container.
> 
> But what I did wonder is, all the Rails apps are also writing to a local tmp by default - do you have a solution for them already? Since this app is a bit of an odd-one-out being Sinatra I wonder if there's a risk of establishing an approach here that may need tinkering again for Rails.

I could update the cache location to say as an example `/tmp/cache/<name of app>/<used by>`  which would look like /tmp/cache/search-api/rummager.

I would also look at a ENV value which would disable Bootsnap on run time, say by adding `USE_BOOTSNAP`ENV var that we can set to TRUE/FALSE in k8s but defaults to true.

sengi:
If we need to add an env var anyway, adding one that lets us disable bootsnap would be much preferable to just changing the directory 👍 (since bootsnap adds no value in a containerised environment)

kevindew:
Are you finding this is only a problem for bootsnap? I'd have thought it'd be all uses of the tmp directory and I understand there is a few.

nsabri1:
> Are you finding this is only a problem for bootsnap? I'd have thought it'd be all uses of the tmp directory and I understand there is a few.

We are finding it with Apps which need to write something to WORKDIR in our case /app/.

this could be for Bootsnap tmp cache files or certain apps which write to /app/log/<log file>

sengi:
Yeah that's a good point - it'll almost certainly be a wider problem than just bootsnap. So if there's a simple way we can tell everything to use /tmp/<name of app> (and automatically create that if it doesn't exist) then that'd be nice.

kevindew:
Sure, I'd suggest starting with a different app than this one in that case since it's a Sinatra app. Perhaps frontend? I feel like you'd do best establishing a good way to resolve it for Rails and then applying any practices here.

nsabri1:
Fixed elsewhere

PR #2393, opened 2022-05-24
Update bootsnap cache dir location

nsabri1:
Updating Dockerfile to soft link /tmp as Bootsnap tmp cache location for GOVUK
 kubernetes deploymenit.
 
 This change symlinks /tmp to /app/tmp

I also attempted to use `BOOTSNAP_CACHE_DIR=/tmp/cache ` as ENV value - which seems to not have any affect on the app.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,16 +23,25 @@ COPY . /app
 
 FROM $base_image
 
-ENV RAILS_ENV=production GOVUK_APP_NAME=search-api LOG_TO_STDOUT=true
+ENV RAILS_ENV=production \
+    GOVUK_APP_NAME=search-api \
+    LOG_TO_STDOUT=true
 
 RUN apt-get update -qy && \
     apt-get upgrade -y && \
     apt-get install -y nodejs && \
     apt-get clean
 
+RUN mkdir /app && ln -fs /tmp /app
+
 COPY --from=builder /usr/local/bundle/ /usr/local/bundle/
 COPY --from=builder /app /app/
 
 WORKDIR /app
+
+RUN groupadd -g 1001 app && \
+    useradd app -u 1001 -g 1001 --home /app
+
+USER app
 
 CMD bundle exec puma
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -35,4 +37,10 @@ COPY --from=builder /app /app/
 
 WORKDIR /app
 
+RUN groupadd -g 1001 app && \
+    useradd app -u 1001 -g 1001 --home /app && \
+    ln -fs /tmp /app
```
sengi: Better to create the symlink before copying the rest of the app tree over, so that there's no ambiguity about it potentially putting the symlink inside a preexisting /app/tmp that might have been copied over. e.g. see https://github.com/alphagov/smart-answers/blob/7379189/Dockerfile#L64

Dockerfile
```diff
@@ -23,7 +23,9 @@ COPY . /app
 
 FROM $base_image
 
-ENV RAILS_ENV=production GOVUK_APP_NAME=search-api LOG_TO_STDOUT=true
+ENV RAILS_ENV=production \
+    GOVUK_APP_NAME=search-api \
+    LOG_TO_STDOUT=true
```
sengi, line 28: Shouldn't this be RAILS_LOG_TO_STDOUT?

sengi (approved):

Dockerfile
nsabri1: i did consider having soft link creation prior the copy but as search-api does not have a tmp folder within its repo i thought it would be fine. 
Saying that things may change in the future so i will bring it above the copy. thanks

Dockerfile
nsabri1: this was present prior my chnage - looking at config.ru
```
if log_path
   enable :logging
   unless ENV["LOG_TO_STDOUT"].present?
     log = File.new(log_path, "a")
     log.sync = true
     STDOUT.reopen(log)
     STDERR.reopen(log)
   end
 end
```
so i guess it should be as this

Dockerfile
nsabri1: updated and tested

Dockerfile
sengi: Oh, good shout, thanks!

PR #2397, opened 2022-05-25
Railties 7

KludgeKML:
Search-API doesn't really use Rails, but it uses ActionPack, ActionView, ActiveSupport and RailTies, so a good idea to update the gems to the 7.0.3 versions. We will also follow the normal rails 7 upgrade procedure by updating all other possible gems at the same time.

https://trello.com/c/nkPvgfQy/1206-rails-7-update-search-api

Notes: Pinning OJ to 3.11.3, the maximum level it can go to without failing tests.

3.11.4 is a massive update, introducing an unknown breaking change for this app:
https://github.com/ohler55/oj/compare/v3.11.3...v3.11.4


1pretz1 (approved):
Nice work 👍 That is a beast of a patch version (oj) with codes changes mixed in with lots of linting

PR #2416, opened 2022-07-21
Upgrade to Python 3.8

gclssvglx:
## What

Upgrade Python from 3.7 to 3.8

## Why

There are outstanding `dependabot` PR's for `numpy` and `tensorflow` which are failing to build as the versions of these tools requires Python 3.8 and is incompatible with Python 3.7.

`dependabot` issues:

- https://github.com/alphagov/search-api/pull/2407
- https://github.com/alphagov/search-api/pull/2406
- https://github.com/alphagov/search-api/pull/2396

[Trello](https://trello.com/c/VwBT9uFl/163-upgrade-python-in-search-api)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,7 +9,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: "3.7"
+          python-version: "3.8"
       - name: Check LTR dependencies install
         run: |
           set -ex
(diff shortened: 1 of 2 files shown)


kevindew (approved):
This looks good to me. I can't find any other references to a Python version and it seems reasonable to expect CI to be testing it sufficiently.

I know very little about this Python add-on to this repo, but I expect there are few people who do institutionally - none I can think of that aren't on leave today.

PR #2417, opened 2022-07-21
Remove title from test

1pretz1:
We're no longer planning to populate the manual title on a section.

Trello:
https://trello.com/c/ItYKGMqF/1464-revert-add-parent-manuals-title-to-the-hmrc-section-content-item


KludgeKML (approved):
LGTM

PR #2422, opened 2022-07-29
Remove ability to filter by facet groups and facet values

leenagupte:
Related to: 
* https://github.com/alphagov/finder-frontend/pull/2835

Trello: https://trello.com/c/EOLde8PE

`facet_groups` and `facet_values` were created to support the EU Exit Business Support Finder [1].
This finder was decommissioned in February 2020 [2], so support for these filter items is no longer needed.

[1]: https://github.com/alphagov/content-tagger/blob/main/docs/arch/adr-001-links-based-facets.md
[2]: https://github.com/alphagov/finder-frontend/pull/1949

### DIFF
--- a/config/govuk_index/migrated_formats.yaml
+++ b/config/govuk_index/migrated_formats.yaml
@@ -374,9 +374,6 @@ non_indexable:
 - document_collection
 - equality_and_diversity
 - facet
-- facet_group
-- facet_value
-- facet_values
 - fatality_notice
 - field_of_operation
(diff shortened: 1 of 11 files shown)


kevindew (approved):
Great stuff. Do we have to do a re-index?

leenagupte:
> Great stuff. Do we have to do a re-index?

That's a good question. I would say no, because we haven't been able to tag anything to these values in over 2 years.

PR #2441, opened 2022-09-23
Avoid unnecessary mkdir /app in the Dockerfile.

sengi:
This is needed in order to remain compatible with an upcoming change to govuk-ruby-base.


nsabri1 (approved):
you could move up line 22 to 18 in the dockerfile

sengi:
> you could move up line 22 to 18 in the dockerfile

Good idea — I'll do that in a follow-up (rationale in https://github.com/alphagov/collections/pull/2973#issuecomment-1257736038)

PR #2463, opened 2022-11-22
Explain why boosting announcement curve was chosen

matthewford:
I'm familiarising myself with the `seach-api` again, and I know why this curve was chosen because I chose it.

The current solar doc has moved on; it was an extension of https://solr.apache.org/guide/7_7/function-queries.html#recip-function.

Also, the graph of the curve is no longer there due to WolframAlpha changing.

But you can imagine a curve decaying slowly for two months and then rapidly.

This was so that announcements are ranked higher in the search, but old announcements (we arbitrarily chose two months) are not (which was an issue as all announcements were boosted equally, causing poor results).

### DIFF
--- a/docs/relevancy.md
+++ b/docs/relevancy.md
@@ -100,17 +100,16 @@ field and its number of page views in the `vc_14` field.
 
 #### Recency
 
-This is an implementation of [this curve][], and is applied to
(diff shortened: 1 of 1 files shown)


leenagupte (approved):
I don't think I fully understand the `recip` function, but the link to the documentation does work.

PR #2491, opened 2023-01-25
Freshen up ltr/scripts dependencies

sihugh:
## Regenerate python dependencies for /ltr/scripts

These have got a bit out of date, so I've taken the liberty of deleting the
requirements-freeze.txt file and regenerated it using:

`venv/bin/python -m pip freeze > ltr/scripts/requirements-freeze.txt`

I've tested that the process still works locally by downloading
[train, validate and test files](https://github.com/alphagov/search-api/blob/main/ltr/scripts/train.sh#L4-L6)
from a govuk-<environment>-search-relevancy S3 bucket and putting them in
(for this example) `./tmp/ltr_data/`.

I then built and ran the docker container:

```
cd ltr/scripts
docker build . -f Dockerfile.sagemaker -t sagemaker:1
docker run -v $(pwd)/../../tmp:/tmp sagemaker:1
```
(I dropped the number of steps down a bit from 100,000!)

It built a model successfully, so I think this is worth a go.

## Bump tensorflow container version
This now matches the version of tensorflow pinned in requirements-freeze.txt
where the model is generated. If the version is mismatched you get this
sort of error:

> Encountered an error for servable version {name: ltr version: 1}: Not found: Op type not registered 'ParseExampleV2' in binary running on 8b60a663ecbd. Make sure the Op and Kernel are registered in the binary running in this process.

Verified that this works by running ./serve.sh with the model generated from
./train.sh being placed in the expected location.

Running the example request from the docs works OK:

```
curl -X POST http://localhost:8501/v1/models/ltr:regress -d '{
  "signature_name": "regression",
  "examples": [
    {"1":0.0, "2":0.0, "3":0.0000001, "4":0.01000},
    {"1":0.0, "2":0.0, "3":0.0000002, "4":0.11000},
    {"1":0.0, "2":0.0, "3":0.0000002, "4":0.11000},
    {"1":0.0, "2":0.0, "3":0.0000002, "4":0.01000}
  ]
}'
```

```
{
    "results": [0.0932982638, 0.0941789448, 0.0941789448, 0.0932982638]
}
```


leenagupte (approved):
I have independently (but following Simon's instructions) trained a model using the input data copied from an S3 bucket and served it. 🎉 

Results:

```
leenagupte@GDS10896 search-api % curl -X POST http://localhost:8501/v1/models/ltr:regress -d '{
  "signature_name": "regression",
  "examples": [
    {"1":0.0, "2":0.0, "3":0.0000001, "4":0.01000},
    {"1":0.0, "2":0.0, "3":0.0000002, "4":0.11000},
    {"1":0.0, "2":0.0, "3":0.0000002, "4":0.11000},
    {"1":0.0, "2":0.0, "3":0.0000002, "4":0.01000}
  ]
}'
{
    "results": [-1.07155609, -1.06826496, -1.06826496, -1.07155609]
}%           
```

PR #2519, opened 2023-02-09
Remove text filter from Content Type selection in search/all

Tetrino:
It was causing issues for screen readers and was determined to be unneeded for such a small option set.

https://trello.com/c/ktDSbGXC/1596-set-showoptionselectfilter-to-false-on-content-item-facet-filtered-results-of-option-select-autocomplete-difficult-to-find-for-s


AgaDufrat (approved):
Seems fine

PR #2544, opened 2023-03-14
Upgrade Ruby to 3.2

gclssvglx:
Ruby v2.7.6 is EOL on 31 March 2023, so we need to update to a later version. 

This change:

- Updates the version of Ruby to v3.2(.0)
- Fixes linter issues generated by the update above
- Fixes an issue when updating Psych to >= v3.1.0
- Fixes an issue when updating RSpec-Mocks to >= v3.10.3

[Trello](https://trello.com/c/Ih8jw0Hw/365-update-to-ruby-32)


lib/content_item_publisher/content_item_presenter.rb
```diff
@@ -27,7 +27,7 @@ def present
     end
 
     def present_links
-      { content_id: content_id, links: {} }
+      { content_id:, links: {} }
```
Rosa-Fox, line 30: I'm confused about how this is working... does it just automatically use a value with the same name as the key if its not specified?

Rosa-Fox (approved):

lib/content_item_publisher/content_item_presenter.rb
gclssvglx: Yeah, exactly right! I find it kinda confusing to be honest. But, it does clear up the params nicely though.

gclssvglx:
Thanks Rosa

lib/content_item_publisher/content_item_presenter.rb
csutter: Whoever thought we'd be borrowing syntactic sugar from Javascript!

Also thanks for sorting this out - I dreaded having to do it myself as part of the search improvement work ❤️

PR #2550, opened 2023-03-23
Add native Ruby `debug` gem

csutter:
This is the [current state-of-the-art][1] first-party (extended standard library) Ruby debugging gem with excellent editor and tooling support, replacing the need for e.g. `pry` or `byebug` (except for individual preferences, which is why I'm leaving `pry-byebug` in the Gemfile for now).

[1]: https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,7 @@ gem "warden"
 gem "warden-oauth2"
 
 group :development, :test do
+  gem "debug"
   gem "pry-byebug"
   gem "rubocop-govuk", require: false
 end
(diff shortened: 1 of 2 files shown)


csutter:
Well that's fun: https://bugs.ruby-lang.org/issues/19158#note-10

Might park this until `3.2.2` is out.

csutter:
Closing to get this out of the way given we've not upgraded to `3.2.2` yet and probably won't be doing much work on this app anytime soon anyway 🔐

PR #2552, opened 2023-03-28
Add ability to boost custom fields

1pretz1:
We want to boost licence_transaction_industry field for the new licence finder. As we're putting more onus / importance on the industries the it seems reasonable to assume that a search match against an industry name should be weight higher than, at least, the body of the licence pages. 

This change isn't tied specifically to our use case and can be used against any field with a compatible field type. Currently the weighting is just below the description as this seems a reasonable start.  

I've tested the new field type in Staging and it works as expected (includes industry in main search, weights industries over indexable _content, facet filters work). 

This change requires reindexing of the `govuk` index. 

More info in the commits ->

Sister PR: 
- https://github.com/alphagov/finder-frontend/pull/3030

Trello:
https://trello.com/c/VhsBND5i/1888-boost-industry-field

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -1023,7 +1023,7 @@
     "type": "identifiers"
   },
   "licence_transaction_industry": {
-    "type": "searchable_identifiers"
+    "type": "searchable_multivalued_text"
   },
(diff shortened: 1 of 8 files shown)


KludgeKML (approved):
Looks good to me.

PR #2582, opened 2023-05-15
Retrieve page traffic file from S3 instead of stdin

theseanything:
This modified the load_page_traffic script to retrieve the page traffic dump from S3 instead of reading from stdin. This is because the search_analytics now uploads the file to s3 and no longer directly interacts with search-api.

This mirrors the behaviour of [bin/page_traffic_load](https://github.com/alphagov/search-api/blob/main/bin/page_traffic_load)

### DIFF
--- a/bin/load_page_traffic.sh
+++ b/bin/load_page_traffic.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+rake page_traffic:load
+
+SEARCH_INDEX=page-traffic rake search:clean
--- a/lib/tasks/page_traffic.rake
+++ b/lib/tasks/page_traffic.rake
@@ -1,0 +1,17 @@
+require "aws-sdk-s3"
+
+namespace :page_traffic do
+  desc "Bulk load data from Google Analytics"
+  task :load, :environment do
+    s3 = Aws::S3::Client.new
+
+    puts "Downloading file from S3..."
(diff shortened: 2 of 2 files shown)


bin/load_page_traffic.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+rake page_traffic:load
+
+SEARCH_INDEX=page-traffic rake search:clean
+
+SEARCH_INDEX=detailed PROCESS_ALL_DATA=true rake search:update_popularity
+sleep 2400
```
sengi: Why the sleeps? (can we easily avoid them?)

sengi (approved):

bin/load_page_traffic.sh
theseanything: Sorry, good catch, I was meant re-visit this. They were in [the original script](https://github.com/alphagov/search-analytics/blob/d124229da53d9664512a74b47ded8dd50c27e8dc/nightly-run.sh#L28) that this is replacing.

Looking at the comment it seems each of the rake tasks queues a bunch of jobs. The [sleep attempts to prevent filling up Redis](https://github.com/alphagov/search-analytics/commit/8f24581b95af9b37581f16988f466364d5de5808)?

I think I could just call these as three cronjobs that run at intervals - don't think they necessarily need be guaranteed to run consecutively.

theseanything: Like this: https://github.com/alphagov/govuk-helm-charts/pull/1113

PR #2601, opened 2023-05-31
Containerise learn to rank process

theseanything:
As part of the migration of the learn to rank process (LTR) off Jenkins, this containerises the application code needed and so that it can be run using k8s job. This PR removes the bash scripts used by Jenkins to spin up an EC2 instance and run docker to run the relevant ruby and python commands. Instead this migrates the ruby commands into a single rake task required to generate the datasets, then creates a python image to run the commands needed to train and deploy the Sagemaker model.


lib/tasks/learn_to_rank.rake
```diff
@@ -15,7 +25,7 @@ namespace :learn_to_rank do
 
     data = LearnToRank::DataPipeline::Bigquery.fetch(JSON.parse(Base64.decode64(ENV["BIGQUERY_CREDENTIALS"])), viewcount: args.viewcount || 10)
 
-    CSV.open("tmp/#{args[:output]}.csv", "wb") do |csv|
+    csv_body = CSV.generate(headers: true) do |csv|
```
sengi, line 28: What sort of size are we talking for this file? Asking cos I think this is changing it from streaming to buffering the whole thing in RAM. (If it's like tens of megabytes then no big deal. If hundreds then might be worth a few mins faff to keep it streaming?)

ltr/sagemaker/deploy.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # try to find the current endpoint config
+    try:
+        current_endpoint_config_name = session.boto_session.client(
+            "sagemaker"
+        ).describe_endpoint(EndpointName=endpoint_name)["EndpointConfigName"]
+
+        current_endpoint_config = session.boto_session.client(
+            "sagemaker"
+        ).describe_endpoint_config(EndpointConfigName=current_endpoint_config_name,)
```
sengi: nit: stray comma makes this look a bit odd (like a 1-tuple)
```suggestion
        ).describe_endpoint_config(EndpointConfigName=current_endpoint_config_name)
```

sengi: I'd probably factor out the `session.boto_session.client("sagemaker")` so that we're not constructing it multiple times and throwing it away. Also helps readability. e.g.


```suggestion
    sagemaker_client = session.boto_session.client("sagemaker")
    try:
        current_endpoint_config_name = sagemaker_client.describe_endpoint(EndpointName=endpoint_name)["EndpointConfigName"]

        current_endpoint_config = sagemaker_client.describe_endpoint_config(EndpointConfigName=current_endpoint_config_name)
```

etc.

i.e. move the instantiation on line 53 up here.

ltr/sagemaker/deploy.py
```diff
@@ -0,0 +1,115 @@
...
+
+        update_endpoint = True
+    except Exception:
```
sengi, line 20: This catches absolutely everything, including potentially things like syntax errors, calls to nonexistent methods if there are typos, etc. You probably want to catch some slightly more specific type (probably `botocore.exceptions.ClientError`?)

ltr/sagemaker/learntorank.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM public.ecr.aws/docker/library/python:3.10
```
sengi: Probably worth trying the `-slim` or `-alpine` version for a quick win here (mainly in terms of reducing the number of vulnerability reports for packages that we're probably not using!)

sengi: Any reason why not 3.11? (i.e. current)

ltr/sagemaker/deploy.py
```diff
@@ -0,0 +1,115 @@
...
+    sagemaker_model_name = f"search-ltr-{str(time.strftime('%Y-%m-%d-%H-%M-%S'))}"
+    print(f"Model name will be: {sagemaker_model_name}.", file=sys.stderr)
+    print(f"Fetching model from s3://{s3_bucket}/{model_key}", file=sys.stderr)
+
+    # Upload trained model to Sagemaker
+    response = client.create_model(
+        ModelName=sagemaker_model_name,
+        PrimaryContainer={
+            # See https://docs.aws.amazon.com/deep-learning-containers/latest/devguide/deep-learning-containers-images.html
+            'Image': '763104351884.dkr.ecr.eu-west-1.amazonaws.com/tensorflow-inference:2.0.0-cpu',
```
sengi, line 60: Does this need some way to keep it up to date? (is there a tag that tracks the latest patch version, for example? or is it not a big deal?)

ltr/sagemaker/train.py
```diff
@@ -0,0 +1,29 @@
+import sagemaker
+import sys
+import time
+
+def train(s3_bucket, image, role, instance_count, instance_type):
+    session = sagemaker.Session()
+
+    model_name = f"{str(time.time())}"
```
sengi: Simplifies to
```suggestion
    model_name = str(time.time())
```

sengi (approved):

ltr/sagemaker/deploy.py
theseanything: Sorry should have made it clearer - this code is originally from the "ltr/jenkins/deploy.py" file. I've just wrapped it in the deploy function to so we can call it from main.py. I'll try and tidying this code up in a separate PR to address these comments.

ltr/sagemaker/deploy.py
theseanything: Probably, but won't address in the PR as this from copied code.

ltr/sagemaker/train.py
theseanything: of course 🤦🏻 - I'd just removed the fixed text from the original code. https://github.com/alphagov/search-api/blob/38f8965d1d7cc6288b2fbd1b414b8c4e809eac33/ltr/jenkins/train.py#L23

lib/tasks/learn_to_rank.rake
theseanything: Good point - largest file is 50mb, so will leave for now.

ltr/sagemaker/deploy.py
sengi: Ohh sorry, no probs at all!

ltr/sagemaker/learntorank.Dockerfile
theseanything: Good shout - updated to use the slim version. Alpine is missing some system dependencies needed for pandas.

theseanything: Updated - think was just a typo

PR #2662, opened 2023-07-25
Add OIM Project and SAU Referral case types

ryanb-gds:
The Competition and Markets Authority have recently added Office for the Internal Market projects and Subsidy Advice Unit referrals as new case types to their CMA case specialist document type. This adds the new case types to the search API, and also adds a "Not Applicable" market sector because the market sector field does not apply to the new case types.

Trello: https://trello.com/c/r7sZCKI4

### DIFF
--- a/config/schema/elasticsearch_types/cma_case.json
+++ b/config/schema/elasticsearch_types/cma_case.json
@@ -43,6 +43,10 @@
         "value": "mergers"
       },
       {
+        "label": "OIM Project",
+        "value": "oim-project"
+      },
+      {
         "label": "Consumer enforcement",
         "value": "consumer-enforcement"
       },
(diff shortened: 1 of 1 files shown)


config/schema/elasticsearch_types/cma_case.json
```diff
@@ -148,6 +156,10 @@
         "label": "Motor industry",
         "value": "motor-industry"
       },
+      {
+        "label": "Not applicable",
+        "value": "not-applicable"
+      },
```
leenagupte, line 162: It seems really strange that this is the way that CMA have decided to go, and that they are willing to assign all new cases to a "not applicable" market sector forever more, rather than having market sector be an optional field.

Was that considered as an option?

leenagupte (approved):

config/schema/elasticsearch_types/cma_case.json
ryanb-gds: It was considered, and I tried it quickly. What I found was that with the multi-select input component in specialist publisher, if you made it optional the value when no options were selected was an array with an empty string. I could have written a transformer for this case that handled the `[""]` by removing the key or providing a nil value, but that felt pretty hacky to me, and for all I know that might be other finders where `[""]` is a valid value. So we agreed with the CMA to take this approach.

PR #2678, opened 2023-08-08
Fix allowed locations list for international development fund

KludgeKML:
Countries are missing from the allowed list of locations against International Development Fund records, leading to situations where despite multiple countries being available in the search index (so they can be filtered on), they will not be returned in results and cannot be displayed on the results page.

Update this file to match the corresponding file in specialist publisher.

https://govuk.zendesk.com/agent/tickets/5441472


patrickpatrickpatrick (approved):
LGTM!

PR #2682, opened 2023-08-11
Regenerate requirements-freeze.txt

sihugh:
Deploys for LTR are failing at present because of a dependency mismatch


leenagupte (approved):
Building the LTR model worked using these dependencies: https://github.com/alphagov/search-api/actions/runs/5833206967

PR #2688, opened 2023-08-15
Revert "Bump tensorflow from 2.11.0 to 2.11.1 in /ltr/scripts"

leenagupte:
Reverts alphagov/search-api#2551

Updating dependencies has broken Learn to Rank. Until we figure out which dependencies are at fault, we need to revert all of the LTR dependabot updates.


leenagupte:
Opened in error

PR #2691, opened 2023-08-15
Update sagemaker build dependencies

sihugh:
To resolve the following error pip when building the sagemaker container:

> ERROR: failed to solve

There were incompatible versions of some of the dependencies in requirements-freeze.txt.

We also noticed that although the requests package was in requirements.txt, it wasn't explicitly referenced by any of the code. It remains in the freeze file as a transitive dependency.

We've regenerated the freeze file and have tested this via the Github action.


leenagupte (approved):
Thanks for fixing this!

PR #2695, opened 2023-08-22
Build learn-to-rank image on release.

sengi:
Builds of the LTR images have been manual for quite some time and that's been ok, but it's a good idea to ensure that the image is up-to-date.


sihugh:
Have disabled using LTR in https://github.com/alphagov/search-api/pull/2905, so we can close this if you like?

sengi:
Ah awesome, thanks!

PR #2698, opened 2023-08-24
Add assurance technique approach to AI assurance portfolio technique elasticsearch type

ryanb-gds:
Required as part of addition of new assurance technique approach facet to AI Assurance technique specialist finder

Trello: https://trello.com/c/dk6wY8nD
Zendesk: https://govuk.zendesk.com/agent/tickets/5444446

PR #2703, opened 2023-08-31
Update current elasticsearch schema rake task

ryanb-gds:
### Preamble

I appreciate this is a fundamental change to the way search indexes are managed and I definitely see it as a somewhat "risky" change (though I think in reality Elasticsearch mitigates the risk - see below). However, the current schema update process is causing the Publishing Experience team significant pain when we are maintaining Specialist Publisher (see example below) so we would find this work highly valuable. Accordingly, I would expect to be responsible for thoroughly testing and monitoring the effectiveness of this change, especially while it is still bedding in. We feel the most effective testing process will be to run it out of hours on the first few occasions we run it and work on the basis that we expect we may need to run a full reindex afterwards if anything goes wrong. Despite all of that, ultimately the AI + GA4 team owns this app and consequently if you are not willing to support it then I will close the PR, no questions asked.

### What

Provide a rake task for updating the current Elasticsearch index schema in place when Search API configuration has changed

### Why

This rake task has been added because at the moment even purely additive changes to elasticsearch schema configuration require a full reindex. Full reindexes require locking the index for significant periods of time so have to be run out of hours for production and can block integration for long periods. This is causing pain when developing applications that consume the Search API.

A good current example would be the work the Publishing Experience team has been doing to [add a new facet to the "Portfolio of AI Assurance Techniques" specialist finder](https://trello.com/c/dk6wY8nD) on Specialist Publisher. To test this effectively on integration we had to perform a full reindex even though the change only affected 14 content items and was purely additive in terms of the index schema. In this situation, just syncing the existing index with the [new schema configuration](https://github.com/alphagov/search-api/pull/2698) would have significantly reduced the length of our feedback loop and would also make releasing the facet much easier.

The new rake task is idempotent and Elasticsearch will not permit changes to the index that would attempt to change the type of already existing data. Any such changes result in a “Bad Request” error from Elasticsearch and these will be reported to the rake operator. In such cases a full re-index would be required.

I don't think this requires locking because it only really applies to new fields, so no content migration will be required. If content that uses those fields is already present in the index then those items would need to be re-published.

Trello: https://trello.com/c/fvygdACf


lib/schema_synchroniser.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def sync
+    @errors = @index.sync_mappings
+  end
+
+  def synchronised_types
+    @index.mappings.keys.difference(@errors.keys)
+  end
+end
```
ryanb-gds, line 15: I considered adding a unit test for this but it ended up basically just testing a mock of the index group, which felt very inflexible. Unfortunately I couldn't see a way to inject a schema into the actual index and index group objects so I decided to stick with the integration test only, but it can't cover the error handling.

lib/index.rb
```diff
@@ -84,6 +84,18 @@ def with_lock
       end
     end
 
+    def sync_mappings
+      errors = {}
```
csutter: The Rubyist in me would write this as:

```ruby
{}.tap do |errors|
  mappings.each do |type, mapping|
    # ...
  end
end
```

Saves you from having to declare and return `errors`!

lib/schema_synchroniser.rb
```diff
@@ -0,0 +1,15 @@
+class SchemaSynchroniser
+  attr_reader :errors
+
+  def initialize(index_group)
+    @index = index_group.current
+  end
+
+  def sync
```
csutter: Another general comment: Given this class looks a lot like a command/interactor type thing, I'd name the "main" method of it `#call` which is more conventional these days in the Ruby world (as the class name already says what it does anyway).

This can give you lots of nifty advantages around DI and testing ([this is a good article that mostly aligns with my thoughts](https://www.alchemists.io/articles/command_pattern)) and makes it clearer that this is a single-responsibility command object.

csutter (approved):

lib/index.rb
ryanb-gds: I've been spending too much time in Java world in the last year or so. Thanks!

lib/schema_synchroniser.rb
ryanb-gds: Done. Thanks for the link, interesting read

ryanb-gds:
~~Frustratingly, this has suddenly started failing due to https://github.com/elastic/elastic-transport-ruby/issues/63. Looking to see if we can upgrade/downgrade out of trouble~~

Work around found

ryanb-gds:
This ran successfully on integration and I was able to confirm the schema changes were working (see screenshot below), so I think we're ready to merge this.
Screenshot 2023-09-18 at 13 31 11

PR #2717, opened 2023-09-18
Remove explicit irb dependency

ryanb-gds:
For the last two weeks deployments to integration haven't been working because the irb gem has a sub-dependency that requires lib-yaml to be installed in the docker container, which it isn't.

The irb gem probably isn't required as it is part of the default Ruby install. An explicit dependency on irb was introduced in [this commit](https://github.com/alphagov/search-api/commit/e75c233bbd5e8aa9c840babf91ec0b33bd82058c) which states it was removed from Ruby in Ruby 2.6, but presumably that decision was reversed according to the [gem's readme](https://github.com/ruby/irb)

I deployed this branch to integration and it fixed the build: https://github.com/alphagov/search-api/actions/runs/6221593937

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,6 @@ gem "govuk_message_queue_consumer"
 gem "govuk_schemas"
 gem "govuk_sidekiq"
 gem "httparty"
-gem "irb", require: false
 gem "logging"
 gem "loofah"
 gem "nokogiri"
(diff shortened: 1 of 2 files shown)


ollietreend (approved):
This seems like a perfectly safe change to make. [No other apps] in the alphagov org include `gem "irb"` in their Gemfile.

Looks good to me 👍🏻

PR #2762, opened 2023-11-23
Configure dependabot to ignore elasticsearch beyond 6.x

peteglondon:
We can't upgrade `elasticsearch` to version 7.

This PR should stop dependendabot from opening PRs to upgrade.

See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,10 @@ updates:
     directory: /
     schedule:
       interval: daily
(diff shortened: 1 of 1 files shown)


gclssvglx (approved):
Cool - thanks @peteglondon 👍

PR #2768, opened 2023-11-24
Update govuk sidekiq take 2

peteglondon:
The [first attempt](https://github.com/alphagov/search-api/pull/2743) failed in production due to an error in way requires are handled in specs.
*  I added `require "active_support/time"` to one of the specs to support `1.seconds`
*  this means it was loaded when running the entire suite and specs didn't fail - it did fail when you ran the single spec and the require had not happened.
*  this caused a bug in production because this require was missing from the production code.
*  removing the require from the spec made the test [fail as it should](https://github.com/alphagov/search-api/actions/runs/6983062573/job/19003479337)
* Final commit adds the missing require


lib/auth/gds_sso.rb
```diff
@@ -1,3 +1,5 @@
+require "active_support"
```
gclssvglx, line 1: Do we not also need `require "active_support/time"` here too @peteglondon ?

gclssvglx: Ah, I see - didn't read all the commits 👍

gclssvglx (approved):
Looks good and makes sense to me - my initial thought was add `require "active_support/time"` to GdsSso. Nice write-up.

PR #2840, opened 2024-02-21
Remove specialist topic code from repo

georges1996:
Ticket: https://trello.com/c/gbfKtIeY/2401-remove-specialist-topic-code-from-search-m

## WHAT

Remove some/all references to `specialist_sectors` from Search API. 

## WHY

As we are retiring specialist topics.


hannako (approved):
The code removal looks good to me. I tested the branch on integration, and I can't see any issues with specialist finders as a result of these changes.

kevindew (approved):
Nice one, great to see so much code removed 🏆 

Do we need to do a reindex after this?

hannako:
@kevindew The million dollar question. I don't _think_ so. 
- All specialist topics have been archived now https://www.gov.uk/api/search.json?filter_format=specialist_sector 
- There was no boosting being applied to documents tagged to specialist topic,[ boosts are on document type or supergroup](https://github.com/alphagov/search-api/blob/main/config/query/boosting.yml#L21)
- So if we don't reindex, I think it means that existing documents will still have the specialist-sector field, where as new ones will not (as the mapping has updated). But I don't think there's any negative consequences of that. 
- Plus SearchAPI V1 isn't being used for Site Search anymore, so I guess this reduces the risk further. 

@leenagupte do you agree?

leenagupte:
> @kevindew The million dollar question. I don't _think_ so.
> 
> * All specialist topics have been archived now https://www.gov.uk/api/search.json?filter_format=specialist_sector
> * There was no boosting being applied to documents tagged to specialist topic,[ boosts are on document type or supergroup](https://github.com/alphagov/search-api/blob/main/config/query/boosting.yml#L21)
> * So if we don't reindex, I think it means that existing documents will still have the specialist-sector field, where as new ones will not (as the mapping has updated). But I don't think there's any negative consequences of that.
> * Plus SearchAPI V1 isn't being used for Site Search anymore, so I guess this reduces the risk further.
> 
> @leenagupte do you agree?

This sounds correct to me. A re-index isn't required for new site search, and it doesn't seem worth the effort to remove a field that isn't being used.

PR #2905, opened 2024-05-09
Disable Learning to Rank's involvement in queries

sihugh:
Learning to Rank only changes the ordering of results when the query includes keywords and is ordered by relevance.

It can only affect how results rank within a single page. The maximum effect it could have is to reverse the order of a page of results. The results all still appear on the same page.

Site search searches with keywords (the majority of searches) are now served by Vertex AI search. The cost and complexity of running a Learning to Rank training and serving pipeline for the other use cases no longer seems commensurate with the impact.

In addition to this, the training of our model has not been supervised effectively for a while, leading to a drop in the quality of the model.

I therefore propose disabling the use of Learning to Rank in Search-API (this change) which allows the removal of the code to invoke the model, the pipeline that builds and deploys the model and the Terraform to support the deployment.

### DIFF
--- a/lib/search/query_parameters.rb
+++ b/lib/search/query_parameters.rb
@@ -77,7 +77,7 @@ module Search
     end
 
     def rerank
-      RelevanceHelpers.ltr_enabled? && [nil, "relevance"].include?(order) && ab_tests[:relevance] != "disable"
+      false
     end
 
     def use_shingles?


sihugh:
A side effect of this is a fall in response times for keyword queries (in particular) because we:
- won't have the extra network call to SageMaker (immediate)
- can request [less information from ElasticSearch](https://github.com/alphagov/search-api/commit/3d663819959cf2b4c2544c5b14e0214149c2b8fd#diff-99f6f86067010c5c5f5ec949246520801812df3db191fe53a7ffaaf855f93352R40-R41) on each query because we don't need to send it to the reranker (future?)

csutter (approved):
🙏

KludgeKML:
This also starts clearing up the tech debt where marking the ab_test parameter is the only way to turn off LTR on an individual finder (the finder that replaced licence-finder had to do this to make the results usable)

KludgeKML (approved):

leenagupte (approved):
🎉 🏅 🏆

PR #2956, opened 2024-07-08
Reorder facets on official documents

KludgeKML:
- Non-expandable "of type" facet moved up so that it isn't confused with the Topic drop-down (which is expandable but closed by default so looks like the title of the "of type" radios.

https://trello.com/c/Mg2RDhII/230-confusing-order-of-filters-on-the-official-documents-finder

## Before

Screenshot 2024-07-08 at 11 18 50

## After

Screenshot 2024-07-08 at 11 20 37


andysellick (approved):
Nice work Keith 👍

PR #2964, opened 2024-07-26
TESTING BRANCH - Migrate GA4 logic to search-api

georges1996:
Ticket: https://trello.com/c/SHc04ixR/110-search-analytics-pipeline-ua-to-ga4

The final output is the 'popularity' fields: https://www.gov.uk/api/search.json?fields=popularity_b,popularity

As part of the migration to GA4, the task was to understand the existing Python application and its output https://github.com/alphagov/search-analytics, then to update the Ruby application https://github.com/alphagov/search-analytics-ga4/ in order to get an identical/very similar output to the Python app. 

Once this was correct the next part was to migrate the Search Analytics GA4 code from the Ruby repo to be part search-api repo (rationale: it’s easier to run the task if part of search-api - can be done on a pod in K8s, whereas running search-analytics from GitHub actions was disastrous for catching failures, it’ll be easier to maintain as a single codebase and not a separate thing for teams to look at) this decision will also bring down costs as we no longer need to maintain another code base and we don't need to have an S3 bucket anymore to dump the data to and for search-api to pick up, it can all be done as part of the `rake page_traffic:load` task

## Rollback Plan

If in the unlikely case this fails in production we have a way to roll back. This will be that we revert this PR, and then it will revert back to using the data in the S3 bucket. We will then have to rerun the cron jobs in Argo to reset the popularity back. 

- `search-api-load-page-traffic` which pulls the data from GA and ingests into Elastic Search
- `search-api-update-govuk-index-popularity` will use this new data from Elastic Search and update the search api popularity results here: https://www.gov.uk/api/search.json?fields=popularity_b,popularity


Gemfile
```diff
@@ -3,8 +3,10 @@ source "https://rubygems.org"
 gem "activesupport"
 gem "aws-sdk-s3"
 gem "bootsnap", require: false
+gem "dotenv"
```
csutter: We don't need to (and shouldn't) use `dotenv` in production, so it should live in the development+test group.

lib/analytics/ga4/formatter.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
```
csutter: Think this is probably a hangover from when this code was a separate app, but this repo `freeze`s strings directly instead of having the magic comment.

lib/analytics/ga4/formatter.rb
```diff
@@ -0,0 +1,74 @@
...
+
+module Analytics
+  module Ga4
+    class Formatter
```
csutter: This class and its logic is very verbose, a bit messy, and could do with a cleanup – in particular it:
* seems to take a lot of primitive data out of `PageData` and perform operations on it, instead of `PageData` owning this logic (["feature envy"](https://refactoring.guru/smells/feature-envy))
* have complex state that isn't idempotent (it sets up a Hash in the initializer and then manipulates it in hard to follow ways in `consolidated_page_views` and its callees)
* have lots of methods named after the operations they perform, rather than the result you want (e.g. `#replace_empty_string_with_slash`)
* does some faffy things that have more idiomatic ways of doing them in Ruby

How about amending `PageData` like so:
```ruby
module Analytics
  module Ga4
    class PageData
      PAGE_NOT_FOUND_TITLE = "Page not found".freeze

      attr_reader :path, :title, :page_views

      def initialize(path, title, page_views)
        @path = path
        @title = title
        @page_views = page_views.to_i
      end

      # Whether the page should be excluded from popularity calculations
      def excluded?
        non_relative_path? || smart_answer? || not_found?
      end

      # Returns the path without any query string or trailing slash
      def normalised_path
        path.partition("?").first.chomp("/").presence || "/"
      end

    private

      def non_relative_path?
        !normalised_path.start_with?("/")
      end

      def smart_answer?
        normalised_path.include?("/y/")
      end

      def not_found?
        title.include?(PAGE_NOT_FOUND_TITLE)
      end
    end
  end
end
```

And then this entire class can just be a super simple transformation pipeline:
```ruby
module Analytics
  module Ga4
    class Formatter
      def initialize(ga_data)
        @ga_data = ga_data
      end

      def consolidated_page_views
        @consolidated_page_views ||= ga_data
          .reject(&:excluded?)
          .group_by(&:normalised_path)
          .transform_values { |grouped_pages| grouped_pages.sum(&:page_views) }
          .sort_by { |_, views| -views }
          .to_h
      end

    private

      attr_reader :ga_data
    end
  end
end
```

The tests for these can then be cleaned up/separated as well.

lib/analytics/ga4/client.rb
```diff
@@ -0,0 +1,15 @@
+require "google/analytics/data/v1beta"
```
csutter: This – and all other `require`/`require_relative` calls in this PR should live in `lib/rummager.rb` for consistency with the rest of Search API.

External libraries like this in the [first block on top](https://github.com/alphagov/search-api/blob/main/lib/rummager.rb#L3-L36), the rest in an appropriate spot (maybe at the end of the file).

Gemfile
```diff
@@ -3,8 +3,10 @@ source "https://rubygems.org"
...
 gem "elasticsearch", "~> 6" # We need a 6.x release to interface with Elasticsearch 6
 gem "gds-api-adapters"
+gem "google-analytics-data-v1beta"
```
csutter, line 8: This should use `google-analytics-data` instead of the beta gem (which will still allow you to use beta functionality if needed, and default to the beta internally at the moment anyway)

lib/analytics/ga4/client.rb
```diff
@@ -0,0 +1,15 @@
...
+require "dotenv"
+Dotenv.load
+
+module Analytics
+  module Ga4
+    class Client
+      attr_reader :client
+
+      def initialize
+        @client = ::Google::Analytics::Data::V1beta::AnalyticsData::Client.new
```
csutter: We should use the idiomatic Google™ way of initialising their client here:

```ruby
@client = ::Google::Analytics::Data.analytics_data
```

(note this will need `google/analytics/data` to be required in `lib/rummager.rb` instead of `google/analytics/data/v1beta`)

lib/analytics/ga4/google_analytics_service.rb
```diff
@@ -0,0 +1,59 @@
...
+require_relative "./page_data"
+require_relative "./client"
+
+require "dotenv"
+
+Dotenv.load
+
+module Analytics
+  module Ga4
+    class GoogleAnalyticsService
```
csutter: Again this class is a bit confusing and can do with a cleanup:
* There shouldn't need to be a separate `LIMIT` and `OFFSET_INCREMENT`, a `PAGE_SIZE` or similar will do
* `get_paginated_data` might not be the best name given that method ultimately gets _all_ data
* `get_data` converts the Google objects into a hash for no particular reason so then you have to faff about with hashes instead of the native Google objects

I personally find `get_paginated_data` overly verbose, imperative, and hard to read, and would go for a lazy functional approach for this kind of paginated API fetching:

```ruby
module Analytics
  module Ga4
    class GoogleAnalyticsService
      PER_PAGE = 10_000

      def initialize(ga_client: ::Google::Analytics::Data.analytics_data)
        @ga_client = ga_client
      end

      def data
        # https://developers.google.com/analytics/devguides/reporting/data/v1/basics#navigate_long_reports
        @data ||= (0..).lazy
          .map { |offset| Request.new(offset * PER_PAGE, PER_PAGE) }
          .map { |request| ga_client.run_report(request.analytics_data) }
          .take_while { |data| data.row_count.positive? }
          .flat_map(&:rows)
          .flat_map { |row| format_row(row) }
      end

    private

      attr_accessor :ga_client

      def format_row(row)
        path = row.dimension_values.first.value
        title = row.dimension_values.last.value
        page_views = row.metric_values.first.value

        Analytics::Ga4::PageData.new(path, title, page_views)
      end
    end
  end
end
```

I know different opinions are available though, and the concept of lazy enumerators may send someone's head in a tailspin in the future. Paging @kevindew for a tie breaker!

lib/analytics/ga4/client.rb
```diff
@@ -0,0 +1,15 @@
...
+      end
+    end
+  end
+end
```
kevindew: This class seems rather redundant - why not just create the `::Google::Analytics::Data::V1beta::AnalyticsData::Client.new` in `GoogleAnalyticsService`

lib/analytics/ga4/client.rb
```diff
@@ -0,0 +1,15 @@
+require "google/analytics/data/v1beta"
+require "dotenv"
+Dotenv.load
```
kevindew: It seems peculiar to do this in the midst of library code, you'd expect it to be done in like a boot section of an app.

Rather than move this though, I suggest we just forgo dotenv for this as it would be pretty weird having a .env file only for importing analytics and not the rest of the app. We can just leave a comment in the documentation on how to set the env vars up locally.

lib/analytics/ga4/formatter.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+module Analytics
+  module Ga4
```
kevindew: I think we could benefit from naming this to refer to importing since all the code under this module would be for that purpose. How about Ga4Import ?

lib/analytics/ga4/google_analytics_service.rb
kevindew: This class name looks like it could be improved to express what it does as GoogleAnalyticsService sounds quite generic.

Is it a DataFetcher?

lib/analytics/ga4/google_analytics_service.rb
```diff
@@ -0,0 +1,59 @@
...
+      OFFSET = 0
+      LIMIT = 10_000
+      OFFSET_INCREMENT = 10_000
+
+      def initialize
+        @ga_client = Analytics::Ga4::Client.new
+        @all_data = []
+      end
+
+      def get_paginated_data
```
kevindew: Seems unnecessary for this to have `get_` given the lack of convention for that in Ruby and `paginated` given this seems to be something done within the class and not a concern of a caller of this method. 

I'd probably suggest just calling this "#call" given it's the only public method.

spec/unit/analytics/ga4/formatter_spec.rb
```diff
@@ -0,0 +1,116 @@
+require "spec_helper"
+require_relative "../../../../lib/analytics/ga4/formatter"
+require_relative "../../../../lib/analytics/ga4/page_data"
```
kevindew:
```suggestion
require "analytics/ga4/formatter"
require "analytics/ga4/page_data"
```

same for other tests.

lib/analytics/ga4/google_analytics_service.rb
```diff
@@ -0,0 +1,59 @@
+require "google/analytics/data/v1beta"
+require_relative "./request"
+require_relative "./page_data"
+require_relative "./client"
+
+require "dotenv"
+
+Dotenv.load
```
kevindew: Similar comment to above about an unusual place for Dotenv.load, would suggest removing.

lib/tasks/page_traffic.rake
```diff
@@ -1,17 +1,27 @@
 require "aws-sdk-s3"
+require "analytics/ga4/google_analytics_service"
+require "analytics/ga4/formatter"
+require "analytics/ga4/relevancy_calculator"
+
+require "stringio"
 
 namespace :page_traffic do
   desc "Bulk load data from Google Analytics"
   task :load, :environment do
```
kevindew, line 5: It'd be good to have a test for this as well.

lib/tasks/page_traffic.rake
```diff
@@ -1,17 +1,27 @@
...
+    puts "Processing Ga4 Analytics..."
+    google_analytics_service = Analytics::Ga4::GoogleAnalyticsService.new
+    paginated_data = google_analytics_service.get_paginated_data
+    consolidated_page_views = Analytics::Ga4::Formatter.new(paginated_data).consolidated_page_views
+    relevancy_calculator = Analytics::Ga4::RelevancyCalculator.new(consolidated_page_views)
 
-    puts "Downloading file from S3..."
-    resp = s3.get_object(bucket: ENV["AWS_SEARCH_ANALYTICS_BUCKET"], key: "page-traffic.dump")
+    relevancy = relevancy_calculator.relevance
+    stream = StringIO.new(relevancy.to_s)
```
kevindew: Rather than changing our string into a StringIO I think it'd be better to change `GovukIndex::PageTrafficLoader` to accept a string. Looking at it I think it may even work without changes.

lib/tasks/page_traffic.rake
```diff
@@ -1,17 +1,27 @@
...
   desc "Bulk load data from Google Analytics"
   task :load, :environment do
-    s3 = Aws::S3::Client.new
+    # This task will now start the process of generating the data from Google Analytics and sending it to GovukIndex::PageTrafficLoader for processing
+
+    puts "Processing Ga4 Analytics..."
+    google_analytics_service = Analytics::Ga4::GoogleAnalyticsService.new
+    paginated_data = google_analytics_service.get_paginated_data
+    consolidated_page_views = Analytics::Ga4::Formatter.new(paginated_data).consolidated_page_views
+    relevancy_calculator = Analytics::Ga4::RelevancyCalculator.new(consolidated_page_views)
```
kevindew: All these operations (and `relevancy = relevancy_calculator.relevance`, but I can't include it in this comment) look like they're coupled together. I'd imagine it'd make sense to put them together in a simple transaction object that can pull these together. We can then test that individually and just have a single item here:

`relevancy = Analytics::Ga4::RelevanceReportGenerator.call`

lib/analytics/ga4/google_analytics_service.rb
```diff
@@ -0,0 +1,59 @@
...
+    class GoogleAnalyticsService
+      attr_accessor :ga_client, :all_data
+
+      OFFSET = 0
+      LIMIT = 10_000
+      OFFSET_INCREMENT = 10_000
+
+      def initialize
+        @ga_client = Analytics::Ga4::Client.new
+        @all_data = []
```
kevindew: I'm not sure why this is an instance variable when it's only used in get_paginated_data. We should just set it as a regular variable in there.

Otherwise, if someone was to call `get_paginated_data` twice they'd get the first result normal and the second time it would combine first and second results.

lib/analytics/ga4/google_analytics_service.rb
```diff
@@ -0,0 +1,59 @@
...
+  module Ga4
+    class GoogleAnalyticsService
+      attr_accessor :ga_client, :all_data
+
+      OFFSET = 0
+      LIMIT = 10_000
+      OFFSET_INCREMENT = 10_000
+
+      def initialize
+        @ga_client = Analytics::Ga4::Client.new
```
kevindew:
```suggestion
        @ga_client = Client.new
```

given we're already in this namespace.

lib/analytics/ga4/google_analytics_service.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    private
+
+      def format_all_data(data)
+        data.map do |row|
+          path = row[:dimension_values].first[:value]
+          title = row[:dimension_values].last[:value]
+          page_views = row[:metric_values].first[:value]
+
+          Analytics::Ga4::PageData.new(path, title, page_views)
```
kevindew:
```suggestion
          PageData.new(path, title, page_views)
```

again already in this namespace.

lib/analytics/ga4/google_analytics_service.rb
```diff
@@ -0,0 +1,59 @@
...
+        loop do
+          data = get_data(offset)
+          break if data.nil? || data[:rows].nil? || data[:rows].empty?
+
+          all_data << format_all_data(data[:rows])
+
+          offset += OFFSET_INCREMENT
+        end
+
+        all_data.flatten!
```
kevindew:
```suggestion
        all_data.flatten
```

as we're returning data here it doesn't seem necessary to modify the source data.

lib/analytics/ga4/google_analytics_service.rb
```diff
@@ -0,0 +1,59 @@
...
+require "dotenv"
+
+Dotenv.load
+
+module Analytics
+  module Ga4
+    class GoogleAnalyticsService
+      attr_accessor :ga_client, :all_data
+
+      OFFSET = 0
```
kevindew: I'm not sure what the purpose of this constant is, would we ever want to start at an offset that isn't 0?

spec/unit/analytics/ga4/google_analytics_service_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    allow(::Google::Analytics::Data::V1beta::AnalyticsData::Client).to receive(:new).and_return(ga_client)
+    allow_any_instance_of(Analytics::Ga4::Client).to receive(:client).and_return(client)
+    allow(client)
+      .to receive(:run_report)
+      .and_return(run_report_response, run_report_response_empty_rows)
+  end
+
+  describe "#get_paginated_data" do
+    it "returns the correct data" do
+      data = Analytics::Ga4::GoogleAnalyticsService.new.get_paginated_data
```
kevindew:
```suggestion
      data = described_class.new.get_paginated_data
```

lib/analytics/ga4/request.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      def screen_page_views
+        # https://developers.google.com/analytics/devguides/reporting/data/v1/api-schema#metrics
+        Google::Analytics::Data::V1beta::Metric.new(
+          name: "screenPageViews",
+        )
+      end
+
+      def date_range
+        # Needs to be 13 days once done testing
```
kevindew: Looks like this is done now.

lib/analytics/ga4/request.rb
```diff
@@ -0,0 +1,58 @@
...
+  module Ga4
+    class Request
+      attr_accessor :offset, :limit
+
+      def initialize(offset, limit)
+        @offset = offset
+        @limit = limit
+      end
+
+      def analytics_data
```
kevindew: This seems a little misleadingly named if I understand it correctly. As I understand this doesn't actually return data as per our other uses of data and instead returns an object that can be used to request data - maybe we should call it a report_request?

kevindew: Looking further it looks like this class is just essentially a factory for generating this report_request, perhaps calling this class Request doesn't make sense either - since it isn't modelling a Ga4 request like the naming would imply - and instead call it a ReportRequestCreator or ReportRequestBuilder

lib/analytics/ga4/formatter.rb
kevindew: This class also feels a bit misleadingly named (sorry for yet another) since it's role seems to be to consolidate the already formatted data and not do formatting. Maybe it should be called PageViewConsolidator ?

spec/unit/analytics/ga4/formatter_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+RSpec.describe Analytics::Ga4::Formatter do
+  describe "#consolidated_page_views" do
+    it "doesn't have a slash" do
```
kevindew:
```suggestion
    it "ignores input that isn't prefixed with a slash" do
```
Currently the `it` doesn't read correctly as it doesn't describe how the method behaves in that scenario.

kevindew: Same problems can be seen for most of the it statements here, they seem to be written as if they were contexts. Should be pretty straight forward to fix.

spec/unit/analytics/ga4/formatter_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+        Analytics::Ga4::PageData.new("/example2", "not a smart answer", "10"),
+        Analytics::Ga4::PageData.new("/example2?something=something", "not a smart answer", "30"),
+      ]
+      formatter = Analytics::Ga4::Formatter.new(page_data)
+      normalised_data = formatter.consolidated_page_views
+
+      expect(normalised_data).to eq({ "/example" => 40, "/other" => 20, "/example2" => 40 })
+    end
+
+    it "removes data with the title and path of (other)" do
```
kevindew: I'm confused by this test, we don't seem to have any logic for `(other)`?

lib/analytics/ga4/formatter.rb
```diff
@@ -0,0 +1,74 @@
...
+    class Formatter
+      PAGE_NOT_FOUND = "Page not found"
```
kevindew: Are we confident this will work? I ask because the title on 404s is: "Page not found - GOV.UK" 

Perhaps we want to do a pattern match of /Page\snot\sfound/ ?

lib/analytics/ga4/formatter.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+
+      def page_not_found(page_data)
+        page_data.title.include?(PAGE_NOT_FOUND)
+      end
+
+      def sort_by_page_views_descending
+        sorted_by_page_views_desc = @consolidated_page_views.sort_by { |_key, value| value }.reverse!
+        sorted_by_page_views_desc.to_h
+      end
```
kevindew: It seems unusual that this returns a sorted hash (and I can see a challenge was hit in the tests). 

Would it be better if this returned an array of `[[path, page_views], [path, page_views]]` that feels more idiomatic.

lib/analytics/ga4/relevancy_calculator.rb
```diff
@@ -0,0 +1,57 @@
+require "json"
+module Analytics
+  module Ga4
+    class RelevancyCalculator
```
kevindew: Sorry to be yet again on the class naming, this one seems confusing too as it seems to not be used as a calculator but instead as a means to create data structures that elastic search would understand.

The fact that does generate elastic search data seems rather wild - as you'd expect that to be a separate concern, but that seems impractical to unpick at this point so we may just want to try use naming to reduce surprise.

maybe ElasticSearchRelevancySerialiser ? not sure if that's a good suggestion, I'm struggling a bit on this one.

lib/analytics/ga4/relevancy_calculator.rb
```diff
@@ -0,0 +1,57 @@
...
+          (1..components.length).each do |i|
+            result.append("/#{components.first(i).join('/')}")
+          end
+        end
+        result
+      end
+
+      def total_page_views
+        all_page_views = consolidated_data.values
+        all_page_views.reduce(:+)
```
kevindew:
```suggestion
        @all_page_views ||= consolidated_data.values.reduce(:+)
```

save us having to recalculate this same number each iteration

lib/analytics/ga4/relevancy_calculator.rb
```diff
@@ -0,0 +1,57 @@
...
+          path_components: path_components(base_path),
+          rank_14: index,
+          vc_14: page_views,
+          vf_14: page_views / total_page_views.to_f,
+        }
+      end
+
+      def path_components(path)
+        result = []
+        if path.start_with?("/")
```
kevindew: Is there a scenario when a path won't start with a `"/"` ? I'd guess we never want to return an empty result here.

spec/unit/analytics/ga4/relevancy_calculator_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "spec_helper"
+require_relative "../../../../lib/analytics/ga4/relevancy_calculator"
+
+RSpec.describe Analytics::Ga4::RelevancyCalculator do
+  describe "#relevance" do
+    it "formats relevancy JSON" do
+      consolidated_data = { "/example" => 40, "/example2" => 20, "/example3" => 100, "/example3/child" => 100 }
+
+      relevancy_calculator = Analytics::Ga4::RelevancyCalculator.new(consolidated_data)
```
kevindew:
```suggestion
      relevancy_calculator = described_class.new(consolidated_data)
```

kevindew (commented):
Thanks for this eoges1996. This looks good, it took me a bit of a while to get my head around what is going on, but once past that it doesn't seem too complex. I think it could benefit from some class renaming, as I've commented a few times, to make it easier to understand - though I do understand this naming is what you've ported rather than named yourself. 

Have you been able to test it on integration? I guess a concern to rule out is that the script exhausts the available memory on the pod.

lib/analytics/ga4/request.rb
csutter: Let's make that magic number a constant and/or configurable

lib/tasks/page_traffic.rake
```diff
@@ -1,17 +1,27 @@
...
+
+require "stringio"
 
 namespace :page_traffic do
   desc "Bulk load data from Google Analytics"
   task :load, :environment do
-    s3 = Aws::S3::Client.new
+    # This task will now start the process of generating the data from Google Analytics and sending it to GovukIndex::PageTrafficLoader for processing
+
+    puts "Processing Ga4 Analytics..."
```
csutter: This would probably be better off as logging, and maybe let's add some logging at the end too?

lib/analytics/ga4/client.rb
csutter: Plus what @kevindew said about this class being redundant :+1:

lib/analytics/ga4/relevancy_calculator.rb
```diff
@@ -0,0 +1,57 @@
+require "json"
+module Analytics
+  module Ga4
+    class RelevancyCalculator
+      def initialize(consolidated_data)
+        @consolidated_data = consolidated_data
+      end
+
+      def relevance
+        consolidated_data.map.with_index(1) { |(base_path, page_views), index|
```
csutter: This should be a `flat_map` instead of `map ... flatten`

csutter (changes requested):
Looks like @kevindew pipped me to the post!

Appreciate this looks like a _lot_ of feedback but I hope it's less than it looks like. Thanks for taking up what I'm sure has been a very faffy and thankless piece of work 🙂

It looks like the code works which is the most important bit, now it's just about making it fit in better and tidying up some of the verbosity and readability of some of the classes.

First and foremost we should refactor the code loading to be more in line with Search API overall (i.e. moving it all from the individual source files to `lib/rummager.rb`).

Then I've got some minor comments about the use of the Google client, and some thoughts on how to make the core functionality more functional and Ruby-like.

Gemfile
georges1996: Chatted with Christian, and agreed this can be done at a later date, we will put this in a backlog and note it down when I move of this work.

lib/analytics/ga4/client.rb
georges1996: Chatted with Christian, and agreed this can be done at a later date, we will put this in a backlog and note it down when I move of this work.

Gemfile
csutter: Now that the code using dotenv is gone we need to remove the gem too 😄

lib/analytics/ga4_import/relevance_report_generator.rb
```diff
@@ -0,0 +1,18 @@
...
+        new.call
+      end
+
+      def call
+        google_analytics_service = Analytics::Ga4Import::DataFetcher.new
+        paginated_data = google_analytics_service.call
+        consolidated_page_views = Analytics::Ga4Import::PageViewConsolidator.new(paginated_data).consolidated_page_views
+        elastic_search_relevancy_serialiser = Analytics::Ga4Import::ElasticSearchRelevancySerialiser.new(consolidated_page_views)
+        relevancy = elastic_search_relevancy_serialiser.relevance
+        relevancy.to_s
```
csutter: Trying to trace this through the change – this is an Array that gets serialized into a string that is then used to import into ES using `JSON.parse`? So this should really be `#to_json`, no?

lib/analytics/ga4_import/relevance_report_generator.rb
```diff
@@ -0,0 +1,18 @@
+module Analytics
+  module Ga4Import
+    class RelevanceReportGenerator
+      def self.call
+        new.call
+      end
+
+      def call
+        google_analytics_service = Analytics::Ga4Import::DataFetcher.new
```
csutter: This should probably be called `data_fetcher`, not `google_analytics_service`

lib/analytics/ga4_import/relevance_report_generator.rb
```diff
@@ -0,0 +1,18 @@
+module Analytics
+  module Ga4Import
+    class RelevanceReportGenerator
+      def self.call
+        new.call
+      end
+
+      def call
```
csutter, line 8: This method could probably also use some tidying up/converting to a more pipeline-style approach, e.g.:

```ruby
def initialize
  @data_fetcher = DataFetcher.new
end

def call
  data_fetcher.call
    .then { |paginated_data|
      PageViewConsolidator.new(paginated_data).consolidated_page_views
    }
    .then { |consolidated_page_views|
      ElasticSearchRelevancySerialiser.new(consolidated_page_views)
    }
    .relevance
    .to_json
end

private

attr_reader :data_fetcher
```

lib/analytics/ga4_import/report_request_builder.rb
```diff
@@ -0,0 +1,59 @@
+module Analytics
+  module Ga4Import
+    class ReportRequestBuilder
+      attr_accessor :offset, :limit
+
+      START_DATE = 13
```
csutter: This could use a better name reflecting what it is, and maybe use `#days` to be more descriptive. I'd also say maybe setting it to 14 and then taking the days off of today rather than yesterday to make it more obvious what is going on?

Maybe:

```ruby
REPORT_WINDOW = 14.days
```

lib/analytics/ga4_import/report_request_builder.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      def screen_page_views
+        # https://developers.google.com/analytics/devguides/reporting/data/v1/api-schema#metrics
+        Google::Analytics::Data::V1beta::Metric.new(
+          name: "screenPageViews",
+        )
+      end
+
+      def date_range
+        start_date = Date.today.prev_day - START_DATE
```
csutter: We have ActiveSupport so `Date.yesterday` should be available. Also see my comment above about calculating start date from today instead of yesterday.

```ruby
start_date = Date.today - REPORT_WINDOW
end_date = Date.yesterday
```

lib/tasks/page_traffic.rake
georges1996: I had a go at this and was very painful, to set up due to the dynamic url's being used, and the data being passed around the app from elasticsearch response. I'd be keen to get this work complete and revisit, to add a specific rake task test. I notice search-api does not test rake tasks like this one, however the unit test coverage should ensure this works as expected.

lib/analytics/ga4/formatter.rb
georges1996: I dont have a strong preference to this. I'd be happy to keep as is, as we know this works as expected. However can be revisited, as the scope of this task was just to "migrate", saying this we have defo improved the look of the code to fit more with search-api

config/sidekiq.yml
```diff
@@ -1,6 +1,6 @@
 ---
 tag: search-api.publishing.service.gov.uk
-:concurrency: 5
+:concurrency: 12
```
kevindew, line 3: I assume this was a temp thing you don't mean to merge?

lib/analytics/ga4_import/data_fetcher.rb
```diff
@@ -0,0 +1,49 @@
+module Analytics
+  module Ga4Import
+    class DataFetcher
+      attr_accessor :ga_client
+
+      LIMIT = 10_000
+      OFFSET_INCREMENT = 10_000
```
kevindew: Do we need these as two separate constants? I can't imagine why we'd want them to ever differ.

lib/analytics/ga4_import/data_fetcher.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+
+      def call
+        # https://developers.google.com/analytics/devguides/reporting/data/v1/basics#navigate_long_reports
+        offset = 0
+        all_data = []
+
+        loop do
+          data = get_data(offset)
+          break if data.nil? || data[:rows].nil? || data[:rows].empty?
```
kevindew: There are a lot of different break conditions here, do we need them all - I assume it's predictable?

lib/govuk_index/updater.rb
```diff
@@ -16,6 +16,7 @@ def run(async: true)
         scroll_enumerator(cluster:).each_slice(PROCESSOR_BATCH_SIZE) do |documents|
           if async
             worker.perform_async(documents, @destination_index)
+            sleep 1
```
kevindew: I assume you don't want this merged eitehr.

spec/unit/analytics/ga4/elastic_search_relevancy_serialiser_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "spec_helper"
+require "analytics/ga4_import/elastic_search_relevancy_serialiser"
```
kevindew:
```suggestion
```
I would expect this is loaded by spec_helper

spec/unit/analytics/ga4/google_analytics_service_spec.rb
```diff
@@ -0,0 +1,108 @@
+require "spec_helper"
+require "analytics/ga4_import/data_fetcher"
```
kevindew:
```suggestion
```
as above loaded by spec_helper

spec/unit/analytics/ga4/page_data_spec.rb
```diff
@@ -0,0 +1,68 @@
+require "spec_helper"
+require "analytics/ga4_import/page_data"
```
kevindew:
```suggestion
```

spec/unit/analytics/ga4/page_view_consolidator_spec.rb
```diff
@@ -0,0 +1,108 @@
+require "spec_helper"
+require "analytics/ga4_import/page_view_consolidator"
+require "analytics/ga4_import/page_data"
```
kevindew:
```suggestion
```

spec/unit/analytics/ga4/relevance_report_generator_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "spec_helper"
+require "analytics/ga4_import/relevance_report_generator"
```
kevindew:
```suggestion
```

spec/unit/analytics/ga4/report_request_builder_spec.rb
```diff
@@ -0,0 +1,58 @@
+require "spec_helper"
+require "google/analytics/data/v1beta"
+require "analytics/ga4_import/report_request_builder"
```
kevindew:
```suggestion
```

spec/unit/analytics/ga4/google_analytics_service_spec.rb
kevindew, line 1: looks like this filename is out of date

spec/unit/analytics/ga4/elastic_search_relevancy_serialiser_spec.rb
kevindew, line 1: Looks like you need to rename the ga4 directory in specs

lib/analytics/ga4_import/report_request_builder.rb
```diff
@@ -0,0 +1,59 @@
...
+      def screen_page_views
+        # https://developers.google.com/analytics/devguides/reporting/data/v1/api-schema#metrics
+        Google::Analytics::Data::V1beta::Metric.new(
+          name: "screenPageViews",
+        )
+      end
+
+      def date_range
+        start_date = Date.today - REPORT_WINDOW
+        end_date = Date.yesterday
```
kevindew, line 50: Does this not mean that we're looking at 13 days of data not 14?

kevindew (commented):
Great progress sgeorges1996, 

I've run out of time on my review and not reached reviewing tests, however I thought I should submit at this point as I've clocked in a few comments already.

A lot of the things I've noticed are bits that got missed as things changed along the way. Would you be able to spend some time before I review this next by checking each of the tests for issues, could you in particular be checking that all the code paths are changed and that there aren't any superfluous things in there? that'd make it easier to review them.

lib/analytics/ga4_import/report_request_builder.rb
georges1996: Going off today its grabbing data across 14 days: **Mon, 19 Aug 2024 - Sun, 01 Sep 2024**

PR #2985, opened 2024-08-23
Improve "Using the Search API docs" with expanded examples

ChrisBAshton:
[Rendered version](https://github.com/alphagov/search-api/blob/improve-search-api-example-docs/docs/using-the-search-api.md) ✨ 

Having these examples handy would have made responding to a recent FOI request much simpler!

This commit:

- Frontloads examples to the top of the doc
- Adds several examples of different Search API queries
- Goes on to explain all of the various URL query parameters one might use in a search
- Consolidates information from the sister doc, "Faceted search", which was split out in https://github.com/alphagov/search-api/pull/946 but is actually depended on by this "Using the Search API docs" in order to explain basics like filtering by organisation. So I concluded it's best to merge the two into one doc and just have this doc get gradually more advanced as we go down the page. Note that this removes the link to https://alistapart.com/article/design-patterns-faceted-navigation that was introduced in the linked PR, but given this concept isn't really referenced anywhere else in the repo, its original inclusion seems a little optimistic (we don't really refer to "faceted navigation" as a design pattern, and trying to treat it as a separate piece of documentation was slowing us down).

Trello: https://trello.com/c/enlxCM9D/2795-foi-request-foi2024-10266


ryanb-gds (approved):
These are great Chris, thanks 🙌

PR #2987, opened 2024-09-02
Migrate GA4 logic to search-api 

georges1996:
Ticket: https://trello.com/c/SHc04ixR/110-search-analytics-pipeline-ua-to-ga4

The final output is the 'popularity' fields: https://www.gov.uk/api/search.json?fields=popularity_b,popularity
Testing PR: https://github.com/alphagov/search-api/pull/2964

As part of the migration to GA4, the task was to understand the existing Python application and its output https://github.com/alphagov/search-analytics, then to update the Ruby application https://github.com/alphagov/search-analytics-ga4/ in order to get an identical/very similar output to the Python app. 

Once this was correct the next part was to migrate the Search Analytics GA4 code from the Ruby repo to be part search-api repo (rationale: it’s easier to run the task if part of search-api - can be done on a pod in K8s, whereas running search-analytics from GitHub actions was disastrous for catching failures, it’ll be easier to maintain as a single codebase and not a separate thing for teams to look at) this decision will also bring down costs as we no longer need to maintain another code base and we don't need to have an S3 bucket anymore to dump the data to and for search-api to pick up, it can all be done as part of the `rake page_traffic:load` task

## Rollback Plan

If in the unlikely case this fails in production we have a way to roll back.  We revert this PR, and then it will revert back to using the data in the S3 bucket as this will be kept around until we are sure this change is working as expected. We will then have to rerun the cron jobs in Argo to reset the popularity back. 

- `search-api-load-page-traffic` which pulls the data from GA and ingests into Elastic Search
- `search-api-update-govuk-index-popularity` will use this new data from Elastic Search and update the search api popularity results here: https://www.gov.uk/api/search.json?fields=popularity_b,popularity


spec/unit/analytics/ga4_import/relevance_report_generator_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "spec_helper"
+
+RSpec.describe Analytics::Ga4Import::RelevanceReportGenerator do
+  let(:data_fetcher) { instance_double(Analytics::Ga4Import::DataFetcher) }
+  let(:page_view_consolidator) { instance_double(Analytics::Ga4Import::PageViewConsolidator) }
+  let(:relevancy_calculator) { instance_double(Analytics::Ga4Import::ElasticSearchRelevancySerialiser) }
+
+  before do
+    stub_classes
```
kevindew:
```suggestion
    allow(Analytics::Ga4Import::DataFetcher).to receive(:new).and_return(data_fetcher)
    allow(data_fetcher).to receive(:call).and_return(:paginated_data)

    allow(Analytics::Ga4Import::PageViewConsolidator).to receive(:new).and_return(page_view_consolidator)
    allow(page_view_consolidator).to receive(:consolidated_page_views).and_return(:consolidated_page_views)

    allow(Analytics::Ga4Import::ElasticSearchRelevancySerialiser).to receive(:new).and_return(relevancy_calculator)
    allow(relevancy_calculator).to receive(:relevance).and_return(:relevance)
```

It seems unnecessary to for these to be in a separate method

lib/analytics/ga4_import/page_view_consolidator.rb
```diff
@@ -0,0 +1,22 @@
+module Analytics
+  module Ga4Import
+    class PageViewConsolidator
+      def initialize(ga_data)
+        @ga_data = ga_data
+      end
+
+      def consolidated_page_views
+        @consolidated_page_views ||= ga_data
```
kevindew: It looks inconsistent with the other classes here to do a memo here, I assume there's no practical advantage as we only ever call this once?

spec/unit/analytics/ga4_import/page_view_consolidator_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      expect(normalised_data).to be_empty
+    end
+
+    it "returns a hash with the path and page views when the path starts with a slash and it is not a smart answer" do
+      page_data = [Analytics::Ga4Import::PageData.new("/example", "not a smart answer", "20")]
+      page_view_consolidator = Analytics::Ga4Import::PageViewConsolidator.new(page_data)
+      normalised_data = page_view_consolidator.consolidated_page_views
+
+      expect(normalised_data).to eq({ "/example" => 20 })
+    end
```
kevindew: These don't seem to be testing logic in the PageViewConsolidator class. I assume this is testing around the `#exclude` method? I imagine we just want a test that it removes ga_data that is excluded and only test the actual scenarios in the PageData class.

spec/unit/analytics/ga4_import/page_view_consolidator_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    it "consolidates page views for the same paths, ignoring query parameters" do
+      page_data = [
+        Analytics::Ga4Import::PageData.new("/example", "not a smart answer", "20"),
+        Analytics::Ga4Import::PageData.new("/example?something=something", "not a smart answer", "30"),
+      ]
+      page_view_consolidator = Analytics::Ga4Import::PageViewConsolidator.new(page_data)
+      normalised_data = page_view_consolidator.consolidated_page_views
+
+      expect(normalised_data).to eq({ "/example" => 50 })
+    end
```
kevindew, line 36: I don't think we need both of these tests on consolidation as it's drifting into the unit testing normalised_path which should be done elsewhere. I'd just have the one test here called "it consolidates the normalised paths of page views" and have this test logic with the query params.

spec/unit/analytics/ga4_import/page_view_consolidator_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    it "ignores pages with the title 'Page Not Found - 404 - GOV.UK'" do
+      page_data = [
+        Analytics::Ga4Import::PageData.new("/page-not-found", "Page not found - 404 - GOV.UK", "10"),
+        Analytics::Ga4Import::PageData.new("/example", "not a smart answer", "20"),
+      ]
+      page_view_consolidator = Analytics::Ga4Import::PageViewConsolidator.new(page_data)
+      normalised_data = page_view_consolidator.consolidated_page_views
+
+      expect(normalised_data).to eq({ "/example" => 20 })
+    end
```
kevindew, line 40: Is this also testing excluded?

spec/unit/analytics/ga4_import/page_view_consolidator_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+        Analytics::Ga4Import::PageData.new("/other", "other", "20"),
+        Analytics::Ga4Import::PageData.new("/example?something=something", "not a smart answer", "30"),
+        Analytics::Ga4Import::PageData.new("/example2", "not a smart answer", "10"),
+        Analytics::Ga4Import::PageData.new("/example2?something=something", "not a smart answer", "30"),
+      ]
+      page_view_consolidator = Analytics::Ga4Import::PageViewConsolidator.new(page_data)
+      normalised_data = page_view_consolidator.consolidated_page_views
+
+      expect(normalised_data).to eq({ "/example" => 40, "/other" => 20, "/example2" => 40 })
+    end
```
kevindew: This seems to be testing the same thing as the test on normalised paths, would suggest combining into one test.

lib/analytics/ga4_import/page_view_consolidator.rb
```diff
@@ -0,0 +1,22 @@
...
+          .reject(&:excluded?)
+          .group_by(&:normalised_path)
+          .transform_values { |grouped_pages| grouped_pages.sum(&:page_views) }
+          .sort_by { |_, views| -views }
+          .to_h
```
kevindew: let's not return a hash here and have `ElasticSearchRelevancySerialiser` accept an array. We have to do weird stuff in the test for it anyway.

spec/unit/analytics/ga4_import/page_view_consolidator_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+        Analytics::Ga4Import::PageData.new("/example", "example", "20"),
+        Analytics::Ga4Import::PageData.new("/example-two", "example-two", "10"),
+        Analytics::Ga4Import::PageData.new("/example-three", "example-three", "30"),
+        Analytics::Ga4Import::PageData.new("/example-four", "example-four", "5"),
+      ]
+
+      page_view_consolidator = Analytics::Ga4Import::PageViewConsolidator.new(page_data)
+
+      # Converts consolidated_page_views to an array for the test as RSpec eq doesn't match based on hash order
+      expect(page_view_consolidator.consolidated_page_views.to_a).to eq([["/example-three", 30], ["/example", 20], ["/example-two", 10], ["/example-four", 5]])
```
kevindew: I don't understand why this is returning a hash and we need this work around. Have suggested we return an array.

lib/analytics/ga4_import/data_fetcher.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+      def call
+        # https://developers.google.com/analytics/devguides/reporting/data/v1/basics#navigate_long_reports
+        offset = 0
+        all_data = []
+
+        loop do
+          data = get_data(offset)
+          break if data.nil? || data[:rows].nil?
```
kevindew: Why do we need to match two cases here? 

I notice we lack tests for both of these and I wondered whether it'd save code being written to just handle this with one condition and not need to write tests for two scenarios (one I presume is not even likely?)

spec/unit/analytics/ga4_import/data_fetcher_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+          Google::Analytics::Data::V1beta::DimensionValue.new(value: "/"),
+          Google::Analytics::Data::V1beta::DimensionValue.new(value: "Welcome to GOV.UK"),
+        ],
+        metric_values: [
+          Google::Analytics::Data::V1beta::MetricValue.new(value: "171078"),
+        ],
+      ),
+      Google::Analytics::Data::V1beta::Row.new(
+        dimension_values: [
+          Google::Analytics::Data::V1beta::DimensionValue.new(value: "sign-in-universal-credit"),
```
kevindew: Minor: would we expect this to be returned without a slash prefix?

lib/analytics/ga4_import/elastic_search_relevancy_serialiser.rb
```diff
@@ -0,0 +1,54 @@
+module Analytics
+  module Ga4Import
+    class ElasticSearchRelevancySerialiser
+      def initialize(consolidated_data)
+        @consolidated_data = consolidated_data
+      end
+
+      def relevance
+        consolidated_data.flat_map.with_index(1) do |(base_path, page_views), index|
```
kevindew, line 9: FWIW with the previous comment on the need to convert to a hash in PageViewConsolidator it looks like this code would need no changes to use the array as this task essentially iterates it as an array anyway - which makes the `.to_h` seem more redundant.

I think only the code at the end would need changing to be `@total_page_views ||= consolidated_data.sum(&:last)`

spec/unit/analytics/ga4_import/elastic_search_relevancy_serialiser_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+        { "index" => { "_id" => "/example2", "_type" => "page-traffic" } },
+        { "path_components" => ["/example2"], "rank_14" => 2, "vc_14" => 20, "vf_14" => 0.07692307692307693 },
+        { "index" => { "_id" => "/example3", "_type" => "page-traffic" } },
+        { "path_components" => ["/example3"], "rank_14" => 3, "vc_14" => 100, "vf_14" => 0.38461538461538464 },
+        { "index" => { "_id" => "/example3/child", "_type" => "page-traffic" } },
+        { "path_components" => ["/example3", "/example3/child"], "rank_14" => 4, "vc_14" => 100, "vf_14" => 0.38461538461538464 },
+      ]
+
+      expect(relevancy_data).to eql(expected)
+    end
```
kevindew, line 33: There's a lot going on in this one test - which makes it hard to follow. I think it would be better broken up into a few assertions:

- It returns an array of JSON objects pairs of index and page data
- it specifies the page rank, views and proportion of overall views for a page
- it breaks up the path into an array of components for a page.

spec/unit/analytics/ga4_import/page_data_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    context "when the path is just a slash" do
+      let(:path) { "/" }
+
+      it "returns '/'" do
+        expect(page_data.normalised_path).to eq("/")
+      end
+    end
+  end
+end
```
kevindew, line 67: Looks good 👍

lib/analytics/ga4_import/report_request_builder.rb
```diff
@@ -0,0 +1,59 @@
...
+        @offset = offset
+        @limit = limit
+      end
+
+      def report_request
+        ::Google::Analytics::Data::V1beta::RunReportRequest.new({
+          property: "properties/330577055",
+          date_ranges: [
+            date_range,
+          ],
```
kevindew:
```suggestion
          date_ranges: [date_range],
```
Formatting nit

lib/analytics/ga4_import/report_request_builder.rb
```diff
@@ -0,0 +1,59 @@
...
+      def page_path
+        # https://developers.google.com/analytics/devguides/reporting/data/v1/api-schema#dimensions
+        Google::Analytics::Data::V1beta::Dimension.new(
+          name: "pagePath",
+        )
+      end
+
+      def page_title
+        # https://developers.google.com/analytics/devguides/reporting/data/v1/api-schema#dimensions
+
```
kevindew:
```suggestion
```
Formatting nit

spec/unit/analytics/ga4_import/report_request_builder_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      expect(analytics_data).to have_attributes(
+        property: "properties/330577055",
+        offset: 0,
+        limit: 100,
+        metric_aggregations: [],
+        order_bys: [],
+        currency_code: "",
+        keep_empty_rows: false,
+        return_property_quota: true,
```
kevindew: These attributes don't reference anything in our code so shouldn't be asserting against it.

kevindew (commented):
Thanks eorges1996 this is looking good - Although I've got a few comments they're all pretty minor so can't imagine they'll take you long to address.

I notice you've got no content in your commit messages, it'd be good to use those to explain what/why of this code as it's quite a big implementation.

spec/unit/analytics/ga4_import/elastic_search_relevancy_serialiser_spec.rb
```diff
@@ -0,0 +1,26 @@
+require "spec_helper"
+
+RSpec.describe Analytics::Ga4Import::ElasticSearchRelevancySerialiser do
+  describe "#relevance" do
+    let(:consolidated_data) { { "/example" => 40, "/example2" => 20, "/example3" => 100, "/example3/child" => 100 } }
+    let(:relevancy_calculator) { described_class.new(consolidated_data) }
+    let(:relevancy_data) { relevancy_calculator.relevance.map { |rel| JSON.parse(rel) } }
+
+    it "returns an array of JSON object pairs of index and page data" do
```
kevindew, line 5: It's a bit weird that the code that is under test (the act part of "Arrange, Act, Assert") is outside the test - which feels a bit of an RSpec smell.

I'd suggest you put the code in here:

```
relevancy_data = described_class.new(consolidated_data).relevance
parsed_data = relevancy_data.map { |d| JSON.parse(d) }
```

spec/unit/analytics/ga4_import/elastic_search_relevancy_serialiser_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      relevancy_data.each_slice(2) do |index_data, page_data|
+        expect(index_data).to include("index")
+        expect(page_data).to include("path_components", "rank_14", "vc_14", "vf_14")
```
kevindew: These seem a little light for matching the JSON. Would suggest something like:

```
expect(index_data).to match("index" => hash_including("_type" => "page-traffic", "_id" => /^\/example/))
expect(page_data).to match(hash_including("path_components", "rank_14", "vc_14", "vf_14"))
```

spec/unit/analytics/ga4_import/elastic_search_relevancy_serialiser_spec.rb
```diff
@@ -0,0 +1,26 @@
+require "spec_helper"
+
+RSpec.describe Analytics::Ga4Import::ElasticSearchRelevancySerialiser do
+  describe "#relevance" do
+    let(:consolidated_data) { { "/example" => 40, "/example2" => 20, "/example3" => 100, "/example3/child" => 100 } }
```
kevindew: This should be an array rather than a hash, though I suggest you set individual ones for the tests in question for easier readability.

spec/unit/analytics/ga4_import/elastic_search_relevancy_serialiser_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    it "returns an array of JSON object pairs of index and page data" do
+      relevancy_data.each_slice(2) do |index_data, page_data|
+        expect(index_data).to include("index")
+        expect(page_data).to include("path_components", "rank_14", "vc_14", "vf_14")
+      end
+    end
+
+    it "specifies the page rank, views and proportion of overall views for a page" do
+      example_page_data = relevancy_data.each_slice(2).map { |_, page_data| page_data }.find { |data| data["path_components"].include?("/example") }
+      expect(example_page_data).to include("rank_14" => 1, "vc_14" => 40, "vf_14" => 0.15384615384615385)
```
kevindew: This test would could be a bit easier to follow with tailored data going into it and avoiding the each_slice

```
consolidated_data = [["/example", 30], ["/other", 10]]
relevancy_data = described_class.new(consolidated_data).relevance
example_page_data = JSON.parse(relevancy_data.first)
expect(example_page_data).to include("rank_14" => 1, "vc_14" => 30, "vf_14" => 0.75)
```

spec/unit/analytics/ga4_import/elastic_search_relevancy_serialiser_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    it "breaks up the path into an array of components for a page" do
+      example_page_data = relevancy_data.each_slice(2).map { |_, page_data| page_data }.find { |data| data["path_components"].include?("/example3/child") }
+      expect(example_page_data).to include("path_components" => ["/example3", "/example3/child"])
```
kevindew: This one could also be an easier read with tailored data

spec/unit/analytics/ga4_import/page_view_consolidator_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  describe "#consolidated_page_views" do
+    it "consolidates page views by normalised path and sorts by views in descending order" do
+      expect(consolidator.consolidated_page_views).to eq([
+        ["/example2", 110],
+        ["/example4/blah", 100],
+        ["/example1", 10],
+        ["/example3", 10],
+        ["/example3/blah", 10],
+      ])
+    end
```
kevindew, line 44: We should have more than one assertion here as it's not clear looking at the input how they all fit together. 

I imagine the assertions you want here are:

it outputs a list of page views and paths
it orders the list by page views
it doesn't include page data that is excluded
it combines page views for the same path
it combines page views for page data that normalises to the same path

kevindew (commented):
Great almost there now. I think just a couple of the test files need tweaks before going. Apologies if my communication led you to think I requested only one test for `PageViewConsolidator` as I was surprised we went from the extreme of 10 tests to 1 and wondered if that was my fault.

spec/unit/analytics/ga4_import/elastic_search_relevancy_serialiser_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      consolidated_data = [["/example", 90], ["/other", 10]]
+      relevancy_data = described_class.new(consolidated_data).relevance
+
+      page_data = JSON.parse(relevancy_data.second)
+
+      expect(page_data).to include("rank_14" => 1, "vc_14" => 90, "vf_14" => 0.9)
+    end
+
+    it "breaks up the path into an array of components for a page" do
+      consolidated_data = [["/example", 40], ["/example2", 20], ["/example3", 100], ["/example3/child", 100]]
```
kevindew:
```suggestion
      consolidated_data = [["/example3/child", 100]]
```
If I understand right the previous 3 things don't impact this assertion?

kevindew (commented):
Thanks eorges1996 this looks fine to merge now. I just have a minor comment on a test that is not a blocker.

The only other thing is that the commit message about the pipeline components we should add some more information:

- Should reference that this is for GA4
- That this code is a port of the work from https://github.com/alphagov/search-analytics-ga4
- Why we've ported the code here: rationale: it’s easier to run the task if part of search-api - can be done on a pod in K8s, whereas running search-analytics from GitHub actions was disastrous for catching failures, it’ll be easier to maintain as a single codebase and not a separate thing for teams to look at

kevindew (approved):
Nice one

PR #2996, opened 2024-09-18
All content finder: update section names

csutter:
- Change topic section name to properly capitalised `Topic` (the current topic facet in Finder Frontend has this title hardcoded anyway so the name isn't used right now, but we want to fix that)
- Change content type section name to just "Type" to match new designs

> [!NOTE]
> This will need to be applied after merge:
> ```
> kubectl exec -it deployments/search-api -- env FINDER_CONFIG=all_content_finder.yml rake publishing_api:publish_finder
> ```

### DIFF
--- a/config/finders/all_content_finder.yml
+++ b/config/finders/all_content_finder.yml
@@ -23,7 +23,7 @@ details:
     keys:
     - level_one_taxon
     - level_two_taxon
-    name: topic
+    name: Topic
     preposition: about
     short_name: topic
     type: taxon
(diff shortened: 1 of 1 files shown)


leenagupte (approved):
This looks fine to me. Just to flag up though that this will change on /search/all for all users not just the people who can see the new design.

csutter:
> This looks fine to me. Just to flag up though that this will change on /search/all for all users not just the people who can see the new design.

Yes – worth mentioning for posterity we've agreed on that within the team 👍

PR #3006, opened 2024-09-25
Don't put a large list of the entire documents into Sidekiq arguments

peteglondon:
Trello: https://trello.com/c/Mz9dttPs/1953-fix-redis-memory-intensive-updater-processes-in-search-api

## The problem
Search API is using up all the Redis memory in integration. 
## The Cause
The main culprit is the PopularityWorker - this serialises a large batch of data to update, it includes the entire `_source` json as `document` - this make the payload very large.
It is also completely unnecessary as the item could just as easily be pulled from ElasticSearch.

All it's doing is updating a number of attributes related to popularity, these come from a different source index.

## The solution

This does the following:

1. removes the unused sync worker, updater and rake task
2. updates the updater to only pass a document_id and not the entire document to the PopularityWorker and SupertypeWorker
3. Updates the PopularityWorker and SupertypeWorker to look up the document from the index using the document_ids passed in
4. Uses that record to update the popularity/type fields in the db


georges1996:
This seems reasonable, I think a good way to check this is working as expected and that its still pulling in the data correctly, is to deploy to integration and then run the 4 rake task jobs to see the new output here: **integration.publishing.service.gov.uk/api/search.json?fields=popularity_b,popularity**

kevindew:
Nice work @peteglondon, I second suggestion that this is tried out in integration and that we compare memory usage of Redis before and after.

I am concerned though at only doing this for one of the 3 classes affected as I think we will not be able to achieve the goal of reducing the amount of memory needed if we have two other jobs that need huge amounts of Redis memory. We'll either have to keep the memory the same, so those jobs can run or accept that those jobs would fail.

I did have a quick go to see if it seemed a challenge to apply the same concepts to all 3 jobs and it seemed ok: https://github.com/alphagov/search-api/compare/main...dont-dump-whole-indicies. So I would suggest once you've done a test of this work to confirm it is low Redis memory usage that we proceed with that so we can reduce the pods memory.

davidgisbey:
So i ran the rake task on main just to check it would use a high % of memory on Redis as a sense check. It maxed out for 

```
SEARCH_INDEX='govuk' bundle exec rake search:update_popularity
```

at 55.1%

image



I than killed the jobs and deployed this branch and reran the rake task 

It's maxed out at 0.9% 

image

So that's a ~98% reduction in memory usage


This is the sidekiq queue length for the same time period 

image

It looks like the jobs are completing far quicker too

image

It completed all the jobs in ~20 mins

davidgisbey:
I've deployed the branch out with the new changes and ran [these](https://github.com/alphagov/govuk-helm-charts/blob/12af1c46e8d7987d14bc4d85e672479c0613e921/charts/app-config/values-integration.yaml#L2399) concurrently. 

It went to 1.01% usage of the 8gb 

so 8000/100 = 80mb usage. I think we should be fine to revert Sean's PR once this is reviewed and merged https://github.com/alphagov/govuk-helm-charts/pull/2570/files

kevindew:
Wow "So that's a ~98% reduction in memory usage" 🤩

lib/govuk_index/popularity_worker.rb
```diff
@@ -4,41 +4,52 @@ class PopularityWorker < Indexer::BaseWorker
...
     sidekiq_options queue: QUEUE_NAME
 
-    def perform(records, destination_index)
-      actions = Index::ElasticsearchProcessor.new(client: GovukIndex::Client.new(timeout: BULK_INDEX_TIMEOUT, index_name: destination_index))
+    def perform(document_ids, index_name)
+      actions = Index::ElasticsearchProcessor.new(
+        client: GovukIndex::Client.new(timeout: BULK_INDEX_TIMEOUT, index_name:),
+      )
+
+      source_index = IndexFinder.by_name(index_name)
```
kevindew:
```suggestion
      index = IndexFinder.by_name(index_name)
```

As we no longer have a concept of source and destination

lib/govuk_index/supertype_worker.rb
```diff
@@ -4,26 +4,40 @@ class SupertypeWorker < Indexer::BaseWorker
...
     sidekiq_options queue: QUEUE_NAME
 
-    def perform(records, destination_index)
-      actions = Index::ElasticsearchProcessor.new(client: GovukIndex::Client.new(timeout: BULK_INDEX_TIMEOUT, index_name: destination_index))
+    def perform(document_ids, index_name)
+      actions = Index::ElasticsearchProcessor.new(client: GovukIndex::Client.new(timeout: BULK_INDEX_TIMEOUT, index_name:))
 
-      updated_records = records.reject do |record|
-        record["document"] == update_document_supertypes(record["document"])
+      source_index = IndexFinder.by_name(index_name)
```
kevindew: same point on source no longer having a meaning.

lib/govuk_index/popularity_worker.rb
```diff
@@ -4,41 +4,52 @@ class PopularityWorker < Indexer::BaseWorker
...
+
+      popularities = retrieve_popularities_for(index_name, document_ids)
+      document_ids.each do |document_id|
+        document = source_index.get_document_by_id(document_id)
+        unless document
+          puts "Skipping #{document_id} as it is not in the index"
```
kevindew: I think I probably made a bit of a janky suggestion of mine to use puts in the workers. I think instead we use `logger.warn`

lib/govuk_index/supertype_worker.rb
```diff
@@ -4,26 +4,40 @@ class SupertypeWorker < Indexer::BaseWorker
...
+      documents = document_ids.filter_map do |document_id|
+        document = source_index.get_document_by_id(document_id)
+
+        unless document
+          puts "Skipping #{document_id} as it is not in the index"
```
kevindew: similar point about logger

spec/unit/govuk_index/popularity_worker_spec.rb
```diff
@@ -2,45 +2,73 @@
...
 
   def stub_popularity_data(data = Hash.new({ popularity_score: 0.5, popularity_rank: 0.5 }))
     popularity_lookup = instance_double("popularity_lookup", lookup_popularities: data)
     allow(Indexer::PopularityLookup).to receive(:new).and_return(popularity_lookup)
   end
+
+  def stub_document_lookups(documents)
+    documents.each do |document|
+      allow(index).to receive(:get_document_by_id).with(document["_id"]).and_return(document)
+    end
```
kevindew, line 75: Minor nit as I'm not sure of practical implications.

Rather than setting up tons of individual stubs we could set up a single one:

```
allow(index).to receive(:get_document_by_id) { |document_id| documents[document_id] }
```

kevindew (commented):
Nice one, looks great - amazing work on the memory reduction. Just a few minor comments here.

davidgisbey:
Thanks for the review @kevindew. I think this ones ready for re-review

kevindew (approved):
Nice one, thanks so much for finishing this off.

PR #3025, opened 2024-10-22
All content finder: Change `date` facet short_name

csutter:
This is used to display the facet value as part of metadata, and when used in that context, "Updated: ___" sounds better than "Date: ___".

image

---

image


kevindew (approved):
Makes sense 👍

PR #3051, opened 2024-11-18
Remove deprecation marker on expanded search result field configuration and explain usage

ryanb-gds:
Given that we don't know why this behaviour was marked deprecated, and downstream apps depend on it, it seems reasonable to remove the deprecation notice.

This commit also adds a usage example from Finder Frontend to make it clear what purposes the configuration can be used for. Arguably finder frontend could look up the label from the finder content item instead of requiring the label to be part of the search result, which might explain why this was deprecated, but that is unlikely to be a simple change.

### DIFF
--- a/docs/schemas.md
+++ b/docs/schemas.md
@@ -86,8 +86,7 @@ The files contain JSON object with the following keys:
  - `fields`: An array of field names which are allowed in documents of this
    type.  The field names must be defined in the `field_definitions.json` file.
 
- - `expanded_search_result_fields`: DEPRECATED. Sets up field "expansion" for a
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Seems reasonable. Even if we were to assume it was deprecated with a plan it's been 8 years now and the future of this app is unclear so I'd be astonished if things would be moving on it.

Looking at the commit that originally flagged it as deprecated: https://github.com/alphagov/search-api/pull/580 I'm not sure they even meant to deprecate it, I wonder if they meant to deprecate the previous field (allowed_values) and then just decided to go with a straight rename.

PR #3055, opened 2024-11-19
Fix broken GovukIndex::PageTrafficJob and flaky tests

kevindew:
This change started out as code to fix the flaky test we've been seeing recently [example](https://github.com/alphagov/search-api/actions/runs/11909087045/job/33185768393) and then I discovered that the underlying code related to that flaky test was also broken in production so I've gone ahead and fixed that.

There's more information in the commit messages.

### DIFF
--- a/spec/integration/indexer/amend_job_spec.rb
+++ b/spec/integration/indexer/amend_job_spec.rb
@@ -9,10 +9,6 @@ RSpec.describe Indexer::AmendJob do
   let(:updates) { { "title" => "New title" } }
   let(:cluster_count) { Clusters.count }
 
-  before do
-    Sidekiq::Job.clear_all
-  end
-
   it "amends documents" do
     commit_document(index_name, document)
 
@@ -41,6 +37,8 @@ RSpec.describe Indexer::AmendJob do
         job.perform(index_name, link, updates)
       }.to change { described_class.jobs.count }.by(1)
     end
+    # clear the side effects of Sidekiq::Testing.fake!
+    Sidekiq::Job.clear_all
   end
 
   it "forwards to failure queue" do
--- a/spec/unit/govuk_index/page_traffic_job_spec.rb
+++ b/spec/unit/govuk_index/page_traffic_job_spec.rb
@@ -1,0 +1,92 @@
+require "spec_helper"
+
+RSpec.describe GovukIndex::PageTrafficJob do
+  describe ".wait_until_processed" do
+    before do
+      allow(described_class).to receive(:sleep)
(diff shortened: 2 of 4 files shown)


spec/unit/govuk_index/page_traffic_job_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      allow(described_class).to receive(:sleep).and_call_original
+
+      expect { described_class.wait_until_processed(max_timeout: 0.01) }
+        .to raise_error(Timeout::Error)
+    end
+  end
+
+  describe ".active_jobs?" do
+    let(:key) { SecureRandom.base64(16) }
+    let(:tid) { rand(1000..2000) }
```
jackbot: Wonder if it's worth naming this `thread_id`? I know it's irrelevant to the tests but I only knew that after looking at the Sidekiq source to see what the 2nd param to `Sidekiq::Work.new` was.

spec/integration/indexer/amend_job_spec.rb
```diff
@@ -41,6 +37,8 @@
         job.perform(index_name, link, updates)
       }.to change { described_class.jobs.count }.by(1)
     end
+    # clear the side effects of Sidekiq::Testing.fake!
+    Sidekiq::Job.clear_all
```
jackbot, line 41: It's not clear to me why moving this from a `before` block to here in the test would fix the issue. 

The commit message says:

> This is caused by Sidekiq::Testing.fake! which will leave information in the Sidekiq::Job.jobs array

How come this state isn't cleared in the `before` block, but would be cleared here?

spec/unit/govuk_index/page_traffic_job_spec.rb
kevindew: Good shout, will make that amend.

spec/integration/indexer/amend_job_spec.rb
kevindew: Sure, I can update the commit message on this to explain better.

A more off the cuff explanation is:

A test with Sidekiq::Testing.fake! will leave state on the Sidekiq global object with the information of the items that were queued.

If Sidekiq::Job.clear_all is run before each of the tests in this file it will mean all the tests start with empty Sidekiq state, but still leave state behind for the next test that runs after this file has completed. That could be solved by every test in the whole application suite clearing the state first in a global before block, but one only in a single file isn't sufficient.

jackbot: Got it, makes sense. Thanks for explaining :+1:

jackbot (approved):
Nice one

PR #3057, opened 2024-11-19
Fix test outputting to stdout

kevindew:
Prior to this change running the test suite would produce output to stdout of:

```
2024-11-19T22:01:11.972Z pid=1 tid=8ud  WARN: Skipping {"identifier"=>{"_id"=>"/test/1873d740-a4e8-4960-8a22-8810741847ae", "_version"=>1}, "document"=>{"link"=>"/test/1873d740-a4e8-4960-8a22-8810741847ae", "popularity"=>0.222}} as it is not in the index
```

(could also be stderr not stdout)

Looking into it it seems that test that was producing this output was out of date with the code that exists and not actually asserting on behaviour.

Therefore I've refactored the test to be based on the skipping behaviour.


davidgisbey (approved):
looks good 👍

PR #3059, opened 2024-11-20
Add `host_authorization` setting for dev environment

csutter:
Sinatra 4.1 ships with a breaking change that [enables host allowlisting middleware][1] in local development, which in turn breaks running Search API through GOV.UK Docker.
    
This adds an explicit `permitted_hosts` setting for the middleware in the development environment, which allows access with arbitrary host headers.

[1]: https://github.com/sinatra/sinatra/pull/2053

### DIFF
--- a/lib/rummager/config.rb
+++ b/lib/rummager/config.rb
@@ -3,6 +3,7 @@ require "rummager"
 
 configure :development do
   set :protection, false
+  set :host_authorization, { permitted_hosts: [] }
 end
 
 # Enable custom error handling (eg ``error Exception do;...end``)


lib/rummager/config.rb
```diff
@@ -3,6 +3,16 @@
...
 configure :development do
   set :protection, false
+  set :host_authorization, {
+    permitted_hosts: [
+      "search-api.dev.gov.uk",
+      "localhost",
+      ".localhost",
+      ".test",
+      IPAddr.new("0.0.0.0/0"),
+      IPAddr.new("::/0"),
```
csutter: (Note that these are the default hosts included in Sinatra's standard configuration)

kevindew:
```suggestion
      ".dev.gov.uk",
      "localhost",
      IPAddr.new("0.0.0.0/0"),
      IPAddr.new("::/0"),      
```

I think we can shorten this a bit as it'd be very unexpected for someone to be using a .localhost or .test domain locally given we have the .dev.gov.uk convention.

Do also wonder if this is worth anything in a dev environment and we might want to set it as an empty array to disable it? but maybe I'm missing something? If I understand right the IP addresses will match basically any IPv4 or IPv6 addresses, so it's very broad.

kevindew (commented):
Thanks for picking this up - does look an annoying change. I've made a suggestion but not fully convinced we benefit from enabling the protection - I'm a bit confused why Sinatra applied it to dev when mostly an issue with reverse proxies.

kevindew (approved):
Nice one

PR #3063, opened 2024-11-21
Bump Ruby to 3.2.6

davidgisbey:
## Description 

We've had to revert out Ruby bump to 3.3.6 due to an issue with the load traffic job. We were getting the following [error](https://argo.eks.integration.govuk.digital/applications/search-api?orphaned=false&resource=&node=batch%2FJob%2Fapps%2Fsearch-api-load-page-traffic-28869000%2F0&tab=logs):

```
Google::Cloud::CanceledError: 1:CANCELLED. debug_error_string:
{
  UNKNOWN:Error received from peer {
    created_time:"2024-11-20T22:00:03.584412449+00:00",
    grpc_status:1,
    grpc_message:"CANCELLED"
  }
} (Google::Cloud::CanceledError)
/usr/local/bundle/ruby/3.3.0/gems/google-analytics-data-v1beta-0.13.1/lib/google/analytics/data/v1beta/analytics_data/client.rb:369:
```

Looking at the gems repo we can see that someone else has opened an issue for the same error [here](https://github.com/googleapis/google-cloud-ruby/issues/27448)

govuk-ruby-base (https://github.com/alphagov/govuk-ruby-images/blob/93c59fe31019cc0621f607ac6bbb0683b111e1b6/README.md?plain=1#L8) only takes minor version of Ruby so we can't bump to say 3.3.4 as the production apps will use to the latest stable Ruby version (3.3.6) which will break the rake task.

This bumps Ruby to the latest 3.2 version which is 3.2.6. Once the issue has been resolved we can then bump this to 3.3.6 (or a later version if one is released).

## Testing

I've deployed this out to to integration and manually kicked off the load traffic rake task to check it's working and then deleted the job.

## Trello card

https://trello.com/c/xG4hrLyP/2169-investigate-fixing-the-search-api-loadpagetraffic-job


kevindew (approved):
Nice one, I think you're missing a word in the PR description "Looking **at**"

davidgisbey:
thanks good spot. Have fixed 👍

PR #3112, opened 2024-12-31
Add Sint Maarten

GDSNewt:
From the [Zendesk](https://govuk.zendesk.com/agent/tickets/6000442) ticket - "We require the addition of "Sint Maarten" to the list of destination countries available to select in the Export Health Certificates section of Specialist Publisher. We have a certificate that needs to be published, but we cannot create a page for it whilst the country does not exist in the system."

[Trello](https://trello.com/c/UuMKP5Si/3289-addition-of-country-to-export-health-certificates-metadata-govt-agency-general-issue)


ryanb-gds (approved):
🇳🇱

PR #3125, opened 2025-01-20
Remove expanded search result fields configuration

ryanb-gds:
Search result field expansion was used exclusively to enable human-friendly labels for metadata in Finder Frontend.

However, Finder Frontend already has access to the human friendly labels from the facet options for each field. Therefore it made sense to map the search result value to the facet option label in Finder Frontend and remove the expanded search result configuration from Search API. This enables us to reduce the amount of changes we need to make to Search API configuration when changing specialist finders.

Relevant finder frontend PR for context: https://github.com/alphagov/finder-frontend/pull/3627 

Trello: https://trello.com/c/kFdQhAGr


kevindew:
Sounds great to remove this information 👍 Does seem such a headache/dubious technical decision to hardcode all of these into Search API so fantastic work pulling them out.

We as the AI team aren't going to be in a great position to point out if these have any usages outside the specialist finders so we'll have to fall back to your testing to verify they're not a problem for non-specialist finders or API responses from Search API.

Only other thing is that this PR is going to be pretty hard to review with all of the config removals all in the same commit. What I think might be an easier way to do this (and potentially easier to revert) would be to have this PR remove the use of the config and then have a second PR that removes the config?

ryanb-gds:
> Only other thing is that this PR is going to be pretty hard to review with all of the config removals all in the same commit. What I think might be an easier way to do this (and potentially easier to revert) would be to have this PR remove the use of the config and then have a second PR that removes the config?

I did try that, but the config parser failed because of an unexpected key in the file.

kevindew:
> > Only other thing is that this PR is going to be pretty hard to review with all of the config removals all in the same commit. What I think might be an easier way to do this (and potentially easier to revert) would be to have this PR remove the use of the config and then have a second PR that removes the config?
> 
> I did try that, but the config parser failed because of an unexpected key in the file.

That's no worry, any chance you could break it out so the removal is a separate commit at the end so other commits can be reviewed distinctly?

lib/schema/elasticsearch_type.rb
```diff
@@ -73,29 +67,18 @@ def parse_file
       raise_error %(Duplicate entries in "fields")
     end
 
-    expanded_search_result_fields = raw.delete("expanded_search_result_fields") || {}
+    raw.delete("expanded_search_result_fields")
```
kevindew: Is this needed would we expect this key to still exist in the data?

lib/schema/combined_index_schema.rb
```diff
@@ -16,13 +16,9 @@ def elasticsearch_types
...
   # Get a hash from field_name to FieldDefinition object, for all fields
   # allowed in any document type in the indexes.
-  #
-  # Since FieldDefinitions can contain an "expanded_search_result_fields" member, and this
-  # may differ between document types, the definitions returned here will
-  # have combined "expanded_search_result_fields" fields.
   def field_definitions
     @field_definitions ||= each_field_with_object({}) do |field_name, field_definition, results|
-      results[field_name] = field_definition.merge(results[field_name])
+      results[field_name] = field_definition
```
kevindew, line 21: This strikes me as something that could have an issue - but I don't know enough about the code to be sure.

Are we expecting this to only yield once for a field_name and, if not, last field_definition wins?

Looking at the previous code it doesn't seem to try merge anything other than the expanded_search_results so this looks like it matches behaviour albeit leaving code that looks a little fishy

docs/schemas.md
```diff
@@ -81,34 +81,11 @@ Document types are defined in the `elasticsearch_types` directory, with the
 additional configuration for each type being defined by a JSON file in that
 directory.
 
-The files contain JSON object with the following keys:
+The files contain a JSON object with the following key:
```
kevindew, line 84: I guess now this file being a JSON object is probably overkill though it seems too big a can of worms to try change as part of this.

kevindew (commented):
Looks a great improvement to me, just a couple of comments on the code in the first commit to look at

lib/schema/elasticsearch_type.rb
ryanb-gds: I removed this in the second commit so that each commit had passing tests, without this line they fail. Continuous deployment habits dying hard 😅 I'll move it back to the first commit

*edit: changed third to second

lib/schema/combined_index_schema.rb
ryanb-gds: I think we're ok because the [parser checks when it loads the fields that they're all unique](https://github.com/alphagov/search-api/blob/939e30a114e904d952dd5c24e01537bd8659971d/lib/schema/elasticsearch_type.rb#L72)

Agree it's a slightly odd read

lib/schema/elasticsearch_type.rb
kevindew: Nice one

lib/schema/combined_index_schema.rb
kevindew: Great - thanks for checking

docs/schemas.md
ryanb-gds: I think so, though I must confess I haven't looked into it too closely.

kevindew (approved):
Nice one

PR #3142, opened 2025-02-04
Migrate Whitehall news article formats to GOVUK index

ryanb-gds:
Because news articles from Content Publisher were already indexed using Publishing API events, we don't seem to need to add any additional fields to the search presenters.

Trello: https://trello.com/c/A5uNzU3G


ChrisBAshton (approved):
Makes sense to me!

ryanb-gds:
Closing this as it needs to be done in two stages as per https://docs.publishing.service.gov.uk/repos/search-api/new-indexing-process.html

PR #3147, opened 2025-02-10
Apply format migration per-publishing app filter

ryanb-gds:
Whilst the migrated section of the migrated_formats.yaml config file did specify that news article formats could only be considered migrated if they were published by content publisher, this logic was not actually applied by the format migrator. The format migrator only applies publishing app restrictions for indexing.

This behaviour was a blocker to migrating news article formats published by Whitehall to the govuk index, because without applying the migrated filter to Whitehall-published documents, duplicate search results will appear in GOV.UK finders once we start allowing indexing of Whitehall news article formats in the GOVUK index. We need to be able to filter out the Whitehall documents indexed on the govuk index until we are confident that all the data has been migrated over safely.

There is an alternative, more straightforward approach that may be more appropriate here. We could add a "migrated_apps" key to the config file and simply say that any document with a publishing app of "content-publisher" in the govuk index must appear in the search results. This would have the advantage of not adding an Elasticsearch condition to evaluate per app/format combination, but isn't true to the original intention of the format migrator (which is intended to work on a format-by-format basis rather than for an entire app). 

### DIFF
--- a/config/govuk_index/migrated_formats.yaml
+++ b/config/govuk_index/migrated_formats.yaml
@@ -4,15 +4,6 @@ migrated:
 - mainstream_browse_page
 # Contacts
 - contact
-# Content Publisher - formats not indexable when published by Whitehall
-- government_response:
-    publishing_app: content-publisher
-- news_story:
-    publishing_app: content-publisher
-- press_release:
-    publishing_app: content-publisher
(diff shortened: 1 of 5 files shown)


ryanb-gds:
Closing this in  favour of #3148

PR #3148, opened 2025-02-10
Add migrated app configuration to Search API and always include results from migrated apps

ryanb-gds:
This allows us to safely migrate document formats to the govuk index in Whitehall. With the current configuration that isn't possible because the format migrator doesn't respect publishing app configuration within the migrated section of the migrated_formats.yaml file.

We have considered maintaining the apparent original intention of the file which was to migrate items on a format by format basis, but applying the format migration for specific publishing apps added substnatial complexity to the Elasticsearch filter which might have affected search performance.

Because we can now just declare an app migrated, we no longer need the configuration for restricting indexable formats based on their publishing app either, so that code has been removed.

The Format Migrator spec unnecessarily tested the default query behaviour for each clause in the Elasticsearch query. Repeating that test another time for the migrated apps clause didn't seem sensible, so we now only test the default query behaviour once.

### DIFF
--- a/config/govuk_index/migrated_formats.yaml
+++ b/config/govuk_index/migrated_formats.yaml
@@ -4,15 +4,6 @@ migrated:
 - mainstream_browse_page
 # Contacts
 - contact
-# Content Publisher - formats not indexable when published by Whitehall
-- government_response:
-    publishing_app: content-publisher
-- news_story:
-    publishing_app: content-publisher
-- press_release:
-    publishing_app: content-publisher
-- world_news_story:
-    publishing_app: content-publisher
 # Publisher
 - answer
 - guide
@@ -315,7 +306,15 @@ migrated:
   - '/help/cookies'
   - '/find-local-council'
 
-indexable: []
+# Below configuration can be removed once government_response, news_story, press_release and world_news_story are migrated
+migrated_publishing_apps:
+  - content-publisher
+
+indexable:
+  - government_response
+  - news_story
+  - press_release
+  - world_news_story
 
 non_indexable_path:
 - '/help/cookie-details'
@@ -388,8 +387,6 @@ non_indexable:
 #- finder # migrated
 - foi_release
 - form
-- government_response: # indexable when published by Content Publisher
-    publishing_app: whitehall
 - guidance
 - html_publication
 - impact_assessment
@@ -402,8 +399,6 @@ non_indexable:
 - modern_slavery_statement
 - national_statistics
 - national_statistics_announcement
-- news_story: # indexable when published by Content Publisher
-    publishing_app: whitehall
 - notice
 - official_statistics
 - official_statistics_announcement
@@ -415,8 +410,6 @@ non_indexable:
 - petitions_and_campaigns
 - policy_area
 - policy_paper
-- press_release: # indexable when published by Content Publisher
-    publishing_app: whitehall
 - procurement
 - promotional
 - publication
@@ -440,7 +433,5 @@ non_indexable:
 - welsh_language_scheme
 - working_group
 - world_location
-- world_news_story: # indexable when published by Content Publisher
-    publishing_app: whitehall
 - worldwide_organisation
 - written_statement
--- a/lib/govuk_index/migrated_formats.rb
+++ b/lib/govuk_index/migrated_formats.rb
@@ -2,10 +2,9 @@ module GovukIndex
   module MigratedFormats
     extend self
 
-    def non_indexable?(format, path, app)
-      non_indexable_formats[format] &&
-        (non_indexable_formats[format] == :all || non_indexable_formats[format].include?(path) ||
-          published_by_non_indexable_app?(format, app)) || non_indexable_path.include?(path)
+    def non_indexable?(format, path)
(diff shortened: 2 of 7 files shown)


config/govuk_index/migrated_formats.yaml
```diff
@@ -315,7 +306,14 @@ migrated:
   - '/help/cookies'
   - '/find-local-council'
 
-indexable: []
+migrated_publishing_apps:
```
kevindew, line 310: Shall we add a comment above this to say this functionality can be removed once government_response, news_story, press_release and world_news_story are migrated ?

lib/govuk_index/migrated_formats.rb
```diff
@@ -29,6 +27,10 @@ def migrated_formats
       @migrated_formats ||= convert_to_allowed_hash(data_file["migrated"])
     end
 
+    def migrated_publishing_apps
+      @migrated_publishing_apps ||= data_file["migrated_publishing_apps"] || []
```
kevindew, line 31: This seems quite error tolerant for when the field doesn't exist - is there an advantage to this (perhaps ease of testing)? If not I wonder if we want to take the opposite approach and have this as: `@migrated_publishing_apps ||= data_file.fetch("migrated_publishing_apps")` to ensure someone deletes this code if they remove this entry to discourage leaving behind mess?

lib/govuk_index/publishing_event_job.rb
```diff
@@ -85,9 +85,9 @@ def process_action(processor, routing_key, payload)
       if type_mapper.unpublishing_type?
         logger.info("#{routing_key} -> DELETE #{identifier}")
         processor.delete(presenter)
-      elsif payload.fetch("locale", "en") != "en" || MigratedFormats.non_indexable?(presenter.format, presenter.base_path, presenter.publishing_app)
+      elsif payload.fetch("locale", "en") != "en" || MigratedFormats.non_indexable?(presenter.format, presenter.base_path)
         logger.info("#{routing_key} -> BLOCKLISTED #{identifier}")
```
kevindew, line 89: aside: blocklisted feels such a strange term for this context 😆

lib/search/format_migrator.rb
```diff
@@ -9,7 +9,7 @@ def call
       {
         bool: {
           minimum_should_match: 1,
-          should: [excluding_formats, only_formats],
+          should: [excluding_formats, only_formats, migrated_publishing_apps].compact,
```
kevindew: I find this part a bit confusing and wonder if we could have a logic bug here.

As I understand this minimum_should_match: 1 is perhaps a complicated way of dealing with runtime of migrated_formats being potentially empty or not (I don't fully understand the code).

So it strikes me that this could mean that an earlier condition in this overrides your migrated_publishing_apps condition.

As I can't actually tell if your code has a bug (admittedly from just a skim) I wonder if we might want to refactor this a bit, as it feels like what this code is expressing is relatively simple but perhaps the names are confusing?

kevindew: I guess what we're asking it for is:

give me formats that are only on legacy index
give me formats that are on migrated indexes
give me formats that are on legacy index but published by a migrated app

is that right?

kevindew (commented):
This looks both a nice bugfix and a nice simplification @ryanb-gds - just a few suggestions and questions

lib/govuk_index/migrated_formats.rb
ryanb-gds: To my mind the advantage is that consuming code doesn't have to check for nil, it can just iterate on the result of the method and if the array is empty then it's just a no-op.

I'm going to come back and remove this anyway as soon as I've migrated the news articles.

lib/search/format_migrator.rb
ryanb-gds: That's almost right. It's actually 

give me documents with unmigrated formats that are only on legacy index
give me documents with migrated formats that are on migrated indexes 
give me documents that are on migrated indexes and published by a migrated app irrespective of format

I've renamed the methods and added some comments to hopefully make it a bit clearer what is going on, let me know if it helps at all.

The migrated apps function should purely expand the range of results that the filter will accept. Previously if a document had a format was indexable and not migrated it would be excluded. Now it will be included regardless of the fact that the format is not migrated, if it was published by a migrated app. So I don't think the app condition can be overridden as such.

It's really just a hack to get around this problem of Content Publisher formats already being effectively migrated, and I will remove that condition once Whitehall news articles are migrated

lib/search/format_migrator.rb
```diff
@@ -9,7 +9,9 @@ def call
       {
         bool: {
           minimum_should_match: 1,
-          should: [excluding_formats, only_formats],
+          # match documents meeting any of below conditions
+          # this query excludes documents that are present in the migrated index, but their format is not yet marked as migrated
+          should: [be_an_unmigrated_document, be_a_migrated_document, be_published_by_a_migrated_app].compact,
```
kevindew, line 14: Nice refactoring here 👍 much clearer

kevindew (approved):
Nice one, looks good 👍

PR #3155, opened 2025-02-18
Mark news article formats as migrated

ryanb-gds:
Now that these formats are marked as migrated, we no longer require the override that forced all documents published by Content Publisher to be migrated. This was necessary because the Whitehall versions of those document formats were indexable but not migrated.


ChrisBAshton (approved):
Nice one!

PR #3176, opened 2025-02-26
Use the bulk Sidekiq queue for bulk reindexing jobs

ryanb-gds:
We have a separate "bulk" sidekiq queue for running bulk reindexing loads so that publisher-driven indexing will not be blocked in production.

However, the GOVUK index publishing event processor was not respecting the routing key provided in the RabbitMQ queue message, and therefore the default Sidekiq queue was being blocked by reindexing operations.

This commit fixes the issue at the cost of an extra loop through the message batch to partition the messages into bulk and non-bulk messages.

This will help me to not need to run indexing jobs for [this ticket](https://trello.com/c/upZnqIOL) out of hours. Some publication types need 20 minutes or so to clear the queue.


ChrisBAshton (approved):
Nice one! 🎉

PR #3181, opened 2025-03-05
Use maximum number of threads available for Sidekiq

ryanb-gds:
We're currently running 8 replicas of the worker pods and Sidekiq is configured to run 12 threads per process, so theoretically we have 84 threads available.

Elasticsearch's indexing rate is looking healthy with the current maximum of 48 threads, so we might as well use the extra threads we have available.


kevindew (approved):
🤞

PR #3182, opened 2025-03-05
Reduce Sidekiq thread count

ryanb-gds:
Elasticsearch couldn't cope with the increased thread count, so dropping this down again.


ryanb-gds:
Closing this as no longer required

PR #3183, opened 2025-03-05
Reduce Elasticsearch refresh rate

ryanb-gds:
The refresh rate setting in Elasticsearch controls how often Elasticsearch makes data that it has indexed available to search. This is quite a resource intensive process, so performing it less frequently can help improve indexing throughput.

### DIFF
--- a/config/schema/elasticsearch_schema.yml
+++ b/config/schema/elasticsearch_schema.yml
@@ -121,7 +121,7 @@ index:
     # will still work fine, though.
     number_of_replicas: 2
     number_of_shards: 3
-    refresh_interval: '1s'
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Thanks for this. 

Great that it improves indexing throughput, I guess it comes at the cost of latency in seeing that data that has been written is available. I'm assuming that's not much of a problem as we anticipate up to a 5 min delay on content changes anyway across GOV.UK due to CDN caching but it wouldn't surprise me if there is some scenario we don't know about that makes this either a user facing or internal system issue.

Can you keep alert for if you hear anything?

PR #3184, opened 2025-03-05
Increase Sidekiq throughput again

ryanb-gds:
Sidekiq workers can be effective with more than 12 threads, so let's see if we can squeeze a little bit more throughput from our 8 processes.

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -4,7 +4,7 @@ tag: search-api.publishing.service.gov.uk
 staging:
   :concurrency:  12
 production:
-  :concurrency:  12
+  :concurrency:  18
 :require: ./lib/rummager.rb
 <% if ENV.key?('SIDEKIQ_LOGFILE') %>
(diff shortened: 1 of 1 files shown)


config/sidekiq.yml
```diff
@@ -14,4 +14,4 @@ production:
   - bulk
 :limits:
   bulk: 4
-  default: 84
+  default: 144
```
ChrisBAshton, line 17: is it worth adding an inline comment explaining where this came from?

```suggestion
  default: 144 # = 8 workers multiplied by production `concurrency` rate 
```

ChrisBAshton (approved):

config/sidekiq.yml
ryanb-gds: It depends how many replicas we have as configured in the helm charts, a number I expect to drop once thi sis over, so probably not. I will add some docs after this is over to explain some of the configs though

PR #3185, opened 2025-03-05
Reset Sidekiq concurrency limits to sensible defaults

ryanb-gds:
We are going to revert the helm config to the original two replicas running Sidekiq, with 12 threads each. We now know we can safely use them all without overburdening Elasticsearch, so this offers 12 threads to the default queue and 12 to the bulk queue. There is an argument that we don't really need the limits at all, but it is probably worthwhile to reserve some threads for each queue. We should re-evaluate this choice at a later date.

One thing that helped speed up indexing was to reduce Elasticsearch's refresh interval to once every 30 seconds. Refreshing is the process by which Elasticsearch makes indexed data available for searching and it is moderately resource intensive. It doesn't make sense to run this every second and reducing the frequency helps to improve indexing throughput.

### DIFF
--- a/config/schema/elasticsearch_schema.yml
+++ b/config/schema/elasticsearch_schema.yml
@@ -121,7 +121,7 @@ index:
     # will still work fine, though.
     number_of_replicas: 2
     number_of_shards: 3
-    refresh_interval: '1s'
+    refresh_interval: '30s'
     search:
       slowlog:
         threshold:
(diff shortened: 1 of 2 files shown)


kevindew (commented):
Thanks for this. Yeah I don’t feel too comfortable with these limit values but can understand resetting them back.

Can you suggest reviewing them as an incident action?

ryanb-gds:
Closing this as the current high limit isn't causing any issues and we need to consider what the limits should really be, and test if possible.

PR #3191, opened 2025-03-06
Ensure official document status is included in the govuk index

ryanb-gds:
The document preparer was stripping the unnumbered paper data from the attachments before the elasticsearch presenter could use the information to derive the official document state. This commit makes sure that data is preserved.

This also includes a small refactor to make use of `Hash.slice` where appropriate rather than mapping attachment values across using object assigment.

Trello: https://trello.com/c/upZnqIOL

### DIFF
--- a/lib/govuk_index/presenters/common_fields_presenter.rb
+++ b/lib/govuk_index/presenters/common_fields_presenter.rb
@@ -87,14 +87,7 @@ module GovukIndex
 
     def attachments
       (payload.dig("details", "attachments") || []).map do |attachment|
-        {
-          "content" => attachment["content"],
-          "title" => attachment["title"],
-          "isbn" => attachment["isbn"],
-          "unique_reference" => attachment["unique_reference"],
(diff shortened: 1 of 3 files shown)


lib/govuk_index/presenters/common_fields_presenter.rb
```diff
@@ -87,14 +87,7 @@ def government_name
...
       (payload.dig("details", "attachments") || []).map do |attachment|
-        {
-          "content" => attachment["content"],
-          "title" => attachment["title"],
-          "isbn" => attachment["isbn"],
-          "unique_reference" => attachment["unique_reference"],
-          "command_paper_number" => attachment["command_paper_number"],
-          "hoc_paper_number" => attachment["hoc_paper_number"],
-        }.compact
+        attachment.slice("content", "title", "isbn", "unique_reference", "command_paper_number", "hoc_paper_number").compact
```
ChrisBAshton, line 90: Ooh, nice refactor 👍

ChrisBAshton (approved):

PR #3216, opened 2025-03-24
Disable Sidekiq processing on `search_api_govuk_index` queue

jackbot:
https://trello.com/c/U9M9Og06/2276

This removes the logic to enqueue Sidekiq jobs for messages pulled from the `search_api_govuk_index` queue and instead processes them synchronously.

It adds retrying logic to the consumer to retry a message 5 times, with a 60s delay between retries that is already configured on RabbitMQ (https://github.com/alphagov/govuk-infrastructure/pull/1848).

It also configures the consumer to use multiple threads for concurrent processing, as opposed to the serial processing it does now.

Rather than extract logic out of the `AmendJob` class, I decided to just call it synchronously instead. If other parts of the app didn't use that class to enqueue Sidekiq jobs, I'd have made it a non-Sidekiq class instead, but it feels like doing it this way is the path of least resistance.

See commit messages for more detail.

### DIFF
--- a/lib/indexer/message_processor.rb
+++ b/lib/indexer/message_processor.rb
@@ -9,35 +9,45 @@
 # the links from the publishing-api.
 module Indexer
   class MessageProcessor
-    def process(message)
+    MAX_RETRIES = 5
+
+    def initialize
+      @logger = Logging.logger[self]
+    end
+
+    def process(queue_message)
+      message = ::RetryableQueueMessage.new(queue_message)
+      payload = message.payload
+
       with_logging(message) do
         indexing_status = Indexer::ChangeNotificationProcessor.trigger(message.payload)
         Services.statsd_client.increment("message_queue.indexer.#{indexing_status}")
-        message.ack
+        message.done
       end
-    rescue ProcessingError => e
-      GovukError.notify(e, extra: message.payload)
-      message.discard
     rescue StandardError => e
-      # This is rescue of last resort. If anything goes wrong during the payload
-      # processing, we don't want to retry the message really quickly because
-      # that might overload elasticsearch or other components. This should be
-      # replaced by a retry mechanism with exponential back-off.
-      GovukError.notify(e, extra: message.payload)
-      sleep 1
-      message.retry
+      if message.retries < MAX_RETRIES
+        logger.error("#{payload['content_id']} scheduled for retry due to error: #{e.class} #{e.message}")
+
+        message.retry
+      else
+        logger.error("#{payload['content_id']} ignored after #{MAX_RETRIES} retries")
+        GovukError.notify(e, extra: payload)
+        message.done
+      end
     end
 
   private
 
+    attr_reader :logger
+
     def with_logging(message)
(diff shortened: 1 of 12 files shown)


lib/retryable_queue_message.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def done
+    queue_message.ack
+  end
+
+  def retry
+    # Some of our RabbitMQ exchanges are configured such that discarding a message
+    # pushes it to a queue to retry, so we have to a call an unintiuitvely named
+    # method totrigger a retry.
```
kevindew: I think we could probably a use a comment at the top of the class about usage as I think this comment potentially makes things a bit confusing when it says "Some of our RabbitMQ exchanges"

I imagine we want to be a bit more assertive, something like:

> This class should be used as a wrapper around `GovukMessageQueueConsumer::Message` whenever we use an exchange that uses a dead letter exchange (dlx) as a retry mechanism. Using a dlx in this way means sending counter intuitive responses to RabbitMQ. Where an ack is the approach to mark a job as completed AND to mark a job as failed, whereas discard actually signifies that we're going to retry the job.
> 
> You should only use this class if you're planning to retry via a dlx otherwise it'll be very confusing!

lib/indexer/message_processor.rb
```diff
@@ -9,35 +9,45 @@
 # the links from the publishing-api.
 module Indexer
   class MessageProcessor
-    def process(message)
+    MAX_RETRIES = 5
+
+    def initialize
+      @logger = Logging.logger[self]
```
kevindew, line 15: Nice one adding a logger. What a strange way the logger is accessed though 😖

spec/unit/indexer/message_processor_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        expect(logger).to receive(:info).with(message)
+      end
+
+      described_class.new.process(message)
+    end
+
+    it "increments the statsd counter" do
+      expect(Services.statsd_client).to receive(:increment).with("message_queue.indexer.accepted")
+      described_class.new.process(message)
+    end
```
kevindew, line 32: As I understand, this is actually a no-op now as we're not running statsd anymore. Though I think we may be best to just leave this in as it'd feel inconsistent to chop bits out and there's quite a lot of statsd in the codebase.

kevindew (commented):
Nice one. Just a couple of minor comments, but this looks great and I'm very much hoping we don't hit issues when we ship it.

lib/indexer/message_processor.rb
jackbot: Yeah weird isn't it!

spec/unit/indexer/message_processor_spec.rb
jackbot: Good to know. I'll leave it in place as per your suggestion.

lib/retryable_queue_message.rb
jackbot: Nice one, thanks for taking the time to write out a more informative comment. I've updated it in the original commit.

kevindew (approved):
Nice one

PR #3228, opened 2025-04-01
Remove Sidekiq in remaining RabbitMQ consumer

jackbot:
This continues the work started in #3216 to fully remove Sidekiq from all the RabbitMQ consumers.

It involved a bit more refactoring than the previous PR, due to this consumer being built to handle multiple RabbitMQ messages at once, which (I think) is not actually a scenario which can ever happen.


lib/tasks/message_queue.rake
```diff
@@ -8,23 +8,35 @@ namespace :message_queue do
...
-
-    channel.queue("search_api_bulk_reindex").bind(exch, routing_key: "*.bulk.reindex")
-    channel.queue("search_api_govuk_index").bind(exch, routing_key: "*.*")
+    queues = [
+      { name: "search_api_to_be_indexed", routing_key: "*.links" },
+      { name: "search_api_bulk_reindex", routing_key: "*.bulk.reindex" },
+      { name: "search_api_govuk_index", routing_key: "*.*" },
+    ]
+
+    queues.each do |queue|
```
kevindew, line 17: yay for predictable naming

spec/support/retryable_queue_examples.rb
```diff
@@ -0,0 +1,41 @@
+RSpec.shared_examples "a retryable queue processor" do
```
kevindew, line 1: nice

lib/govuk_index/publishing_event_job.rb
```diff
@@ -40,8 +40,10 @@ class MissingExternalUrl < StandardError; end
       worldwide_office_staff_role
     ].freeze
```
kevindew: Do you happen to know if anything terrible will happen if all this stuff moves inside the class definition? As they're in the wrong place in the namespace. Though I appreciate fixing this could break something else.

lib/govuk_index/publishing_event_message_processor.rb
```diff
@@ -40,12 +40,12 @@ class MissingExternalUrl < StandardError; end
       worldwide_office_staff_role
     ].freeze
 
-  class PublishingEventJob
+  class PublishingEventMessageProcessor
```
kevindew: I wonder if Handler rather than Processor might be a good synonym since we already have PublishingEventProcessor calling this class. So that's a lot of "processor" going on - don't worry about this if you think it should be called processor because of the idiom of `Index::ElasticsearchProcessor`

lib/govuk_index/publishing_event_message_processor.rb
```diff
@@ -45,6 +45,10 @@ def initialize
       @logger = Logging.logger[self]
     end
 
+    def self.process(...)
+      new.process(...)
+    end
```
kevindew: I think to be a bit more idiomatic you should pass the arguments to new and then have messages as an ivar like logger.

spec/integration/govuk_index/publishing_event_message_processor_spec.rb
```diff
@@ -24,7 +22,7 @@
 
       expect(@statsd_client).to receive(:increment).with("govuk_index.elasticsearch.index")
 
-      subject.process([["routing.key", payload]])
+      described_class.process([["routing.key", payload]])
```
kevindew: Nice improvement to the specs.

spec/unit/tasks/message_queue_spec.rb
```diff
@@ -26,6 +26,52 @@
...
+          queue_name: "search_api_bulk_reindex",
+          processor:,
+          logger:,
+          worker_threads: 10,
+          prefetch: 10,
+        ).and_return(consumer)
+
+      Rake::Task["message_queue:bulk_insert_data_into_govuk"].invoke
+    end
+  end
```
kevindew, line 85: Nice one adding these specs.

We might want to add a `before { Rake::Task[task_name].reenable }` before them as they currently leave a bit of a trap if anyone was to add any specs for this task as it'd not execute a second time without it.

spec/unit/govuk_index/publishing_event_processor_spec.rb
```diff
@@ -2,57 +2,47 @@
...
+  context "when an error is raised" do
+    before do
+      expect(GovukIndex::PublishingEventMessageProcessor)
+        .to receive(:process)
+        .with("routing.key", message.payload)
+        .and_raise("oh no")
+    end
 
-    subject.process([message1, message2])
+    it_behaves_like "a retryable queue processor"
```
kevindew, line 41: Shall we move this to be situated next to `it_behaves_like "a message queue processor"` to indicate we intentionally want both?

kevindew (commented):
Looks good, added a few minor comments but didn't spot any logic issues.

spec/unit/govuk_index/publishing_event_processor_spec.rb
jackbot: Do you mean move the whole `context` block up to the top? We need the `before` block in this context so ensure the error is raised, so the two shared examples won't be able to be right next to each other.

Do you mean something like this?

```ruby
RSpec.describe GovukIndex::PublishingEventProcessor do
  ...

  it_behaves_like "a message queue processor"

  context "when an error is raised" do
    before do
      expect(GovukIndex::PublishingEventMessageHandler)
        .to receive(:process)
        .with("routing.key", message.payload)
        .and_raise("oh no")
    end

    it_behaves_like "a retryable queue processor"
  end

  context "when successful" do
    ...
  end
end
```

To me that doesn't read much better, as the two `it_behaves_like` are still reasonably far away from each other.

kevindew: Oh I hadn't clocked that this was within a context - yeah leave as is.

lib/govuk_index/publishing_event_job.rb
jackbot: Excellent spot, I'd not noticed that. I've added the change as a separate commit (9e07ded1b44e6db8579c02f41f50cf5947ad69d6)

lib/govuk_index/publishing_event_message_processor.rb
jackbot: Yeah I was always a bit iffy on having `PublishingEventProcessor` and `PublishingEventMessageProcessor`. `Handler` is better so I've updated the original commits.

lib/govuk_index/publishing_event_message_processor.rb
jackbot: Yep good shout, done.

spec/unit/tasks/message_queue_spec.rb
jackbot: Ah I didn't know about this little gotcha. Added it as a separate commit in fb00b626c8e6149ee89f72d897eb42fd139f71ec for clarity.

lib/govuk_index/publishing_event_message_handler.rb
```diff
@@ -15,57 +13,56 @@ class NotIdentifiable < StandardError; end
 
   class MissingExternalUrl < StandardError; end
```
kevindew: These classes should also be within the PublishingEventMessageHandler namespace (or if they can't be there should probably be in a lib/govuk_index.rb file)

lib/govuk_index/publishing_event_message_handler.rb
```diff
@@ -15,57 +13,56 @@ class NotIdentifiable < StandardError; end
...
 
-      messages.each do |routing_key, payload|
-        process_action(processor, routing_key, payload)
-      end
+    def self.process(...)
+      new(...).process
+    end
 
-      responses = processor.commit
+    def process
```
kevindew: Sorry I think a comment I made last time got lost (or I messed it up somehow).

I think we should rename these methods to be `call` rather than process to have a more generic entrypoint.

kevindew (commented):
Sorry just a couple of things

lib/govuk_index/publishing_event_message_handler.rb
jackbot: Addressed in a4ff5ce2bc3cb3eb977b3e0e886f011e12094b5d

lib/govuk_index/publishing_event_message_handler.rb
jackbot: No probs, I've updated the original commit.

kevindew (approved):
Nice one, sorry for the delay in finishing this. Exciting to see this go through.

PR #3234, opened 2025-04-07
Add ADR for removing Sidekiq in RabbitMQ consumers

jackbot:


docs/arch/adr-013-rabbitmq-sidekiq-removal.md
```diff
@@ -0,0 +1,33 @@
+# Decision record: Removing Sidekiq from RabbitMQ queue consumers
+
+**Date:** 2025-04-07
+
+This supersedes [ADR-001][].
```
kevindew, line 5: Thanks for flagging this, can we add a note in ADR-001 that it was superseded.

docs/arch/adr-013-rabbitmq-sidekiq-removal.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Context
+After investigation following a recent incident where Redis was running out of memory, it was determined that the problem lay in the queue consumer code in this codebase.
+
+When consuming a job off the RabbitMQ queue, we immediately push the payload into a Sidekiq job for asynchronous processing, essentially moving the processing from one queue to another. This poses a particular problem when bulk reindexing of documents takes place, as thousands of jobs are pushed to Sidekiq (and therefore Redis).
```
kevindew: I think this rationale isn't quite correct - I think it might be easier to suggest a rephrase than try articulate why though.

In [ADR-001] a decision was recorded that queue messages from RabbitMQ would be then pushed onto a Sidekiq queue for handling. We have decided to revisit this decision in light of a number of problems:

1. The data in an individual queue message is often large and bulk pushing of data by Publishing API is relatively common; this had made it hard to establish the appropriate size for Redis as it needs substantial memory to handle a bulk republishing. This has contributed to two incidents.
2. At the time of the initial decision the message queue was processed by a single thread so had very low concurrency. There are now concurrency features provided by govuk_message_queue_consumer so this can be scaled in a similar way to Sidekiq.
3. A confusing architecture that is inconsistent with other GOV.UK apps. Other GOV.UK apps that listen to queues process them directly, which reduces what you need to understand to monitor queue statuses. By combining RabbitMQ and Sidekiq it is much harder to understand the actual processing of messages from the message queue, with data in multiple places interlinked with other Sidekiq processing.

docs/arch/adr-013-rabbitmq-sidekiq-removal.md
```diff
@@ -0,0 +1,33 @@
...
+When the queue consumer code was written, there were no concurrency features available in our RabbitMQ consumer gem, so it made sense to push everything to Sidekiq where jobs could be processed concurrently. But now we have the ability to do this within the consumer itself, so we can remove the Sidekiq dependency and process jobs directly from RabbitMQ.
+
+## Decision
+We have decided to remove Sidekiq from the RabbitMQ queue consumer code, and instead leverage the concurrency features of the [govuk_message_queue_consumer][] gem.
+
+## Status
+Approved
+
+## Consequences
+One useful feature of Sidekiq is that it automatically handles retries. When a job fails, Sidekiq will catch the error and try again multiple times with exponential backoff. The tradeoff we've made here is that we've now had to write code to do this ourselves in the queue consumers, which involves a bit of extra complexity both with regards to the code in this repo and the setup of RabbitMQ.
```
kevindew: I expect it's worth flagging that we've done the extra complexity because RabbitMQ provides no other retry system than a basic automatically retry straight away without extra queue configuration

kevindew (commented):
Looks good, just a few suggestions.

docs/arch/adr-013-rabbitmq-sidekiq-removal.md
jackbot: Yep, done.

docs/arch/adr-013-rabbitmq-sidekiq-removal.md
jackbot: Thanks for this, I've updated the context.

docs/arch/adr-013-rabbitmq-sidekiq-removal.md
jackbot: Yep, added :+1:

kevindew (approved):
Nice one

PR #3237, opened 2025-04-08
Don't raise exception when base_path is missing

jackbot:
Currently we raise an error when processing a message from the queue if
the message payload is missing a `base_path` attribute.

However according to the content schema spec[1] this attribute is
optional, so there's no need to raise an error.

[1]: https://github.com/alphagov/publishing-api/blob/7b945cb0a7762b6326708143c7283df6a96daece/content_schemas/formats/contact.jsonnet#L3

### DIFF
--- a/lib/govuk_index/presenters/elasticsearch_delete_presenter.rb
+++ b/lib/govuk_index/presenters/elasticsearch_delete_presenter.rb
@@ -14,12 +14,6 @@ module GovukIndex
       base_path
     end
 
-    def valid!
-      unless payload["base_path"] || payload["content_id"]
-        raise(NotIdentifiable, "base_path and content_id missing from payload")
-      end
-    end
-
     def type
       raise NotFoundError if existing_document.nil?
 
--- a/lib/govuk_index/presenters/elasticsearch_presenter.rb
+++ b/lib/govuk_index/presenters/elasticsearch_presenter.rb
@@ -254,12 +254,6 @@ module GovukIndex
       common_fields.publishing_app
     end
 
-    def valid!
-      if format == "recommended-link" && !details.url
-        raise(MissingExternalUrl, "url missing from details section")
-      end
-    end
-
     def image_url
       details.image_url || (expanded_links.default_news_image if newslike?)
     end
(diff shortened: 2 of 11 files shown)


kevindew:
Thanks for this. If I understand right I don't think this PR does everything we want.

My understanding is that we want: 
- Search API to not index content that doesn't have a base path
- Search API to not raise an exception when it encounters such an occurence.

If I understand right, I think this would stop raising the exception yet it would try index content without a base path?

kevindew:
Thanks @jackbot - of course it seems to be more complicated than expected.

Hmm I guess what we want to ascertain is exactly what the behaviour is supposed to be. My understanding is that we can't index items without a base_path because we use the base_path as an id, see: https://www.gov.uk/api/search.json - but that doesn't answer how external_content should work, is there a conditional in the code for them to work?

kevindew:
Ahh I see now that this is the "recommended-link" exception already in #valid! - I'm not sure I quite understand how the code in https://github.com/alphagov/search-api/blob/927db5b001567b82ae5ba4c8818a08bb3c502ad5/spec/integration/govuk_index/external_content_spec.rb#L18-L27 gets a details.url but it seems to for the assertion on line 35.

So it sounds like it doesn't want base_paths unless it's the `external_content` type. I suspect we also don't want that to raise an error, but it seems a more unusual error given external_content requires a url. 

So I guess for us we have to decide where to do that filtering. Is this `GovukIndex::ElasticsearchPresenter` class used in multiple scenarios so we need to code to be on that? if not we could probably add a conditional that excludes items from the message processor (or I suppose at a push we could even keep the old exceptions in `GovukIndex::ElasticsearchPresenter` and also filter them out at the message processor level?

kevindew:
Thanks @jackbot this looks pretty close to what I meant but the aspect I'd like to do to finish this would be to handle both base_paths and external_content the same so handling is consistent.

So I'd suggest we remove the `valid!` method from the presenter and have two methods on the processor:

```
      if external_content_without_details_url?(payload)
        logger.info("#{payload['content_id']} ignored due to missing details.url")
        message.done
        return
      end

      if without_base_path?(payload)
        logger.info("#{payload['content_id']} ignored due to no base_path")
        message.done
        return
      end
```

kevindew (approved):
Nice one, sorry this went round the houses so much.

PR #3270, opened 2025-05-20
Changes to Countryside Stewardship Grant finder

lauraghiorghisor-tw:
🚨 These changes should be made live on the 27th of June. ⏰

- Changes only required removing 2 facets: `tiers_or_standalone_items` and `funding_amount` but there seem to be many fields registered here. Will remove them and test in integration.
- This seems to be the case for many other finders - I cannot find those fields in the Specialist Publisher schemas, but removing those is out of scope.

[Trello](https://trello.com/c/UXXwOxKQ/3728-specialist-finder-edit-request-countryside-stewardship-grants)


config/schema/elasticsearch_types/countryside_stewardship_grant.json
```diff
@@ -1,16 +1,6 @@
 {
   "fields": [
-    "business_activity",
```
lauraghiorghisor-tw, line 3: I guess I hope these fields were not doing anything here. Some exist on other finders too, with no equivalent in specialist publisher, or without being mentioned elsewhere in the config.

GDSNewt (approved):

PR #3271, opened 2025-05-20
Parts with link

hannako:
Trello: https://trello.com/c/L3TUHrsG/2564-fix-broken-search-api-links

### What

A SearchAPI document contains an array of `parts` which are built from information stored in the `attachments` key of `details` fetched from Publishing API. Previously, the parts contained only three attributes: `slug`,`title`,`body`. This PR adds an additional field, `link` to parts.

### Why?

[Finder-frontend](https://github.com/alphagov/finder-frontend/blob/a074bb293d2b4ce2c57a147842439a60af415faa/app/presenters/search_result_presenter.rb#L93) uses the value of the `slug` key to build the link to parts returned in search results. It does this by appending the slug to the parent document's URL. In most cases this is fine. However in the case of consultation outcomes' their html_attachments have an additional sub-directory, `/outcome/`, which is not accounted for. 

One known issue this has caused in on the official documents finder, where the part link returns a 404

Screenshot 2025-05-21 at 20 47 33

Correct link to the attachment: https://www.gov.uk/government/consultations/uks-future-exhaustion-of-intellectual-property-rights-regime/outcome/government-response-to-the-consultation-on-the-uks-future-exhaustion-of-intellectual-property-rights-regime

Broken link to an attachment from search results: https://www.gov.uk/government/consultations/uks-future-exhaustion-of-intellectual-property-rights-regime/government-response-to-the-consultation-on-the-uks-future-exhaustion-of-intellectual-property-rights-regime

### Todo

- [x] Add a conditional to Finder Frontend - if a part doesn't have a slug then skip
  - PR: https://github.com/alphagov/finder-frontend/pull/3795
- [x] Add link to Search API Document parts array (this PR)
- [x] Run rake task to represent all documents to SearchAPI
  - Can start by checking docs that were updated overnight, or represent a doc or two
  -  Run for all documents with attachments so we can deprecate slug field
    -  https://github.com/alphagov/publishing-api/blob/main/lib/tasks/represent_downstream.rake#L2
- [x] Use SearchAPI to eyeball results
- [x] Update finder-frontend to read the new value of link
- [x] Document the deprecation of the slug field and hand back to the team

### DIFF
--- a/lib/govuk_index/payload_preparer.rb
+++ b/lib/govuk_index/payload_preparer.rb
@@ -24,24 +24,32 @@ module GovukIndex
     def prepare_parts(payload)
       parts = payload["details"]["parts"]
 
-      if parts && parts.any?
-        updated_parts = parts.map do |part|
-          next part if part["link"]
+      updated_parts = if parts && parts.any?
+                        create_part_links(payload, parts)
+                      else
+                        promote_html_attachments_to_parts(payload)
+                      end
 
-          link = "#{payload['base_path']}/#{part['slug']}"
-          part.merge("link" => link)
-        end
+      updated_parts ? merge_details(payload, "parts", updated_parts) : payload
+    end
 
-        return merge_details(payload, "parts", updated_parts)
+    def create_part_links(payload, parts)
+      parts.map do |part|
+        next part if part["link"]
+
+        link = "#{payload['base_path']}/#{part['slug']}"
+        part.merge("link" => link)
       end
+    end
 
-      details = Indexer::PartsLookup.prepare_parts(
+    def promote_html_attachments_to_parts(payload)
+      details_with_html_attachments_as_parts = Indexer::PartsLookup.prepare_parts(
         payload["details"].merge("link" => payload["base_path"]),
         return_raw_body: true,
       )
-      return payload if details.fetch("parts", []).empty?
+      return if details_with_html_attachments_as_parts.fetch("parts", []).empty?
 
-      presented_parts = details["parts"].map do |part|
+      details_with_html_attachments_as_parts["parts"].map do |part|
         {
           "slug" => part["slug"],
           "title" => part["title"],
@@ -49,8 +57,6 @@ module GovukIndex
           "body" => [{ "content_type" => "text/html", "content" => part["body"] }],
         }
       end
-
-      merge_details(payload, "parts", presented_parts)
     end
 
     def merge_details(payload, field, value)
--- a/lib/govuk_index/payload_preparer.rb
+++ b/lib/govuk_index/payload_preparer.rb
@@ -22,7 +22,18 @@ module GovukIndex
     end
 
     def prepare_parts(payload)
-      return payload unless payload["details"].fetch("parts", []).empty?
+      parts = payload["details"]["parts"]
+
+      if parts && parts.any?
+        updated_parts = parts.map do |part|
+          next part if part["link"]
+
+          link = "#{payload['base_path']}/#{part['slug']}"
+          part.merge("link" => link)
+        end
+
+        return merge_details(payload, "parts", updated_parts)
+      end
 
       details = Indexer::PartsLookup.prepare_parts(
         payload["details"].merge("link" => payload["base_path"]),
--- a/lib/govuk_index/payload_preparer.rb
+++ b/lib/govuk_index/payload_preparer.rb
@@ -33,6 +33,7 @@ module GovukIndex
       presented_parts = details["parts"].map do |part|
         {
           "slug" => part["slug"],
+          "link" => part["link"],
           "title" => part["title"],
           "body" => [{ "content_type" => "text/html", "content" => part["body"] }],
         }
--- a/lib/govuk_index/payload_preparer.rb
+++ b/lib/govuk_index/payload_preparer.rb
@@ -33,8 +33,8 @@ module GovukIndex
       presented_parts = details["parts"].map do |part|
         {
           "slug" => part["slug"],
-          "link" => part["link"],
           "title" => part["title"],
+          "link" => part["link"],
           "body" => [{ "content_type" => "text/html", "content" => part["body"] }],
         }
       end
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -144,6 +144,7 @@
     "type": "objects",
     "children": {
       "slug": { "type": "searchable_identifier" },
+      "link": {"type": "searchable_identifier"},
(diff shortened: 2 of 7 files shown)


lib/govuk_index/payload_preparer.rb
```diff
@@ -35,7 +35,7 @@ def prepare_parts(payload)
           "slug" => part["slug"],
           "title" => part["title"],
           "body" => [{ "content_type" => "text/html", "content" => part["body"] }],
-        }
+        }.tap { |h| h["url"] = part["url"] if part["url"] }
```
kevindew: What are the problems that occur if we have a url field that is nil? I can see from the commit you reference there are spec issues but I was wondering if there were application issues?

I did think using compact on the issue might be better than the if condition but I somewhat wonder if we can have the whole thing?

Finally, is this a temporary change that'd be resolved after migrating or something we'd want permanently?

kevindew (commented):
Change looks pretty good - nice job getting into this.

I do think I'd benefit from understanding a bit of a broader plan of this work. For example I'm wondering does a field_definition change need to go out first and be reindexed before we can start writing these links to search

lib/govuk_index/payload_preparer.rb
chaecramb: Thanks for bringing this up, with some more investigation I think we're actually safe to just leave it as is, I've updated the specs to match.

kevindew (approved):
This looks good to me 👍 nice one.

The thing I'm not clear on at this point (I think I flagged it somewhere but can't see it on my earlier comments on this PR) is whether we need to do a reindex to use this field. My expectations are that we can start writing to the field without a reindex but it won't be available as a searchable field until we reindex, which should be fine. What I'm not clear on is whether we'll be able to access the data in a query without a reindex. I imagine you could learn this yourself by deploying to integration and seeing what happens. You may also be able to get some help from someone like @sihugh or someone from publishing like @ChrisBAshton (with experience of Specialist Publisher reindexing).

lib/govuk_index/payload_preparer.rb
```diff
@@ -22,7 +22,18 @@ def prepare_attachments(payload)
...
+
+      if parts && parts.any?
+        updated_parts = parts.map do |part|
+          next part if part["link"]
+
+          link = "#{payload['base_path']}/#{part['slug']}"
+          part.merge("link" => link)
+        end
+
+        return merge_details(payload, "parts", updated_parts)
```
kevindew: I'm not sure I understand why this has to exit early?

I somewhat wonder if you need this code and line 48 could be:

```
"link" => "#{payload['base_path']}/#{part['slug']}",
```

?

---

I did wonder is it because parts["link"] may already be populated somehow? 

if so would it be a case that we need line 48 to be:

```
"link" => part.fetch("link", "#{payload['base_path']}/#{part['slug']}"),
```

?

kevindew (commented):
Thanks @chaecramb, I do remain a bit confused about this so stabbed a little in the dark at my review here.

Would you be able to write up a bit of an explainer with code links as to the routes that either a Whitehall document with attachment ends up with attachments as parts AND how a travel advice documents with parts - I feel like I've still not groked how the system properly works.

Also sorry this is going on so long, I know you must be super sick of it now :( we can have a chat if it's annoying to try work out how we can put it to bed.

lib/govuk_index/payload_preparer.rb
chaecramb: @kevindew I could it without the early return, but would need another condition instead. The idea is that `if parts && parts.any?` is true, then all `prepare_parts` needs to do, is add a `link` field to the pre-existing parts. However, if there aren't any parts we instead promote attachments to parts, which is what the rest of the function after that conditional does.

kevindew: Ahh so maybe we need a conditional and break into two functions? perhaps we could play with variable naming to make it clearer. Can even be cheeky and use variables like actual_parts and attachments_pretending_to_be_parts if it helps understanding - I feel like 90% of the challenge here is understanding what is going on.

chaecramb:
@kevindew I've refactored `PayloadPreparer` a little, so hopefully it's clearer what it does now.

lib/govuk_index/payload_preparer.rb
```diff
@@ -22,23 +22,41 @@ def prepare_attachments(payload)
...
     def prepare_parts(payload)
-      return payload unless payload["details"].fetch("parts", []).empty?
+      parts = payload["details"]["parts"]
 
-      details = Indexer::PartsLookup.prepare_parts(
+      updated_parts = if parts && parts.any?
+                        create_part_links(payload, parts)
+                      else
+                        promote_html_attachments_to_parts(payload)
+                      end
```
kevindew, line 31: Great - this is much easier to understand 👍

spec/unit/govuk_index/payload_preparer_spec.rb
```diff
@@ -0,0 +1,55 @@
+require "spec_helper"
+
+RSpec.describe GovukIndex::PayloadPreparer do
```
kevindew: I'm thinking it might be best to avoid adding this file and to test these conditions through the existing "integration" test file (which really seems like it's more of a unit test than an integration test).

Tricky of course, because I think the way the tests are laid out here have tech debt and I think our challenge is to just not increase surprises. I somewhat suspect, despite good intentions, that a partial unit test file may add a new surprise.

What I'd suggest doing is:

For "when details.parts does not exist and an attachment URL matches the slug but is not under the base path"

modify one of the example attachments in: https://github.com/alphagov/search-api/blob/42ed135722aaf6a56f3463249b660c166490656f/spec/integration/govuk_index/payload_preparer_spec.rb#L159-L161 to include this scenario

I think the second scenario you have, "when details.parts exist and there are no attachments" is already covered in: https://github.com/alphagov/search-api/blob/42ed135722aaf6a56f3463249b660c166490656f/spec/integration/govuk_index/payload_preparer_spec.rb#L115-L117

kevindew (commented):
Super, thanks for your changes here I find it much easier to understand and it looks like it's just about ready to merge.

spec/unit/govuk_index/payload_preparer_spec.rb
chaecramb: Ok, yeah, that's cleaner. I've updated to take this approach.

chaecramb:
@kevindew I've pushed a change to address your comment regarding the tests

kevindew (approved):
Nice one

PR #3356, opened 2025-09-09
Add HMRC manuals finder

richardTowers:
This is a new finder that lets users search through all HMRC manuals / manual sections.

I've been working behind the scenes with HMRC to agree the content and settings for this, which is now done. We're planning on deploying it to production as a trial, and HMRC are going to get feedback on its effectiveness from external the users of their manuals.

As this is only configuration, and I haven't had to change any code or reindex anything in search, I don't think this will add to our maintenance burden significantly. I've also done everything I can to manage down HMRC's expectations of ongoing support - they understand that there's no promise of ongoing feature improvements to this finder, and they're still keen to have it over the current situation (where it's difficult to search across HMRC manuals).

[You can currently see the finder deployed to integration here](https://www.integration.publishing.service.gov.uk/find-hmrc-manuals)

Published using `FINDER_CONFIG=hmrc_manuals_finder.yml rake publishing_api:publish_finder`

### DIFF
--- a/config/finders/hmrc_manuals_finder.yml
+++ b/config/finders/hmrc_manuals_finder.yml
@@ -1,0 +1,75 @@
+---
+base_path: "/find-hmrc-manuals"
+content_id: f11819b7-1c54-4b54-a8e8-56e0004b2dbf
+description: Find and search HMRC technical guidance for HMRC staff and tax professionals.
+details:
+  beta_message: Help us improve this page - email your feedback to <a class="govuk-link" href="mailto:hmrcmanualsteam@hmrc.gov.uk">hmrcmanualsteam@hmrc.gov.uk</a>
+  summary:
+    - content: |-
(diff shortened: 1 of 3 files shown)


leenagupte (commented):
The changes themselves look good, but as mentioned in the slack thread, could you update the commit message to say why organisations and taxons need to be added.

As this page is being configured in the same place as the non-specialist finders, e.g. news, guidance etc, it didn't occur to me that they were being used to render the breadcrumb and the organisation metadata, as none of those other finders do. I was looking at the filter options, and wondered if they needed to be added there. It's likely someone else could trip over that too.

richardTowers:
Thanks @leenagupte  - commit message updated in 239f147a56ec3b015a59a6fab760409a70e1ea1a

leenagupte (approved):
Thanks for updating the commit message 👍

PR #3379, opened 2025-10-13
Ensure tests always run in UTC

koetsier:
Previously, the test suite relied on the host operating system's timezone, which could cause failures when the local timezone was not UTC.

This change explicitly sets ENV['TZ'] to "UTC" before the test environment is loaded, ensuring consistent time-based behaviour across all environments.

part of Jira: https://gov-uk.atlassian.net/browse/SCH-1486

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -75,6 +75,10 @@ RSpec.configure do |config|
     Services.cache.clear
   end
 
+  config.around(:each) do |example|
+    ClimateControl.modify(TZ: "UTC") { example.run }
+  end
+
   if config.files_to_run.one?
     config.default_formatter = "doc"
   end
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -75,6 +75,14 @@
     Services.cache.clear
   end
 
+  config.around do |example|
```
hannako: Is the only way to do this to overwrite the environment variable? Can we not use something like `Time.use_zone('London') ...`?

hannako: Follow up! If that's not possible for some reason, can we use climate control instead just for the site map spec? [This](https://github.com/alphagov/search-api/blob/99ddaca24103ea0bec5fe7528a13d82bd34fa71b/spec/integration/app/authorization_spec.rb#L7-L9) kind of thing.

koetsier: Nice one. I didn't know about climatecontrol. That looks a lot neater

We can't use 'London' because in summertime it is UTC+1 which causes the failures

hannako (approved):

emmalowe (approved):

PR #3394, opened 2025-10-28
Remove redundant rake task [WHIT-1833]

ChrisBAshton:
We've now run the tasks migrating finders from Whitehall to Search API. There is no longer any need for this rake task, as all the routes are now reserved by search-api.

Jira: https://gov-uk.atlassian.net/browse/WHIT-1833


kevindew (approved):
Thank you

PR #3399, opened 2025-11-03
Handle empty filter parameters

koetsier:
Requests with empty filter parameters, such as '/search.json?filter_public_timestamp' should return an HTTP code 422 (Unprocessable Entity) and not raise an exception

Jira:  https://gov-uk.atlassian.net/browse/SCH-1579

### DIFF
--- a/lib/parameter_parser/search_parameter_parser.rb
+++ b/lib/parameter_parser/search_parameter_parser.rb
@@ -287,12 +287,12 @@ private
     # operation is :reject or :filter
     # multivalue_query is :all or :any
     def initialize(field_name, values, operation, multivalue_query)
+      @errors = values.empty? ? [%(no value for parameter "#{field_name}")] : []
       @field_name = field_name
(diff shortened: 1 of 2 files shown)


hannako:
This looks good, would you mind adding a link to the review app and an equivalent link to integration search showing the different behaviour?

lib/parameter_parser/search_parameter_parser.rb
```diff
@@ -287,12 +287,12 @@ class Filter
     # operation is :reject or :filter
     # multivalue_query is :all or :any
     def initialize(field_name, values, operation, multivalue_query)
+      @errors = values.empty? ? [%(no value for parameter "#{field_name}")] : []
```
hannako, line 290: When finder-frontend renders search/all it sends a request to search-api with empty params, so this line is raising a 422 when trying to load the page. 

https://github.com/alphagov/finder-frontend/blob/main/app/controllers/search_controller.rb#L28

Screenshot 2025-12-02 at 11 00 08

koetsier:
see comment by @hannako

PR #3411, opened 2025-11-12
Allow indexing of organisations in the govuk index [WHIT-1799]

ryanb-gds:
This requires adding a number of organisation-specific fields to handle logo, status, branding etc.

[JIRA](https://gov-uk.atlassian.net/browse/WHIT-1799)

### DIFF
--- a/lib/govuk_index/presenters/details_presenter.rb
+++ b/lib/govuk_index/presenters/details_presenter.rb
@@ -59,6 +59,34 @@ module GovukIndex
       details["attachments"]&.any? { |attachment| attachment["hoc_paper_number"].present? || attachment["unnumbered_hoc_paper"] }
     end
 
+    def acronym
+      details["acronym"]
+    end
+
+    def logo_formatted_title
+      details.dig("logo", "formatted_title")
+    end
+
+    def logo_url
+      details.dig("logo", "image", "url")
+    end
+
+    def organisation_brand
+      details["brand"]
+    end
+
+    def organisation_crest
+      details.dig("logo", "crest")
+    end
+
+    def organisation_state
+      details.dig("organisation_govuk_status", "status")
+    end
+
+    def organisation_type
+      details["organisation_type"]
+    end
+
   private
 
     def service_manual
--- a/lib/govuk_index/presenters/elasticsearch_presenter.rb
+++ b/lib/govuk_index/presenters/elasticsearch_presenter.rb
@@ -15,6 +15,7 @@ module GovukIndex
 
     def document
       {
+        acronym: details.acronym,
         ai_assurance_technique: specialist.ai_assurance_technique,
         aircraft_category: specialist.aircraft_category,
         aircraft_type: specialist.aircraft_type,
@@ -140,6 +141,8 @@ module GovukIndex
         life_saving_maritime_appliance_manufacturer: specialist.life_saving_maritime_appliance_manufacturer,
         link: common_fields.link,
         location: specialist.location,
+        logo_formatted_title: details.logo_formatted_title,
+        logo_url: details.logo_url,
         mainstream_browse_page_content_ids: expanded_links.mainstream_browse_page_content_ids,
         mainstream_browse_pages: expanded_links.mainstream_browse_pages,
         manual: details.parent_manual,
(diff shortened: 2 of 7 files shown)


lib/govuk_index/presenters/details_presenter.rb
```diff
@@ -38,6 +38,7 @@ def parent_manual
     def image_url
       details.dig("image", "url")
     end
+    alias_method :logo_url, :image_url
```
hannako: Looking at the [OrganisationPresenter](https://github.com/alphagov/whitehall/blob/main/app/presenters/publishing_api/organisation_presenter.rb#L83-L87) in Whitehall, I would have thought this should be.

```
def logo_url
  details.dig("logo","image","url")
end
```

 If I'm wrong please could you explain why for my understanding!

spec/unit/govuk_index/presenters/details_presenter_spec.rb
```diff
@@ -256,4 +256,37 @@
...
   end
+
+  context "organisation" do
+    let(:format) { "organisation" }
+    let(:details) do
+      {
+        "acronym" => "Companies House",
+        "brand" => "department-for-business-trade",
+        "image" => {
+          "url" => "https://assets.publishing.service.gov.uk/media/67f65f2bb7e44efc70acc3db/test_logo.png",
```
hannako: I think the image key should be inside the logo hash, rather than at the top level.

hannako (commented):
I've not quite finished Ryan, I'm up to the expanded_links presentation. I'll come back to it shortly!

lib/govuk_index/presenters/details_presenter.rb
ryanb-gds: No, you're absolutely spot on, great catch. I think because I'm working on topical events I looked at the presenter for that. Topical events also have logos so I didn't notice the difference 🤦 

What's weird is when I tested on integration there were three govuk indexes. One of them did get a logo URL for the record I tested, so I just assumed that was the active one and the other indexes weren't currently active (although they were newer than the one that got the URL, so that seemed odd too). Very confusing, I'll need to dig into that

hannako: Did you get to the bottom of the multiple indexes?

ryanb-gds: Just writing up Slack thread here as well for posterity - we have confirmed that the index that did have a logo URL was the active one, and that the others were inactive through running `rake search:list_indices`. So I think this is now correct, having made the change suggested in @hannako's first comment above

lib/govuk_index/presenters/expanded_links_presenter.rb
```diff
@@ -16,6 +16,10 @@ def organisation_content_ids
       content_ids("organisations")
     end
 
+    def parent_organisations
```
hannako: nitpick

I wondered if this method should be `ordered_parent_organisations` to match the pattern on line 35 for `primary_publishing_organisation`

lib/govuk_index/presenters/elasticsearch_presenter.rb
```diff
@@ -149,9 +152,14 @@ def document
         market_sector: specialist.market_sector,
         medical_specialism: specialist.medical_specialism,
         opened_date: specialist.opened_date,
+        organisation_brand: details.organisation_brand,
+        organisation_crest: details.organisation_crest,
         organisation_content_ids: expanded_links.organisation_content_ids,
+        organisation_state: details.organisation_state,
+        organisation_type: details.organisation_type,
         organisations: expanded_links.organisations,
```
hannako, line 160: I am _very_ sketchy on how this works, but it looks when organisations are added to the government index, they [are tagged to themselves](https://github.com/alphagov/search-api/blob/1886b52bf67b38afb081942b5a825d3a8448859b/lib/indexer/document_preparer.rb#L127). 

So I **think** that's why https://www.integration.publishing.service.gov.uk/api/search.json?filter_organisations=department-for-work-pensions&filter_format=organisation returns the organisation page itself. 

Do we need to do the same here?! Or when publishing API publishes an organisation, does the expanded links contain the organisation itself?

hannako (commented):
Sorry to ask more questions!

lib/govuk_index/presenters/elasticsearch_presenter.rb
ryanb-gds: I can't find any use of this in the alphagov organisation, so I haven't ported it over. Publishing API doesn't automatically add the link so I'd have to do it from Whitehall now

hannako: Let me just check in with Tara on this, she might have some context for why it was needed.

hannako (approved):
It looks like changing the name of the `parent_organisations` key to `ordered_parent_organisations` has broken your specs. If it's less work to ignore my suggestion and use `parent_organisations` instead that's fine - it was only a nitpick suggestion.

Approving this PR in advance of you fixing that, as don't want to be a blocker. And on the assumption that you're planning to add an organisation to its own links as discussed offline in Slack.

PR #3412, opened 2025-11-17
[SCH-1598] Add tests for sitemap Uploader class

leenagupte:
# What?
Add tests for Sitemap::Uploader

# Why?
Increases the test coverage for Sitemap::Uploader from 44% to 100%.

# Coverage report

## Before
Screenshot 2025-11-17 at 11 43 25

## After
Screenshot 2025-11-17 at 11 44 42

### DIFF
--- a/lib/rummager.rb
+++ b/lib/rummager.rb
@@ -190,6 +190,7 @@ require "services"
 require "sitemap/property_boost_calculator"
 require "sitemap/generator"
 require "sitemap/presenter"
+require "sitemap/uploader"
 require "retryable_queue_message"
 
 require "core_extensions/time/to_string"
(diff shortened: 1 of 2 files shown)


lib/rummager.rb
```diff
@@ -190,6 +190,7 @@
 require "sitemap/property_boost_calculator"
 require "sitemap/generator"
 require "sitemap/presenter"
+require "sitemap/uploader"
```
hannako, line 193: I don't understand the consequences of adding this require. Highlighting my lack of experience around sinatra apps. I would have expected it to be already required, but just not be tested. Do you know?

leenagupte: It's needed to be able to run the test file on it's own. If you don't add this line, you can run `rake` and the tests will pass, but running an individual test fails. The other option is to require the file in the spec file, but I added it here to follow the same pattern as the other files.

hannako: Looks good then - thanks for explaining. Shall we set it as draft until the integration test is over?

leenagupte: I'll add a Do not merge label in the title.

hannako (approved):
Looks good 🏅

PR #3416, opened 2025-11-19
Add more examples of how to query search-api

leenagupte:
Add an explanation breaking down how aggregate queries are constructed.

Co-authored-by: Simon Hughesdon <simon.hughesdon@digital.cabinet-office.gov.uk>

### DIFF
--- a/docs/using-the-search-api.md
+++ b/docs/using-the-search-api.md
@@ -104,7 +104,7 @@ This can be useful to find all documents that are tagged to two taxons (use `fil
 
 ### aggregate_*
 
-Aggregations look at all the values of a field and count up the number of times each one appears in documents matching the search. For example, the `aggregate_organisations` parameter will group search results by organisation, if it is set to a valid value.
+Aggregations is just a SQL "GROUP BY" in other words. Aggregations look at all the values of a field and count up the number of times each one appears in documents matching the search. For example, the `aggregate_organisations` parameter will group search results by organisation, if it is set to a valid value. (See: [full list of fields](https://github.com/alphagov/search-api/blob/5a47f2147b071c0e78d7be94faf14d395b15936e/lib/parameter_parser/base_parameter_parser.rb#L47-L74) that can be aggregated on)
 
 The value of an `aggregate_*` parameter is a comma separated list of options:
 
@@ -140,8 +140,50 @@ The value of an `aggregate_*` parameter is a comma separated list of options:
 
 - `example_fields` (optional, and only used if `examples` is provided): colon-separated list of fields.
   If the examples option is supplied, this lists the fields which are returned for each example. By default, only a small number of fields are returned for each.
+  
+#### Examples
 
-Example of aggregate query with multiple options: <https://www.gov.uk/api/search.json?count=0&aggregate_organisations=1,scope:all_filters,order:filtered:-count>
+##### A simple example
+
+The best way to understand what an aggregate query does is to read it right to left, for example:
+
+<https://www.gov.uk/api/search.json?aggregate_rendering_app=20,examples:1,example_scope:query&count=0&filter_publishing_app=publisher>
+
+The query string can be divided into two sections:
+
+- the search results: `count=0&filter_publishing_app=publisher`
+- the aggregation: `aggregate_rendering_app=20,examples:1,example_scope:query` 
+
+###### The search results
+This returns a set of results for the filters provided.
+ - `filter_publishing_app=publisher` - is a query for results published by Publisher.
+ - `count=0` - The results would normally display 10 documents where the filter is true, but this counts sets it for zero actual results. If `count` is set to another number, e.g. [`count=2`](https://www.gov.uk/api/search.json?aggregate_rendering_app=20,examples:1,example_scope:query&count=2&filter_publishing_app=publisher), then two regular search results would be returned above the aggregation results.
+ 
+###### The aggregation
+This returns back a separate set of results and is not affected by `count=0`.
+- `example_scope:query` - The scope of a query can be either global (where it doesn't have filters applied), or `query`, which applies whatever filters are included. So in our example we will filter the results by `publishing_app=publisher`
+- `examples:1` - for each each aggregate that it identifies (in this case "rendering app" as `aggregate_rendering_app` is applied), it will provide 1 example document
+- `aggregate_rendering_app=20` - Requests all documents that have a publishing app matching the filter and groups them by rendering app. It will select up to 20 different rendering apps to group by. On GOV.UK there are only 2 rendering applications that render content published by Publisher, `frontend` and `government_frontend`.
+
+##### An example with a query
+
+<https://www.gov.uk/api/search.json?q=biscuit&aggregate_content_store_document_type=20,examples:2,example_scope:query,example_fields:title:description&count=5>
+
+- Searches for biscuit
+- Groups the results by the content_store_document_type field
+  - Gets the first 20 content store document types
+  - Provides two examples of each
+  - The examples use the query too (they match the search for biscuit)
+  - Include title, description fields in the example
+- Also get 5 normal search results
+
+##### An example with multiple options
+
+<https://www.gov.uk/api/search.json?count=0&aggregate_organisations=1,scope:all_filters,order:filtered:-count>
+
+- Returns the organisation with the most documents tagged to it
+- If you change the `aggregate_organisations` to 10, you'll see the top 10 organisations with the most documents.
+- The organisation with the most documents is displayed first because `-count` means start with the highest, i.e. in reverse. 
 
 ### ab_tests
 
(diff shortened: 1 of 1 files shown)


docs/using-the-search-api.md
```diff
@@ -91,6 +106,25 @@ The filters are grouped by field name: documents will only be returned if they m
 
 `filter_*`/`reject_*` works with date fields too, although unlike string fields, it is not permitted to provide multiple values for a single date field filter. The date field filter value should be either a `from:<date>`, `to:<date>` or both (comma separated), where `<date>` is an ISO formatted date (with no timezone: UTC is assumed). Date ranges are inclusive: for example, `from:2014-04-01 00:00,to:2014-04-02 00:00` is a range of 24 hours from midnight at the start of April the 1st 2014. If the time is omitted, the `from:` parameter defaults to `00:00` and the `to:` parameter defaults to `23:59`, i.e. `from:2014-04-01,to:2014-04-02` covers a full 48 hour period.
 
+#### Examples
+
+<https://www.gov.uk/api/search.json?filter_organisations=hm-revenue-customs&fields=title&order=-public_timestamp>
+
+- Only includes results from the hm-revenue-customs organisation.
+- Only includes the title (over the minimum)
```
emmalowe: I think the "minimum" list of fields is specified [here](https://github.com/alphagov/search-api/blob/main/lib/search/presenters/result_presenter.rb#L66). We could include that detail under the "fields" sub-heading.

emmalowe (commented):
Wow, aggregates make so much more sense now with examples and explanations! 🤯 

Thanks @leenagupte ❤️

docs/using-the-search-api.md
```diff
@@ -140,8 +174,45 @@ The value of an `aggregate_*` parameter is a comma separated list of options:
...
+<https://www.gov.uk/api/search.json?aggregate_rendering_app=20,examples:1,example_scope:query&count=0&filter_publishing_app=publisher>
+
+`filter_publishing_app` is a query for results published by Publisher. It would normally display all documents where this is true, but it's set for zero actual results (the count=0)
+Then you're into the aggregate/facet bit which brings back a separate set of results and is not affected by the count=0:
+- the scope of this query is the above bit, so it will also filter these results by `publishing_app=publisher`
+  - (the other option is global where it doesn't have filters applied)
+- it'll get up to 20 rendering apps (in this case there are only 2, `frontend` and `government_frontend`) that have content matching the filter
+- for each rendering app that it identifies, it'll provide 1 example document
+
+If count was set to another number in the first filter, e.g. 2, then you would see two sets of results, the aggregate as broken above, and two regular search results:
```
emmalowe: typo? "aggregate as broken above" -> "aggregate as broken **down** above"

docs/using-the-search-api.md
```diff
@@ -140,8 +174,45 @@ The value of an `aggregate_*` parameter is a comma separated list of options:
 
 - `example_fields` (optional, and only used if `examples` is provided): colon-separated list of fields.
   If the examples option is supplied, this lists the fields which are returned for each example. By default, only a small number of fields are returned for each.
+  
+#### Examples
+
+##### A simple example
+
+The best way to understand what an aggregate query does is to read it left to right, for example:
```
emmalowe: Do you mean "right to left" instead of "left to right"? Your explanation seems to go right to left.

docs/using-the-search-api.md
```diff
@@ -140,8 +174,45 @@ The value of an `aggregate_*` parameter is a comma separated list of options:
...
+
+<https://www.gov.uk/api/search.json?q=biscuit&aggregate_content_store_document_type=20,examples:2,example_scope:query,example_fields:title:description&count=5>
+
+- Searches for biscuit
+- Groups the results by the content_store_document_type field
+  - Gets the first 20 content store document types
+  - Provides two examples of each
+  - The examples use the query too (they match the search for biscuit)
+  - Include title, description fields in the example
+Also get 5 normal search results
```
emmalowe: This line has funny formatting. Maybe you wanted to include it in the list?

docs/using-the-search-api.md
```diff
@@ -104,7 +138,7 @@ This can be useful to find all documents that are tagged to two taxons (use `fil
 
 ### aggregate_*
```
emmalowe, line 141: It looks like we can only aggregate on certain fields, which are listed here:
https://github.com/alphagov/search-api/blob/main/lib/parameter_parser/base_parameter_parser.rb#L83

docs/using-the-search-api.md
hannako: Suggestion:

The best way to understand what an aggregate query does is to read it right to left, for example:

<https://www.gov.uk/api/search.json?aggregate_rendering_app=20,examples:1,example_scope:query&count=0&filter_publishing_app=whitehall>

- `&count=0&filter_publishing_app=whitehall` is a query for results published by Whitehall. It would normally display all documents where this is true, but it's set for zero actual results (the `count=0`)

- `aggregate_rendering_app=20,examples:1,example_scope:query` is the aggregation and filter. It brings back a separate set of results and is not affected by the `count=0`. 
  - `example_scope:query` The scope of a query can be either `global` (where it doesn't have filters applied), or `query`, which applies whatever filters are included. So in our example we will filter the results by `publishing_app=whitehall`
  - `examples:1` For each rendering app that it identifies, it'll provide 1 example document
  - `aggregate_rendering_app=20` Requests all documents that match the filters, grouped by rendering application. It will get only select 20 of the available values of `rendering_app` to group by. On gov.uk there are only actually 2 frontend applications that render content published by Whitehall, `frontend` and `government_frontend`, so giving a value of 20 is a way of making sure we group by all available applications.

hannako (approved):
Looks great! Thanks so much for doing it!

docs/using-the-search-api.md
leenagupte: I've removed this line and merged it into the paragraph about the count.

docs/using-the-search-api.md
leenagupte: Thanks for the suggestion. I've broken down the parts of the query a bit more. Hopefully it makes more sense now.

leenagupte:
@emmalowe @hannako Thanks for your reviews. I've amended the sections and fixed up. Hopefully the bit on aggregates is easier to follow now.

docs/using-the-search-api.md
```diff
@@ -73,12 +75,27 @@ It uses a 0-based index.
...
+#### Example
+
+<https://www.gov.uk/api/search.json?q=tax&count=20&start=10>
+
+Gets 20 results starting at the tenth.
+
 ### fields
 
-Only a subset of fields are returned by default. You can override the fields returned using the `fields` parameter. Refer to the fields in [field_definitions.json](/config/schema/field_definitions.json).
+Only a subset of [fields are returned by default](https://github.com/alphagov/search-api/blob/276bcba361bd334ebd9302a635d9be4e68920208/lib/parameter_parser/base_parameter_parser.rb#L20-L31). You can override the fields returned using the `fields` parameter. Refer to the fields in [field_definitions.json](/config/schema/field_definitions.json).
```
emmalowe: I just realised this links to the wrong section. I think you want:

https://github.com/alphagov/search-api/blob/276bcba361bd334ebd9302a635d9be4e68920208/lib/parameter_parser/base_parameter_parser.rb#L127

docs/using-the-search-api.md
emmalowe: Oops, sorry, this is the correct link:
https://github.com/alphagov/search-api/blob/5a47f2147b071c0e78d7be94faf14d395b15936e/lib/parameter_parser/base_parameter_parser.rb#L47

What I gave before is fields that you can request aggregate examples for.

docs/using-the-search-api.md
```diff
@@ -140,8 +176,50 @@ The value of an `aggregate_*` parameter is a comma separated list of options:
...
+
+The query string can be divided into two sections:
+
+- the search results: `count=0&filter_publishing_app=publisher`
+- the aggregation: `aggregate_rendering_app=20,examples:1,example_scope:query` 
+
+###### The search results
+This returns a set of results for the filters provided.
+ - `filter_publishing_app` - is a query for results published by Publisher.
+ - `count=0` - The results would normally display all documents where the filter is true, but this counts sets it for zero actual results. If `count` is set to another number, e.g. [`count=2`](https://www.gov.uk/api/search.json?aggregate_rendering_app=20,examples:1,example_scope:query&count=2&filter_publishing_app=publisher), then two regular search results would be returned above the aggregation results.
```
emmalowe: I think it displays 10 by default, rather than everything.

docs/using-the-search-api.md
emmalowe: Thanks @leenagupte , I do find this easier to follow now!

leenagupte:
@emmalowe Thanks. I've amended based on your last set of comments and fixed up.

docs/using-the-search-api.md
emmalowe: @leenagupte this one still needs fixing I think

leenagupte:
Thanks @emmalowe. The link has been fixed and fixed up.

emmalowe (approved):

PR #3419, opened 2025-11-21
[SCH-1598] Add autocomplete presenter tests

leenagupte:
# What?
Add tests for Autocomplete class. 
Also adds documentation about how to see autocomplete suggestions for a query.

# Why?
Increases the test coverage for Indexer from 53% to 100%.

# Coverage report

## Before
Screenshot 2025-11-21 at 14 00 37

## After
Screenshot 2025-11-21 at 16 29 02

### DIFF
--- a/docs/autocomplete.md
+++ b/docs/autocomplete.md
@@ -1,0 +1,339 @@
+# Autocomplete
+
+Autocomplete is a feature that suggests search phrases as you type. In search-api autocomplete is based on popularity data from GA4.
+
+## Does autocomplete work?
+
+Unclear. It is possible to see autocomplete suggestions from elasticsearch in the payload but these suggestions are not shown to the user in the supergroup or specialist finders.
+
+## How do I see autocomplete suggestions for a query?
+
+To see autocomplete suggestions you must add `suggest=autocomplete` to the query string. For example: 
+
+In production visit: 
+<https://www.gov.uk/api/search.json?q=tax&suggest=autocomplete&count=1>
+
+Or locally go to:
+<http://search-api.dev.gov.uk/search?q=tax&suggest=autocomplete&count=1>
+
+You should see something similar to:
+
+```json
+"suggested_autocomplete": [
+  "Tax your vehicle",
+  "Tax-Free Childcare",
+  "Tax overpayments and underpayments",
+  "Tax codes",
+  "Tax on savings interest",
+  "Tax your vehicle without a vehicle tax reminder",
+  "Tax on dividends",
+  "Tax when you sell property"
+],
+```
+
+<details>
+<summary>Full example response</summary>
+
+```json
+{
+  "results": [
+    {
+      "description": "Renew or tax your vehicle for the first time using a reminder letter, your log book or the green 'new keeper' slip - and how to tax if you do not have any documents",
+      "format": "transaction",
+      "link": "/vehicle-tax",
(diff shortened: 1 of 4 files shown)


Tetrino (approved):
Always hate how lengthy tests can get in search, but LGTM - sensible documentation and testing to cover a notable gap. No notes to add on my part, though the others may want to weigh in.

hannako:
Do we actually use autocomplete on the frontend anywhere? Finder frontend only seems to call the v2 autocomplete endpoint https://www.gov.uk/search/guidance-and-regulation?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0

Oh but maybe an external service is using it.

docs/autocomplete.md
```diff
@@ -0,0 +1,339 @@
+# Autocomplete
+
+Autocomplete is a features that suggests search phrases as you type. In search-api autocomplete is based on popularity data from GA4.
```
emmalowe: typo: `features` -> `feature`

emmalowe:
I was doing a little bit more digging into how elasticsearch comes up with the suggestions for a query. It looks like, when the search API endpoint is hit, we check whether we want to return suggestions [here](https://github.com/alphagov/search-api/blob/main/lib/search/query.rb#L89) and then query elasticsearch for suggestions [here](https://github.com/alphagov/search-api/blob/main/lib/search/query.rb#L155).

As far as I can tell we're using the elasticsearch Suggester API: https://www.elastic.co/docs/reference/elasticsearch/rest-apis/search-suggesters

Would that be helpful info to add (potentially in a separate PR to this one)?

emmalowe (approved):
Tiny typo to fix and a suggestion for more details in docs (potentially in a follow up), but otherwise this looks good to me!

leenagupte:
> Do we actually use autocomplete on the frontend anywhere? Finder frontend only seems to call the v2 autocomplete endpoint https://www.gov.uk/search/guidance-and-regulation?parent=%2Feducation&topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0
> 
> Oh but maybe an external service is using it.

I couldn't see that it is being used by GOV.UK in the frontend, but you're right, an external user of the api could be using it. They'd need to pass `suggest=autocomplete` into the query string. So perhaps an analyst can see if there are any hits for that to help us decide whether to keep this functionality.

leenagupte:
> I was doing a little bit more digging into how elasticsearch comes up with the suggestions for a query. It looks like, when the search API endpoint is hit, we check whether we want to return suggestions [here](https://github.com/alphagov/search-api/blob/main/lib/search/query.rb#L89) and then query elasticsearch for suggestions [here](https://github.com/alphagov/search-api/blob/main/lib/search/query.rb#L155).
> 
> As far as I can tell we're using the elasticsearch Suggester API: https://www.elastic.co/docs/reference/elasticsearch/rest-apis/search-suggesters
> 
> Would that be helpful info to add (potentially in a separate PR to this one)?

I think we're missing a whole section on Elasticsearch and how we're using their API for the various suggestions. But if we decide to keep the autocomplete feature, it would be good to add those docs.

I think there's a bit about how `suggest=autocomplete` is parsed from the query as well to join up the whole journey. 

I focused on getting the payload so we can see what's being sent at various stages to help with testing.

PR #3426, opened 2025-11-27
Increase test coverage of special route publisher

hannako:
Add small refactor to use existing Services.publishing_api method. 
#### Before:
Screenshot 2025-11-27 at 20 01 16

#### After:
Screenshot 2025-11-27 at 20 00 55

https://gov-uk.atlassian.net/browse/SCH-1598


leenagupte (approved):
👍

PR #3427, opened 2025-11-28
Document how Search represents parts and attachments

koetsier:
This documents: https://github.com/alphagov/search-api/pull/2044

diff: https://github.com/alphagov/search-api/pull/3427/files?short_path=1116562#diff-1116562c64131afb8fdc923f67a5a258dc3870293c277ee71fc8d974bfb79f2a

https://gov-uk.atlassian.net/browse/SCH-1624

### DIFF
--- a/docs/indexing.md
+++ b/docs/indexing.md
@@ -65,3 +65,9 @@ See [schemas](schemas.md) for more detail.
 After changing the schema, you'll need to recreate the index. This reindexes documents from the existing index.
 
   SEARCH_INDEX=all bundle exec rake search:migrate_schema
+
+### Representing parts and attachments
+Parts are subpages within a single GOV.UK content item, each with its own title, body, and slug. 
+They allow one piece of content to be split into multiple sections (e.g., /parent/section-name) without creating separate content items. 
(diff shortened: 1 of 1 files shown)


docs/indexing.md
```diff
@@ -65,3 +65,6 @@ See [schemas](schemas.md) for more detail.
 After changing the schema, you'll need to recreate the index. This reindexes documents from the existing index.
 
   SEARCH_INDEX=all bundle exec rake search:migrate_schema
+
+### Representing parts and attachments
```
emmalowe, line 69: I had to look up what "parts" are usually used for for this to make sense to me 😅 Perhaps adding something on that would help? For example:

> "Parts" are subpages of a document. These are included in the parts field in the elasticsearch schema. The Search API also uses the parts field to index HTML attachments, as well as parts, as if they were sections of the main document. Instead of treating HTML attachments like file downloads, they are also stored as additional “parts” of the parent content item. This keeps everything in one indexed document while still allowing each HTML attachment’s title and body to be searchable.

koetsier: I changed it to add a bit more information about parts

emmalowe: Looks great, thanks!

emmalowe (approved):

PR #3432, opened 2025-12-03
Refactor connectivity integration tests

stringiest:
Increases test coverage for lib/healthcheck/elasticsearch_connectivity_check.rb to 96.88% - leaves one line of untested code.
Screenshot 2025-12-03 at 16 05 54

Stub the requests to `_cluster/health` in a similar way to the stubbing used for diskspace check.

Previously although the tests were raising a Faraday Error it was raising with a message about incorrect number of arguments rather than not being able to connect to clusters.


koetsier (approved):
looks good to me

PR #3436, opened 2025-12-08
Add tests for missing metadata

emmalowe:
Adds testing for:

- lib/missing_metadata/fetcher.rb: 37.04 % -> 100%
- lib/missing_metadata/runner.rb: 75.00 % -> 100%

Increases overall test coverage to 96.09% 🥳 

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-1598

I've added some context on the existing tests so they're easier to follow, but it does mean there are lots of formatting changes in this PR. I've split up the commits so that the first one is formatting/context changes and the rest are functional changes, so I'd recommend reviewing this PR by looking at individual commits separately.

### DIFF
--- a/spec/integration/missing_metadata_spec.rb
+++ b/spec/integration/missing_metadata_spec.rb
@@ -16,6 +16,20 @@ RSpec.describe "MissingMetadataTest" do
       expect([{ _id: "/path/to_page", index: "government_test" }]).to eq results
     end
 
+    it "ignores external links" do
+      commit_document(
+        "government_test",
+        {
+          "link" => "https://www.nhs.uk",
+        },
+      )
+
+      runner = MissingMetadata::Runner.new("content_id", search_config: SearchConfig.default_instance, logger: io)
+      results = runner.retrieve_records_with_missing_value
+
+      expect(results).to be_empty
+    end
+
     it "ignores already set content_id" do
       commit_document(
         "government_test",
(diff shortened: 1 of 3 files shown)


spec/integration/missing_metadata_spec.rb
```diff
@@ -63,6 +77,153 @@
...
+
+      context "when fetch content_id times out" do
+        it "sleeps for 1 second and retries, until content_id is returned" do
+          commit_document(
+            "government_test",
+            {
+              "link" => @base_path,
+            },
+          )
+          allow(io).to receive(:puts)
```
koetsier: StringIO can already accept 'puts', so I don't think this is necessary. for line 216, you can check if io.string contains your string.

spec/integration/missing_metadata_spec.rb
```diff
@@ -63,6 +77,153 @@
...
+    context "when add metadata returns an error" do
+      it "logs a message and skips the result" do
+        commit_document(
+          "government_test",
+          {
+            "link" => "/path/to_page",
+          },
+        )
+        allow_any_instance_of(MissingMetadata::Fetcher).to receive(:add_metadata).and_raise(StandardError)
+        allow(io).to receive(:puts)
```
koetsier: same as below

hannako (commented):
I've not gone through the bulk of the changes, ie the fourth commit, but would you mind adding some more context to why you're making the change in the logger commit, as I'm not sure I follow. Is it just to make it easier to test?

emmalowe:
> I've not gone through the bulk of the changes, ie the fourth commit, but would you mind adding some more context to why you're making the change in the logger commit, as I'm not sure I follow. Is it just to make it easier to test?

It was mostly for consistency. The change on line 25 of `runner.rb` seemed like a no brainer, since you'd want to be using the same logger across that whole class. Then in `fetcher.rb`, I figured you'd want to also be using the same logger across the whole rake task. To be honest though, I'm not really sure why we're passing in a logger to `MissingMetadata::Runner` in the first place - that seems a bit unusual to me. If there's a better way of doing this, I'm all ears.

It was a nice side effect that this makes it easier to test as well, but I'm sure there are ways to do it with other patterns.

spec/integration/missing_metadata_spec.rb
```diff
@@ -63,6 +77,153 @@
...
+              "link" => @base_path,
+              "content_id" => @content_id,
+            },
+          )
+          stub_publishing_api_has_item(@publishing_content_item)
+
+          MissingMetadata::Runner.new("content_store_document_type", search_config: SearchConfig.default_instance, logger: io).update
+
+          expect(a_request(:get, @get_content_url)).to have_been_made.once
+          expect(Indexer::AmendJob).to have_received(:perform_async).with(
```
koetsier: I think it is clearer not to mock this, but to just check if the right document has been updated with: 
```
          stub_publishing_api_has_expanded_links({ content_id: @content_id }, with_drafts: false)
          MissingMetadata::Runner.new("publishing_app", search_config: SearchConfig.default_instance, logger: io).update
          expect_document_is_in_rummager({"publishing_app" => "publisher"}, index: "government_test", id: @base_path)
```

koetsier: Actually, this shows there's an error in the code at https://github.com/alphagov/search-api/blob/653df3822efc08b879d9b095e666b07c02a9ceb9/lib/missing_metadata/fetcher.rb#L40. These arguments should not be symbols, but strings instead (i.e. ```"content_store_document_type" => response["document_type"]) 

Otherwise it throws: ```failed with the exception 'Job arguments to Indexer::AmendJob must be native JSON types```

spec/integration/missing_metadata_spec.rb
```diff
@@ -1,64 +1,227 @@
...
+          stub_request(:get, @get_content_url).with(query: hash_including({}))
+            .to_timeout
+            .then
+            .to_return(status: 200, body: @publishing_content_item.to_json, headers: {})
+
+          MissingMetadata::Runner.new("content_store_document_type", search_config: SearchConfig.default_instance, logger: io).update
+
+          expect(a_request(:get, @get_content_url)).to have_been_made.twice
+          expect(Kernel).to have_received(:sleep).with(1).exactly(1).times
+          expect(io).to have_received(:puts).with("Publishing API timed out getting content... retrying").exactly(:once)
```
koetsier: same

koetsier (changes requested):

spec/integration/missing_metadata_spec.rb
emmalowe: Oh, great catch - thank you! Let me fix that.

spec/integration/missing_metadata_spec.rb
emmalowe: Ah ok, makes sense. I'll change that.

emmalowe:
Thanks @koetsier - I've made the changes you suggested.

koetsier (approved):
looks good!

PR #3437, opened 2025-12-08
Add integration tests for supertype_updater

stringiest:
Adding integration tests for supertype_updater. Bringing test coverage of this class to 100% from 63.64%.

Re [Jira ticket 1598](https://gov-uk.atlassian.net/browse/SCH-1598) - increasing test coverage of Search API.

### DIFF
--- a/spec/integration/govuk_index/supertype_updater_spec.rb
+++ b/spec/integration/govuk_index/supertype_updater_spec.rb
@@ -1,0 +1,18 @@
+require "spec_helper"
+
+RSpec.describe GovukIndex::SupertypeUpdater do
+  let(:index) { "government_test" }
+
+  before do
+    allow(GovukDocumentTypes).to receive(:supertypes)
+      .with(document_type: "testgroup")
+      .and_return("supertype1" => "type1", "supertype2" => "type2")
+  end
+
+  it "calls the SupertypeJob for all documents in the index" do
+    commit_document(index, { link: "link/path", content_store_document_type: "testgroup" })
+
+    GovukIndex::SupertypeUpdater.update(index)
+    expect_document_is_in_rummager({ "link" => "link/path", "supertype1" => "type1", "supertype2" => "type2" }, index:)
+  end
+end


spec/unit/govuk_index/supertype_updater_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      data = { "_id" => "abc", "_type" => "generic-document", "_source" => { "custom" => "data", "document_type" => "stuff" } }
+      stub_client_for_scroll_enumerator(return_values: [[data], []])
+
+      described_class.update("govuk_test")
+
+      expect(GovukIndex::SupertypeJob).to have_received(:perform_async).with(%w[abc], "govuk_test")
+    end
+  end
+
+  def stub_client_for_scroll_enumerator(return_values:, search_body: nil, search_type: "query_then_fetch")
```
emmalowe: I find this factoring confusing, because it's more than just a stub and it feels odd that some checks are pulled out in this way. Should it be a separate test altogether, or not pulled out at all?

spec/unit/govuk_index/supertype_updater_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+RSpec.describe GovukIndex::SupertypeUpdater do
+  describe ".update" do
+    let(:supertype_job) { double(GovukIndex::SupertypeJob) }
+    let(:scroll_enumerator) { instance_double(ScrollEnumerator) }
+
+    before do
+      allow(GovukIndex::SupertypeJob).to receive(:perform_async).and_return(true)
+    end
+
+    it "updates supertypes from govuk_document_types gem" do
```
emmalowe: If I'm reading this correctly, this test doesn't actually check that the supertypes have been updated (we seem to be relying on the tests in https://github.com/alphagov/search-api/blob/main/spec/integration/govuk_index/supertype_job_spec.rb for that), which I think is okay, but the name is misleading. Maybe something like `it "calls the SupertypeJob" for all documents in the index`?

spec/unit/govuk_index/supertype_updater_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      data = { "_id" => "abc", "_type" => "generic-document", "_source" => { "custom" => "data", "document_type" => "stuff" } }
+      stub_client_for_scroll_enumerator(return_values: [[data], []])
+
+      described_class.update("govuk_test")
+
+      expect(GovukIndex::SupertypeJob).to have_received(:perform_async).with(%w[abc], "govuk_test")
```
emmalowe: Just to say 👍 for checking the arguments this is called with, because I didn't do that in one of my testing PRs and if I had, it would have revealed an issue with the code 😵 (Thanks to Jos for catching that for me [here](https://github.com/alphagov/search-api/pull/3436#discussion_r2601980418)!)

emmalowe (commented):
Generally looks okay to me. I've added some minor suggestions to help with clarity. You might want to get a review from someone who has more experience with Ruby though :)

spec/unit/govuk_index/supertype_updater_spec.rb
stringiest: That's a good point. I fell into the trap of naming the test similarly to another one that did a similar thing, but you're right that I'm only testing that it calls the SupertypeJob. Will change :)

spec/unit/govuk_index/supertype_updater_spec.rb
stringiest: I'll have a think and see if I can refactor. 

The scroll_enumerator is a separate service called within the updater, so it should be correct to stub it, but I agree this particular stub feels like overkill. It's cribbed from the [compare_enumerator_spec](https://github.com/alphagov/search-api/blob/main/spec/unit/indexer/compare_enumerator_spec.rb#L127) (not that that excuses it). 

Possibly it needs to be pulled out into a helper if it's going to be used across two specs, but I'll see if I can stub it more straightforwardly tomorrow first 🤞🏻

koetsier (commented):
I think I'd go for an integration test here, to avoid too much complex mocking, something like:
```
commit_document(index, { link: "link/path", content_store_document_type: "testgroup" })
GovukIndex::SupertypeUpdater.update(index)
expect_document_is_in_rummager({ "link" => "link/path", "supertype1" => "type1", "supertype2" => "type2" }, index:)
```
and then mock out the ```GovukDocumentTypes``` class like here: https://github.com/alphagov/search-api/blob/b8b36ada6670263d4e8e8951c24b3c590da6a5a0/spec/integration/govuk_index/supertype_job_spec.rb#L8

hannako:
@stringiest I salute your efforts to opt for a unit test, and for wrapping your head around all this. But I think having to mock a lot of the implementation of the classes that the updater interacts with will leave us with a brittle test.

- It looks like we have decent coverage of the ScrollEnumerator [here](https://github.com/alphagov/search-api/blob/main/spec/integration/scroll_enumerator_spec.rb)
- and of the GovukIndex::SupertypeJob [here](https://github.com/alphagov/search-api/blob/main/spec/integration/govuk_index/supertype_job_spec.rb).

I haven't checked the line coverage report, so I could be wrong, but it seems like what we're missing is testing the integration of all of these things working together. All the supertype_updater is really doing is passing the index to through to the supertype job. So the code Jos has suggested above as an integration test would give us that. 

If it's not straight forward we could perhaps all work on it together in a dev catchup!

koetsier (approved):
Looks good to me

hannako (approved):
Thanks for picking this up Lucy, looks great 🏅

PR #3440, opened 2025-12-10
Add a rake task to handle duplicate documents with the same content_id

koetsier:
Our Elasticsearch indices use the `link` property as the _id, uniquely
identifying a document. However, when a link changes, multiple
documents can end up sharing the same content_id, which can cause
issues.

Rake duplicates:find[index_name] finds all duplicates and displays them
Rake duplicates:remove[index_name] removes the duplicates

https://gov-uk.atlassian.net/browse/SCH-1654


spec/integration/search/duplicate_remover_spec.rb
```diff
@@ -0,0 +1,59 @@
+require "spec_helper"
+
+RSpec.describe Search::DuplicateRemover do
+  let(:index) { "government_test" }
+  let(:io) { StringIO.new }
+  let(:logger) { Logger.new(io) }
+  let(:duplicates) { Search::DuplicateFinder.new(index:).find_duplicates }
+  subject(:remover) { described_class.new(index:, logger:) }
+
+  it "logs an error if there is no updated_at attribute in the duplicates and skips" do
```
hannako: I tripped over the first three tests. Could you add some context blocks eg

```
Context "when all of the duplicated documents are missing an updated_at attribute"
  it "logs and error and does not delete the documents" do
...

Context "when some of the duplicated documents are missing an updated_at attribute"
  it "deletes the documents without an updated_at at field" do
  it "keeps the latest document" do

File: spec/integration/search/duplicate_remover_spec.rb [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+  it "keeps the latest document " do
+    commit_document(index, { link: "link/path_a", content_id: "same", updated_at: Time.utc(2024, 1, 1) })
+    commit_document(index, { link: "link/path_b", content_id: "same", updated_at: Time.utc(2023, 1, 1) })
+    remover.remove_duplicates(duplicates:)
+
+    expect_document_is_in_rummager({ "link" => "link/path_a" }, index:)
+    expect_document_missing_in_rummager(id: "link/path_b", index:)
+  end
+
+  context "when the duplicates have different content_ids" do
```
hannako: This really tripped me up! I read it as two documents being classed as duplicates even thought they have different content ids. Whereas I think what you're saying is it can handle more than one set of duplicates at a time? Ie.

context "when there are multiple sets of duplicated documents in the index" do

it "deletes all duplicates" do

lib/search/duplicate_remover.rb
```diff
@@ -0,0 +1,36 @@
...
+        end
+
+        _, *tail = sorted_docs
+        tail.each do |doc|
+          Services.elasticsearch.delete_by_query(
+            index: index,
+            body: { query: { term: { link: doc["link"] } } },
+          )
+          logger.info "Deleted duplicate document: #{doc['link']}"
+        end
```
hannako: I find this method really hard to follow - the sort function in particular is melting my brain! Could you refactor it into smaller methods and add tests for them?

lib/search/duplicate_finder.rb
```diff
@@ -0,0 +1,46 @@
...
+      )
+
+      buckets = response.dig("aggregations", "dupes", "buckets") || []
+      buckets.map do |bucket|
+        hits = bucket.dig("docs", "hits", "hits") || []
+        {
+          content_id: bucket["key"],
+          documents: hits.map { |hit| hit["_source"].slice("title", "link", "updated_at") },
+        }
+      end
```
hannako, line 43: I am not familiar with the format of the response from elasticsearch. Could you split this into two methods? One that fetches the aggregations, and another that transforms them with a test for each. That way I can see in the test what we get back, and what we're doing to it?

hannako (commented):
I'm sure this works great, but I'm finding it hard to review so I've made some refactoring and test coverage requests.

Thanks for digging into all this (and for adding a test for the rake tasks!)

lib/search/duplicate_finder.rb
koetsier: In the tests I'm trying to test the behaviour of this particular method, i.e. finding duplicates. If I'd split this up I would have to make both methods private, which makes me unable to test them anyway.

I think this is actually a good thing, because, if we ever upgrade Elasticsearch to a future version and ES implemented a change in the response, these tests would signal that by breaking. 

Let me know what you think!

spec/integration/search/duplicate_remover_spec.rb
koetsier: Good point. In hindsight this could have been a lot clearer 😊

spec/integration/search/duplicate_remover_spec.rb
koetsier: fixed

lib/search/duplicate_remover.rb
koetsier: I see what you mean yes. I split it up as much as possible and added some comments. I tried to refactor the logic but everything I tried ended up being more complicated. Let me know if this makes things clearer

lib/search/duplicate_remover.rb
```diff
@@ -0,0 +1,48 @@
...
+          next
+        end
+
+        sorted_docs = sort_by_updated_at(duplicate[:documents])
+
+        # Delete all duplicate documents except the most recent one.
+        sorted_docs.drop(1).each do |doc|
+          delete_document(doc["link"])
+        end
+      end
```
hannako, line 23: Thanks for doing this 👍

spec/integration/search/duplicate_finder_spec.rb
```diff
@@ -0,0 +1,50 @@
+require "spec_helper"
+
+RSpec.describe Search::DuplicateFinder do
+  let(:index) { "government_test" }
+  describe "there are no documents in Elasticsearch" do
+    it "returns an empty array" do
+      expect(Search::DuplicateFinder.new(index:).find_duplicates).to eq([])
```
hannako, line 7: nit pick

On line 15 you use to be_empty, here you check eq([]). Can we use the same?

hannako (approved):
LGTM

lib/search/duplicate_finder.rb
hannako: Makes sense 👍

lib/search/duplicate_remover.rb
hannako: Thanks - much clearer ❤️

PR #3446, opened 2025-12-17
Add more organisation fields [WHIT-1799]

ryanb-gds:
These are some fields that I missed first time around when adding all of the necessary fields to the govuk index to index organisations in #3411.

[JIRA](https://gov-uk.atlassian.net/browse/WHIT-1799)


hannako:
Sorry I missed this too! Thanks for fixing. Would you mind adding some more detail to the commit messages to join up the dots ie add links to the code in collections and Whitehall that makes these changes necessary. Links below to save you the time! 

https://github.com/alphagov/collections/blob/main/app/presenters/organisations_api_presenter.rb#L26-L52
https://www.gov.uk/api/organisations
https://github.com/alphagov/whitehall/blob/c1e79c8b2eaf764b4769199125fd20625bfa0ad7/app/models/organisation.rb#L168

hannako (approved):
Approving as code looks good, but please update the commit messages before merging. Thanks again!

PR #3448, opened 2025-12-18
Limit GA4 path length before indexing into Elasticsearch

koetsier:
We have been seeing a large number of errors from Sidekiq::GovukIndex::PageTrafficJob. The root cause is malformed data retrieved from GA4, which the job attempts to index into Elasticsearch.

Elasticsearch rejects these documents because the base_path field exceeds its 512-byte limit.

GOV.UK paths should never be this long, so values exceeding 512 bytes can be safely treated as invalid. This change filters out those paths before indexing.

https://gov-uk.atlassian.net/browse/SCH-1663

### DIFF
--- a/lib/analytics/ga4_import/page_view_consolidator.rb
+++ b/lib/analytics/ga4_import/page_view_consolidator.rb
@@ -1,6 +1,8 @@
 module Analytics
   module Ga4Import
     class PageViewConsolidator
+      MAX_PATH_LENGTH = 512
+
       def initialize(ga_data)
         @ga_data = ga_data
       end
@@ -8,6 +10,7 @@ module Analytics
       def consolidated_page_views
         ga_data
           .reject(&:excluded?)
+          .reject { |page_data| page_data.path.bytesize >= MAX_PATH_LENGTH }
           .group_by(&:normalised_path)
           .transform_values { |grouped_pages| grouped_pages.sum(&:page_views) }
           .sort_by { |_, views| -views }
--- a/spec/unit/analytics/ga4_import/page_view_consolidator_spec.rb
+++ b/spec/unit/analytics/ga4_import/page_view_consolidator_spec.rb
@@ -1,6 +1,8 @@
 require "spec_helper"
 
 RSpec.describe Analytics::Ga4Import::PageViewConsolidator do
+  let(:really_long_path) { "/#{'a'.b * (Analytics::Ga4Import::PageViewConsolidator::MAX_PATH_LENGTH - 1)}" }
+
   let(:ga_data) do
     [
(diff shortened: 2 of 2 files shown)


lib/analytics/ga4_import/page_view_consolidator.rb
```diff
@@ -1,13 +1,16 @@
 module Analytics
   module Ga4Import
     class PageViewConsolidator
+      MAX_PATH_LENGTH = 512
```
hannako, line 4: Is there are similar limit in publishing API so we know with certainty this is fine?

leenagupte: I've been having a look and can't see that publishing-api enforces a limit.

`base_path` is defined by [absolute_path](https://github.com/alphagov/publishing-api/blob/aed2a85a216790a3ba4c6707562f12c7aecc7de4/content_schemas/formats/shared/default_properties/frontend.jsonnet#L3), so my theory it's the [Ruby URI class](https://github.com/alphagov/publishing-api/blob/aed2a85a216790a3ba4c6707562f12c7aecc7de4/app/validators/absolute_path_validator.rb#L13) that sets the limit.

It would be good to get confirmation though.

koetsier: I could not find a hard limit in GOV.UK that enforces the <512 bytes limit and there are actually 4 live articles in there that exceed this limit. These are basepaths that are encoded to represent other alphabets such as cyrillic, which means they have far more bytes than characters.

However, Elasticsearch does not accept _id fields with length>512. (https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/mapping-id-field) which means they can't be indexed. This goes for the govuk and government indices as well

hannako: Could you provide the 4 live article urls pls?

I think we need to ship this, but we should let the relevant publishing team know so that there can be some validation added at the point of publishing.

hannako (approved):
Looks good - but as per comments we should let the relevant publishing team know.

lib/analytics/ga4_import/page_view_consolidator.rb
hannako: @koetsier Could you provide the 4 live article urls pls?

koetsier: Actually, we don't index non-english content so we can just ignore these

https://github.com/alphagov/search-api/blob/3698a34ac1c3f516c72d5cb87cb0bc95276b6c49/lib/govuk_index/publishing_event_message_handler.rb#L73

koetsier: Also, all four are unpublished.

PR #3457, opened 2026-01-05
Mark organisation document type as migrated to the govuk index [WHIT-1799]

ryanb-gds:
The data for all organisations has been successfully indexed within the `govuk` index, so it's time to switch over Search API to return it's results for organisations from there.

This also requires us to use the `govuk` index for entity expansion when presenting results. This change required making sure that the test organisation document for the search spec is present in the govuk index. However, doing this across the test suite caused a number of unrelated tests to fail, so we have overriding the test setup on a case-by-case basis.

JIRA: https://gov-uk.atlassian.net/browse/WHIT-1799

### DIFF
--- a/config/govuk_index/migrated_formats.yaml
+++ b/config/govuk_index/migrated_formats.yaml
@@ -354,6 +354,7 @@ migrated:
 - open_call_for_evidence
 - open_consultation
 - oral_statement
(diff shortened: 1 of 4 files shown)


hannako (approved):
🥳 :shipit:

PR #3466, opened 2026-01-09
[SCH-1679] Setup pact tests with gds-api-adapters

stringiest:
## What
This PR is the first step in adding contract tests with gds-api-adapters. See [Jira ticket SCH-1679](https://gov-uk.atlassian.net/browse/SCH-1679)

## Details
- The initial new pact is defined in this [PR](https://github.com/alphagov/gds-api-adapters/pull/1375) and has been published to [our pact broker](https://govuk-pact-broker-6991351eca05.herokuapp.com/).
- You can then verify this pact locally by running `env PACT_CONSUMER_VERSION=branch-search-api-pact-testing bundle exec rake pact:verify`.

## Next steps
- Add provider states for other Search API behaviours
- Add pact-verify.yml workflow - e.g. [collections/.github/workflows/pact-verify.yml](https://github.com/alphagov/collections/blob/main/.github/workflows/pact-verify.yml)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,14 @@ To replicate real data into your local development environment, use the [replica
 bundle exec rake
 ```
 
+### Pact tests
+
+Search API has a pact with [GDS API Adapters](https://github.com/alphagov/gds-api-adapters).
+
+See the [guidance on Pact testing](https://docs.publishing.service.gov.uk/manual/pact-testing.html) for how to run and modify the tests.
+
+Note that when you run the tests locally logs are output to `log/production.log` rather than to the console.
+
 ### Additional Docs
 
 - [Search API documentation](docs/using-the-search-api.md)
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -1,0 +1,63 @@
+ENV["PACT_DO_NOT_TRACK"] = "true"
+ENV["RACK_ENV"] = "test"
+
+require "pact/provider/rspec"
+require "webmock/rspec"
+require "gds_api"
+
+require_relative "../../spec/support/index_helpers"
+require_relative "../../spec/support/integration_test_helper"
+
+Pact.configure do |config|
+  config.reports_dir = "spec/reports/pacts"
+  config.include WebMock::API
+  config.include WebMock::Matchers
+  config.include IntegrationTestHelper
+end
+
+def pact_broker_base_url
+  "https://govuk-pact-broker-6991351eca05.herokuapp.com"
+end
+
+Pact.service_provider "Search API" do
+  include ERB::Util
+
+  honours_pact_with "GDS API Adapters" do
+    if ENV["PACT_URI"]
+      pact_uri(ENV["PACT_URI"])
+    else
+      path = "pacts/provider/#{url_encode(name)}/consumer/#{url_encode(consumer_name)}"
+      version_modifier = "versions/#{url_encode(ENV.fetch('PACT_CONSUMER_VERSION', 'branch-main'))}"
+      pact_uri("#{pact_broker_base_url}/#{path}/#{version_modifier}")
+    end
+  end
+end
+
+Pact.provider_states_for "GDS API Adapters" do
+  set_up do
+    WebMock.enable!
+    WebMock.reset!
+    IntegrationTestHelper.allow_elasticsearch_connection_to_test(pact_broker_base_url)
+    IndexHelpers.setup_test_indexes
+  end
+
+  tear_down do
+    IndexHelpers.clean_all
+    WebMock.disable!
+  end
+
+  provider_state "there are search results for universal credit" do
+    set_up do
+      document_params = {
+        "title" => "Universal credit",
+        "link" => "/universal-credit",
+      }
+      second_document_params = {
+        "title" => "Universal credit too",
+        "link" => "/universal-credit-too",
+      }
+      commit_document("government_test", document_params)
+      commit_document("government_test", second_document_params)
+    end
+  end
+end
--- a/spec/support/integration_spec_setup_helper.rb
+++ b/spec/support/integration_spec_setup_helper.rb
@@ -1,0 +1,43 @@
+require "spec/support/index_helpers"
+require "spec/support/integration_test_helper"
+
+module IntegrationSpecSetupHelper
+  def self.included(base)
+    base.around do |example|
+      IntegrationSpecSetupHelper.allow_connection_during_test do
+        example.run
+        # clean up after test run
+        IndexHelpers.all_index_names.each do |index|
+          clean_index_content(index)
+        end
+      end
+    end
+
+    # we only want to setup the before suite code once, in addition this this we only want to
+    # set it up when we are running integration tests (hence the reason we do it here).
+    @included ||= setup_before_suite
+  end
+
+  def self.setup_before_suite
+    RSpec.configure do |config|
+      config.before(:suite) do
+        IntegrationSpecSetupHelper.allow_connection_during_test do
+          IndexHelpers.setup_test_indexes
+        end
+      end
+
+      config.after(:suite) do
+        IntegrationSpecSetupHelper.allow_connection_during_test do
+          IndexHelpers.clean_all
+        end
+      end
+    end
+  end
+
+  def self.allow_connection_during_test
+    IntegrationTestHelper.allow_elasticsearch_connection_to_test
+    yield
+  ensure
+    IntegrationTestHelper.disable_net_connections
+  end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ tmp/*
 /data
 /log
 search_check_results.csv
+spec/reports/pacts
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,8 @@ gem "warden"
 gem "warden-oauth2"
 
 group :development, :test do
+  gem "pact", require: false
+  gem "pact_broker-client"
   gem "pry-byebug"
   gem "rubocop-govuk", require: false
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -34,6 +34,7 @@ GEM
       public_suffix (>= 2.0.2, < 8.0)
     amq-protocol (2.3.3)
     ast (2.4.3)
+    awesome_print (1.9.2)
     aws-eventstream (1.4.0)
     aws-partitions (1.1195.0)
     aws-sdk-core (3.240.0)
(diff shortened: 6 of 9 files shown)


spec/service_consumers/pact_helper.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+end
+
+Pact.provider_states_for "GDS API Adapters" do
+  set_up do
+    WebMock.enable!
+    WebMock.reset!
+    IntegrationSpecHelper.allow_elasticsearch_connection_to_test(pact_broker_base_url)
+    # need test environment index names
+    stub_index_names if ENV["RACK_ENV"] == "development"
```
stringiest: The methods in `IntegrationSpecHelper` only work if the indexes have the suffix `_test`, and this only happens if `ENV["RACK_ENV"] == "test"`. I tried passing in `ENV["RACK_ENV"] = "test"` at the top of this file, in the command line when running the pact tests and via `require_relative "../../spec/spec_helper"`, but the pact tests just hung when I tried to run them. I've therefore stubbed the index names here when `ENV["RACK_ENV"] == "development"` so that the pact tests can be run locally. Interested to know if there's a better way of doing this, though.

README.md
```diff
@@ -37,6 +37,12 @@ To replicate real data into your local development environment, use the [replica
 bundle exec rake
 ```
 
+### Pact tests
+
+Search API has a pact with [GDS API Adapters](https://github.com/alphagov/gds-api-adapters).
+
+See the [guidance on Pact testing](https://docs.publishing.service.gov.uk/manual/pact-broker.html) for how to run and modify the tests.
```
emmalowe: This seems to redirect so it's fine, but I think the link you want is:
https://docs.publishing.service.gov.uk/manual/pact-testing.html

Thanks for putting this in the README!

spec/service_consumers/pact_helper.rb
hannako: I think we should try and understand why the pact tests are being run with a  RACK_ENV of development, as that seems to be the source of the problem. Rather than this work around with the index names.

When I set the RACK_ENV to be test, then I am getting:

```
SPEC_OPTS='' /Users/jessica.jones/.rbenv/versions/3.2.6/bin/ruby -S pact verify --pact-helper /Users/jessica.jones/govuk/search-api/spec/service_consumers/pact_helper.rb
INFO: Reading pact at https://govuk-pact-broker-6991351eca05.herokuapp.com/pacts/provider/Search%20API/consumer/GDS%20API%20Adapters/versions/branch-search-api-pact-testing
not starting Prometheus metrics server: address already in use
```

Which I think is maybe why you have got `ENV["PACT_DO_NOT_TRACK"] = "true"` at the top of your pact_helper?

hannako: Update - I get the Prometheus metrics server error when I run tests with docker in other applications, so actually the issue is that setting RACK_ENV to `test` results in the pact tests not being run.

emmalowe: I've been trying to run these Pact tests locally (with Docker) too today. What I'm seeing looks like a difference in logging configurations where if you run in the dev environment, the logs get output to the command line, but if you run in test, they get saved to `log/production.log`. At least for me, it looks like this is all working fine (if a little unintuitively) locally.

Logging configuration is set here for reference: https://github.com/alphagov/search-api/blob/main/config.ru

emmalowe: So I think we can add `ENV["RACK_ENV"] = "test"` to the top of `pact_helper.rb` and then either change the logger output (which we might not want to do for CI), or write the expected behaviour in the README.

stringiest: Thanks both! I've rerun the tests locally, and confirmed that the test logs output to `log/production.log`, so have set `ENV["RACK_ENV"] = "test"` at the top of `pact_helper.rb` and put the expected logger output behaviour in the README.

README.md
stringiest: Oooh good spot thanks! Now updated :)

spec/support/integration_spec_setup_helper.rb
emmalowe, line 1: Is there any value in splitting this out anymore, now that we've sorted out the environment issue? Since you're requiring `integration_spec_helper` to the `pact_helper` anyway, having it split like this makes it more confusing to follow to me.

stringiest: I think we do still need to keep this separate, unfortunately. It's been separated out because `around` isn't a valid method for Pact, so it wasn't able to handle the setup and tear down instructions contained within `#self.included`. I did try splitting them into before and after hooks, but didn't get the right behaviour for the integration tests, and in pact the webmock kept killing the connection after the `honors_pact_with` but before the `provides_state`s were checked (though that could be my error!).

emmalowe: Oh I see, I think I'd misunderstood this. So to clarify:

- The integration tests use both `integration_spec_setup_helper.rb` and `integration_spec_helper.rb`.
- The pact tests use `integration_spec_helper.rb`.
- `integration_spec_setup_helper.rb` contains RSpec specific stuff, that isn't relevant for Pact and will cause the Pact tests to break, so they tests don't use it.
- The Pact specific set up / tear down (which is the Pact equivalent of the RSpec stuff in `integration_spec_setup_helper.rb`) is in `pact_helper.rb`

Is that right?

In which case, I think it's probably just a naming issue. I find `integration_spec_helper.rb` and `integration_spec_setup_helper.rb` confusing, since `integration_spec_helper.rb` is only for Pact but `integration_spec_setup_helper.rb` is for both.

emmalowe: I assume the "spec" in integration_spec_helper.rb is a reference to RSpec so maybe:

`integration_spec_setup_helper.rb` -> `integration_helper.rb`

if we're considering Pact tests as a type of integration test?

stringiest: Yes, that's right:
- `integration_spec_setup_helper.rb` is only used by the RSpec integration tests
- `integration_spec_helper.rb` is used by both the RSpec integration tests and the Pact tests

So the bullet points of your comment on Monday were correct, but I think you switched round where the files were used in the final paragraph.

I agree that the names of the files are confusing! 

I think your comment yesterday was about renaming `integration_spec_helper.rb` rather than renaming `integration_spec_setup_helper.rb` (reinforcing the need to rename one of them!)

How about `integration_test_helper.rb` rather than `integration_spec_helper.rb`? That would keep the information that the helper was for tests, but remove the RSpec-centric 'spec'? I think that Pact tests are a type of integration test, yes.

Then `integration_spec_setup_helper.rb` would keep its existing name.

Does that sound okay?

emmalowe: Ah yes, I did confuse that last paragraph - sorry! You're right:

- `integration_spec_setup_helper.rb` is only used by the RSpec integration tests
- `integration_spec_helper.rb` is used by both the RSpec integration tests and the Pact tests

My confusion at the end meant my suggestion didn't make sense, but I'm glad you got the gist of what I was trying to say!

Changing `integration_spec_helper.rb` to `integration_test_helper.rb` to make it more generic to cover both sounds good. You could change `integration_spec_setup_helper.rb` to `integration_spec_helper.rb` to make it shorter (I'm not sure there's much value in the "setup" part there) but I'd be fine with it either way.

stringiest: Great, thanks! Now changed. I've left 'setup' in the other file name, as I feel like it's setting up for the integration specs.

spec/service_consumers/pact_helper.rb
```diff
@@ -0,0 +1,63 @@
+ENV["PACT_DO_NOT_TRACK"] = "true"
+ENV["RACK_ENV"] = "test"
+
+require "pact/provider/rspec"
+require "webmock/rspec"
+require "gds_api"
+
+require_relative "../../spec/support/index_helpers"
+require_relative "../../spec/support/ion_test_helper"
```
emmalowe: Typo?

`ion_test_helper` -> `integration_test_helper`

Did you run the Pact test to make sure they're still working with the latest changes? I'd be surprised if this is working.

stringiest: Ooh good spot - thanks! 

I've also had to change back to using `government_test` to commit the documents to rather than `govuk_test`. I'd been thinking to use `govuk_test` as `government` index will hopefully be retired shortly, but I think we currently block posting documents directly to `govuk` index.

emmalowe: Ah nice! I was literally about to ask you what you just fixed to make the tests run, since I was getting a failure before your latest push.

I'll give this a quick once over now and then hopefully we're good to go 🤞

emmalowe:
Thanks @stringiest, these changes look great! The only thing is your first commit labelled as merge conflicts. It might be clearer if the description was about adding gems for the pact tests. I don't think there should be any merge conflict if you rebase from main.

stringiest:
> Thanks @stringiest, these changes look great! The only thing is your first commit labelled as merge conflicts. It might be clearer if the description was about adding gems for the pact tests. I don't think there should be any merge conflict if you rebase from main.

Thanks @emmalowe! I think the merge conflicts came from a rebase with main, but have renamed the commit to say it's adding the pact gem & dependencies. I'm pretty sure that's what the commit was called to start with, but there's been a lot of rebasing!

spec/service_consumers/pact_helper.rb
```diff
@@ -0,0 +1,63 @@
...
+
+Pact.service_provider "Search API" do
+  include ERB::Util
+
+  honours_pact_with "GDS API Adapters" do
+    if ENV["PACT_URI"]
+      pact_uri(ENV["PACT_URI"])
+    else
+      path = "pacts/provider/#{url_encode(name)}/consumer/#{url_encode(consumer_name)}"
+      version_modifier = "versions/#{url_encode(ENV.fetch('PACT_CONSUMER_VERSION', 'master'))}"
```
emmalowe: Should this be `branch-main` instead of `master`?

https://github.com/alphagov/places-manager/blob/main/spec/service_consumers/pact_helper.rb#L37
https://github.com/alphagov/email-alert-api/blob/main/spec/service_consumers/pact_helper.rb#L35

stringiest: Ooh yes! Thanks! Looks like there are some other repos still using `master`, but `branch-main` looks like the right thing. Now changed :)

emmalowe (approved):
We did it! 🎊 😂

PR #3470, opened 2026-01-14
Correct the non-indexable Whitehall formats

ryanb-gds:
At some point the formats Whitehall indexes have drifted apart from this configuration file. In order to eventually migrate these formats onto the 'govuk' index, we need to make sure the config file matches the format actually publishes.

Some of the formats have been removed as they are no longer indexed by Whitehall.

### DIFF
--- a/config/govuk_index/migrated_formats.yaml
+++ b/config/govuk_index/migrated_formats.yaml
@@ -431,19 +431,9 @@ non_indexable:
   - "/tour"
   - "/ukwelcomes"
 
-# whitehall formats
-#- contact # migrated
-- facet
-- field_of_operation
-#- finder # migrated
-- html_publication
-- imported
-- national_statistics_announcement
-- official_statistics_announcement
-- policy_area
-- services_and_information
-- take_part
-- topical_event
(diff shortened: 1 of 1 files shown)


config/govuk_index/migrated_formats.yaml
```diff
@@ -432,18 +432,8 @@ non_indexable:
...
-- html_publication
-- imported
-- national_statistics_announcement
-- official_statistics_announcement
-- policy_area
-- services_and_information
-- take_part
+- operational_field
+- statistics_announcement
+- policy_group
```
hannako: Would you mind adding some links to the relevant files in Whitehall to support the inclusion of the remaining 5 formats.

ryanb-gds: Would you prefer a comment in the file or in the commit message @hannako ?

hannako: I was assuming commit message, but comment in the file is a good idea. I don't mind really

hannako (approved):
Looks great - thanks for the additional comments. Lovely tidying up!

PR #3471, opened 2026-01-14
Retire the detailed index

koetsier:
The detailed index was previously used to store ‘detailed-guidance’ documents,
  which are now called ‘detailed-guide’. These document types are now handled
  by the govuk index and so the detailed index can be removed

- Remove references to the detailed index
- Add error messaging if the detailed index is being referenced
- Add a rake task to remove the index from Elasticsearch

Jira: https://gov-uk.atlassian.net/browse/SCH-1692

### DIFF
--- a/lib/rummager/app.rb
+++ b/lib/rummager/app.rb
@@ -65,9 +65,9 @@ class Rummager < Sinatra::Application
     halt(403, "You do not have permission to access this endpoint") unless u["permissions"].include? permission
   end
 
-  def prevent_access_to_govuk
-    if index_name == "govuk"
-      halt(403, "Actions to govuk index are not allowed via this endpoint, please use the message queue to update this index")
+  def prevent_access_to_govuk_and_detailed
+    if %w[govuk detailed].include?(index_name)
+      halt(403, "Actions to the govuk or detailed indices are not allowed via this endpoint.")
     end
   end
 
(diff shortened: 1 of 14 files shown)


lib/tasks/delete_detailed_index.rake
```diff
@@ -0,0 +1,13 @@
+require "rummager"
+desc "Delete the detailed index"
+task :delete_detailed_index do
+  client = Services.elasticsearch
+
+  indices = client.indices.get_alias(name: "detailed").keys
+
+  puts "Deleting indices: #{indices.join(', ')}"
+
+  client.indices.delete(index: indices)
```
hannako, line 10: Does the delete endpoint accept an array?

koetsier: Yes: https://rubydoc.info/gems/elasticsearch-api/Elasticsearch/API/Indices/Actions#delete-instance_method

lib/rummager/app.rb
```diff
@@ -65,9 +65,9 @@ def require_authentication(permission)
     halt(403, "You do not have permission to access this endpoint") unless u["permissions"].include? permission
   end
 
-  def prevent_access_to_govuk
-    if index_name == "govuk"
-      halt(403, "Actions to govuk index are not allowed via this endpoint, please use the message queue to update this index")
+  def prevent_access_to_govuk_and_detailed
+    if %w[govuk detailed].include?(index_name)
```
hannako, line 69: Maybe we should have two methods: `prevent_access_to_legacy_indices` - which we could extend when we retire other indices and raises a relevant message and 403. And the original `prevent_access_to_govuk`?

koetsier: Could do, but this is a lot simpler. And this is all supposed to be removed anyway. I can do it if you feel strongly? Let me know

hannako (approved):

PR #3475, opened 2026-01-20
Ensure rake tasks are included in the coverage report

koetsier:
Ensure all relevant files are included in the coverage report

  - track all ruby and rake files
  - load all rake tasks before the test suite startsa
  - remove 'load' commands from rake spec files to avoid double loading

There are 15 rake files and 160 ruby files.
```
find ./lib -type f -name "*.rake" | wc -l   => 14
find ./lib -type f -name "*.rb" | wc -l    => 156
```

The coverage report now correctly shows 170 files: 

image

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -2,7 +2,13 @@ ENV["RACK_ENV"] = "test"
 require "pry"
 
 require "simplecov"
-SimpleCov.start { add_filter "/spec/" }
+SimpleCov.start do
+  add_filter "/spec/"
+  add_filter "/config/"
+  add_filter "/env.rb"
+  track_files "{lib}/**/*.rb"
+  track_files "{lib}/**/*.rake"
+end
 
 $LOAD_PATH << File.expand_path("..", __dir__)
 $LOAD_PATH << File.expand_path("../lib", __dir__)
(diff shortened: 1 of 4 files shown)


hannako (approved):

emmalowe:
I haven't figured out why yet, but it seems like some Ruby files are not appearing in the coverage report when I run it locally. For example: https://github.com/alphagov/search-api/blob/main/lib/analytics/overall_ctr.rb

spec/spec_helper.rb
```diff
@@ -2,7 +2,11 @@
 require "pry"
 
 require "simplecov"
-SimpleCov.start { add_filter "/spec/" }
+SimpleCov.start do
```
emmalowe, line 5: Removing this change and adding the following in the RSpec configuration seemed to do the trick to add rake tasks and the files they require to the coverage report:

```
  config.before :suite do
    Rake.application = Rake::Application.new
    Dir.glob("lib/tasks/**/*.rake").each { |r| load r }
  end
```

This actually seems to improve our coverage, such all the definition lines then get included in the coverage, which they weren't with this tracking methods.

emmalowe (changes requested):
As per my other comments, we need to make sure the files that are required by the rake tasks are also included in the coverage report/stats.

spec/spec_helper.rb
```diff
@@ -2,7 +2,14 @@
...
+  add_filter "/spec/"
+  add_filter "/config/"
+  add_filter "/rakefile"
```
emmalowe: Do we need this line with `rakefile`? I can't see that it's doing anything.

Also, would it need to be capitalised?

koetsier: removed it. I thought it was necessary but turns out not to be

emmalowe (approved):
Looks good to me - thanks!

PR #3480, opened 2026-01-21
Remove delete_detailed_index rake task

koetsier:
This has been run, the detailed index has been removed, so this task is no longer needed

https://gov-uk.atlassian.net/browse/SCH-1692


hannako (approved):
🥳

PR #3483, opened 2026-01-27
Map remaining Whitehall document types to Elasticsearch document types

ryanb-gds:
All Whitehall documents map to Search API editions.

In several cases the document type Whitehall sends to Publishing API does not match the model name in the Whitehall code, so I have added the names of the Whitehall models in a comment.


koetsier (approved):
lgtm

PR #3484, opened 2026-01-27
Add topical event fields

ryanb-gds:
Enable the govuk indexing process to extract the slug, start and end date fields from the Publishing API payload for topical events. We reuse the existing Elasticsearch field definitions instead of declaring new fields specifically for topical events.

### DIFF
--- a/spec/unit/govuk_index/presenters/details_presenter_spec.rb
+++ b/spec/unit/govuk_index/presenters/details_presenter_spec.rb
@@ -310,4 +310,37 @@ RSpec.describe GovukIndex::DetailsPresenter do
       expect(presented_details.closed_at).to eq("2024-06-01T00:00:00.000+01:00")
     end
   end
+
+  context "any format" do
+    let(:format) { "any" }
+    context "it has a start and an end date" do
+      let(:details) do
+        {
+          "start_date" => "2024-06-01T00:00:00.000+01:00",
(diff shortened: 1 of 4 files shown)


spec/unit/govuk_index/presenters/details_presenter_spec.rb
```diff
@@ -310,4 +310,37 @@
...
+  context "any format" do
+    let(:format) { "any" }
+    context "it has a start and an end date" do
+      let(:details) do
+        {
+          "start_date" => "2024-06-01T00:00:00.000+01:00",
+          "end_date" => "2024-06-02T00:00:00.000+01:00",
+          "opening_date" => "2024-06-03T00:00:00.000+01:00",
+          "closing_date" => "2024-06-04T00:00:00.000+01:00",
+        }
```
hannako, line 323: How did you come to this knowledge, ie that start date should be selected if both start date and opening date are available?

ryanb-gds: I've made an assumption, based on the naming similarity. I introduced the mapping from opening and closing date to start and end date in https://github.com/alphagov/search-api/pull/3162, which was done so that I didn't break the frontend. But I don't think there are any other content types using start and end date this way, and I think in the unlikely event both should appear it would make more sense that start and end date should take precedence in the mapping.

hannako (approved):

PR #3485, opened 2026-01-27
Corrections to document mapping config for some Whitehall document types

ryanb-gds:
I previously believed that the keys for these configuration values should match the `document_type` attribute in the Publishing API payload, but in fact Search API requires that they should match the format configured in the `migrated_formats.yml` configuration for the `govuk` index.


ryanb-gds:
Closing this as it doesn't solve the problem. https://github.com/alphagov/search-api/pull/3486 should do the trick.

PR #3486, opened 2026-01-29
Map Whitehall document types to Search API format

ryanb-gds:
The `document_type` value in the Publishing API payload differs from the Search API format. Mapping the two via the common fields presenter enables indexing of these types in the GOVUK index.

_This isn't currently covered by a unit test and I haven't added one because the format method only does a trivial hash lookup, but happy to add one if desired._ - edit: there is a test, I missed it


hannako:
It's covered by this test isn't it? https://github.com/alphagov/search-api/blob/60d35f13bf401ce02fe52c6bc5ffdb6f9a5e68da/spec/unit/govuk_index/presenters/common_fields_presenter_spec.rb#L49

hannako (approved):

PR #3488, opened 2026-01-30
Sch 1737 search api rake tasks delete analytics rake task and related code

emmalowe:
Remove rake task and related code to upload elasticsearch content to Universal Analytics (which is no longer used).

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-1737


stringiest (commented):
This looks good to me! 

Can I check I've understood correctly that these bits of code just uploaded a list of indexed pages to UA - i.e. no popularity data etc, just the pages themselves?

Follow-up question if the above is right: we don't seem to have an equivalent job in search-api-v2. I can see where the documents are sync'd from publishing-api to vertex, but not the pages being uploaded to GA. Is this no longer necessary? (Or does it originate outside of the repo?)

emmalowe:
> Can I check I've understood correctly that these bits of code just uploaded a list of indexed pages to UA - i.e. no popularity data etc, just the pages themselves?

That's my understanding, except that I don't think it's the whole page - it appears to be a subset of fields as shown here:
https://github.com/alphagov/search-api/blob/main/lib/analytics/extract.rb#L32

> Follow-up question if the above is right: we don't seem to have an equivalent job in search-api-v2. I can see where the documents are sync'd from publishing-api to vertex, but not the pages being uploaded to GA. Is this no longer necessary? (Or does it originate outside of the repo?)

I don't know enough background to give you a good answer on this. My very high level understanding is that GA4 (our "new" way of doing analytics) has a different data pipeline so this is no longer needed, but @tarastockford might know more details. I'm not sure why we would have needed to upload search documents specifically to analytics, given that search documents are just pages on the website, so I would have thought those would be tracked anyway.

stringiest:
> I'm not sure why we would have needed to upload search documents specifically to analytics, given that search documents are just pages on the website, so I would have thought those would be tracked anyway.

Yes, that's the part that was confusing me!

stringiest (approved):
Thanks for confirming my understanding! Happy to approve now that I know what the code was doing :)

emmalowe:
@stringiest FYI I found some info [here](https://github.com/alphagov/ai-ruby-test-app-2023/blob/4cd7f3ec2b8938de115519c010fc46c82047d5a4/dev-docs/source/manual/alerts/enhanced-ecommerce.html.md?plain=1#L23) on why this task was needed. In summary, we wanted to supplement the page data that was tracked by default by UA.

PR #3489, opened 2026-02-02
Add specs for indices.rake

koetsier:
So we increase our test coverage

image

https://gov-uk.atlassian.net/browse/SCH-1743


spec/unit/tasks/indices_spec.rb
```diff
@@ -0,0 +1,401 @@
+require "spec_helper"
+require "rake"
+load "tasks/indices.rake"
+
+RSpec.describe "indices" do
+  let(:elasticsearch_client) { double("Elasticsearch::Client") }
+  let(:indices_client) { double("Elasticsearch::API::Indices::IndicesClient") }
+  let(:base_name) { index_names.first }
+  let(:index_name) { base_name }
+  let(:index_names) { SearchConfig.all_index_names }
```
hannako, line 19: nitpick

Could you reorder so index_names is defined first?

spec/unit/tasks/indices_spec.rb
```diff
@@ -0,0 +1,401 @@
...
+    ) { example.run }
+  end
+
+  describe "list_indices" do
+    let(:task_name) { "search:list_indices" }
+    let(:index1) { "#{base_name}-2026-01-27t08-24-32z-aaaaaaaa-1111-1111-1111-aaaaaaaaaaaa" }
+    let(:index2) { "#{base_name}-2026-01-27t08-24-32z-bbbbbbbb-2222-2222-2222-bbbbbbbbbbbb" }
+    let(:indices) { { index1 => {}, index2 => {} } }
+    let(:aliases) do
+      { index1 => { "aliases" => { base_name => {} } } }
```
hannako, line 39: So if I'm understanding this correctly, it's this line 

```
let(:aliases) do
      { index1 => { "aliases" => { base_name => {} } } }
end
```

that's determining that index2 is a closed index, and therefore not returned by the first test?

But the rake task itself isn't calling the indices_client directly, it's using [SearchIndices:Index's real_name ](https://github.com/alphagov/search-api/blob/d7312a6297b8b6adf84b5354a80e8ae103d1ee94/lib/tasks/indices.rake#L12) method to do that, defined [here](https://github.com/alphagov/search-api/blob/d7312a6297b8b6adf84b5354a80e8ae103d1ee94/lib/index.rb#L39). 

We should probably add a test for the `real_name` method to the [index_spec](https://github.com/alphagov/search-api/blob/main/spec/unit/index_spec.rb) unit test.

If we did that, so we felt confident that if the elasticsearch api change a test would break, could we stub the real_name response here instead? Would it be easier to have some setup methods defined in a spec helper?

spec/unit/tasks/indices_spec.rb
```diff
@@ -0,0 +1,401 @@
...
+  end
+
+  describe "create_all_indices" do
+    let(:task_name) { "search:create_all_indices" }
+    before do
+      allow(indices_client).to receive(:get_alias).and_return({})
+      allow(indices_client).to receive(:update_aliases).and_return({})
+      allow(indices_client).to receive(:create).with(hash_including({}))
+    end
+    it "creates all indices" do
```
hannako, line 70: Similar to above, this rake task using the create_indices method on index_groups, for which we already have [a test](https://github.com/alphagov/search-api/blob/main/spec/unit/index_group_spec.rb#L26). So can we simplify this rake task test to check that the `create_indices` method is being called?

hannako (commented):
I've not got to the end, but I think all of my comments are going to be similar. In that we're sort of using indices_spec like an integration test, but there are existing unit test files we could add to, and in some files there are already unit tests that cover much of the functionality. 

In an ideal world this file would just be checking that the underlying classes the rake task calls are sending the right messages. Have you gone for the integration test style just for speed or am I missing something?

spec/unit/tasks/indices_spec.rb
koetsier: This line determines that index1 has an alias 'base_name'

The Search-api uses the presence of an alias to determine if an index is actively used by the search-api.

Something needs to be mocked to test this. Mocking 'real_name' would work too but I feel that gets into the implementation a bit too much.

You're probably right that real_name needs a spec though

koetsier: I moved the stubs around a bit and moved the 'allows' to the creation of the indices client, so it is a bit clearer what we're stubbing/expecting

spec/unit/tasks/indices_spec.rb
koetsier: I feel this is maybe the most implementation-independent way though, whatever the implementation, the elasticsearch index create method should be called. Let me know what you think

spec/unit/tasks/indices_spec.rb
hannako: Nice, thanks! That does help.

hannako (approved):

PR #3491, opened 2026-02-04
Sitemap request validation

hannako:
We get relatively frequent vulnerability scans to the sitemaps endpoint which is causing [sentry](https://govuk.sentry.io/issues/?environment=production&project=1461905&query=is%3Aunresolved%20sitemap&referrer=issue-list&sort=freq&statsPeriod=90d) noise.

🚧 WIP 

These are the errors:

:x: Sinatra::BadRequest 
:x: Aws::S3::Errors::BadRequest
:x: NoMethodError
:x: ArgumentError
:x: Aws::S3::Errors::NotFound

### DIFF
--- a/lib/rummager/app.rb
+++ b/lib/rummager/app.rb
@@ -317,6 +317,12 @@ class Rummager < Sinatra::Application
     serve_from_s3(sitemap)
   end
 
+  post "/sitemaps/*" do
+    headers = { "Allow" => "GET" }
+    body = { message: "Method Not Allowed: Use GET to access the sitemap." }.to_json
+    halt(405, headers, body)
+  end
+
   def serve_from_s3(key)
     o = Services.s3_client.get_object(bucket: ENV["AWS_S3_SITEMAPS_BUCKET_NAME"], key:)
 
--- a/lib/rummager/app.rb
+++ b/lib/rummager/app.rb
@@ -210,6 +210,10 @@ class Rummager < Sinatra::Application
     serve_from_s3("#{sitemap}.xml")
   end
 
+  post %r{/sitemap} do
+    halt 404
+  end
+
   def serve_from_s3(key)
     o = Services.s3_client.get_object(bucket: ENV["AWS_S3_SITEMAPS_BUCKET_NAME"], key:)
 
--- a/lib/rummager/app.rb
+++ b/lib/rummager/app.rb
@@ -78,7 +78,7 @@ class Rummager < Sinatra::Application
   end
 
   def xml_only(request_format)
-    halt 404 if request_format != "xml"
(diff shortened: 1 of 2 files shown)


lib/rummager/app.rb
```diff
@@ -317,6 +317,12 @@ def get_type_from_request_body(request_body)
     serve_from_s3(sitemap)
   end
 
+  post "/sitemaps/*" do
```
koetsier, line 320: I'm not sure this will fix it. The error in Sentry is ```Invalid multipart/form-data: Rack::Multipart::EmptyContentError (Sinatra::BadRequest)``` I think this is a bot sending a broken http request to the sitemaps endpoint

It's probably better to catch the Sinatra::BadRequest error like:
```
error Sinatra::BadRequest do
  status 400
  "Bad request"
end
```

koetsier (changes requested):

lib/rummager/app.rb
hannako: Maybe we need both? If a bot sends a valid body it would hit the `post` route right?

koetsier: I'm actually not sure what would happen. Maybe it would return a 404? We could try it out using curl

emmalowe:
Marking this as draft, since it's not being actively worked on.

There is a Jira ticket that captures this work, so we can reprioritise later: https://gov-uk.atlassian.net/browse/SCH-2004

lib/rummager/app.rb
```diff
@@ -195,12 +199,19 @@ def json_only
...
-  get "/sitemap.xml" do
+  get %r{/sitemap(?:\.(\w+))?} do |request_format|
+    xml_only(request_format)
     serve_from_s3("sitemap.xml")
   end
 
-  get "/sitemaps/:sitemap" do |sitemap|
-    serve_from_s3(sitemap)
+  # matches /sitemaps/sitemap_1.xml and /sitemaps/sitemap_1
+  get %r{/sitemaps/(\w+)(?:\.(\w+))?} do |sitemap, request_format|
```
hannako, line 208: There's probably not much point doing this, because it won't match the majority of the spammy requests.

We could either:
- whackamole capture the various errors on the pr description
- update this regex to capture everything after the /sitemaps or /sitemap and pass it through a custom validator so we only send valid keys to aws rather than sending junk and rescuing the error
- we could add a before filter to this controller to check the user agent and filter out the bots

emmalowe: The [sitemap file names are hardcoded](https://github.com/alphagov/search-api/blob/main/lib/sitemap/generator.rb#L24), so is there any harm in using a tighter validation for this? That seems like a pragmatic solution to me.

PR #3501, opened 2026-02-06
Sch 1738 search api rake tasks delete unused and add tests for debug.rake

emmalowe:
First half of work to add test coverage for the debug rake tasks. This PR covers:

- show_old_index_link
- show_govuk_link
- fetch_best_bets

It takes code coverage from 91.33% to 91.59%.

### DIFF
--- a/spec/integration/helpers/best_bet_helpers.rb
+++ b/spec/integration/helpers/best_bet_helpers.rb
@@ -1,0 +1,46 @@
+require "spec_helper"
+
+module BestBetIntegrationTestHelpers
+  def get_links(path)
+    get(path)
+    parsed_response["results"].map { |result| result["link"] }
+  end
+
+  def add_best_bet(args)
+    payload = build_sample_bet_hash(
+      query: args[:query],
+      type: args[:type],
+      best_bets: [args.slice(:link, :position)],
+      worst_bets: [],
+    )
+
+    post "/metasearch_test/documents", payload.to_json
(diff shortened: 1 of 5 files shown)


stringiest (commented):
This looks good! Thanks for the clean commits extracting the reusable helpers. 

Do you also need tests for the unhappy paths? E.g. when there are no matching best/worst bets, or when the link searched for doesn't return a document.

emmalowe:
> Do you also need tests for the unhappy paths? E.g. when there are no matching best/worst bets, or when the link searched for doesn't return a document.

Maybe... I'm on the fence about it. I decided not to do it initially because the unhappy paths and other happy paths seem quite well tested by unit tests sitting a layer down from the rake tasks, and I didn't want to repeat these tests. For example:

- Best bets: https://github.com/alphagov/search-api/blob/main/spec/unit/search/best_bets_checker_spec.rb
- Get document by link: https://github.com/alphagov/search-api/blob/main/spec/unit/indexer/change_notification_processor_spec.rb#L27

I thought it was better to focus on testing what is just the functionality of the rake tasks, which is essentially printing stuff. But I can see an argument for making sure the rake task doesn't error when there's nothing to print.

I suppose it won't take much to add the unhappy paths, so I'll do that.

koetsier (approved):
Really nice tests 👍

stringiest (approved):
Looks good to me! 🎉

PR #3505, opened 2026-02-10
Add tests for synonyms rake task

emmalowe:
Adds integration tests for synonyms rake task.

Takes code coverage from 91.57% to 92.07%.

### DIFF
--- a/lib/tasks/debug.rake
+++ b/lib/tasks/debug.rake
@@ -25,8 +25,9 @@ namespace :debug do
   end
 
   desc "New synonyms test"
-  task :show_new_synonyms, [:query] do |_, args|
-    model = Debug::Synonyms::Analyzer.new
+  task :show_new_synonyms, [:query, :index_name] do |_, args|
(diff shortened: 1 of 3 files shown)


lib/tasks/debug.rake
```diff
@@ -26,7 +26,8 @@ namespace :debug do
 
   desc "New synonyms test"
   task :show_new_synonyms, [:query] do |_, args|
-    model = Debug::Synonyms::Analyzer.new
```
hannako, line 29: I think we discussed being able to pass in an index as a param to the rake task? Which would just need a tweak here eg
```
task :show_new_synonyms, [:query, :index_name] do |_, args|
index = args.index || SearchConfig.govuk_index_name
...
```

emmalowe: Ah yes, I forgot about that. Thanks for the reminder! Now done.

hannako (approved):
Looks good!

PR #3507, opened 2026-02-11
Add specs for the delete:by_format Rake task

koetsier:
So we increase the total test coverage

https://gov-uk.atlassian.net/browse/SCH-1739


hannako (approved):
Looks good!

PR #3510, opened 2026-02-12
[SCH-1679] Add pact-verify workflow to ci

stringiest:
Now that we've added the necessary pact tests and tested them locally against the published pacts we can add the pact tests to our CI pipeline. 

There is a [companion PR to add the search-api pact tests to the gds-api-adapters ci workflow](https://github.com/alphagov/gds-api-adapters/pull/1382), but this PR will need to be merged first.

### DIFF
--- a/.github/workflows/pact-verify.yml
+++ b/.github/workflows/pact-verify.yml
@@ -1,0 +1,68 @@
+# Pact verify workflow
+#
+# This workflow asserts that Pact contract tests are valid against this
+# codebase. It is trigged when changes are made to this project and it
+# is explicitly called by GDS API Adapters when changes are made there.
+
+name: Run Pact tests
+
+on:
+  workflow_call:
+    inputs:
+      # The commit / tag of this repo to test against
+      ref:
+        required: false
+        type: string
+        default: main
+      # A GitHub Action artifact which contains the pact definition files
+      # Search API calls this action to test new pacts against this
+      # workflow
+      pact_artifact:
+        required: false
+        type: string
+      # When using an artifact this is the file path to the pact that is verified
+      # against
+      pact_artifact_file_to_verify:
+        required: false
+        type: string
+        default: gds_api_adapters-search_api.json
+      # Which version of the pacts to use from the Pact Broker service
+      # This option will be ignored if pact_artifact is set
+      pact_consumer_version:
+        required: false
+        type: string
+        default: branch-main
+
+jobs:
+  pact_verify:
+    name: Verify pact tests
+    runs-on: ubuntu-latest
+    env:
+      RAILS_ENV: test
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/pact-verify.yml
```diff
@@ -0,0 +1,68 @@
...
+on:
+  workflow_call:
+    inputs:
+      # The commit / tag of this repo to test against
+      ref:
+        required: false
+        type: string
+        default: main
+      # A GitHub Action artifact which contains the pact definition files
+      # Publishing API calls this action to test new pacts against this
```
koetsier: You probably meant search-api?

.github/workflows/pact-verify.yml
```diff
@@ -0,0 +1,68 @@
...
+
+jobs:
+  pact_verify:
+    name: Verify pact tests
+    runs-on: ubuntu-latest
+    env:
+      RAILS_ENV: test
+    steps:
+      - name: Setup Elasticsearch
+        uses: alphagov/govuk-infrastructure/.github/actions/setup-elasticsearch@main
```
koetsier, line 44: this action uses the old version of elasticsearch. we use 6.8.x now.

Maybe this is better as a service?

like: 
```
   services:
      elasticsearch:
        image: docker.elastic.co/elasticsearch/elasticsearch:6.8.23
        ports:
          - 9200:9200
        env:
          discovery.type: single-node
          bootstrap.memory_lock: "true"
          ES_JAVA_OPTS: -Xms512m -Xmx512m
          xpack.security.enabled: "false"
        options: >-
          --ulimit nofile=65536:65536
          --ulimit memlock=-1:-1
          --health-cmd="curl -f http://localhost:9200/_cluster/health || exit 1"
          --health-interval=10s
          --health-timeout=5s
          --health-retries=10
```
that also avoids the 'wait' in the infrastructure script

koetsier (changes requested):
looks good I think, one comment, let me know what you think

.github/workflows/pact-verify.yml
stringiest: Whoops! 🙈 Thanks! :)

.github/workflows/pact-verify.yml
stringiest: Interesting! It looks like it's only us that use that action. Maybe it's time to update the action itself.

The service above looks good, but I'm afraid I don't understand exactly where it would go in the GHA workflows. Perhaps you meant that the above would update the action in govuk-infrastructure! Will DM you 😃

koetsier: We'll leave it for now then, we can make a change in alphagov/govuk-infrastructure/.github/actions/setup-elasticsearch@mainat a later date

koetsier (approved):

PR #3511, opened 2026-02-12
Remove relevancy rake task

koetsier:
The Relevancy Rake task is no longer used and so we can remove it and all dependencies

https://gov-uk.atlassian.net/browse/SCH-1769

Related PR removing unused env vars: https://github.com/alphagov/govuk-helm-charts/pull/4074


docs/search-quality-metrics.md
```diff
@@ -1,159 +0,0 @@
...
-2. Sign in to your HMRC account
-3. Harry Potter
-
-The first situation is good and the second is bad. But what we really
-want is a metric to say *how bad* the results are.
-
-Rank evaluation provides us with a metric which tells us how good or bad
-the results are for a set of queries that we have already manually ranked.
-
-So we have a rake task `debug:rank_evaluation` that does this.
```
hannako, line 51: I'm wondering if these docs are actually talking about [debug:ranking_evaluation](https://github.com/alphagov/search-api/blob/main/lib/tasks/debug.rake#L61) here?

@emmalowe and I were chatting about that task and thought it was worth keeping. If we get a new judgement list from Tara it would be useful for the elasticsearch upgrade OKR - to run as a sense check after upgrading. 

So can you leave those bits in this file.

emmalowe: I think you're right @hannako . I had in mind that I was going to update this doc as part of my PR on the rank evaluation rake task.

koetsier: ok I reverted this 👍

emmalowe:
@koetsier do you know how the authentication works for the `page_traffic` job that pulls data from GA4? I can't see how `GoogleAnalyticsExportCredentials` would be connected to the `analytics/ga4_import` stuff, but I also can't see how else the `page_traffic` would be authenticated, which makes me a bit nervous to delete the GA credentials.

lib/analytics/google_analytics_export_credentials.rb
emmalowe, line 1: If this is being taken out, with `reporting_requester.rb` I _think_ we can remove the "googleauth" gem in the Gemfile - is that right?

lib/tasks/relevancy.rake
emmalowe: Should commits be self-contained? In your first commit there's still quite a few references to deleted classes in this file. You could delete the whole rake file in this commit - that seems like a logical part of step 1 to me.

koetsier: I tried to give a logical order I followed to find all dependent files that are unused by other files. I don't know if they should be self-contained

emmalowe: I think they should be, but I don't know how strongly we want to enforce that:
https://gds-way.digital.cabinet-office.gov.uk/standards/source-code/working-with-git.html#commits

koetsier:
> @koetsier do you know how the authentication works for the `page_traffic` job that pulls data from GA4? I can't see how `GoogleAnalyticsExportCredentials` would be connected to the `analytics/ga4_import` stuff, but I also can't see how else the `page_traffic` would be authenticated, which makes me a bit nervous to delete the GA credentials.

This seems to work with environment variables set in the helm charts (https://github.com/alphagov/govuk-helm-charts/blob/96d3f74a4ad3a90e5baea6cf601d62b185cbe82c/charts/app-config/values-integration.yaml#L3027) . I initially thought they could be removed as well, but they seem to be used to get the data from GA4

emmalowe:
> This seems to work with environment variables set in the helm charts (https://github.com/alphagov/govuk-helm-charts/blob/96d3f74a4ad3a90e5baea6cf601d62b185cbe82c/charts/app-config/values-integration.yaml#L3027) . I initially thought they could be removed as well, but they seem to be used to get the data from GA4

As we chatted about, it seems like these are being used under the hood by `googleauth` when we use the google analytics data client. 

Do you think we should add a note for this in the README or somewhere, so it's obvious we need these?

I think we could still delete `GOOGLE_ANALYTICS_GOVUK_VIEW_ID`.

lib/analytics/google_analytics_export_credentials.rb
koetsier: I removed it because that file is the only one directly requiring it. But it doesn't really impact anything because it is still a dependency of the `google-cloud-bigquery` gem for instance

emmalowe (approved):
Looks good to me - thanks for adding the docs on the env vars.

PR #3513, opened 2026-02-12
Add tests for rank eval rake task

emmalowe:
Tidy up and add tests for rank evaluation rake task. There's a few things going on in here. I'd recommend reviewing the commits in order.

Takes test coverage from 92.07% to 93.28% 🎉 

I've tested the rake task locally, by running it with a judgements csv file in the repo's temp folder. I've also tested the branch on integration, connecting to the AWS bucket to get the judgement csv from there.

### DIFF
--- a/docs/search-quality-metrics.md
+++ b/docs/search-quality-metrics.md
@@ -1,28 +1,8 @@
 # Search Quality Metrics
 
-We assess the quality of Search API search results with two metrics:
-Click Through Rate (an online metric) and Normalised Discounted Cumulative Gain
-([nDCG][]).
-
-## Online metrics
-
-We monitor changes to search relevance over time in a [Search Relevancy Dashboard][].
-We call the metrics displayed there *online metrics*.
-
-Our main online metric is Click Through Rate on the top results
-(top 1, 3, and 5). This isn't a great metric, since users might
-click on something that isn't what they were looking for. But this
-serves our needs in the absence of a more sophisticated way of
-measuring user success following a search.
-
-## Offline metrics
-
-Our main offline metric is nDCG.
-
 We use Elasticsearch's [Ranking Evaluation API](ranking_evaluation_api)
-to assess the quality of results retrieved from Elasticsearch prior
-to re-ranking.
-The API enables us to score how well search-api ranks results for a given query.
+to assess the quality of results retrieved from Elasticsearch prior to re-ranking. 
+The API enables us to score how well search-api ranks results by relevancy for a given query.
 
 ### What is rank evaluation?
 
@@ -71,8 +51,7 @@ returning better results than the query for `harry potter`.
 Moreover, if the score for `harry potter` was `0.2392687105963024` before
 we made a change, then that means we've made a good change for that query.
 
-This can be measured over time, and it is! See the section on 'What do we do
-with the query scores?' below.
+This could be measured over time, though we don't do this systematically at the moment.
 
 ## How do we compute a score for a query?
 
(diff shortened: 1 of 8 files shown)


docs/search-quality-metrics.md
```diff
@@ -124,36 +103,13 @@ in time.
...
-
-The implicit relevance judgements are derived from user click data.
-
-The explicit judgements are provided by experts across government via
-the search relevance tool: https://github.com/alphagov/govuk-search-relevance-tool.
-
-Relevance judgements are uploaded in CSV format to an S3 bucket,
-which then gets pulled by search-api when the scheduled job runs.
+We use explicit judgements based on expert opinion. The relevance judgements are uploaded in 
+CSV format to an S3 bucket, which then gets pulled by search-api when the rake task runs.
```
stringiest: Out of interest, how are the relevance judgements uploaded to the S3 bucket? I wonder if it would be useful to include this here, or to signpost if it's explained elsewhere?

spec/integration/tasks/debug_spec.rb
```diff
@@ -122,6 +124,32 @@
     end
   end
 
+  describe "debug:ranking_evaluation" do
```
stringiest, line 127: Do we also need to test that an error is raised if no `AWS_S3_RELEVANCY_BUCKET_NAME` env var?

stringiest (commented):
This is really clear, thanks! The commit timeline really explains what's been changed and why. Nice movement of the bet bets helper too! 

Just had a documentation question, and I think maybe an additional test.

docs/search-quality-metrics.md
emmalowe: I'm pretty sure this would be manually updated. I'll add that.

spec/integration/tasks/debug_spec.rb
emmalowe: Ah yes, good idea. I'll add that.

stringiest (approved):
This looks great! 🎉

spec/integration/tasks/debug_spec.rb
```diff
@@ -122,6 +124,42 @@
...
+    let(:task_name) { "debug:ranking_evaluation" }
+
+    context "the bucket is provided and contains a judgement csv" do
+      let(:bucket) { "test-bucket" }
+      let(:filename) { "judgements.csv" }
+      let(:s3_client) { Aws::S3::Client.new(stub_responses: true) }
+
+      before do
+        s3_client.stub_responses(:get_object, body: mock_judgement_csv)
+        s3_object = Aws::S3::Object.new(bucket_name: bucket, key: filename, client: s3_client)
```
koetsier: it may be easier to inject the s3 client in the rake task (line 67):

```      o = Aws::S3::Object.new(bucket_name:, key: "judgements.csv", client: Services.s3_client) ```

Then you can either stub the responses of Services.s3_client or use the FakeS3 services as in: https://github.com/alphagov/search-api/blob/8a49efc7f28cd214ecc732395135248000a785ac/spec/integration/app/sitemap_spec.rb#L18

you can then 'put' the file in a fake s3 bucket before the test

koetsier (commented):
Nice! One comment

spec/integration/tasks/debug_spec.rb
emmalowe: Thanks @koetsier, I've made that change now.

The actual change was quick but when I re-tested on integration the request to the `rank_eval` endpoint was timing out. I think when I tested it before, I didn't wait long enough for the branch to deploy (even though it was shown as deployed in the release app, I've since noticed that the actual workflow in argo can lag behind that), so really I was testing on main. Because of that, I've reverted my revert of the [rank_eval workaround](https://github.com/alphagov/search-api/blob/a8488413c2e89a0479c276164486ab4a1f7cd833/lib/debug/rank_eval.rb#L73), increased the timeout and edited the comments to warn people in the future that this isn't as simple as it looks to remove (sadly).

koetsier (approved):
looks good! 👍

PR #3524, opened 2026-03-03
[SCH-1638] Configure auto-merging of some dependencies

stringiest:
Now that we've turned on continuous deployment for search-api we can configure dependabot to allow automatic merging of patch and minor dependency update PRs. This is in line with guidance from [RFC-167](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-167-auto-patch-dependencies.md). It has been confirmed that the repo meets the conditions of this RFC.

See [Jira ticket SCH-1638](https://gov-uk.atlassian.net/browse/SCH-1638).

Other useful links:
- [Guidance for how to configure Dependabot](https://docs.publishing.service.gov.uk/repos/govuk-dependabot-merger.html#content)
- [Search v2 Dependabot configuration for comparison](https://github.com/alphagov/search-api-v2/blob/main/.govuk_dependabot_merger.yml)

### DIFF
--- a/.govuk_dependabot_merger.yml
+++ b/.govuk_dependabot_merger.yml
@@ -1,0 +1,7 @@
+api_version: 2
+defaults:
+  update_external_dependencies: true
+  auto_merge: true
+  allowed_semver_bumps:
+    - patch
+    - minor


.govuk_dependabot_merger.yml
```diff
@@ -0,0 +1,11 @@
+api_version: 2
+defaults:
+  update_external_dependencies: true
+  auto_merge: true
+  allowed_semver_bumps:
+    - patch
+    - minor
+overrides:
+  - dependency: rails
```
emmalowe: We don't really need this, since this isn't a rails app. Are you keeping it in for consistency? (I suppose it doesn't hurt.)

emmalowe:
One thought I had when I was making sure you meet the requirements on the RFC is that it's difficult to see what's going on with the Pact tests because we're not printing to the console. I think it would be worth changing that (later) - do you?

.govuk_dependabot_merger.yml
stringiest: Whoops! Too much copy-paste...

stringiest:
> One thought I had when I was making sure you meet the requirements on the RFC is that it's difficult to see what's going on with the Pact tests because we're not printing to the console. I think it would be worth changing that (later) - do you?

Yes that would definitely make it easier! I couldn't see how to do it at the time of adding the pact tests, but I'll make a note to have a go 💪🏻

emmalowe:
> I couldn't see how to do it at the time of adding the pact tests, but I'll make a note to have a go 💪🏻

I think the settings for the logs are in the [config.ru file](https://github.com/alphagov/search-api/blob/main/config.ru).

emmalowe (approved):

PR #3548, opened 2026-03-24
Bump the Elasticsearch client gem to version 7.10.x.

koetsier:
Bump the Elasticsearch client gem to version 7.10.x.

We’re planning an upcoming upgrade of our cluster to 7.10, so this
change allows us to adopt the newer client early and ensure the codebase
remains compatible.

https://gov-uk.atlassian.net/browse/SCH-1868

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source "https://rubygems.org"
 gem "activesupport"
 gem "aws-sdk-s3"
 gem "bootsnap", require: false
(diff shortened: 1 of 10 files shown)


hannako:
This looks good based on my reading of the docs, but it will be interesting to see how it fares on integration as it's hard to know for sure if we've missed anything!

koetsier:
This is no longer needed as it is now part of https://github.com/alphagov/search-api/pull/3714

PR #3566, opened 2026-04-09
[SCH-1975] Add rake task to get docs from one index and check against another

stringiest:
Adding a rake task to retrieve `content_id`s for all docs in one index, and then to check another index to ensure they exist there. Intended to be used to check that all docs from `government` index were copied over to `govuk` index.

WIP: 
- added initial attempt at search query to get all docs
- Need to set a batch size then re-request until all docs are retrieved
- Haven't checked yet whether digging gives correct output.

### DIFF
--- a/lib/search/document_finder.rb
+++ b/lib/search/document_finder.rb
@@ -1,0 +1,26 @@
+module Search
+  class DocumentFinder
+    TIMEOUT = 60
+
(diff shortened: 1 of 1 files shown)


stringiest:
We've checked for any missing documents another way. Closing this PR as not needed.

PR #3571, opened 2026-04-14
Update search:update_popularity to use govuk index

emmalowe:
The govuk index will be the only relevant index to run this rake task on, once the government index has been retired (this is in progress). The government popularity job has already been [removed from helm charts](https://github.com/alphagov/govuk-helm-charts/pull/4218), which unblocks this change.

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-1986

To test the changes, I've checked:
- govuk only includes migrated formats, so the `search_body` filter in `popularity_updater.rb` is redundant (see csv on Jira ticket)
- `search:update_popularity` task locally: logs show scroll enumerator batching fetching of documents, and I added some logging locally to check we get to the point where we call the async job with batches of 100 docs [here](https://github.com/alphagov/search-api/blob/main/lib/govuk_index/updater.rb#L18)
- `search:update_supertypes` task locally: logs show scroll enumerator batching fetching of documents, and I added some logging locally to check we get to the point where we call the async job with batches of 100 docs [here](https://github.com/alphagov/search-api/blob/main/lib/govuk_index/updater.rb#L18)


lib/tasks/indices.rake
```diff
@@ -69,17 +69,15 @@ namespace :search do
...
 Update all data in the index inplace (without locks) with the new popularity
 data using sidekiq jobs.
 
 This does not update the schema.
 "
   task :update_popularity do
-    index_names.each do |index_name|
-      GovukIndex::PopularityUpdater.update(index_name, process_all: ENV.key?("PROCESS_ALL_DATA"))
-    end
+    GovukIndex::PopularityUpdater.update(SearchConfig.govuk_index_name, process_all: ENV.key?("PROCESS_ALL_DATA"))
```
emmalowe: The `PROCESS_ALL_DATA`  flag was being used for the government and detailed index popularity jobs, so maybe we can take this out now?

See: https://github.com/alphagov/govuk-helm-charts/pull/4218/changes

stringiest: Ooh good spot, yes, with some related refactoring in `lib/govuk_index/popularity_updater.rb` too.

stringiest (approved):
This looks great! I agree that `PROCESS_ALL_DATA` can probably come out. Approving as-is as I'm guessing you were envisaging a separate PR for that?

emmalowe:
Moving this to draft because I'm going to add the refactoring of PROCESS_ALL_DATA to this PR.

lib/govuk_index/popularity_updater.rb
```diff
@@ -1,39 +1,7 @@
...
-      )
-    end
-
-  private
-
-    def search_body
-      return { query: { match_all: {} } } if @process_all
-
-      # only sync migrated formats as the rest will be updated via the sync job.
-      {
```
emmalowe, line 30: This filter has no effect because everything in `govuk` is a migrated format. I checked the latest snapshot (see Jira ticket for spreadsheet of document counts against formats).

lib/govuk_index/popularity_updater.rb
```diff
@@ -1,39 +1,7 @@
...
-      super(
-        source_index:,
-        destination_index:,
-      )
-    end
-
-  private
-
-    def search_body
-      return { query: { match_all: {} } } if @process_all
```
emmalowe, line 27: `process_all` used to be set to true for the government and detailed indexes, and defaulted to false for `govuk`. See https://github.com/alphagov/govuk-helm-charts/pull/4218

However, the filter that is used when `process_all` is false, isn't useful anyway (see my other comment).

lib/govuk_index/updater.rb
```diff
@@ -6,6 +6,13 @@ class Updater
 
     class ImplementationRequired < StandardError; end
 
+    def self.update(index_name)
```
koetsier: The only difference between the updater classes is the `self.job` method. Could we not just pass that as a parameter to the update method so we can get rid of the subclassing?

koetsier (commented):
looks good. one comment

lib/govuk_index/updater.rb
emmalowe: Oh good call! I've added two commits to do that.

koetsier (approved):
Great! 🥇

PR #3574, opened 2026-04-15
Update Pact tests to use govuk index

emmalowe:
Prep work for retiring the government index.

In updating the Pact tests to use the govuk index, we need to add a migrated format to the documents, so the documents don't get removed in the post query filter in Search::QueryBuilder.
https://github.com/alphagov/search-api/blob/main/lib/search/query_builder.rb#L22

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-1987


stringiest (approved):
Amazing! Well done for working out how to get this working for the pact tests! 🦸🏻‍♀️

PR #3580, opened 2026-04-17
[PP-7588] Fix max path length

yndajas:
In 5ed2e55, content with a path of 512 bytes or larger was filtered out before being indexed to prevent errors. This requirement comes from Elasticsearch:
https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/mapping-id-field

However, per the Elasticsearch documentation and the introducing commit's message, it should only be paths of greater than 512 bytes that get rejected. This therefore fixes the logic to use greater than rather than greater than or equal to


stringiest (approved):
Looks good - thanks! 🙌🏻

PR #3581, opened 2026-04-17
Sch 1987 refactor unit tests to use govuk test index

emmalowe:
Update unit tests to refer to the govuk index, since the government index is being retired.

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-1987


stringiest (approved):
This looks good to me!

PR #3589, opened 2026-04-21
Remove listen to publishing queue

koetsier:
Remove listen to publishing queue

There is a rake task that reads events from the publishing queue and updates documents in government index with any changes to links. As the government index has been decommissioned, we remove this task and all dependent code and tests

Jira: https://gov-uk.atlassian.net/browse/SCH-1985


emmalowe (dismissed):
LGTM!

emmalowe:
Sorry - just spotted. Do we need to remove the reference to the rake task here?
https://github.com/alphagov/search-api/blob/main/Procfile

koetsier:
> Sorry - just spotted. Do we need to remove the reference to the rake task here? https://github.com/alphagov/search-api/blob/main/Procfile

good spot. removed it

emmalowe (approved):

emmalowe (approved):

PR #3593, opened 2026-04-22
Remove the fetch_missing_metadata rake task

koetsier:
We don't use this task anymore and so we can delete it.

It was used to populate possible missing values in documents directly from the Publishing API. Should this occur, we can simply republish the affected content.

https://gov-uk.atlassian.net/browse/SCH-2051


emmalowe (approved):
❤️ 🔥

PR #3594, opened 2026-04-23
deprecate the DELETE /:index/documents.* endpoint

koetsier:

All updates to the Elasticsearch index are now done through RabbitMQ so
this endpoint can be removed.

Log any attempt to access this endpoint


emmalowe (approved):
LGTM!

PR #3596, opened 2026-04-23
deprecate the POST /:index/documents/* endpoint

koetsier:

All updates to the Elasticsearch index are now done through RabbitMQ so
this endpoint can be removed.

Log any attempt to access this endpoint

There are a lot of dependent classes that have also been removed


lib/index.rb
stringiest: Do we also need to remove this method if we're removing the closing_spec.rb?

spec/integration/indexer/locking_spec.rb
```diff
@@ -5,33 +5,13 @@
...
-    index = search_server.index_group("govuk_test").current
-    index.add([sample_document])
-
-    with_lock(index) do
-      expect {
-        index.amend(sample_document.link, "title" => "New title")
-      }.to raise_error(SearchIndices::IndexLocked)
-    end
-  end
-
```
stringiest, line 27: For my understanding, would we want to add some tests to give confidence that we weren't able to write to the `page_traffic` index (and maybe also the `govuk` index) while they're locked (or closed)? (Though I accept that they couldn't now raise this error, as we've removed the code that raises it, so maybe those tests would live elsewhere.)

stringiest (commented):
Wow, this is a lot! 🚀 

Possibly mostly for my understanding, but where we're removing tests related to locked or closed indices, do we need to rewrite the existing tests to work to remaining indices? I can see reference in the code to the `page_traffic` index being locked at a certain point, for example.

spec/integration/indexer/locking_spec.rb
koetsier: I'm unsure we even need the locking functionality. according to https://github.com/alphagov/search-api/pull/153 this was used during a migration. We don't need to do that anymore. Thanks for mentioning this - I'll raise it on Monday.

For now, I think testing that deletion works is enough - testing other operations would test Elasticsearch itself which  I feel is a bit pointless.

lib/index.rb
koetsier: Good point. it isn't being used anymore

spec/integration/indexer/locking_spec.rb
stringiest: That makes sense - thanks for explaining!

stringiest (approved):
This looks great!

PR #3597, opened 2026-04-24
Sch 2034 update search api rake tasks that reference government index

emmalowe:
Remove references to government index in search-api rake tasks where relevant, in preparation of retiring the government index.

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-2034


lib/tasks/duplicates.rake
```diff
@@ -2,8 +2,8 @@ require "rummager"
 
 namespace :duplicates do
   desc "Find all documents with the same content_id"
-  task :find, [:index] do |_t, args|
-    index = args[:index] || "government"
+  task :find do
+    index = SearchConfig.govuk_index_name
 
     duplicates = Search::DuplicateFinder.new(index:).find_duplicates
```
emmalowe: I did wonder about removing index as an input here and in Search::DuplicateRemover and just having it hardcoded inside those classes. Curious to hear what others think about that.

koetsier: It's not really a big deal in my opinion but it would probably make sense to do that

koetsier (approved):
Nice! 👍

lib/tasks/duplicates.rake
emmalowe: Done!

emmalowe:
@koetsier I've refactored the duplicates code - do you mind reviewing this again please?

PR #3552, opened 2026-03-27
Add patent decision specialist document configuration [WHIT-3027]

eYinka:
Adds support for the new patent decision specialist document.

Related PRs:

- https://github.com/alphagov/publishing-api/pull/3978
- https://github.com/alphagov/specialist-publisher/pull/3590

[JIRA](https://gov-uk.atlassian.net/browse/WHIT-3027)

PR #3598, opened 2026-04-24
Update field definitions for patent decision schema to use identifiers

eYinka:
This PR updates the filter schema for the patent decision finder added in https://github.com/alphagov/search-api/pull/3552

Following [field_types.json](https://github.com/alphagov/search-api/blob/main/config/schema/field_types.json),`identifiers` is the correct type for fields with a small, fixed set of values used only for exact-match filtering and aggregation.

In this case, `patent_decision_type_of_hearing` only has two fixed, controlled values (inter-partes-only, ex-parte-only) and is used exclusively for filtering. See [patent_decision.json](https://github.com/alphagov/specialist-publisher/blob/56d83d495cfeabaf4d6278aac32a1b179f0422e8/lib/documents/schemas/patent_decisions.json#L8-L31)

### DIFF
--- a/config/schema/field_definitions.json
+++ b/config/schema/field_definitions.json
@@ -1234,6 +1234,6 @@
     "type": "searchable_multivalued_text"
   },
   "patent_decision_type_of_hearing": {
-    "type": "searchable_multivalued_text"
+    "type": "identifiers"
   }
 }


hannako:
Please could you add a bit of context for why you're making this change to the PR description to help us review future specialist finder PR's?

eYinka:
> Please could you add a bit of context for why you're making this change to the PR description to help us review future specialist finder PR's?

Thanks! I've added more information.

hannako (approved):
Thanks for adding the additional information to the PR description - much appreciated 👍

PR #3599, opened 2026-04-24
[SCH-2036] Use govuk index to serve spelling suggestions.

stringiest:
Previously we used the government index to serve spelling suggestions. This was a necessary workaround when we had multiple content indices. Now all content is read from a single index, we can serve spelling suggestions from govuk index instead.

See Jira ticket [SCH-2036](https://gov-uk.atlassian.net/browse/SCH-2036)

Deployed to integration & checked that spelling suggestions still worked ✅ 

[SCH-2036]: https://gov-uk.atlassian.net/browse/SCH-2036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/search/query.rb
+++ b/lib/search/query.rb
@@ -7,14 +7,12 @@ module Search
 
     class QueryTooLong < Error; end
 
-    attr_reader :index, :registries, :spelling_index, :suggestion_blocklist
+    attr_reader :index, :registries, :suggestion_blocklist
 
-    def initialize(registries:, content_index:, metasearch_index:, spelling_index:)
+    def initialize(registries:, content_index:, metasearch_index:)
       @index = content_index
-
       @registries = registries
       @metasearch_index = metasearch_index
-      @spelling_index = spelling_index
       @suggestion_blocklist = SuggestionBlocklist.new(registries)
     end
 
--- a/lib/search/query.rb
+++ b/lib/search/query.rb
@@ -130,7 +130,7 @@ module Search
     # some indexes contain very few words, Elasticsearch returns too many spelling
     # suggestions for common terms. For example, using the suggester on all indices
     # will yield a suggestion for "PAYE", because it's mentioned only in the
-    # `government` index, and not in other indexes.
+    # `govuk` index, and not in other indexes.
     #
     # This issue is mentioned in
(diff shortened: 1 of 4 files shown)


lib/search/query.rb
```diff
@@ -130,7 +130,7 @@ def present_aggregates_with_examples(search_params, es_response, builder)
     # some indexes contain very few words, Elasticsearch returns too many spelling
     # suggestions for common terms. For example, using the suggester on all indices
     # will yield a suggestion for "PAYE", because it's mentioned only in the
-    # `government` index, and not in other indexes.
```
koetsier, line 133: This comment seems to suggest that the only reason they submit a separate suggest query here is because we use multiple indices. Because we now only use the govuk index, it may be possible to merge the suggest query into the main one, so that only one query is submitted. 

So, in QueryBuilder you could add 

```
    def suggest
      QueryComponents::Suggest.new(search_params).payload
    end
```

and then modify `payload` to add this to the main query:

```
 def payload
      hash_without_blank_values(
        from: search_params.start,
        size: search_params.count,
        _source: {
          includes: fields.uniq,
        },
        query:,
        post_filter: filter,
        sort:,
        aggs: aggregates,
        highlight:,
        explain: search_params.debug[:explain],
        suggest:,
      )
    end

```

so the second query can be removed

koetsier (commented):
Looks good. I have one suggestion to maybe simplify things a bit. Let me know what you think

lib/search/query.rb
stringiest: I've had a go at this this morning. I think that, at the moment, too much of the downstream code relies on there being multiple content indices. When I take out the second query I have to move the `return unless suggestion_blocklist.should_correct?(search_params.query)` to `query_builder.rb#suggest` too. That then creates a parsing exception with unexpected token for about 40 of the tests. 

I'll make a separate ticket for refactoring this, as I think that we'll be removing the second content_index in other PRs and that will make this easier to refactor.

koetsier (approved):
👍

PR #3600, opened 2026-04-24
Deprecate the /content and 'delete "/:index/documents"' endpoints

koetsier:
Deprecate get "/content, delete "/content" and delete "/:index/documents" endpoints

These routes are no longer used and now raise and error when accessed.

In order to preserve the functionality of removing a document given a link and to show the raw elasticsearch document given a link, the debug:show_govuk_link and delete:by_link rake tasks were added


lib/rummager/app.rb
stringiest: Does this use the `#delete` that has been removed in `lib/index.rb`?

lib/tasks/delete.rake
stringiest: I really like the `abort` for missing argument in `delete:by_link` - could we also use it here for consistency?

stringiest (commented):
This is looking great! One possible remaining use of a deleted method, and a suggestion for consistency 🚀

lib/rummager/app.rb
koetsier: Yes you're right! I fixed it by removing that endpoint as well, as it was on the list and it doesn't have any other dependencies.

lib/tasks/delete.rake
koetsier: I added a commit for that

stringiest (approved):
This looks great! 🎉

PR #3603, opened 2026-04-28
Deprecate post "/commit" endpoint

koetsier:
This route is no longer in use

Remove some previously deprecated endpoints and reorder some routes 

jira: https://gov-uk.atlassian.net/browse/SCH-1989


stringiest (approved):
This looks great! Nice plan grouping all the endpoints deprecated at the same time 🌟

PR #3604, opened 2026-04-28
Ensure queues are bound with durable: true option

koetsier:
Modern versions of RabbitMQ do not allow queues that are not durable and not exclusive. This change ensures that the queues we define are now all durable to fix this issue.

We also ensure we only create each queue once and then bind two exchanges

This fixes an error when make-ing the search-api in govuk-docker

### DIFF
--- a/lib/tasks/message_queue.rake
+++ b/lib/tasks/message_queue.rake
@@ -20,21 +20,21 @@ namespace :message_queue do
       retry_dlx = channel.fanout("#{name}_retry_dlx")
       discarded_dlx = channel.fanout("#{name}_discarded_dlx")
 
-      channel
-        .queue(name, arguments: { "x-dead-letter-exchange" => retry_dlx.name })
-        .bind(exch, routing_key: routing_key)
-
       # messages are queued on {queue}_discarded_dlx for 30s before their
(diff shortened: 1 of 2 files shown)


lib/tasks/message_queue.rake
```diff
@@ -20,21 +20,21 @@ namespace :message_queue do
...
         .queue(
           "#{name}_wait_to_retry",
+          durable: true,
           arguments: { "x-dead-letter-exchange" => discarded_dlx.name, "x-message-ttl" => 10 * 3000 },
         )
         .bind(retry_dlx)
 
-      # messages on the {queue}_discarded_dlx are routed back to the original queue
-      channel.queue(name).bind(discarded_dlx)
+      channel.queue(name, durable: true, arguments: { "x-dead-letter-exchange" => retry_dlx.name }).tap do |q|
```
stringiest, line 33: Oooh I haven't come across `tap` before! It's very pleasing!

stringiest (approved):
This looks great, thanks!

Adding a [link to the RabbitMQ docs re durability](https://www.rabbitmq.com/docs/queues#durability) for future reference.

PR #1492, opened 2019-03-27
Remove rake task

MuriloDalRi:
This changes how the advanced search finder is published. This will now be done using the same rake task as the other finders.

https://trello.com/c/WwLFwsX4/332-remove-publishdocumentfinder-rake-task

PR #3605, opened 2026-04-28
Delete publishing_api_finder_publisher.rb

emmalowe:
Classes in this file are no longer referenced in search-api and can be removed. They used to be used in a rake task that published the "advanced search finder" until this was updated to be published in the same way as other finders [1][2].

The deprecated rake task used a class called
PublishingApiFinderPublisher that was removed [3] but the other classes in this file that PublishingApiFinderPublisher referenced were not deleted at that time. This is overdue clean up.

[1] https://github.com/alphagov/search-api/pull/1492
[2] https://github.com/alphagov/search-api/blob/main/lib/tasks/publishing_api.rake#L10
[3] https://github.com/alphagov/search-api/pull/1967


hannako (commented):
Excellent detective work!

For posterity, could you add some additional information to the PR description.

https://github.com/alphagov/search-api/pull/1967 to give context to the commit I think would be helpful.

And then could you link to https://github.com/alphagov/search-api/blob/main/lib/tasks/publishing_api.rake#L10 which eventually calls https://github.com/alphagov/search-api/blob/main/lib/content_item_publisher/finder_presenter.rb#L2 as the route to publish finders from this application.

hannako (approved):

emmalowe:
> #1967 to give context to the commit I think would be helpful.

Thanks @hannako - I struggled to find the PR that the commit was part of! ⭐

PR #3610, opened 2026-04-30
Sch 1990 remove registry index

emmalowe:
In elasticsearch.yml there is a mapping of a "registry_index" to the government index. This reference is used in code that looks up linked documents to supported expanded links for four fields that could exist on an elasticsearch document:

- policy areas
- document_series
- document_collections
- world_locations

The `policy_areas` and `document_series` fields have been deprecated, meaning we no longer need the link expansion and so it's being removed. For the other two, we're switching to look up additional information from the govuk index instead of the government index. We're then deleting the reference to the registry_index, in preparation of deprecating the government index.

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-1990

This PR can be merged now that https://github.com/alphagov/search-api/pull/3613 is merged and documents with the format "document_collection" have been republished to populate the slug field in elasticsearch. This ensures the registry look up will stay functional, once we switch to using the govuk index. This has been tested on integration.

### DIFF
--- a/elasticsearch.yml
+++ b/elasticsearch.yml
@@ -3,7 +3,6 @@ production: &default
   content_index_names: ["government"]
   govuk_index_name: "govuk"
   auxiliary_index_names: ["page-traffic", "metasearch"]
-  registry_index: "government"
   metasearch_index_name: "metasearch"
   page_traffic_index_name: "page-traffic"
   popularity_rank_offset: 10
@@ -21,7 +20,6 @@ test:
   content_index_names: ["government_test"]
   govuk_index_name: "govuk_test"
   auxiliary_index_names: ["page-traffic_test", "metasearch_test"]
-  registry_index: "government_test"
   metasearch_index_name: "metasearch_test"
   page_traffic_index_name: "page-traffic_test"
   popularity_rank_offset: 10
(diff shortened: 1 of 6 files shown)


hannako (commented):
This looks great, thank you for splitting into nice commits with relevant links - really helpful.

Do we have a card to track the workaround we've left in place here: https://github.com/alphagov/search-api/pull/3610/changes/d1d7150e36b2532111b23a7b759be41104f94f39

It looks solid, but given the complexity of v1 shall we put this branch on integration and publish a couple of relevant document types just to make sure nothing unexpected happens?

emmalowe:
> Do we have a card to track the workaround we've left in place here: [d1d7150](https://github.com/alphagov/search-api/commit/d1d7150e36b2532111b23a7b759be41104f94f39)

Not yet, but I've added a checklist item on the Jira ticket to make follow up cards.

> It looks solid, but given the complexity of v1 shall we put this branch on integration and publish a couple of relevant document types just to make sure nothing unexpected happens?

I did some testing already today with @stringiest . We deployed https://github.com/alphagov/search-api/pull/3613 to integration, republished document_collections so that they had the slugs, then deployed this branch to integration to check that documents were being expanded properly. I ran https://github.com/alphagov/search-api/blob/main/lib/tasks/report_inconsistent_aggregate_values.rake to check that aggregates for document_collections were being expanded correctly and didn't get any errors. Are we missing some checks?

hannako (approved):
:shipit:

PR #3613, opened 2026-05-05
Populate slugs in document_collections

emmalowe:
Documents that have the format document_collection are no longer populating the slug field in elasticsearch. This prevents the document_collections field being expanded correctly via the registries look up [1]. This change populates the slug field using the base path, so that the document_collections field can be expanded correctly.

[1] https://github.com/alphagov/search-api/blob/main/lib/search/registries.rb#L19

Jira link: https://gov-uk.atlassian.net/browse/SCH-1990

Have tested this on integration, republished document_collection documents and now all document_collections in the govuk index have slugs 🎉

### DIFF
--- a/lib/govuk_index/presenters/common_fields_presenter.rb
+++ b/lib/govuk_index/presenters/common_fields_presenter.rb
@@ -47,6 +47,8 @@ module GovukIndex
 
     def slug
       case format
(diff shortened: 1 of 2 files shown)


stringiest (approved):
This looks great! 🎉

PR #3614, opened 2026-05-05
[SCH-2065] address failing comparer specs

stringiest:
In order to remove `government` and `government_test` from `elasticsearch.yml` we need to review the failing tests of `spec/integration/comparer_spec.rb`. See [Jira ticket SCH-2065](https://gov-uk.atlassian.net/browse/SCH-2065).

The `spec/integration/comparer_spec.rb` tests `Indexer::Comparer` and `Indexer::CompareEnumerator`, which are only used as the end result of the `search:migrate_schema` rake task, which is no longer used.

The `search:migrate_schema` rake task was originally added as a way of updating popularity on an index in place, and was run nightly as a cronTask. This function was taken over by `search:update_popularity`, which is now run nightly to update popularity scores on the `govuk` index. The new rake task only updates popularity, it does not update the schema. (If the schema needs to be updated in place the rake task `search:update_schema` can be used. This is sufficient unless the change being made is not applicable to existing data.) 

Changes to the schema are rare. With the exception of Ryan removing expanded search result fields configuration a year ago, the other changes to the schema in the last 7 years have all been linting changes. 

If we wanted to leave it in situ and for it to remain fully tested we would have to create a new dummy index for testing purposes, which would be frustrating as we're trying to delete an index. Removing all code in the path as it is not needed. This code could be reinstated in the future if there were configuration changes that were not applicable to existing data.


stringiest:
Closing this following a discussion with @koetsier, as it looks like there's a way to remove the comparer functionality and still be able to check that the reindexing has worked - and therefore to keep the `search:migrate_schema` rake task.

PR #3615, opened 2026-05-06
(CM-934) Make Content Blocks available to Search API

pezholio:
This adds the various content blocks as supported document types in Search API

### DIFF
--- a/config/govuk_index/mapped_document_types.yaml
+++ b/config/govuk_index/mapped_document_types.yaml
@@ -25,6 +25,10 @@ complaints_procedure: edition
 consultation: edition
 consultation_outcome: edition
 contact: contact
+content_block_contact: edition
+content_block_pension: edition
+content_block_tax: edition
+content_block_time_period: edition
 coronavirus_landing_page: edition
 corporate_report: edition
 correspondence: edition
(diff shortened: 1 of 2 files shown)


hannako:
This PR popped up in our tech channel today and we are not clear on the consequences given our limited understanding of what a content block is. Do you have a link to a content block content item so we can understand what a search result might look like, and what page it would link to? Or is this PR part of a bigger piece of work and further commits are to come, if so feel free to set it as draft and we will hold off with the questions until you're done!

pezholio:
So, this is an example of a content block:

https://www.gov.uk/api/content/content-blocks/content_block_pension/new-state-pension

They are not used in pages in isolation, but embedded in other pages. The thinking behind having them indexed in Search API is to provide a way for a tool we're building to be able to list content blocks without having to build a new API or publish an "index" page.

Please let me know if this has any unintended consequences!

edavey:
The Content Block Manager is developing its own API in order to allow content blocks to be discovered and rendered. The immediate use-case is the development of a "content block picker" which will allow editors to insert blocks into their content without having to leave those publishing apps. 

https://github.com/alphagov/content-block-manager/blob/681c56c8f2ad98979f322088fda02dd0fd85d4cb/docs/architecture/decisions/0013-add-an-api-for-content-block-manager.md

PR #3617, opened 2026-05-08
Sch 2066 address sitemap generator spec test failures

emmalowe:
Update sitemap tests to use the govuk index in preparation of decommissioning the government index.

There is still one sitemap test remaining that uses the government index. This is testing the filter on returned documents, so should be removed at the same time that the filter is updated, or when we delete the government index.

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-2066


koetsier (approved):
lgtm

PR #3621, opened 2026-05-09
Remove separate Elasticsearch suggestion query

koetsier:
The separate suggestion query was originally introduced to work around Elasticsearch returning spelling suggestions for terms missing from some indices but present in others.

Now that search uses a single index, this workaround is no longer necessary. Suggestions can be fetched as part of the main query, removing the extra Elasticsearch request.

According to the comment:

    Elasticsearch tries to find spelling suggestions for words that don't occur in
    our content, as they are probably mispelled. However, currently it is
    returning suggestions for words that do not occur in *every* index. Because
    ome indexes contain very few words, Elasticsearch returns too many spelling
    suggestions for common terms. For example, using the suggester on all indices
    will yield a suggestion for "PAYE", because it's mentioned only in the
    `govuk` index, and not in other indexes.

    This issue is mentioned in
    https://github.com/elastic/elasticsearch/issues/7472.

    Our solution is to run a separate query to fetch the suggestions, only using
    the indices we want.

https://gov-uk.atlassian.net/browse/SCH-2059

### DIFF
--- a/lib/search/query.rb
+++ b/lib/search/query.rb
@@ -38,11 +38,14 @@ module Search
     attr_reader :metasearch_index
 
     def timed_build_query(search_params)
+      include_suggestions = search_params.suggest_spelling? && suggestion_blocklist.should_correct?(search_params.query)
+
       GovukStatsd.time("build_query") do
         builder = QueryBuilder.new(
           search_params:,
           content_index_names:,
           metasearch_index:,
+          include_suggestions:,
         )
 
         payload = process_elasticsearch_errors { builder.payload }
@@ -77,10 +80,6 @@ module Search
 
     def process_es_response(search_params, builder, payload, es_response)
       # Augment the response with the suggest result from a separate query.
-      if search_params.suggest_spelling?
-        es_response["suggest"] = run_spell_checks(search_params)
-      end
-
       if search_params.suggest_autocomplete?
         es_response["autocomplete"] = run_autocomplete_query(search_params)
       end
@@ -116,35 +115,6 @@ module Search
       AggregateResultPresenter.merge_examples(presented_aggregates, examples)
 
       presented_aggregates
-    end
-
-    # Elasticsearch tries to find spelling suggestions for words that don't occur in
-    # our content, as they are probably mispelled. However, currently it is
-    # returning suggestions for words that do not occur in *every* index. Because
-    # some indexes contain very few words, Elasticsearch returns too many spelling
-    # suggestions for common terms. For example, using the suggester on all indices
-    # will yield a suggestion for "PAYE", because it's mentioned only in the
-    # `govuk` index, and not in other indexes.
-    #
-    # This issue is mentioned in
-    # https://github.com/elastic/elasticsearch/issues/7472.
-    #
-    # Our solution is to run a separate query to fetch the suggestions, only using
-    # the indices we want.
-    def run_spell_checks(search_params)
-      return unless suggestion_blocklist.should_correct?(search_params.query)
-
-      GovukStatsd.increment "suggest.spelling"
-      GovukStatsd.time("suggest.spelling") do
-        query = {
-          size: 0,
-          suggest: QueryComponents::Suggest.new(search_params).payload,
-        }
-
-        response = index.raw_search(query)
-
-        response["suggest"]
-      end
     end
 
     def run_autocomplete_query(search_params)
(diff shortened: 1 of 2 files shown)


lib/search/query.rb
```diff
@@ -118,35 +117,6 @@ def present_aggregates_with_examples(search_params, es_response, builder)
       presented_aggregates
     end
 
-    # Elasticsearch tries to find spelling suggestions for words that don't occur in
-    # our content, as they are probably mispelled. However, currently it is
-    # returning suggestions for words that do not occur in *every* index. Because
```
emmalowe, line 117: Just checking that this means every index we're searching through, rather than every index? Otherwise, we'd need to keep it, I think.

lib/search/query.rb
```diff
@@ -118,35 +117,6 @@ def present_aggregates_with_examples(search_params, es_response, builder)
...
-    # will yield a suggestion for "PAYE", because it's mentioned only in the
-    # `govuk` index, and not in other indexes.
-    #
-    # This issue is mentioned in
-    # https://github.com/elastic/elasticsearch/issues/7472.
-    #
-    # Our solution is to run a separate query to fetch the suggestions, only using
-    # the indices we want.
-    def run_spell_checks(search_params)
-      return unless suggestion_blocklist.should_correct?(search_params.query)
```
emmalowe: If we are removing this (which I'm not sure we should, yet), I think you can get rid of the SuggestionBlocklist class.

lib/search/query_builder.rb
```diff
@@ -71,7 +73,14 @@ def filter
 
   private
 
-    attr_reader :content_index_names, :metasearch_index
+    attr_reader :content_index_names, :metasearch_index, :include_suggestions
+
+    def suggest
+      GovukStatsd.increment "suggest.spelling"
```
emmalowe: I didn't think we were using GovukStatsd anymore.

See: https://docs.publishing.service.gov.uk/repos/govuk_app_config.html#statsd-deprecated

lib/search/query.rb
```diff
@@ -118,35 +117,6 @@ def present_aggregates_with_examples(search_params, es_response, builder)
       presented_aggregates
     end
 
-    # Elasticsearch tries to find spelling suggestions for words that don't occur in
-    # our content, as they are probably mispelled. However, currently it is
-    # returning suggestions for words that do not occur in *every* index. Because
-    # some indexes contain very few words, Elasticsearch returns too many spelling
-    # suggestions for common terms. For example, using the suggester on all indices
-    # will yield a suggestion for "PAYE", because it's mentioned only in the
```
emmalowe, line 120: Do we know how sensitive the spelling suggester is? For example: if an acronym appears once, is that enough for it to not get suggested spelling suggestions, or is it more complex than that? (I don't think anything in the PR would functionally effect this, I'm just curious.)

lib/search/query.rb
koetsier: I assumed this referred to searching through indices included in the search, i.e. government, detailed govuk etc. We now only have gov.uk we search with, so this should no longer be an issue

lib/search/query.rb
koetsier: In here: https://www.elastic.co/docs/reference/elasticsearch/rest-apis/search-suggesters it says" `missing: Only provide suggestions for suggest text terms that are not in the index (default).` So it seems that is only suggests words that are not in the index

lib/search/query_builder.rb
koetsier: nice. I removed it.

lib/search/query.rb
koetsier: I deployed the branch to integration and I tried a bunch of queries, i.e. https://www.gov.uk/api/search.json?q=crodit&suggest=spelling or https://www.gov.uk/api/search.json?q=credit&suggest=spelling and it the results are the same as on production.

emmalowe: Makes sense. We're not querying the auxiliary indexes here, so they shouldn't come into it.

emmalowe:
I'm surprised there are no test changes. Is this feature adequately tested? Do you think we should add anything?

emmalowe:
Changes look fine to me in principle, but should we wait until we've actually decommissioned the government index to deploy them?

emmalowe (approved):
LGTM!

I noticed there are a bunch of integration tests that cover spell checking in https://github.com/alphagov/search-api/blob/main/spec/integration/search/search_spec.rb so we should be good there.

PR #3624, opened 2026-05-14
Remove content indices

koetsier:
The [elasticsearch.yml](https://github.com/alphagov/search-api/blob/main/elasticsearch.yml) file defines our elasticsearch indices. We need to remove references to government and government_test from this file.

This is one of the final steps in the retirement of the government index and should be picked up once we have reduced the quantity of legacy code (and tests) that call this index.

Jira: https://gov-uk.atlassian.net/browse/SCH-2038


lib/search/query.rb
```diff
@@ -38,17 +38,11 @@ def run(search_params)
     attr_reader :metasearch_index
 
     def timed_build_query(search_params)
-      GovukStatsd.time("build_query") do
-        builder = QueryBuilder.new(
-          search_params:,
-          content_index_names:,
-          metasearch_index:,
-        )
+      builder = QueryBuilder.new(search_params:, metasearch_index:)
```
stringiest: I know the `GovukStatsd` metrics aren't working, but in the interests of consistency is it worth either renaming this method to just `build_query`, or leaving in the `GovukStatsd` metric to make it easier to replace all instances of it with prometheus metrics in a later PR?

spec/integration/search/search_spec.rb
```diff
@@ -35,118 +30,95 @@
...
-  end
-
-  it "can filter and reject" do
-    build_sample_documents_on_content_indices(documents_per_index: 2)
+      expect(result_links.sort).to match_array(["/one", "/missing"])
+    end
 
-    get "/search?reject_mainstream_browse_pages=1&filter_document_type[]=edition"
+    it "can filter and reject" do
+      get "/search?reject_format=aaib&filter_mainstream_browse_pages[]=browse/page/2"
```
stringiest, line 111: Is this `reject_format=aaib` doing anything? I couldn't see a document being created with format `aaib`

koetsier: fixed

stringiest (approved):
What a mammoth! Looks good to me - thanks!

spec/integration/search/search_spec.rb
emmalowe: I'm still not sure this is doing anything:

1. Wouldn't we need to commit the document before running the request?
2. Wouldn't the mainstream_browse_pages filter need to match?

spec/integration/search/search_spec.rb
```diff
@@ -286,95 +251,66 @@
...
+    it "can filter to date" do
+      get "/search?filter_opened_date=to:2000-02-20"
 
-    expect(last_response).to be_ok
-    expect_response_includes_matching_date_and_datetime_results(parsed_response.fetch("results"))
-  end
+      expect(result_links).to match_array(["/january", "/february"])
+    end
 
-  it "can filter times in different time zones" do
```
emmalowe, line 359: Do we still need to be testing this?

emmalowe: @stringiest there's just a couple of little things I've noticed going along like this, that I was going to put in a follow up PR, since they're not blocking.

Overall this is looking pretty good to me though, so I think we could move onto integration testing tomorrow?

stringiest: Sounds good to me @emmalowe !

spec/integration/search/search_spec.rb
emmalowe: I'll put this in a follow up PR.

emmalowe (approved):

PR #3627, opened 2026-05-15
[SCH-2065] Refactor schema migrator to remove index comparer

stringiest:
`index_comparer` was being used to confirm that reindexing had been successful. Elasticsearch provide a field in the tasks API response that we can use for this instead.

Refactoring to use failures means that we can remove the `index_comparer` & `index_compare_enumerator` code, and associated tests. This removes tests that were relying on there being a second content index, helping to progress the work to decommission the government index.

I've added `close` and `open` index methods, so that we can stop an index from being able to be reindexed in tests. I've then exposed `dest_index`, so that the destination index can be closed, generating an error in the elasticsearch response and allowing me to test indexing failures.

See [Jira ticket SCH-2065](https://gov-uk.atlassian.net/browse/SCH-2065).


spec/integration/schema_migrator_spec.rb
```diff
@@ -26,69 +26,24 @@
     }
     commit_document("govuk_test", document)
 
-    described_class.new("govuk_test", wait_between_task_list_check: 0.2) do |migrator|
-      migrator.reindex
-      migrator.switch_to_new_index
-    end
+    migrator = described_class.new("govuk_test", wait_between_task_list_check: 0.2)
+    migrator.reindex
+    migrator.switch_to_new_index
```
koetsier, line 33: I think one issue is that these tests create new indices after each test and so the state before and after the tests isn't the same. You could maybe think about setting the indices to the original state?

The IndexHelpers class has `#clean_all` and `#create_all` for that. It might be slow though. what do you think? (By the way, I think that the old tests also didn't clean up)

koetsier (commented):
I added a suggestion about cleaning up indices. Also, I believe that facet_config.yml and metadata.csv are unused but are not directly related to removing comparer code. Maybe a separate commit for those might be useful?

spec/integration/schema_migrator_spec.rb
stringiest: Ooh interesting! I'd thought that the indices were being created before the test suite and deleted after it already, by [setup_before_suite](https://github.com/alphagov/search-api/blob/main/spec/support/integration_spec_setup_helper.rb#L21)

stringiest: Resolved after chat - thanks! ⭐

koetsier (approved):
Great! 👍

PR #3638, opened 2026-05-26
Clean up from remove content indices

emmalowe:
Some clean up following on from https://github.com/alphagov/search-api/pull/3624, as well as other tidy up related to removing the government index.


stringiest (approved):
Very tidy! 🧹

PR #3639, opened 2026-05-26
Delete schema for detailed index

emmalowe:
The detailed index has been decommissioned so this is no longer required.

See https://github.com/alphagov/search-api/pull/3471


stringiest (approved):
🧹 Nice!

PR #3647, opened 2026-06-03
Remove auxiliary index names

koetsier:
Auxiliary index names are no longer needed now that search uses a single content index, page traffic index, and metasearch index.

Simplify the code by removing support for auxiliary index names and refactor SearchServer to read index names from SearchConfig instead of accepting govuk_index_name and auxiliary_index_names as arguments.

### DIFF
--- a/spec/unit/search_server_spec.rb
+++ b/spec/unit/search_server_spec.rb
@@ -1,7 +1,6 @@
 require "spec_helper"
 
 RSpec.describe SearchIndices::SearchServer do
-  let(:auxiliary_index_names) { SearchConfig.auxiliary_index_names }
   let(:govuk_index_name) { SearchConfig.govuk_index_name }
   def schema_config
     schema = double("schema config")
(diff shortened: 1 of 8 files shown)


spec/unit/search_server_spec.rb
```diff
@@ -11,37 +10,23 @@ def schema_config
...
 
   it "raises an error for unknown index" do
-    search_server = described_class.new("http://l", schema_config, auxiliary_index_names, govuk_index_name, SearchConfig.default_instance)
+    search_server = described_class.new("http://l", schema_config, SearchConfig.default_instance)
     expect {
       search_server.index("unknown")
     }.to raise_error(SearchIndices::NoSuchIndex)
   end
-
-  it "can get multi index" do
```
emmalowe: Why are we deleting these tests? `index_for_search` should still be able to handle arrays.

emmalowe (approved):
🧹 ❤️

PR #3650, opened 2026-06-05
Remove legacy client code

koetsier:
The LegacyClient::IndexForSearch and LegacyClient::MultiValueConverter are left-overs from a time the search-api migrated from SOLR to Elasticsearch. The IndexForSearch class almost overlap entirely with the SearchIndices::Index class and so are better merged. The LegacyClient::MultiValueConverter is entirely obsolete because it was geared towards a specific quirk of legacy SOLR. 

Doing this gets rid of a confusing ‘LegacyClient’ namespace that isn’t really used anymore. It also increases integration test coverage because we no longer need to test duplicate code

https://gov-uk.atlassian.net/browse/SCH-2135


lib/index.rb
```diff
@@ -28,6 +28,10 @@ def initialize(base_uri, index_name, base_index_name, mappings, search_config)
       @is_content_index = !(SearchConfig.auxiliary_index_names.include? base_index_name)
     end
 
+    def schema
```
emmalowe, line 31: Are we even using this? I can't see where we are.

spec/unit/search/aggregate_example_fetcher_spec.rb
emmalowe, line 1: A link to https://github.com/alphagov/search-api/pull/3634 and a bit more description in the commit message on these test changes could be helpful.

lib/search/registries.rb
```diff
@@ -51,7 +51,7 @@ def govuk_registry_for_document_format(format)
     end
 
     def govuk_index
-      search_server.index_for_search([SearchConfig.govuk_index_name])
```
emmalowe, line 54: Changes look good but commit message is unfinished.

lib/index.rb
```diff
@@ -124,7 +124,7 @@ def documents_by_format(format, field_definitions)
       }
 
       ScrollEnumerator.new(client: @client, index_names: @index_name, search_body:, batch_size:) do |hit|
-        LegacyClient::MultivalueConverter.new(hit["_source"], field_definitions).converted_hash
+        hit["_source"]
```
emmalowe, line 127: I'm a bit confused about whether "multivalued" as part of the field type (in field_types.json) is an elasticsearch concept or a search-api concept. If a fields that is "multivalued" has only one element in the array, does elasticsearch still return that as an array and if not, do we need it to?

lib/index.rb
koetsier: yes, here: https://github.com/alphagov/search-api/blob/82a3934b589147fec5e966d9f71737fdcb0f877a/lib/search/query.rb#L89

lib/index.rb
koetsier: This is not an ES concept as far as I can see. Elasticsearch retrieves attributes in documents in the '_source' field as they were indexed. Attributes in the documents are returned as a single valued if they're indexed as such and as an array if they were multi valued. 

On the indexing side a similar piece of code exists here:  (https://github.com/alphagov/search-api/blob/616c621fc06e5d581cc566053020a1726d1f84b9/lib/search/presenters/result_presenter.rb#L42)

So it seems to me that if documents are indexed according to the schema we then don't need to check again when retrieving them (if that is even necessary)

spec/unit/search/aggregate_example_fetcher_spec.rb
emmalowe: Thanks for putting the link in the commit message, but I think you put it on the wrong line?

lib/index.rb
emmalowe: That _should_ make sense.

However... I went digging and found a counter example:
https://www.gov.uk/api/search.json?filter_link=/find-licences/safety-certificates-for-sports-grounds-northern-ireland&fields=licence_transaction_licence_identifier

`licence_transaction_licence_identifier` has type "identifiers", which is supposedly multivalued, but it's not returned as an array.

For fields that are **not** supposed to be multivalued, the [ResultsPresenter](https://github.com/alphagov/search-api/blob/616c621fc06e5d581cc566053020a1726d1f84b9/lib/search/presenters/result_presenter.rb#L42) would handle making these single valued. I think there is a gap in the `ResultsPresenter` logic that would previously have been captured by the `MultivalueConverter` where results that are supposed to be multivalued but aren't an array would be indexed into elasticsearch as whatever they started out as and then returned as an array because of the `MultivalueConverter`.

I noticed that the `MultivalueConverter` is not used by a regular search request - it's only used when we get [documents_by_format](https://github.com/alphagov/search-api/blob/main/lib/index.rb#L115), which in turn is only used by the [BaseRegistry](https://github.com/alphagov/search-api/blob/main/lib/search/registry.rb#L34). This is why the gov.uk link above doesn't return `licence_transaction_licence_identifier` as an array. When we do a registry look up, we either return [default fields](https://github.com/alphagov/search-api/blob/main/lib/search/registry.rb#L5), or [specific fields](https://github.com/alphagov/search-api/blob/main/lib/search/registries.rb#L25-L45) if it's an organisation. Out of all of these specified fields that we return on a registries look up, only [organisation related ones](https://github.com/alphagov/search-api/blob/main/lib/search/registries.rb#L41-L44) are multivalued and from a manual check, these all seem to be returned as arrays anyway.

So, in conclusion: I think this is fine to remove 😅

emmalowe (approved):
Looks good to me.

Just a small commit message tidy up if you can; and it would probably be good to read through my logic on the MultivalueConverter to sense check that before merging.

emmalowe:
Such a nice bit of tidy up! 🧹

PR #3659, opened 2026-06-10
Update ranking evaluation rake task to work with BigQuery clickstream dataset 

hannako:
## What

Update the existing rank_evaluation rake task so that it can consume a clickstream judgement list. 

```
kubectl -n apps exec -it deploy/search-api -- bash
rake ranking_evaluation
```

| Old csv format | New csv format |
|-|-|
|Screenshot 2026-06-16 at 13 59 20 | Screenshot 2026-06-16 at 14 00 06| 

## Why

To support the upgrade of elasticsearch from 6.8 to 7.10 we would like to be able to obtain an nDCG score for our search results, so that we can compare relevance pre and post upgrade. The clickstream judgement list is taken from BigQuery, and is the list used to evaluate site search results (powered by search-api-v2/google Agent Search). 

JIRA ticket: https://gov-uk.atlassian.net/browse/SCH-1976

### DIFF
--- a/lib/evaluation/rank_eval.rb
+++ b/lib/evaluation/rank_eval.rb
@@ -4,9 +4,9 @@ require "json"
 
 module Evaluation
   class RankEval
-    def initialize(datafile, ab_tests)
+    def initialize(datafile)
       @data = load_from_csv(datafile)
-      @search_params = ab_tests.nil? ? {} : { "ab_tests" => [ab_tests] }
+      @search_params = {}
       @search_config = SearchConfig.parse_parameters(@search_params).search_config
     end
 
--- a/lib/evaluation/rank_eval.rb
+++ b/lib/evaluation/rank_eval.rb
@@ -4,10 +4,6 @@ require "json"
 
 module Evaluation
   class RankEval
-    def initialize(datafile)
-      @data = load_from_csv(datafile)
-    end
-
     def load_from_csv(datafile)
       data = {}
       last_query = ""
@@ -29,8 +25,8 @@ module Evaluation
       ignore_extra_judgements(data)
     end
 
-    def evaluate
-      requests = queries.each_with_object([]) do |(query, data), acc|
+    def evaluate(csv_data)
+      requests = queries(csv_data).each_with_object([]) do |(query, data), acc|
         acc << {
           id: query,
           request: {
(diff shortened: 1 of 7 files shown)


spec/unit/lib/debug/rank_eval_spec.rb
```diff
@@ -36,24 +49,56 @@
       it_behaves_like "a malformed CSV row validation", "missing content id for row 'harry potter,,3\n'", ["harry potter", nil, 3]
     end
 
-    context "when a valid csv is provided" do
+    context "when the publishing api is available" do
```
stringiest: Should this be 'when the content item is found'?

stringiest (approved):
This looks great!

There are a few links to specific commits rather than main branch. I wasn't sure if this was a deliberate choice so that the links still worked if the code they referred to was removed from main? E.g.
- commit message 'Fetch uri from Publishing API', link 2
- commit message 'Support external content', link 2
- `spec/unit/lib/debug/rank_eval_spec.rb`

I've got a general feeling of unease about our continued use of Harry Potter-related examples. Could we change to something less controversial? Doesn't have to be within this PR - I could make a Jira ticket for it?

hannako:
@stringiest Thank you for the review.

> There are a few links to specific commits rather than main branch. I wasn't sure if this was a deliberate choice so that the links still worked if the code they referred to was removed from main?

Yes deliberate choice so that the link doesn't break.

> I've got a general feeling of unease about our continued use of Harry Potter-related examples. Could we change to something less controversial? Doesn't have to be within this PR - I could make a Jira ticket for it?

I don't think you need to make a ticket, just feel free to open PR's to update them when you feel inclined/are in the area.

lib/evaluation/rank_eval.rb
emmalowe, line 1: Can we rename this folder? It's been (de)bugging me

emmalowe: And the module name too.

hannako: To what?

emmalowe: `evaluation`?

So we have `lib/evaluation/rank_eval.rb` instead of `lib/debug/rank_eval.rb` and `Evaluation::RankEval` instead of `Debug::RankEval`. 

I figured if we're taking this rake task out of `debug.rake` it doesn't make sense for it to be associated with the 'debug' folder anymore.

hannako: Done!

emmalowe (approved):

PR #3660, opened 2026-06-10
Handle errors in the Query class at the top level instead

koetsier:
The Query class previously caught InternalServerError (500).  Elasticsearch seems to have changed
its behaviour and throws BadRequest (400) errors now, so this code does not work.

Furthermore, errors involving numbers out of range or query size being too long should be handled in the validation of input parameters

Instead of catching errors in the Query class, we can catch both InternalServerError and  BadRequest errors at the top level in app.rb instead with makes error handler more straightforward

### DIFF
--- a/lib/rummager/app.rb
+++ b/lib/rummager/app.rb
@@ -97,7 +97,11 @@ class Rummager < Sinatra::Application
     halt(503, "Redis queue timed out")
   end
 
-  error Search::Query::Error do
+  error Elasticsearch::Transport::Transport::Errors::BadRequest do
+    halt(400, env["sinatra.error"].message)
+  end
+
+  error Elasticsearch::Transport::Transport::Errors::InternalServerError do
     halt(400, env["sinatra.error"].message)
   end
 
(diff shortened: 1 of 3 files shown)


lib/rummager/app.rb
```diff
@@ -97,7 +97,11 @@ def json_only
     halt(503, "Redis queue timed out")
   end
 
-  error Search::Query::Error do
+  error Elasticsearch::Transport::Transport::Errors::BadRequest do
+    halt(400, env["sinatra.error"].message)
```
hannako, line 101: What error message would we get here? How would we know if the cause was query too long or the other one?

koetsier: You get the message of the error being caught whatever elasticsearch puts in there. something like: 

```[400] {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"No enum constant org.elasticsearch.index.query.Operator.INVALID_OPERATOR"}],"type":"illegal_argument_exception","reason":"No enum constant org.elasticsearch.index.query.Operator.INVALID_OPERATOR"},"status":400```

koetsier: In reality it is really hard to hit this error because the parameter validator catches most errors like this: https://github.com/alphagov/search-api/blob/main/spec/unit/parameter_parser/search_parameter_parser_spec.rb

hannako (approved):

PR #3663, opened 2026-06-11
Remove unused CoreQuery#match_phrase method

koetsier:
This is not called anywhere

Improves integration test coverage:  https://gov-uk.atlassian.net/browse/SCH-2101

### DIFF
--- a/lib/search/query_components/core_query.rb
+++ b/lib/search/query_components/core_query.rb
@@ -148,18 +148,6 @@ module QueryComponents
       }
     end
 
-    def match_phrase(field_name, query, boost = 1.0)
-      {
-        match_phrase: {
-          synonym_field(field_name) => {
-            boost:,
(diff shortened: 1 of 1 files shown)


hannako (commented):
Can you amend the docs with this PR? 

https://github.com/alphagov/search-api/blob/main/docs/relevancy.md?plain=1#L406

koetsier:
> Can you amend the docs with this PR?
> 
> https://github.com/alphagov/search-api/blob/main/docs/relevancy.md?plain=1#L406

I think the documentation is still relevant. This bit is about how we don't use the 'slop' parameter or the 'match_phrase' queries that Elasticsearch supports and that that may cause problems. I don't think this refers to the '#match_phrase' method I deleted but to the generic Elastisearch 'match_phrase' functionality

hannako (approved):

PR #3664, opened 2026-06-11
Move and fix search:check_recovery rake task

koetsier:
Move search:check_recovery rake task to the debug namespace

This rake task checks whether a restored index has recovered and therefore fits better under debug
 
The fix involves using the 'real' name and not the alias to find the recovery status in the hash returned by @client.indices.recovery

Add an integration test to verify correct behaviour: https://gov-uk.atlassian.net/browse/SCH-2101

### DIFF
--- a/spec/integration/tasks/debug_spec.rb
+++ b/spec/integration/tasks/debug_spec.rb
@@ -170,4 +170,16 @@ RSpec.describe "debug" do
       end
     end
   end
+
+  describe "debug:check_recovery" do
+    let(:task_name) { "debug:check_recovery" }
+    it "returns the recovery status, which on the test database should be 'ok'" do
+      output = capture_stdout { Rake::Task[task_name].invoke(SearchConfig.govuk_index_name) }
+      expect(output).to match(/Recovery status of/)
+      expect(output).to match(/true/)
+    end
+    it "needs an argument" do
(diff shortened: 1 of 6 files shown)


spec/integration/tasks/debug_spec.rb
```diff
@@ -170,4 +170,16 @@
       end
     end
   end
+
+  describe "debug:check_recovery" do
```
hannako, line 174: We don't have a test for passing in a cluster argument. I can see https://github.com/alphagov/search-api/blob/main/lib/tasks/task_helper.rb#L9 that we're defaulting to the default cluster. 

But as we might want to run this rake task as part of the upgrade (presuming that's why you've added this additional coverage) I can see a need to pass in a cluster argument. So I think it's probably worth adding a test case to make sure it works.

koetsier: The different clusters are a support for AB testing (having an 'A' cluster and a 'B' one). We don't really use that so I thought it wasn't worth testing.

I added the tests because I found out the implementation seems to be broken and the unit tests didn't catch that

hannako: Ah gotcha. I was going to say remove the argument but actually let's just move on!

hannako (approved):

PR #3665, opened 2026-06-11
Add integration tests for retrieving examples for aggregate queries

koetsier:
Include examples for example_scope=global and example_scope=query

Jira: https://gov-uk.atlassian.net/browse/SCH-2101

### DIFF
--- a/spec/integration/search/aggregates_spec.rb
+++ b/spec/integration/search/aggregates_spec.rb
@@ -105,15 +105,35 @@ RSpec.describe "search queries" do
         )
       end
 
-      it "returns examples" do
-        get "/search?q=important&aggregate_mainstream_browse_pages=1,examples:5,example_scope:global,example_fields:link:title:mainstream_browse_pages"
+      describe "requesting examples" do
+        context "There is a document with a 'mainstream_browse_pages' attribute that does not match the query" do
+          before do
+            commit_document(index_name, build(:document, :all,
+                                              indexable_content: "doesnotmatch",
+                                              mainstream_browse_pages: "browse/page/1",
+                                              link: "/govuk-3"))
+          end
 
-        expect(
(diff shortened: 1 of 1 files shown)


spec/integration/search/aggregates_spec.rb
```diff
@@ -105,15 +105,33 @@
...
-      it "returns examples" do
-        get "/search?q=important&aggregate_mainstream_browse_pages=1,examples:5,example_scope:global,example_fields:link:title:mainstream_browse_pages"
-
-        expect(
-          parsed_response["aggregates"]["mainstream_browse_pages"]["options"]
-            .first["value"]["example_info"]["examples"]
-            .map { |h| h["link"] }
-            .sort,
-        ).to eq(["/govuk-1"])
+      context "There is a document with a mainstream browse page that does not match the query" do
```
hannako: I think this would be easier to follow. What do you think?
```
context "When requesting examples" ...

it "returns documents that do not match the query if the scope is global" do ...
it "will only return documents that match the query if the scope is query" do ...
```

spec/integration/search/aggregates_spec.rb
```diff
@@ -105,15 +105,35 @@
...
+      describe "requesting examples" do
+        context "There is a document with a 'mainstream_browse_pages' attribute that does not match the query" do
+          before do
+            commit_document(index_name, build(:document, :all,
+                                              indexable_content: "doesnotmatch",
+                                              mainstream_browse_pages: "browse/page/1",
+                                              link: "/govuk-3"))
+          end
+
+          it "returns example documents including those does not match the query if scope is 'global'" do
```
hannako: typo: including those that do not

hannako (approved):
Looks good apart from a small typo, thanks for updating

PR #3666, opened 2026-06-11
Remove SearchConfig#search_servers

koetsier:
These methods are unused

https://gov-uk.atlassian.net/browse/SCH-2101

### DIFF
--- a/lib/search_config.rb
+++ b/lib/search_config.rb
@@ -25,12 +25,6 @@ class SearchConfig
       instance(Clusters.default_cluster)
     end
 
-    def search_servers
-      Cache.get(Cache::SEARCH_SERVERS) do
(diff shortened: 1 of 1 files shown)


lib/search_config.rb
```diff
@@ -118,13 +112,6 @@ def get_index_for_alias(alias_name)
     client.indices.get_alias(index: alias_name).keys.first
   end
 
-  def rank_eval(requests:, metric:, indices: "*")
```
hannako, line 121: I might use this as part of a refactor of the ranking_evaluation debug rake task. I think it makes sense to have it here to centralise request to the client, rather than how it currently is. Could you drop this change? The deletion of search_server looks good

hannako (approved):

PR #3669, opened 2026-06-12
Reduce WebMock usage in the healthcheck integration spec.

koetsier:
Update the live healthcheck to use a real Elasticsearch index.

For connectivity failure scenarios, simulate an unreachable health endpoint by raising Errno::ECONNREFUSED, which better reflects real-world behavior than returning an empty response.

Jira: https://gov-uk.atlassian.net/browse/SCH-2117


emmalowe (approved):
Looks good to me!

PR #3670, opened 2026-06-12
[SCH-2072] Ruby file to format csv input to gor format

stringiest:
Add a rake task to reformat a csv file of kibana logs into gor format. These transactions can then be replayed to using GoReplay.

The rake task is currently written to be run locally from search-api folder on a kibana logs template file saved to `/tmp` using `bundle exec rake traffic_replay:format_logs\[tmp/search_router_traffic_20260618_9am-10am.csv\]`

This is one step in the search traffic replay for elasticsearch upgrade. See [Jira ticket SCH-2072](https://gov-uk.atlassian.net/browse/SCH-2072) for the full process.

### DIFF
--- a/lib/kibana_log_formatter.rb
+++ b/lib/kibana_log_formatter.rb
@@ -1,0 +1,35 @@
+require "csv"
+require "random/formatter"
+
+# Yup that's the line separator format https://github.com/buger/goreplay/wiki/Saving-and-Replaying-from-file#file-format
+GOR_LINE_SEPARATOR = "\n🐵🙈🙉\n".freeze
+
+kibana_logs_file = ARGV[0]
+gor_traffic_replay_file = "tmp/traffic_replay_#{Time.now.strftime('%Y%m%d%H%M%S')}.gor"
+
+def log_timestamp(row)
+  DateTime.iso8601(row["message"][15, 25], Date::ENGLAND).strftime("%Q")
+end
+
+def save_as_gor(kibana_logs_file, output_file)
+  File.open(output_file, "a") do |f|
+    CSV.foreach(kibana_logs_file, headers: true) do |row|
+      f.print get_gor_entry(row)
+      f.print GOR_LINE_SEPARATOR
+    end
+  end
+end
+
+def get_gor_entry(log_entry)
+  # Protocol header
+  # {Protocol Mode} {24 Random Bytes} {timing} {latency}
+  # "1" means it is Making a request
+  gor_entry = []
+  gor_entry << "1 #{Random.hex(12)} #{log_timestamp(log_entry)}000000 0"
+  gor_entry << "GET #{log_entry['request_uri'].strip} HTTP/1.1\r"
+  gor_entry << "Host: www.gov.uk\r"
+
+  "#{gor_entry.join("\n")}\n\n"
+end
+
+save_as_gor(kibana_logs_file, gor_traffic_replay_file)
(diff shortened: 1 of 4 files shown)


lib/kibana_log_formatter.rb
```diff
@@ -0,0 +1,35 @@
...
+end
+
+def get_gor_entry(log_entry)
+  # Protocol header
+  # {Protocol Mode} {24 Random Bytes} {timing} {latency}
+  # "1" means it is Making a request
+  gor_entry = []
+  gor_entry << "1 #{Random.hex(12)} #{log_timestamp(log_entry)}000000 0"
+  gor_entry << "GET #{log_entry['request_uri'].strip} HTTP/1.1\r"
+  gor_entry << "Host: www.gov.uk\r"
```
koetsier: should there be an additional \r\n after this section?

koetsier (commented):
Would it be possible to write a test for this?

What about making this a rake task? 

Let me know what you thinkg

lib/kibana_log_formatter.rb
stringiest: The replayed files work okay without it, so I think not.

stringiest:
> Would it be possible to write a test for this?
> 
> What about making this a rake task?
> 
> Let me know what you thinkg

I think I got overexcited and put this up for review without remembering that I was supposed to make it a rake task. Sorry! Will look at it after the prioritisation meeting.

spec/unit/kibana_log_formatter_spec.rb
```diff
@@ -0,0 +1,36 @@
+require "spec_helper"
+require "tempfile"
+
+RSpec.describe KibanaLogFormatter do
+  let(:file) { Tempfile.create(%w[logs .csv]) }
+  let(:output_gor_file) { Tempfile.create(%w[traffic_replay_ .gor]) }
+
+  input_csv_file = CSV.generate do |csv|
```
koetsier: Would this not have to be a `let` block?

spec/unit/kibana_log_formatter_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  describe "#save_as_gor" do
+    it "formats the output correctly" do
+      formatter.save_as_gor
+      expect(File.open(output_gor_file.path).read).to include(gor_formatted_first_request)
+    end
+  end
+
+  def gor_formatted_first_request
```
koetsier: this could also be a `let`

lib/kibana_log_formatter.rb
```diff
@@ -0,0 +1,41 @@
+require "csv"
+require "random/formatter"
+
+class KibanaLogFormatter
+  attr_accessor :csv_logs_file, :gor_traffic_replay_file
```
koetsier: is this needed? I don't think these variables are accessed outside the class?

koetsier (changes requested):
Looks good. few small comments

spec/unit/kibana_log_formatter_spec.rb
stringiest: Good point - now changed!

spec/unit/kibana_log_formatter_spec.rb
stringiest: Also changed - thanks!

lib/kibana_log_formatter.rb
stringiest: Ah good spot - thanks! Now removed.

koetsier (approved):
lgtm

PR #3671, opened 2026-06-12
Remove redundant TARGETPLATFORM setting

stringiest:
Flagged as a warning in Deploy action e.g. https://github.com/alphagov/search-api/actions/runs/27418020366.

RedundantTargetPlatform: Setting platform to predefined $TARGETPLATFORM in FROM is redundant as this is the default behavior
More info: https://docs.docker.com/go/dockerfile/rule/redundant-target-platform/


hannako (approved):
Thanks for spotting and fixing! 🏅

PR #3677, opened 2026-06-17
fix Elasticsearch version handling 

koetsier:
Use the `version` attribute instead of `_version` for concurrent update handling in Elasticsearch; `_version` is only used in responses.

Replace heavily mocked popularity job tests with integration tests to better verify real behavior rather than implementation details.

see also: 
 https://www.rubydoc.info/gems/elasticsearch-api/Elasticsearch/API/Actions#bulk-instance_method
https://www.elastic.co/guide/en/elasticsearch/reference/6.8/docs-bulk.html#bulk-versioning

Jira: https://gov-uk.atlassian.net/browse/SCH-2164

### DIFF
--- a/lib/govuk_index/popularity_job.rb
+++ b/lib/govuk_index/popularity_job.rb
@@ -29,9 +29,12 @@ module GovukIndex
     def process_document(document, popularities)
       base_path = document.fetch("_id")
       title = document.dig("_source", "title")
-      identifier = document.slice("_id", "_version")
+      identifier = { _id: document["_id"],
+                     _type: "generic-document",
+                     version: document["_version"],
+                     version_type: "external_gte" }
       OpenStruct.new(
-        identifier: identifier.merge("version_type" => "external_gte", "_type" => "generic-document"),
+        identifier:,
         document: document.fetch("_source").merge(
           "popularity" => popularities.dig(base_path, :popularity_score),
           "popularity_b" => popularities.dig(base_path, :popularity_rank),
--- a/spec/integration/govuk_index/popularity_job_spec.rb
+++ b/spec/integration/govuk_index/popularity_job_spec.rb
@@ -2,76 +2,45 @@ require "spec_helper"
 
 RSpec.describe GovukIndex::PopularityJob do
   subject(:job) { described_class.new }
-  let(:index) { instance_double("index") }
+  let(:traffic_index) { SearchConfig.page_traffic_index_name }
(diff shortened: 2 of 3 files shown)


lib/govuk_index/popularity_job.rb
```diff
@@ -29,9 +29,12 @@ def perform(document_ids, index_name)
     def process_document(document, popularities)
       base_path = document.fetch("_id")
       title = document.dig("_source", "title")
-      identifier = document.slice("_id", "_version")
```
emmalowe, line 32: Are there any docs on this? (For my own understanding and the commit message!)

emmalowe: Maybe this: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-bulk#operation-bulk-body-application-json

emmalowe: I also found this helpful, for documentation on our specific version:
https://www.elastic.co/guide/en/elasticsearch/reference/6.8/docs-bulk.html#bulk-versioning

lib/govuk_index/popularity_job.rb
```diff
@@ -29,9 +29,12 @@ def perform(document_ids, index_name)
...
+      identifier = { _id: document["_id"],
+                     _type: "generic-document",
```
emmalowe, line 33: Just to check my understanding: `_type` is deprecated in 7.10 onwards, right?

spec/integration/govuk_index/popularity_job_spec.rb
emmalowe, line 1: Nice! Added benefit that these are much shorter and easier to follow now too 🌟

emmalowe (approved):
Looks good to me! Just a suggestion for some documentation links you could add to the first commit message.

lib/govuk_index/popularity_job.rb
koetsier: this one? https://www.rubydoc.info/gems/elasticsearch-api/Elasticsearch/API/Actions#bulk-instance_method

lib/govuk_index/popularity_job.rb
koetsier: yes it is, that's for the next PR

lib/govuk_index/popularity_job.rb
emmalowe: I was looking for something that mentioned the `version` part of the identifier. Not so easy to find!

PR #3680, opened 2026-06-22
Simplify schema synchroniser code.

koetsier:
Elasticsearch no longer supports multiple mapping types, so the schema synchroniser code has been simplified accordingly.

Jira: https://gov-uk.atlassian.net/browse/SCH-2169


stringiest (approved):
This makes sense to me! Out of an abundance of caution, is it worth deploying to integration just to check it's all okay? 🧹

PR #3686, opened 2026-06-24
Add integration test for page_traffic rake task and small refactors

koetsier:
* Add integration spec for page_traffic task
* Remove ElasticseachProcessor#raw and small refactors of page_traffic_job.rb

Add an integration test for the page_traffic task that runs the code all at once and checks if the relevant documents are created in the page traffic index.

The ElasticseachProcessor#raw method is only used by the PageTrafficJob, and the '#save' method is a better fit. The ElasticseachProcessor#raw is removed and a few small refactors are implemented

https://gov-uk.atlassian.net/browse/SCH-2173

### DIFF
--- a/lib/analytics/ga4_import/elastic_search_relevancy_serialiser.rb
+++ b/lib/analytics/ga4_import/elastic_search_relevancy_serialiser.rb
@@ -8,7 +8,7 @@ module Analytics
       def relevance
         consolidated_data.flat_map.with_index(1) do |(base_path, page_views), index|
           [
-            elastic_search_index(base_path).to_json,
+            elastic_search_identifier(base_path).to_json,
             elastic_search_rank(base_path, index, page_views).to_json,
           ]
         end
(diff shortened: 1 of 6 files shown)


lib/analytics/ga4_import/elastic_search_relevancy_serialiser.rb
```diff
@@ -18,13 +18,8 @@ def relevance
...
       attr_reader :consolidated_data
 
-      def elastic_search_index(base_path)
-        {
-          index: {
-            _type: "page-traffic",
-            _id: base_path,
-          },
-        }
+      def elastic_search_identifier(base_path)
```
hannako, line 21: I don't understand the changes in the second commit - maybe it would help if you split it? Seems like there are three changes and I don't think they are all related.

koetsier: I tried splitting it up, but I think it's doing essentially one thing, which is to get the PageTrafficJob to use ElasticsearchProcessor#save instead of ElasticsearchProcessor#raw and it's really hard to split it up in a meaningful way

hannako: Based on an offline conversation, approving PR

hannako (approved):

PR #3695, opened 2026-07-10
Add rake task to reindex an elasticsearch index from another instance

koetsier:
We are about to upgrade our Elasticsearch cluster from version 6.8 to 7.10. As part of the upgrade we need to reindex the data from the old to the new cluster. This is a rather complex ‘curl’ command and so a rake task is better so we have fewer chances to make a mistake

This PR adds a rake task that reindexes an index from one Elasticsearch cluster to another. 

https://gov-uk.atlassian.net/browse/SCH-2234

### DIFF
--- a/lib/index/remote_reindexer.rb
+++ b/lib/index/remote_reindexer.rb
@@ -1,0 +1,103 @@
+module Index
+  class RemoteReindexer
+    class MissingArguments < ArgumentError; end
+    class ReindexFailed < StandardError; end
+
+    def initialize(source_url:, dest_url:, index:, dest_client: nil, poll_interval: 10, out: $stdout)
+      @source_url = source_url
+      @dest_url = dest_url
+      @index = index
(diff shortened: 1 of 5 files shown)


hannako (approved):
This looks great. Great idea turning it into a rake task with tests 😅 

I was just skimming the [docs](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-reindex)

- `Reindex requires _source to be enabled for all documents in the source` is that fine?
- `If the Elasticsearch security features are enabled, you must have the following security privileges: The read index privilege for the source data stream, index, or alias.
The write index privilege for the destination data stream, index, or index alias.
To automatically create a data stream or index with a reindex API request, you must have the auto_configure, create_index, or manage index privilege for the destination data stream, index, or alias.
If reindexing from a remote cluster, the source.remote.user must have the monitor cluster privilege and the read index privilege for the source data stream, index, or alias.` Do we have that?

koetsier:
Don't need this anymore. We're using Elasticdump instead

PR #3696, opened 2026-07-13
Add shell script to use elasticdump

koetsier:
This script installs 'elasticdump' in a temporary directory, runs it and removes it afterwards.

Example Usage:

bin/reindex.sh --input=https://source:443 --output=https://target:443 --index-source=govuk --retryAttempts=3


koetsier:
superseded

PR #3700, opened 2026-07-17
Refactor rank eval

emmalowe:
Two key changes:

1. Simplifies SearchConfig by moving the elasticsearch client and related methods into Evaluation, since this was the only thing using them. That reduces complexity around clients in the main application.
2. Replace direct HTTP call to rank evaluation endpoint with the client call, now that the relevant client has been isolated to rank evaluation and we can therefore safely increase the timeout on it.

This has been tested on integration.


hannako (approved):
😍

PR #3710, opened 2026-07-22
Add 'elasticdump' rake task

koetsier:
Add rake task to call 'elasticdump' (https://github.com/elasticsearch-dump/elasticsearch-dump) to transfer data from one Elasticsearch cluster to another. It expects an environment variable called ELASTICDUMP_PARAMETERS which should contain a JSON representation of a Hash containing parameters that get passed to elasticdump.

https://gov-uk.atlassian.net/browse/SCH-2234

### DIFF
--- a/lib/elasticdump_runner.rb
+++ b/lib/elasticdump_runner.rb
@@ -1,0 +1,43 @@
+# app/services/elasticdump_runner.rb
+
+class ElasticdumpRunner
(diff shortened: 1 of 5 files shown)


stringiest (approved):
This is very exciting! 

Out of interest, can we test the green cluster's emptiness status before and after the index dump? Haven't made this blocking, as I think it's mostly for my own understanding.

PR #3714, opened 2026-07-24
Upgrade to Elasticsearch 7

koetsier:
This PR ensures Search API can use the legacy Elasticseach 6.8 cluster as well as a newer Elasticsearch 7.10 cluster.

It uses the ElasticsearchClient class to query the cluster to determine the Elasticsearch version and then provides some simple abstractions to ensure compatibility.

Once we fully migrated to Elasticsearch 7.10, the ElasticsearchClient class can be removed.

This PR builds on https://github.com/alphagov/search-api/pull/3548, which upgrades the client gem

Jira: https://gov-uk.atlassian.net/browse/SCH-2026

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ gem "activesupport"
 gem "aws-sdk-s3"
 gem "bootsnap", require: false
 gem "csv"
-gem "elasticsearch", "~> 6" # We need a 6.x release to interface with Elasticsearch 6
+gem "elasticsearch", "7.10"
 gem "gds-api-adapters"
 gem "google-analytics-data-v1beta"
 gem "google-api-client"
--- a/lib/elasticsearch_client.rb
+++ b/lib/elasticsearch_client.rb
@@ -1,5 +1,9 @@
 module ElasticsearchClient
   class << self
+    def search(index_name:, body:, client: Services.elasticsearch)
+      client.search(compatible_params(index: index_name, body:))
+    end
+
     def es7?
       return true if ENV["USE_ELASTICSEARCH_7"]
       return false if ENV["USE_ELASTICSEARCH_6"]
@@ -12,6 +16,12 @@ module ElasticsearchClient
     end
 
   private
+
+    def compatible_params(params)
+      return params if es7?
+
+      params.merge(type: "generic-document")
+    end
 
     def es_version
       @es_version ||= Gem::Version.new(Services.elasticsearch.info.dig("version", "number"))
--- a/lib/elasticsearch_client.rb
+++ b/lib/elasticsearch_client.rb
@@ -12,6 +12,10 @@ module ElasticsearchClient
       client.indices.put_mapping(index: index_name, type: "generic-document", body: mapping)
     end
 
+    def delete(id:, index_name:, client: Services.elasticsearch)
+      client.delete(compatible_params(index: index_name, id: id))
+    end
+
     def search(index_name:, body:, client: Services.elasticsearch)
       client.search(compatible_params(index: index_name, body:))
     end
--- a/lib/elasticsearch_client.rb
+++ b/lib/elasticsearch_client.rb
@@ -1,5 +1,11 @@
 module ElasticsearchClient
   class << self
+    def compatible_mappings(properties)
+      return { "properties" => properties } if es7?
+
+      { "generic-document" => { "properties" => properties } }
+    end
+
     def search(index_name:, body:, client: Services.elasticsearch)
       client.search(compatible_params(index: index_name, body:))
     end
--- a/lib/elasticsearch_client.rb
+++ b/lib/elasticsearch_client.rb
@@ -1,5 +1,11 @@
 module ElasticsearchClient
   class << self
+    def mappings_properties(mappings)
+      return mappings if es7?
+
+      mappings["generic-document"]
+    end
+
     def compatible_mappings(properties)
       return { "properties" => properties } if es7?
 
--- a/lib/elasticsearch_client.rb
+++ b/lib/elasticsearch_client.rb
@@ -6,6 +6,12 @@ module ElasticsearchClient
       { "generic-document" => { "properties" => properties } }
     end
 
+    def put_mapping(index_name:, mapping:, client: Services.elasticsearch)
+      return client.indices.put_mapping(index: index_name, body: mapping) if es7?
+
+      client.indices.put_mapping(index: index_name, type: "generic-document", body: mapping)
+    end
+
     def search(index_name:, body:, client: Services.elasticsearch)
       client.search(compatible_params(index: index_name, body:))
     end
--- a/lib/elasticsearch_client.rb
+++ b/lib/elasticsearch_client.rb
@@ -1,0 +1,20 @@
+module ElasticsearchClient
+  class << self
+    def es7?
+      return true if ENV["USE_ELASTICSEARCH_7"]
+      return false if ENV["USE_ELASTICSEARCH_6"]
+
+      es_version >= Gem::Version.new("7.0.0")
+    end
+
+    def reload_version
+      @es_version = nil
+    end
+
+  private
+
+    def es_version
+      @es_version ||= Gem::Version.new(Services.elasticsearch.info.dig("version", "number"))
+    end
+  end
+end
--- a/lib/elasticsearch_client.rb
+++ b/lib/elasticsearch_client.rb
@@ -40,6 +40,8 @@ module ElasticsearchClient
       return true if ENV["USE_ELASTICSEARCH_7"]
       return false if ENV["USE_ELASTICSEARCH_6"]
 
+      return true if opensearch?
+
       es_version >= Gem::Version.new("7.0.0")
     end
 
@@ -53,6 +55,10 @@ module ElasticsearchClient
       return params if es7?
 
       params.merge(type: "generic-document")
+    end
+
+    def opensearch?
+      Services.elasticsearch.info.dig("version", "distribution") == "opensearch"
     end
 
     def es_version
--- a/lib/elasticsearch_client.rb
+++ b/lib/elasticsearch_client.rb
@@ -12,6 +12,12 @@ module ElasticsearchClient
       { "generic-document" => { "properties" => properties } }
     end
 
+    def compatible_identifier(params)
+      return params if es7?
+
+      params.merge("_type" => "generic-document")
+    end
+
     def put_mapping(index_name:, mapping:, client: Services.elasticsearch)
       return client.indices.put_mapping(index: index_name, body: mapping) if es7?
 
--- a/lib/elasticsearch_client.rb
+++ b/lib/elasticsearch_client.rb
@@ -4,6 +4,10 @@ module ElasticsearchClient
       client.search(compatible_params(index: index_name, body:))
     end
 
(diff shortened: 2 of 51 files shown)


Gemfile
```diff
@@ -4,7 +4,7 @@ gem "activesupport"
 gem "aws-sdk-s3"
 gem "bootsnap", require: false
 gem "csv"
-gem "elasticsearch", "~> 6" # We need a 6.x release to interface with Elasticsearch 6
+gem "elasticsearch", "7.10"
```
hannako, line 7: The [commit](https://github.com/alphagov/search-api/pull/3714/changes/92c3c712669f6e1df911ba826102daa34aa50f3a) message is weird, do you need to update it?

spec/unit/elasticsearch_index_spec.rb
```diff
@@ -51,7 +51,7 @@
   end
 
   it "can be searched" do
-    stub_get = stub_request(:get, "http://example.com:9200/govuk_test/generic-document/_search").with(
+    stub_get = stub_request(:post, "http://example.com:9200/govuk_test/generic-document/_search").with(
```
hannako: Do you have a reference for this you could add to the [commit](https://github.com/alphagov/search-api/pull/3714/changes/2f105f087c2922fe9b0c39d85167ba66bca16421)? I can't find anything that lists this as a breaking change. Is this not the [right](https://www.elastic.co/guide/en/elasticsearch/client/ruby-api/7.17/examples.html#ex-get) documentation?

lib/search/query_components/booster.rb
```diff
@@ -12,7 +12,7 @@ def wrap(core_query)
           score_mode: :multiply,
           query: {
             bool: {
-              should: [core_query],
+              should: [core_query].flatten,
```
hannako, line 15: For this [commit](https://github.com/alphagov/search-api/pull/3714/changes/12301a92e24542fec4ae6ddfb5c47f1e5bd6fd33) you have a link to documentation? For breaking changes, or to the deprecations? I feel like I'm reviewing it a bit blind 😂

spec/integration/scroll_enumerator_spec.rb
```diff
@@ -24,7 +24,7 @@
     results = ScrollEnumerator.new(
       client:,
       index_names: "govuk_test",
-      search_body: { query: { match_all: {} }, sort: [{ _uid: { order: "asc" } }] },
+      search_body: { query: { match_all: {} }, sort: [{ _id: { order: "asc" } }] },
```
hannako: You could also reference https://www.elastic.co/guide/en/elasticsearch/reference/7.0/breaking-changes-7.0.html#uid-meta-field-removed in https://github.com/alphagov/search-api/pull/3714/changes/dc65a18c94ca2cc8a909cfb58342182682212e0c

lib/elasticsearch_response.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class ElasticsearchResponse
```
hannako, line 3: This is nice! [Link](https://www.elastic.co/guide/en/elasticsearch/reference/7.0/breaking-changes-7.0.html#hits-total-now-object-search-response) to documentation for the [commit](https://github.com/alphagov/search-api/pull/3714/changes/471086050c68598aa052b1c73de90ac1d9d66553) message or inline - whichever easier.

lib/govuk_index/presenters/elasticsearch_delete_presenter.rb
```diff
@@ -33,7 +33,7 @@ def type
     def existing_document
       @existing_document ||=
         begin
-          Client.get(type: "_all", id:)
```
hannako, line 36: Just on https://github.com/alphagov/search-api/pull/3714/changes/3fb174e192dc4ff996381221d44d8b3663802c7f, I thought it was less about the "_all" not being supported in 7.10, I thought it was that a `type` parameter of any value - is no longer supported?

spec/support/integration_test_helper.rb
```diff
@@ -22,6 +22,7 @@ def self.allow_elasticsearch_connection_to_test(testing_url = nil)
     allowed_paths << "_nodes"
     allowed_paths << "_cluster"
     allowed_paths << "_msearch"
+    allowed_paths << "\\z"
```
hannako, line 25: Is this needed? It's added in https://github.com/alphagov/search-api/pull/3714/changes/af8a400c71fe6b598a1b80b6f978b8f83eced87d

lib/elasticsearch_client.rb
```diff
@@ -0,0 +1,20 @@
+module ElasticsearchClient
```
hannako, line 1: This is nice 😍

spec/spec_helper.rb
```diff
@@ -96,6 +100,12 @@
     ClimateControl.modify(TZ: "UTC") { example.run }
   end
 
+  config.around(:each, :unit) do |example|
+    ClimateControl.modify(USE_ELASTICSEARCH_7: "true") do
```
hannako, line 106: But given we are planning to ship this PR, and continue to run SearchAPI against a 6.8 cluster for 2 weeks, is setting all tests to run against 7.10 a bit risky? Should we remove [this commit ](https://github.com/alphagov/search-api/pull/3714/changes/8eb18ca7534d839d3940c93ba41577228aa011c6#r3676292614) for now - now you know the tests will pass - and add it back in just before we do the upgrade?

spec/unit/schema/index_schema_parser_spec.rb
```diff
@@ -23,8 +23,7 @@ def schema_dir
 
     it "include configuration for the `manual section` type in the `govuk` index" do
       es_mappings = @index_schemas["govuk"].es_mappings
-      expect(es_mappings.keys).to include("generic-document")
-      expect(es_mappings["generic-document"]["properties"]).to match(
+      expect(es_mappings["properties"]).to match(
```
hannako, line 26: As per https://github.com/alphagov/search-api/pull/3714/changes/8eb18ca7534d839d3940c93ba41577228aa011c6#r3676292614 given we're not upgrading straight to 7.10 I think we should probably run these tests against both versions don't you? i.e. add 6.8 and 7.10 context blocks to these specs and amend the expectations. I know that's a faff - but if something delays the upgrade after your leave, we need to feel confident that the test suite is still reliable and not in a sort of half way state.

lib/elasticsearch_client.rb
```diff
@@ -4,6 +4,10 @@ def search(index_name:, body:, client: Services.elasticsearch)
       client.search(compatible_params(index: index_name, body:))
     end
 
+    def index(id:, index_name:, atts:, params: {}, client: Services.elasticsearch)
```
hannako, line 35: Minor typo in https://github.com/alphagov/search-api/pull/3714/changes/487981ed375fa5f24c432877f072ba028c779c4c message `Introduce ElasticsearchClient.search` -> `Introduce ElasticsearchClient.index`

lib/services.rb
```diff
@@ -31,15 +31,25 @@ def self.s3_client
...
-      hosts: cluster ? cluster.uri : hosts,
+      hosts: with_default_port(raw_uri),
       request_timeout: timeout,
       logger: Logging.logger[self],
       retry_on_failure:,
       transport_options: { headers: { "Content-Type" => "application/json" } },
     )
   end
 
+  def self.with_default_port(uri)
```
hannako, line 45: This change looks different to what you've explained in the commit message. Do both 6.8 and 7.10 default to 9200 for http and 443 for https?

spec/unit/elasticsearch_client_spec.rb
```diff
@@ -87,6 +87,46 @@
         end
       end
 
+      context "when the server is OpenSearch" do
+        let(:version_info) do
+          { "version" => { "distribution" => "opensearch" } }
```
hannako, line 92: Just checking that it's correct that the key is `distribution` rather than `number`?

hannako (commented):
So much work here Jos, thank you for breaking it into really easy to follow commits.

I've made a bunch of comments which are mainly a request for links to documentation and a question about running unit tests against 6.8 as well as 7.0

spec/unit/elasticsearch_index_spec.rb
koetsier: I added this link: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-search

This explains that GET and POST are both valid. The es7 gem chooses POST instead of GET. I don't think that there's any documentation explicitly stating that the gem produces a different method.

lib/search/query_components/booster.rb
koetsier: I added a link to general documentation about this, which explains 'should' has to be an array. There is no explicit documentation that arrays of arrays used to work and now doesn't. It is likely ES7.10 is simply stricter.

spec/support/integration_test_helper.rb
koetsier: Yes because `Services.elasticsearch.info.dig("version", "number")` calls this path. (i.e. it does a GET /)

spec/spec_helper.rb
koetsier: This change only sets the Elasticsearch version to 7 for unit tests, not integration tests. The reason we have to set this is because the unit tests don't use a real instance of Elasticsearch, they use canned request-responses and they need to be aligned to one or the other.  This PR has already updated the unit tests to make them compatible with ES7, so setting this parameter to ES6 would cause them to break.

This is the reason we had to add extra integration tests so that we can be more confident this PR works.

spec/unit/schema/index_schema_parser_spec.rb
koetsier: That should be covered by integration tests. I'm not sure if adding more tests make this better. This particular test deals with mappings, which is not used when running the search api, only when creating a new index.

spec/unit/elasticsearch_client_spec.rb
koetsier: It looks like this
```
{
  "name" : "d81c2896760e",
  "cluster_name" : "docker-cluster",
  "cluster_uuid" : "KIWQ_pmZTEK9NcfSfvCclA",
  "version" : {
    "distribution" : "opensearch",
    "number" : "3.7.0",
    "build_type" : "tar",
    "build_hash" : "72121f014083f9ca010fd5a7da83b2ec4886027f",
    "build_date" : "2026-06-03T23:29:44.853498367Z",
    "build_snapshot" : false,
    "lucene_version" : "10.4.0",
    "minimum_wire_compatibility_version" : "2.19.0",
    "minimum_index_compatibility_version" : "2.0.0"
  },
  "tagline" : "The OpenSearch Project: https://opensearch.org/"
}
```
The `version->number` here is lower than 6.8, so we have to look at the `distribution` value instead

lib/services.rb
koetsier: This bit adds an explicit port number if there is none, so 9200 for http and 443 for https. This means that if we don't specify a port number it uses the correct one.

hannako: But do both versions of Elasticsearch use the same defaults?

hannako (approved):
This looks good - do you want to set it to draft and we can merge when you're back?

PR #3720, opened 2026-07-28
Enable branch coverage

emmalowe:
Adds branch coverage to coverage report.

Also updates simplecov configuration to use new method names since the update to version 1.0.2. This removes deprecation warnings brought in by the upgrade: 
https://github.com/alphagov/search-api/actions/runs/29866977029/job/88757552706


stringiest (approved):
Ah nice!

PR #3739, opened 2026-08-20
Track total hits in search API

koetsier:
Set track_total_hits: true so the Search API returns the exact number of matching documents.

Elasticsearch 7 changed the default to only count hits up to 10,000, returning relation: "gte" above that limit.

See: https://www.elastic.co/docs/solutions/search/the-search-api?track-total-hits

https://gov-uk.atlassian.net/browse/SCH-2304


emmalowe (approved):
LGTM!

PR #3742, opened 2026-08-24
Add 'general rules' for when to use each field type

ChrisBAshton:
The descriptions in field_types.json were useful, but it was still difficult to know where to start in choosing a field type for each of the new fields I was adding for a Specialist Finder. After a fair bit of sense-checking with an LLM, this is the rule of thumb I've come up with.

Many of the fields for existing finders won't be quite of the correct type under this guidance, but there's little value in fixing-forward those (and the subsequent re-index that would be necessary if we did). But we can try to be consistent with new fields going forward.

### DIFF
--- a/docs/schemas.md
+++ b/docs/schemas.md
@@ -67,6 +67,23 @@ value has the following properties:
    the same format as the top-level field definitions in the file (and could
    even be recursive).
 
+### Choosing a field type
+
+The key distinction is **what the field represents** and **whether it should contribute to general search**:
+
+| Type                          | Meaning                                                              | Example                                 |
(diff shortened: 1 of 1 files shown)


hannako (approved):
Thanks for opening this PR Chris 🌟

eYinka (commented):
Nice one! Just wondering if it's worth mentioning that the `date` field also doubles as a duration field on a finder's filters. See Decision date facet on [Residential property tribunal decisions](https://www.gov.uk/residential-property-tribunal-decisions)

ChrisBAshton:
Thanks @eYinka - I think I'll omit that as it's a finder-frontend specific implementation detail, whereas this is more about _how_ we want the data itself to be indexed.

PR #3745, opened 2026-08-25
Sch 1964 document search analytics pipeline custom service account

emmalowe:
Adds a reference to the [search-analytics-pipeline service account](https://github.com/alphagov/govuk-data-infrastructure/blob/main/ga4-analytics/service_account.tf#L20) in the search-api docs. This service account gives permissions to search-api to access data stored in the GA4 Analytics GCP project, for the page_traffic job to run. Adding a reference here, because it's not clear from the name of the service account what it is used for, and we want that information to be more easily findable.

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-1964


stringiest (approved):
✨

